PR #168, opened 2017-06-30
EPIPE and ESTRPIPE may be equal

### LINKED ISSUE #133: EPIPE and ESTRPIPE may be the same things
Compiling on FreeBSD today -- with ALSA-driver enabled -- I got the following error:

    FAILED: /usr/bin/cc  -DHAVE_CONFIG_H -Dlibfluidsynth_EXPORTS -O2 -pipe -march=opteron  -I/opt/include/portaudio2 -I/opt/include  -fstack-protector -fno-strict-aliasing -O2 -pipe -march=opteron  -I/opt/include/portaudio2 -I/opt/include  -fstack-protector -fno-strict-aliasing -fPIC -I. -Isrc -Isrc/drivers -Isrc/synth -Isrc/rvoice -Isrc/midi -Isrc/utils -Isrc/sfloader -Isrc/bindings -Iinclude -I/opt/include/glib-2.0 -I/opt/lib/glib-2.0/include -I/opt/include -I/opt/include/alsa -MMD -MT src/CMakeFiles/libfluidsynth.dir/drivers/fluid_alsa.c.o -MF src/CMakeFiles/libfluidsynth.dir/drivers/fluid_alsa.c.o.d -o src/CMakeFiles/libfluidsynth.dir/drivers/fluid_alsa.c.o   -c src/drivers/fluid_alsa.c

    ...
    src/drivers/fluid_alsa.c:357:8: error: duplicate case value '-32'
      case -ESTRPIPE:
           ^
    src/drivers/fluid_alsa.c:350:8: note: previous case defined here
      case -EPIPE:

Indeed, both `EPIPE` and `ESTRPIPE` are -32 here, which breaks switch/case construct.

The proposed patch moves the `EPIPE` case next to `ESTRPIPE` (which makes a tad more sense) and makes sure, the two do not clash if they are defined to the same value.

The fluid_alsa.c driver still generates a large number of _deprecated_ warnings (`g_static_mutex_get_mutex_impl` is deprecated in modern glib2), but it compiles now. The patch makes no attempt to address the warnings.

Reported by: kot

Original Ticket: [fluidsynth/tickets/135](https://sourceforge.net/p/fluidsynth/tickets/135)

derselbst:
fixes #133

### DIFF
--- a/fluidsynth/src/drivers/fluid_alsa.c
+++ b/fluidsynth/src/drivers/fluid_alsa.c
@@ -349,7 +349,7 @@ static int fluid_alsa_handle_write_error (snd_pcm_t *pcm, int errval)
     break;
 // on some BSD variants ESTRPIPE is defined as EPIPE.
 // not sure why, maybe because this version of alsa doesnt support
-// suspending pcm stream. anyway, since EPIPE seems to be more 
+// suspending pcm streams. anyway, since EPIPE seems to be more 
 // likely than ESTRPIPE, so ifdef it out in case.
 #if ESTRPIPE != EPIPE
   case -ESTRPIPE:
--- a/fluidsynth/src/drivers/fluid_alsa.c
+++ b/fluidsynth/src/drivers/fluid_alsa.c
@@ -351,7 +351,9 @@ static int fluid_alsa_handle_write_error (snd_pcm_t *pcm, int errval)
 // not sure why, maybe because this version of alsa doesnt support
 // suspending pcm streams. anyway, since EPIPE seems to be more 
 // likely than ESTRPIPE, so ifdef it out in case.
-#if ESTRPIPE != EPIPE
+#if ESTRPIPE == EPIPE
+  #warning "ESTRPIPE defined as EPIPE. This may cause trouble with ALSA playback."
+#else
   case -ESTRPIPE:
     if (snd_pcm_resume(pcm) != 0) {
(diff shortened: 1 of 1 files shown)


fluidsynth/src/drivers/fluid_alsa.c
```diff
@@ -347,19 +347,25 @@ static int fluid_alsa_handle_write_error (snd_pcm_t *pcm, int errval)
...
+// suspending pcm stream. anyway, since EPIPE seems to be more 
+// likely than ESTRPIPE, so ifdef it out in case.
+#if ESTRPIPE != EPIPE
+  case -ESTRPIPE:
+    if (snd_pcm_resume(pcm) != 0) {
+      FLUID_LOG(FLUID_ERR, "Failed to resume the audio device");
+      return FLUID_FAILED;
+    }
+  /* fall through, since the stream got resumed, but still has to be prepared */
+#endif
```
ensonic: So on systems that have ESTRPIPE == EPIPE you would skip the snd_pcm_resume(pcm) call?

derselbst: Yes, ESTRPIPE says the pcm stream is suspended and needs to be resumed(). I consider the condition `ESTRPIPE == EPIPE` an error in itself, dunno why such a define is possible at all. The only explanation is that these systems dont support suspended streams thus, ignore it at all, as EPIPE is more likely.

PR #177, opened 2017-07-12
License Upgrade to LGPL-2.1+

derselbst:
This pull request addresses two issues:

1. relicense fluid_chorus.c under the LGPL
1. consistently upgrade fluidsynth from `GNU Library GPL 2` to `GNU Lesser GPL 2.1`

### Reasoning
1. Packagers at debian have detected that fluidsynth has infringed GPL due to the copyright notice of `fluid_chrous.c` (see http://lists.nongnu.org/archive/html/fluid-dev/2017-04/msg00000.html). I think we are allowed to relicense this file under LGPL because as [pointed out by Peter Hanappe](http://lists.nongnu.org/archive/html/fluid-dev/2017-04/msg00003.html):
    * all changes done by fluidsynth contributors were made under the terms of the LGPL
    * SoX/chorus.c is part of libsox, a library licensed under the terms of the LGPL

    Also due to the integration of `SoX/chorus.c` into fluidsynth and the implementation of a variable delay line there is no trace of the original implementation anymore, [see my comment below](https://github.com/FluidSynth/fluidsynth/pull/177#issuecomment-315391773).
1. There was a dissent under which license fluidsynth ships: Although all file headers point out that it's licensed under the `GNU Library GPL 2`, some places link to the `GNU Lesser GPL 2.1`. This is not dramatic as these licenses are compatible to each other, however it may cause confusions. For the sake of consistency we should get rid of the very outdated `Library GPL` and go with the `Lesser GPL`.

### DIFF
--- a/fluidsynth/CMakeLists.txt
+++ b/fluidsynth/CMakeLists.txt
@@ -3,14 +3,14 @@
 # Copyright (C) 2003-2011 Peter Hanappe and others.
 #
 # This library is free software; you can redistribute it and/or
-# modify it under the terms of the GNU Library General Public License
+# modify it under the terms of the GNU Lesser General Public License
 # as published by the Free Software Foundation; either version 2 of
 # the License, or (at your option) any later version.
 #
 # This library is distributed in the hope that it will be useful, but
 # WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-# Library General Public License for more details.
+# Lesser General Public License for more details.
 #
 # You should have received a copy of the GNU Library General Public
 # License along with this library; if not, write to the Free
--- a/fluidsynth/CMakeLists.txt
+++ b/fluidsynth/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public License
-# as published by the Free Software Foundation; either version 2 of
+# as published by the Free Software Foundation; either version 2.1 of
 # the License, or (at your option) any later version.
 #
 # This library is distributed in the hope that it will be useful, but
--- a/fluidsynth/CMakeLists.txt
+++ b/fluidsynth/CMakeLists.txt
@@ -12,7 +12,7 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 # Lesser General Public License for more details.
 #
-# You should have received a copy of the GNU Library General Public
+# You should have received a copy of the GNU Lesser General Public
 # License along with this library; if not, write to the Free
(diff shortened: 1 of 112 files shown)


derselbst:
@hanappe @diwic Could you please look into this and let me know whether you (dis)agree with this change?

tresf:
> get rid of the very outdated Library GPL and go with the Lesser GPL.

As far as I understand, you can go LGPL -> GPL, not the other way around (unless sole copyright holder, or permission from all contributors).  As much as it pains me to say this, I think you're stuck with GPL.  You may also receive criticism for changing this text without explicit reference.  To alter the copyright [as you have](https://github.com/FluidSynth/fluidsynth/pull/177/commits/2b7d2bf10831569a28640a396dccbe8418530153#diff-1550e03d37c4b26fc1affccee69356e7R3) can get quite a bit of push back.

elementgreen:
@tresf Your statement is incorrect.  What is being changed "GNU Library GPL 2 to GNU Lesser GPL 2.1".  That is LGPL 2 to LGPL 2.1.  It used to be called Library GPL.  To clarify matters more.  The "fluidsynth" executable is GPL, while libfluidsynth is LGPL.

Again, it sounds like the change which is being requested is to change libfluidsynth from LGPL 2 -> LGPL 2.1.

derselbst:
> Again, it sounds like the change which is being requested is to change libfluidsynth from LGPL 2 -> LGPL 2.1.

Correct.

tresf:
> It used to be called Library GPL.

Thanks for clarification. 👍

derselbst:
To better comprehend the changes fluid_chorus.c underwent over time, please see the attached zip file. It contains [fluid_chorus.c from recent git](https://github.com/FluidSynth/fluidsynth/blob/b456e88efce3eebafd9cf0ae4c2d5743552903ab/fluidsynth/src/rvoice/fluid_chorus.c) and [SoX/chorus.c  from sox-12.17.3](https://sourceforge.net/projects/sox/files/sox/12.17.3/). The diff was generated by

`diff -NauriwB chorus.c fluid_chorus.c`

So far I cant see any functionality written by Juergen Mueller that survived the changes. What I see is a complete rewrite of the original file (except for the comments) which IMO justifies adopting the copyright notice as suggested by this PR.

[sox-12.17.3_chorus_diff.zip](https://github.com/FluidSynth/fluidsynth/files/1148873/sox-12.17.3_chorus_diff.zip)

diwic:
I don't know of any significant changes between LGPL 2.0 and LGPL 2.1 that would make me prefer one or the other. Basically this is a one-way operation, so if LGPL 2.1 has any drawbacks, we can't go back to LGPL 2.0. But if this has already happened to some files, then that transition might already have happened in practice.

As for the `chorus.c` investigation, I'm just going to say IANAL and hope that the rest of you know what you're doing :-)

Anyhow, thanks for taking care of this boring stuff.

diwic:
Which means that if you're looking for a formal review comment, it's going to be along the lines of "don't know, don't care, go ahead and leave me out of it" essentially :-)

Issue #179, opened 2017-07-21
Mixed whitespace


markh-de:
My OCD (just kidding) is asking me to fix FluidSynth's source-code whitespace, as I have noticed that it contains mixed-style tabs and spaces. We even have one line with "misleading indentation", which will be fixed when cleaning up.
What is your project-specific whitespace/indentation policy? Personally (and professionally) I prefer 2 spaces per level, no tabs.
Also, any interest to pass the code through clang-format or uncrustify, while we're at it?
Or are you worried about messing up the file history (e.g. for "blame")?
I already made a fork and would send you PRs when I got something. Is that OK for you?

derselbst:
Yes, I agree it's inconsistent and it's good to see that someone cares.

> What is your project-specific whitespace/indentation policy?

Not sure, there doesnt seem to exist any written style guide and I dont want to bother the former maintainers with that topic. What I can tell from the sources is that it seems we are using K&R brace placement and 4 spaces indentation. Though I personally prefer Allman brace placement, I think it's better to stick to fluidsynths conventions.

> Also, any interest to pass the code through clang-format or uncrustify, while we're at it?

clang-format sounds like a really good idea. Feel free to prepare the source and submit a PR. However: I will not accept any code re-formatting before the next major release (i.e. 1.2 or 2.0, whatever it'll be called). This is because I want to merge in the changes done by several custom forks here (SF3 support, lyrics support, API for manip def. mod., ...) and whitspace changes will make it harder to trace back changes.

Anyway, this issue is suited to be done before the next major release, after all the dev work, API adjustment and making-fluidsynths-folder-structure-github-friendly work is done :)

PR #180, opened 2017-07-27
Fix GCC 7 and Clang 4 compiler warnings

markh-de:
Fixes compiler warnings of GCC 7 and Clang 4 and very few minor code format issues.

### DIFF
--- a/fluidsynth/src/sfloader/fluid_defsfont.h
+++ b/fluidsynth/src/sfloader/fluid_defsfont.h
@@ -435,7 +435,7 @@ struct _fluid_defpreset_t
 {
   fluid_defpreset_t* next;
   fluid_defsfont_t* sfont;                  /* the soundfont this preset belongs to */
-  char name[21];                        /* the name of the preset */
+  char name[22];                        /* the name of the preset */
   unsigned int bank;                    /* the bank number */
   unsigned int num;                     /* the preset number */
   fluid_preset_zone_t* global_zone;        /* the global zone of the preset */
--- a/fluidsynth/src/bindings/fluid_rtkit.c
+++ b/fluidsynth/src/bindings/fluid_rtkit.c
@@ -146,7 +146,7 @@ finish:
 }
 
 int rtkit_get_max_realtime_priority(DBusConnection *connection) {
-        long long retval;
+        long long retval = 0;
         int err;
 
         err = rtkit_get_int_property(connection, "MaxRealtimePriority", &retval);
@@ -154,7 +154,7 @@ int rtkit_get_max_realtime_priority(DBusConnection *connection) {
 }
 
 int rtkit_get_min_nice_level(DBusConnection *connection, int* min_nice_level) {
-        long long retval;
+        long long retval = 0;
         int err;
 
         err = rtkit_get_int_property(connection, "MinNiceLevel", &retval);
@@ -164,7 +164,7 @@ int rtkit_get_min_nice_level(DBusConnection *connection, int* min_nice_level) {
 }
 
 long long rtkit_get_rttime_nsec_max(DBusConnection *connection) {
-        long long retval;
+        long long retval = 0;
         int err;
 
         err = rtkit_get_int_property(connection, "RTTimeNSecMax", &retval);
--- a/fluidsynth/src/drivers/fluid_alsa.c
+++ b/fluidsynth/src/drivers/fluid_alsa.c
@@ -359,8 +359,9 @@ static int fluid_alsa_handle_write_error (snd_pcm_t *pcm, int errval)
       FLUID_LOG(FLUID_ERR, "Failed to resume the audio device");
       return FLUID_FAILED;
     }
-  /* fall through, since the stream got resumed, but still has to be prepared */
 #endif
+  /* fall through ... */
+  /* ... since the stream got resumed, but still has to be prepared */
   case -EPIPE:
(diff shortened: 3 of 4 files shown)


markh-de:
Related to #179.

fluidsynth/src/sfloader/fluid_defsfont.h
```diff
@@ -435,7 +435,7 @@ struct _fluid_defpreset_t
 {
   fluid_defpreset_t* next;
   fluid_defsfont_t* sfont;                  /* the soundfont this preset belongs to */
-  char name[21];                        /* the name of the preset */
+  char name[22];                        /* the name of the preset */
```
derselbst: The SF2 spec says a preset name has 20 characters without having counted the terminal '\0' (see `The PHDR Sub-chunk`, page 21). So changing this to 22 actually is a mistake IMO. Where got this format overflow warning triggered?

derselbst: Found it: fluid_defsfont.c:1045

And since fluid_defpreset_t fluidsynths private preset struct, I think we can increase it to 22.

derselbst:
Ok, rest looks fine. Could you perhaps consider to associate the email you used for these commits with your github account? Merging it anyway.

fluidsynth/src/sfloader/fluid_defsfont.h
markh-de: Hm, I agree with your first comment. Maybe we should really fix the actual `FLUID_SPRINTF`, where GCC assumes the overflow to appear. Of course the compiler assumes that the arguments `bank` and `prenum` can become up to 5 bytes in size, each. I'm currently not sure about the valid range of `bank` (max. 128 for drums?). But if I just read correctly in the spec, at least preset numbers are in the range 0..127.
So, can you confirm that the `prenum` value cannot get greater than 127? Then I would change that argument to `sfpreset->prenum % 128`. GCC will recognize this and know that the string will always fit. And I'll revert the array size change.

derselbst: > the compiler assumes that the arguments bank and prenum can become up to 5 bytes in size, each.

4 bytes, but 5 decimal digits.

> 7.2 The PHDR Sub-chunk
> ...
> If the value in [wBank and wPreset] is not a valid MIDI value of zero through 127, or 128 for wBank, the preset cannot be played but should be maintained.

Which means they can have any value of 0 to 65536 and must be maintained (i.e. loaded; and from the code it seems fluidsynth would really load such presets...) although they become unusable because a midi prog change only ranges from 0 to 127. However for the bank selection having CC0 and CC32 we should be able to address up to `127*128+127=16383` banks in mma mode.

Anyway, internally I would prefer to have uniquely named preset, i.e. not doing `sfpreset->prenum % 128`.

markh-de: > 4 bytes, but 5 decimal digits.

Really 2 bytes (unsigned short). But yes, 5 digits. By "become" I meant the actual output of the sprintf, where digits become bytes.

> Anyway, internally I would prefer to have uniquely named preset, i.e. not doing sfpreset->prenum % 128.

OK, right. If input values are valid throughout the full `unsigned short` range, then we must of course represent their values properly. But then, sadly, GCC is right: We had a buffer overflow here. We'd need to shorten the `Bank%d,Preset%d` format string by at least one byte, in order to fit into 20(+1) bytes, as per the SF2 spec. And I think you're right, we shouldn't use longer names than the SF2 spec permits (at least not only for the purpose of fitting this default string).

derselbst: > Really 2 bytes (unsigned short).

Yep, sry, noticed too late.

> we shouldn't use longer names than the SF2 spec permits (at least not only for the purpose of fitting this default string).

`SFPreset` is the struct used for reading soundfonts. This correctly uses [char name[21]](https://github.com/FluidSynth/fluidsynth/blob/master/fluidsynth/src/sfloader/fluid_defsfont.h#L129). What we make up internally with `_fluid_defpreset_t` shouldnt matter that much, should it? One would always use `strlen() and its friends` and not just silently assume there are max. 21 chars IMO.

markh-de: Well, of course we can internally maintain names longer than what we actually read from the Soundfont file (as long as we only read and never write SF2). But do you really want such a special case **only** for the purpose of assigning a default name for that preset? Any other preset names will always fit in 20+1 bytes. I think this default preset name assignment is such a minor special case that it would be nice if it would simply stand to the rules.
What about shortening the default name to `Bank%d,Pre%d` or simply assign or leave it an empty string? I mean, if the incoming SF2 preset name is empty, why tinker with it and not simply leave it empty?

derselbst: > I mean, if the incoming SF2 preset name is empty, why tinker with it and not simply leave it empty?

I dont know why it was done like this, but I'm not feeling comfortable breaking with this convention either. I can also understand your point, it IS a very rare case. So taking all of that into account again we might be best doing to change the default string as you suggested. Feel free to create another PR.

derselbst: @markh-de I'll apply the change at the weekend if you dont mind and havent done it yet.

Issue #140, opened 2015-06-23
Support of vorbis-compressed sf3 sound fonts 


derselbst:
Musescore 2.0 introduced a new soundfont format sf3
See: https://musescore.org/en/node/20818

I took the liberty to post this request from Fabian Greffrath <fabian@greffrath.com> from
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=740710 into the ticket system of fluidsynth.

Musescore is widely used and created this new format as part of their version 2.0 delivery.
Would be great if this feature would be officially distributed by fluidsynth.

Reinhold

Reported by: z34rgfj35

Original Ticket: [fluidsynth/tickets/142](https://sourceforge.net/p/fluidsynth/tickets/142)

fabiangreffrath:
See:
https://github.com/fabiangreffrath/fluidsynth-code-git/commit/6a9e05ee3e486ac1c91fda3d0df535422934a578

derselbst:
@fabiangreffrath Looks good. Feel free to create a pull request so this can be discussed. However, I currently dont think that this will make it into the upcoming 1.1.7 maintenance release.

PR #183, opened 2017-08-02
Add support for sound fonts in SF3 format that contain Ogg Vorbis compressed samples

fabiangreffrath:
Author: Fabian Greffrath <fabian@debian.org>
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=740710
Forwarded: https://sourceforge.net/p/fluidsynth/tickets/142/
Last-Update: 2015-11-26

Fixes https://github.com/FluidSynth/fluidsynth/issues/140

### DIFF
--- a/fluidsynth/src/sfloader/fluid_defsfont.c
+++ b/fluidsynth/src/sfloader/fluid_defsfont.c
@@ -1838,7 +1838,7 @@ sfvio_get_filelen(void* user_data)
   fluid_sample_t *sample = (fluid_sample_t *)user_data;
 
   return (sf_count_t)(sample->end + 1 - sample->start);
-};
+}
 
 static sf_count_t
 sfvio_seek(sf_count_t offset, int whence, void* user_data)
@@ -1859,7 +1859,7 @@ sfvio_seek(sf_count_t offset, int whence, void* user_data)
   }
 
   return (sf_count_t)sample->userdata;
-};
+}
 
 static sf_count_t
 sfvio_read(void* ptr, sf_count_t count, void* user_data)
@@ -1873,7 +1873,7 @@ sfvio_read(void* ptr, sf_count_t count, void* user_data)
   sample->userdata = (void *)((intptr_t)sample->userdata + count);
 
   return count;
-};
+}
 
 static sf_count_t
 sfvio_tell (void* user_data)
@@ -1881,7 +1881,7 @@ sfvio_tell (void* user_data)
   fluid_sample_t *sample = (fluid_sample_t *)user_data;
 
   return (sf_count_t)sample->userdata;
-};
+}
 #endif
 
 int
--- a/fluidsynth/src/sfloader/fluid_defsfont.c
+++ b/fluidsynth/src/sfloader/fluid_defsfont.c
@@ -1851,7 +1851,7 @@ sfvio_seek(sf_count_t offset, int whence, void* user_data)
       sample->userdata = (void *)offset;
       break;
     case SEEK_CUR:
-      sample->userdata = (void *)((intptr_t)sample->userdata + offset);
+      sample->userdata = (void *)((sf_count_t)sample->userdata + offset);
       break;
     case SEEK_END:
       sample->userdata = (void *)(sfvio_get_filelen(user_data) + offset);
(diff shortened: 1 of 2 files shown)


fluidsynth/src/sfloader/fluid_defsfont.c
```diff
@@ -1817,6 +1829,61 @@ fluid_sample_in_rom(fluid_sample_t* sample)
 /*
  * fluid_sample_import_sfont
  */
+#if LIBSNDFILE_SUPPORT
+// virtual file access rountines to allow for handling
+// samples as virtual files in memory
+static sf_count_t sfvio_pos;
```
derselbst: I dont think introducing this global var here is correct. It may work because we only open one sample at a time, do all the necessary work and never come back to it again. However `_fluid_sample_t` has a member `void* userdata`, I wouldnt mind at all abusing this one for storing a `sf_count_t sfvio_pos` for that specific sample. Imagine something like
```
case SEEK_SET:
      sample->userdata = (void*) offset;
...
sfvio_tell (void* user_data)
{
   fluid_sample_t *sample = (fluid_sample_t *)user_data;
   return (sf_count_t)sample->userdata;
}
```
... not nice but more suited IMO.

fluidsynth/src/sfloader/fluid_defsfont.c
```diff
@@ -1817,6 +1829,61 @@ fluid_sample_in_rom(fluid_sample_t* sample)
...
+
+static sf_count_t
+sfvio_get_filelen(void* user_data)
+{
+  fluid_sample_t *sample = (fluid_sample_t *)user_data;
+
+  return (sf_count_t)(sample->end + 1 - sample->start);
+};
```
derselbst: Why have you placed these lonesome `;` here and below? Could you remove them? They may cause compiler warnings.

fabiangreffrath:
> Could you provide me / point me to some example sf3 files so I can do some testing?

Sure, there is one in the latest musescore sources. It can be considered the reference implementation, since these guys originally came up with the idea of the SF3 format:

https://github.com/musescore/MuseScore/tree/master/share/sound

>  I wouldnt mind at all abusing this one for storing a sf_count_t sfvio_pos for that specific sample

Ok, will do so, but it was *your* idea, in case anyone asks! ;)

> Why have you placed these lonesome ; here and below?

They are there by accident, will remove them!

fluidsynth/src/sfloader/fluid_defsfont.c
```diff
@@ -1817,6 +1829,61 @@ fluid_sample_in_rom(fluid_sample_t* sample)
...
+{
+  fluid_sample_t *sample = (fluid_sample_t *)user_data;
+
+  switch (whence)
+  {
+    case SEEK_SET:
+      sample->userdata = (void *)offset;
+      break;
+    case SEEK_CUR:
+      sample->userdata = (void *)((intptr_t)sample->userdata + offset);
```
derselbst: Why are you using `(intptr_t)` here and below?

fabiangreffrath: I am using it for intermediate casts between pointer and int. Shouldn't make a difference to `sf_count_t`, though.

derselbst: But it confuses as you're not mangling with pointers, only with `sf_count_t`. So I'd prefer `sf_count_t`.

derselbst:
Ok, thanks so far for your contribution! After getting #184 to work, I'll have some tests in the next days. If it all goes well we might already have this in 1.1.7, lets see.

derselbst:
@fabiangreffrath I merged it in and made some changes to mangling with incorrect sample loops. Hopefully this is correct now, at least I tested with FluidR3.sf3 and it sounds really nice. So we'll have this feature in 1.1.7. Anyway, feel free to build recent git and test it yourself.

Thanks again!

fabiangreffrath:
Thank you very much for merging this! I am really looking forward to testing the new feature, though this will have to wait until next week, as I am currently on vacation. ;)

fabiangreffrath:
Just checked out some Doom music from the Alien Vendetta PWAD with Musescore's SF3 sound font and it worked as expected! \o/

trebmuh:
Awesome. Thanks for that guys !

mmontag:
I've been very happy to find out about SF3 format today and see this recently merged PR here. 
Bravo @fabiangreffrath !

fabiangreffrath:
@mmontag Thanks! :grin:

fabiangreffrath:
Hm, unfortunately I didn't make it neither into the AUTHORS nor into the THANKS file.

derselbst:
If you want to be credited, you are welcome to open a PR and add yourself to either one of the files you prefer.

fabiangreffrath:
> If you want to be credited, you are welcome to open a PR and add yourself to either one of the files you prefer.

Sorry, I forgot about this but will open a PR shortly.

Issue #163, opened 2017-05-25
Calculation of lower boundary for attenuation should also consider channel pressure, key pressure and pitch wheel


derselbst:
While testing the patch for ticket #50, I noticed that modulators which affect initial attenuation and use the general controllers channel pressure, key pressure and pitch wheel prevent note on events from happening when the modulator is set so that the initial volume is 0.

The reason is that the function fluid_voice_get_lower_boundary_for_attenuation only considers continuous controllers but ignores the three controllers mentioned above. With this patch, key pressure, channel pressure and pitch wheel can be used in initial attenuation modulators and behave just like other continuous controllers, like the expression controller for example.

Reported by: mawe242

Original Ticket: [fluidsynth/tickets/165](https://sourceforge.net/p/fluidsynth/tickets/165)

mawe42:
I noticed that the patch I sent to the old sourceforge tracker contains a bug, it should therefore not be merged as is. Will fix it and create a proper pull request in the next few days.

derselbst:
@mawe42 Thx for joining. Could you take care of this issue whenever you're free?

mawe42:
Sure, will look at it today.

Issue #50, opened 2010-05-26
Key pressure (polyphonic aftertouch) is not implemented


derselbst:
Key pressure (or polyphonic aftertouch, midi message A0-AF), is not implemented in the synth. It seems like it can be used in the MIDI router though.  

Reported by: diwic

Original Ticket: [fluidsynth/tickets/50](https://sourceforge.net/p/fluidsynth/tickets/50)

derselbst:
fixed by merging #185

PR #185, opened 2017-08-03
Implement polyphonic key pressure (aftertouch) handling

mawe42:
This is the pull request for the patch in #50. It contains some references to a "1.1.7" version, which is probably not correct anymore. @derselbst: you mentioned a 1.2 maintenance release, so this new feature would probably have to wait until after that? Please let me know what you think, I'm happy to make changes and update the pull request.

The bugfix in #163 is not included in this pull request. I will send a separate (and corrected) one later on.

### DIFF
--- a/fluidsynth/include/fluidsynth/event.h
+++ b/fluidsynth/include/fluidsynth/event.h
@@ -56,6 +56,7 @@ enum fluid_seq_event_type {
   FLUID_SEQ_TIMER,		/**< Timer event (DOCME) */
   FLUID_SEQ_ANYCONTROLCHANGE,	/**< DOCME (used for remove_events only) */
   FLUID_SEQ_CHANNELPRESSURE,    /**< Channel aftertouch event @since 1.1.0 */
+  FLUID_SEQ_KEYPRESSURE,        /**< Polyphonic aftertouch event @since 1.1.7 */
   FLUID_SEQ_SYSTEMRESET,        /**< System reset event @since 1.1.0 */
   FLUID_SEQ_UNREGISTERING,      /**< Called when a sequencer client is being unregistered. @since 1.1.0 */
   FLUID_SEQ_LASTEVENT		/**< Defines the count of event enums */
@@ -103,6 +104,7 @@ FLUIDSYNTH_API void fluid_event_volume(fluid_event_t* evt, int channel, short va
 FLUIDSYNTH_API void fluid_event_reverb_send(fluid_event_t* evt, int channel, short val);
 FLUIDSYNTH_API void fluid_event_chorus_send(fluid_event_t* evt, int channel, short val);
 
+FLUIDSYNTH_API void fluid_event_key_pressure(fluid_event_t* evt, int channel, short key, short val);
 FLUIDSYNTH_API void fluid_event_channel_pressure(fluid_event_t* evt, int channel, short val);
 FLUIDSYNTH_API void fluid_event_system_reset(fluid_event_t* evt);
 
--- a/fluidsynth/src/synth/fluid_chan.h
+++ b/fluidsynth/src/synth/fluid_chan.h
@@ -41,7 +41,7 @@ struct _fluid_channel_t
   int sfont_bank_prog;                  /**< SoundFont ID (bit 21-31), bank (bit 7-20), program (bit 0-6) */
   fluid_preset_t* preset;               /**< Selected preset */
 
-  int key_pressure[128];                /**< MIDI polyphonic key pressure */
+  char key_pressure[128];               /**< MIDI polyphonic key pressure */
   int channel_pressure;                 /**< MIDI channel pressure */
   int pitch_bend;                       /**< Current pitch bend value */
   int pitch_wheel_sensitivity;          /**< Current pitch wheel sensitivity */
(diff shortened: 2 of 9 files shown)


derselbst:
Yes I wanted the 1.1.7 to be as stable as possible. But all these minor feature additions sum up to a huge bunch of things that can go wrong and need to be tested (either in production by the community or by the developers i.e. currently only me). So to avoid that I'm humping up even more things, I think we'll wait with that until next release, whether this will be a minor a major one... we'll see.

mawe42:
Ok, then I leave this as it is for now and wait for further comments. Thanks for all your work on FluidSynth, btw! It's fantastic to have an active maintainer again!

fluidsynth/include/fluidsynth/event.h
```diff
@@ -56,6 +56,7 @@ enum fluid_seq_event_type {
   FLUID_SEQ_TIMER,		/**< Timer event (DOCME) */
   FLUID_SEQ_ANYCONTROLCHANGE,	/**< DOCME (used for remove_events only) */
   FLUID_SEQ_CHANNELPRESSURE,    /**< Channel aftertouch event @since 1.1.0 */
+  FLUID_SEQ_KEYPRESSURE,        /**< Polyphonic aftertouch event @since 1.1.7 */
```
derselbst: We are in a public header. Placing the key pressure next to channel pressure is absolutely comprehensible. Unfortunately this breaks the ABI. Applications compiled against older versions of fluidsynth will have different values for e.g. `FLUID_SEQ_SYSTEMRESET`.

Infact adding it anywhere else will always change the value of `FLUID_SEQ_LASTEVENT`, which might be used downstream.

Long story short: you can leave everything as it is, but this change makes fluidsynth require a `SOVERSION` bump, which will only happen upon a major release.

fluidsynth/src/synth/fluid_chan.h
```diff
@@ -41,7 +41,7 @@ struct _fluid_channel_t
   int sfont_bank_prog;                  /**< SoundFont ID (bit 21-31), bank (bit 7-20), program (bit 0-6) */
   fluid_preset_t* preset;               /**< Selected preset */
 
-  int key_pressure;                     /**< MIDI key pressure */
+  int key_pressure[128];                /**< MIDI polyphonic key pressure */
```
derselbst: Do we really need `int` here? Makes a worst case memory usage of `8*128 bytes = 1KiB`. I think internally we can use `signed char`.

Note however that the types you've chosen for the API calls can stay as they are, i.e. `fluid_event_key_pressure` with `short` and `fluid_synth_key_pressure` with `int`.

fluidsynth/include/fluidsynth/event.h
mawe42: Ok, as it's a new feature, I guess waiting for a major release makes sense anyway.

fluidsynth/src/synth/fluid_chan.h
mawe42: You are right, the pressure value is always in the range 0-127, so char makes absolute sense.

mawe42:
Changed the key_pressure array type to char. Do you want me to do anything with the version number now, or wait until you're ready to merge this after the next maintenance release?

Oh, and how do you want changes to pull requests to happen. Simply as additional commit (like I've done just now), or should I rewrite the history?

derselbst:
Ok, thanks for your contribution!

> Do you want me to do anything with the version number now, or wait until you're ready to merge this after the next maintenance release?

Dont worry, I'll take care of that.
### LET ME JUST PLACE THIS VERY OBVIOUS REMINDER TO ACTUALLY DOING THIS AND NOT SIMPLY FORGET ABOUT :)

> Simply as additional commit (like I've done just now), or should I rewrite the history?

I'm a friend of KISS. No obfuscated squashes / rebases, but simple additional commits as you did ;)

ReinholdH:
I have been using the patch #50 on top of 1.1.6 with the interface
fluid_synth_channel_pressure
Because it did not make it into 1.1.7 - which is OK to start with a solid base -  I unfortunately need to wait using the latest and greatest package.

derselbst:
> with the interface fluid_synth_channel_pressure

You mean `fluid_synth_key_pressure`?

> I unfortunately need to wait using the latest and greatest package.

As explained before this feature introduces `FLUID_SEQ_KEYPRESSURE` which changes the value of `FLUID_SEQ_LASTEVENT` and therefore breaks ABI compatibility. `FLUID_SEQ_LASTEVENT` has been deprecated in 1.1.7, so it will be removed in the next major release. However other API/ABI breaking changes also wait to be adopted for that major release, which will probably take a couple of months to be completed.

For now all I can offer you is to merge this feature into master, so you can use a build from recent git then.

ReinholdH:
Correct: fluid_synth_key_pressure
Thanks you very much for the offer for a merge as a pre-solution.  But this is not necessary. We will go for our upcoming release with the patched 1.1.6 and will bring an final solution when it will  officially available.

weshi:
Is this feature actually working now (I'm using 2.3.2, but it seems it has no effect) ?

derselbst:
> Is this feature actually working now (I'm using 2.3.2 on Android, but it seems it has no effect) ?

Pls. refer to the wiki, I've just fixed it: https://github.com/FluidSynth/fluidsynth/wiki/FluidFeatures

PR #187, opened 2017-08-05
Fix channel fine tune RPN to use correct range of +/- 100 cent

mawe42:
FluidSynth currently uses +/- 50 cent for the channel fine tune range. But the MIDI standard defines channel fine tune as follows:

> RPN 01 - FINE TUNING:
Resolution: 100/8192 cents
Range: 100/8192* (-8192) to 100/8192* (+8191)
Control Value Displacement in cents from A440
MSB LSB
00 00 100/8192* (-8192)
40H 00 100/8192* (0)
7FH 7FH 100/8192* (+8191) 

This change brings FluidSynth in line with the specs.


derselbst:
Hm, looks uncomplicated, obvious and non side effect prone. What a welcome change, thanks!

PR #193, opened 2017-08-22
workaround timing issues

### LINKED ISSUE #192: Weird tempo issues when writing to buffer
I've been having weird tempo issues with fluidsynth when writing to a buffer with `fluid_synth_write_s16`.  I was wanting to use fluidsynth to write to a buffer and then use that buffer with another library.  However, for some midis, the timing seems to be off.  Here's code that can reproduce the issue:

```
#include <fluidsynth.h>

#include <iostream>
#include <fstream>
#include <vector>

const int SAMPLE_LENGTH = 44100;
const int CHANNEL_NUMBER = 2;
const int SECONDS = 30;

std::vector<int16_t> samples(SAMPLE_LENGTH*CHANNEL_NUMBER*SECONDS);

fluid_settings_t* settings;
fluid_synth_t* synth;
fluid_player_t* player;

int main(int argc, char* argv[])
{
    if (argc < 3) {
        std::cout << "Please input a soundfont and midi file.";
        return 1;
    }
    if (fluid_synth_sfload(synth, argv[1], 1) == FLUID_FAILED) {
        std::cerr << "Unable to open soundfount" << std::endl;
        return 1;
    }
    player = new_fluid_player(synth);
    if (fluid_player_add(player, argv[2]) == FLUID_FAILED) {
        std::cerr << "Unable to open midi" << std::endl;
        return 1;
    }
    if (fluid_player_play(player) == FLUID_FAILED) {
        std::cerr << "Unable to play midi" << std::endl;
        return 1;
    }
    if (fluid_synth_write_s16(synth, SAMPLE_LENGTH*SECONDS, &samples[0], 0, 2, &samples[0], 1, 2) == FLUID_FAILED) {
        std::cerr << "Unable to write buffer" << std::endl;
        return 1;
    }

    // Not what I actually want to do, but it will let you hear it straight from the buffer
    std::ofstream file("sound.raw", std::ios::binary);
    file.write((char*)&samples[0], samples.size() * 2);
    file.close();

    return 0;
}
```

I have attached a few extra files.  001-Battle01.mid works perfectly.  However, 015-Theme04.mid gets messed up.  I've converted the output of this program with that midi to an ogg and attached that as sound.ogg.

Here's a few things I've managed to figure out myself:
- When playing songs normally from the command line or using an audio driver, it outputs correctly.
- It definitely seems to be fluidsynth, since I'm not using any other audio drivers or anything.  I've also reproduced this problem on Debian Testing (at this time Buster) with g++ and on Windows 10 using mingw64's g++.
- I've tried a few different soundfonts and they all have the same problem.
- It seems to be when less instruments are playing.  Some songs are fine until one channel has a solo and then it suddenly speeds up.
- The way the tempo messes up is exactly the same every time.

This bug makes fluidsynth unusable for what I'm trying to do.  Any help would be greatly appreciated.  Thank you!

[prog.zip](https://github.com/FluidSynth/fluidsynth/files/1232477/prog.zip)

EDIT: I forgot to mention I'm using fluidsynth 1.1.6.

derselbst:
closes #192

### DIFF
--- a/fluidsynth/doc/fluidsynth-v11-devdoc.txt
+++ b/fluidsynth/doc/fluidsynth-v11-devdoc.txt
@@ -527,7 +527,9 @@ to call fluid_synth_write_s16, fluid_synth_write_float,
 fluid_synth_nwrite_float or fluid_synth_process in parallel with the
 rest of the calls, and it won't be blocked by time intensive calls to
 the synth. Turn it off if throughput is more important than latency, e g
-in rendering-to-file scenarios where underruns is not an issue.</td>
+in rendering-to-file scenarios where underruns is not an issue.
+<br /><br /><strong>Deprecated:</strong><br />As of 1.1.7 this option is deprecated. The current implementation relies on the fact that this option is set to TRUE to correctly render audio. Also in fact it enforces thread safety for rvoice_mixer instead of enabling parallel-render, which is mainly controlled by "synth.cpu-cores".
+    </td>
   </tr>
 
 
--- a/fluidsynth/doc/fluidsynth-v11-devdoc.txt
+++ b/fluidsynth/doc/fluidsynth-v11-devdoc.txt
@@ -528,7 +528,7 @@ fluid_synth_nwrite_float or fluid_synth_process in parallel with the
 rest of the calls, and it won't be blocked by time intensive calls to
 the synth. Turn it off if throughput is more important than latency, e g
 in rendering-to-file scenarios where underruns is not an issue.
-<br /><br /><strong>Deprecated:</strong><br />As of 1.1.7 this option is deprecated. The current implementation relies on the fact that this option is set to TRUE to correctly render audio. Also in fact it enforces thread safety for rvoice_mixer instead of enabling parallel-render, which is mainly controlled by "synth.cpu-cores".
+<br /><br /><strong>Deprecated:</strong><br />As of 1.1.7 this option is deprecated. This option enforces thread safety for rvoice_mixer, which causes rvoice_events to be queued internally. The current implementation relies on the fact that this option is set to TRUE to correctly render any amount of requested audio. Also calling fluid_synth_write_* in parallel is not considered to be a use-case. It would cause undefined audio output, as it would be unpredictable for the user which rvoice_events specifically would be dispatched to which fluid_synth_write_* call.
     </td>
   </tr>
 
--- a/fluidsynth/src/midi/fluid_midi.c
+++ b/fluidsynth/src/midi/fluid_midi.c
@@ -1624,7 +1624,7 @@ fluid_player_callback(void *data, unsigned int msec)
         player->cur_msec = msec;
         player->cur_ticks = (player->start_ticks
                 + (int) ((double) (player->cur_msec - player->start_msec)
-                        / player->deltatime));
+                        / player->deltatime + 0.5)); /* 0.5 to average overall error when casting */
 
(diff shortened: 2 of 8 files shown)


derselbst:
@diwic Even after the fix of 6c3061cba30b8a79913ee49f67671009358a5b97 I had timing issues for some MIDIs. It seems fluidsynth is not designed to process multiple blocks and then render what we've got. For instance: if we process 100 blocks through the sampletimers, and a noteon is triggered in block 99, then we call `fluid_rvoice_mixer_render()` that noteon would appear at the very beginning of this rendered block. Though it should appear somewhere at the end. Because notes are scheduled for rendering immediately after `fluid_synth_noteon()`, disregard how long we lag behind with the render() call. That's why rendering blocks should be as small as 64 frames so the difference not hearable. Correct so far?

I wanted to work that around by hardcoding `blockcount=1` in `fluid_synth_render_blocks()`, see fa3894c97f0bd2f13c7e56cbbfd4575e4fd8a2f1. However that causes glitchy audio after around 50s for both: with and without parallel render. I consider this to be a bug somewhere in `rvoice_mixer`, but not sure where to look. Can you give me any advice?

[broken_sound.zip](https://github.com/FluidSynth/fluidsynth/files/1246671/broken_sound.zip)

diwic:
@derselbst 

Hmm, so suppose we got 10 blocks to render and there's an event at block 6. What's supposed to happen is the following:

 * We call `fluid_sample_timer_process` 6 times.
 * After the 6th block, `fluid_rvoice_eventhandler_dispatch_count` returns > 0
 * We render 6 blocks *without* dispatching the events in `fluid_rvoice_mixer_render`, so while we've detected that event, we have not applied it yet.
 * We get another call to `fluid_synth_render_blocks` with 4 blocks remaining to render
 * We call `fluid_rvoice_eventhandler_dispatch_all` to apply the changes
 * We call `fluid_sample_timer_process` 4 times.
 * We render the remaining 4 blocks, now with the event applied.

I e, the event gets correctly timed at block 6.

diwic:
@derselbst 
As for the 50 s glitch bug, I don't have time to look into that, at least not right now. Sorry.

derselbst:
@diwic 
Form my understanding what you wrote only applies to threadsafe-api enabled. Without it, events would be dispatched immediately, not being queued, so that `fluid_rvoice_eventhandler_dispatch_count` never returns > 0 ?

> As for the 50 s glitch bug, I don't have time to look into that

No problem, I'll be digging a little more in the code.

diwic:
> Form my understanding what you wrote only applies to threadsafe-api enabled. Without it, events would be dispatched immediately, not being queued, so that fluid_rvoice_eventhandler_dispatch_count never returns > 0 ?

Hmm, so maybe that was a mis-thinking on my part when I wrote that code. It obviously looks like I'm using the fact that the messages gets in the queue but not applied all the way.

Not sure about the best way to fix it.

If you hard-code blockcount=1 then you can forget about parallel render; rendering a few hundred voices for just one block is likely to take less time than waking up another thread. IIRC; things might have changed as well, with never OS versions etc.

Not sure if we can enable queueing temporarily and then disable it after render_blocks completes? There needs to be at least some indication of how far we can render before new events need to be applied.

derselbst:
I would even go so far to always enable queuing. Should guarantee that no events happen too early. I would also be curious whether using openMP for naive parallel rvoice rendering and mixing makes any difference (would at least simplify `rvoice_mixer`)... maybe I'll play around a little with that in the next days.

PR #194, opened 2017-08-23
refactor modulator handling

derselbst:
Refactors the calculation and treatment of modulators:

* fixes incorrect convex and biploar linear calculations for source2
* fixes calculations if pitchbend used as (non-linear) source


tim-janik:
Thanks a lot @swesterfeld.
Please preserve your comments next time, so things are easier to understand then next time someone reads the code. And if you found a related bug link like in this case, that's useful information as well.
I've added this for you in commit f84a96b743b967d4e763887c8978a2e1cc470cef.

Issue #195, opened 2017-08-29
fluid_sequencer_set_time_scale causes timing problems


imhcyx:
I'm using FluidSynth's sequencer and recently got a problem. I need to change the time scale several times and then send note events. I have noticed <code>fluid_sequencer_set_time_scale</code> limits scale<=1000, but I need to work on 1280 ticks/s. As a workaround, I reduced both the scale and the delays between events to one half, that is scale=640. So the situation is first I set the time scale to 960, then change it to 640, and finally play several notes. However, the timing is not correct and some of the notes seem to be eaten! I've no idea how this could happen.

Here is the code which can reproduce the problem. Using <code>#define REMOVE_SCALE_LIMIT 0</code> will demostrate the problem. If the limit in <code>fluid_sequencer_set_time_scale</code> is removed and <code>#define REMOVE_SCALE_LIMIT 1</code> is used, there will be no such problem.
[play.zip](https://github.com/FluidSynth/fluidsynth/files/1259183/play.zip)
I've tested it on MSYS2 MinGW-w64 and Ubuntu 16.04 x86_64 using the 1.1.6 release and the latest commit on GitHub.

derselbst:
> I have noticed fluid_sequencer_set_time_scale limits scale<=1000

Yes, the smallest unit fluidsynth is able to render are 64 audio frames. Given a samplerate of 48kHz this makes about 1ms.

> but I need to work on 1280 ticks/s. As a workaround, I reduced both the scale and the delays between events to one half, that is scale=640.

The logic behind that seems to be correct. To avoid misunderstandings: Even if you would change the scale to 1280 (or to 640 and halve the delays), wont give you any higher resolution (although `scale` being *double* in `fluid_sequencer_set_time_scale` might suggest so). E.g if you want an event to happen at

    1/1280 = 0.5/640 = 0.78ms
 
it will actually happen at 0ms, because being casted to int (yeah, it should probably happen at 1ms to reduce the error...).

So as I understand this concept, using `fluid_sequencer_set_time_scale` is only a comfortable way that prevents you from having to recalculate event ticks, when sending them to the seq. Since you cant set scale to 1280, you have to fiddle with event timings anyway, so using `fluid_sequencer_set_time_scale` might be a questionable approach here.

> However, the timing is not correct and some of the notes seem to be eaten!

[...]

PR #196, opened 2017-08-31
fix seq time scale issue

imhcyx:
fix issue #195


derselbst:
Ok, seems to work as expected. Thanks!

PR #204, opened 2017-09-14
Add a setting for volume envelope processing

### LINKED ISSUE #70: Add a setting for volume envelope processing (compliant or EMU)
Starting with FluidSynth 1.0.9, the volume envelope was changed to be compatible with the EMU10K1. I propose that a setting is added, such as synth.volenv with the options "compliant" or "emu", with "emu" as the default. "compliant" would use the previous behavior which is compliant with the SoundFont specification. This ticket was created in response to a message from Sebastian Biallas:  

<http://lists.gnu.org/archive/html/fluid-dev/2010-01/msg00011.html>  

Reported by: elementgreen

Original Ticket: [fluidsynth/tickets/70](https://sourceforge.net/p/fluidsynth/tickets/70)

derselbst:
fix #70

### DIFF
--- a/doc/fluidsynth.1
+++ b/doc/fluidsynth.1
@@ -221,6 +221,9 @@ Must always to be true for usage by fluidsynth executable.
 .TP
 .B synth.verbose            BOOL  [def=False]
 Print received MIDI events to stdout.
+.TP
+.B synth.volenv             STR   [def='emu' vals:'compliant', 'emu']
+Specifies the kind of volume envelope processing. This esp. influences the way fluidsynth responses to noteon velocity. The default setting 'emu' causes the envelope to be highly dynamic (i.e. compatible with the EMU10K1). Alternatively this may be set to 'compliant' for a less dynamic envelope, as it was done before fluidsynth 1.0.9.
 
 .TP
 .B GENERAL AUDIO
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -224,6 +224,7 @@ void fluid_synth_settings(fluid_settings_t* settings)
   fluid_settings_add_option(settings, "synth.midi-bank-select", "xg");
   fluid_settings_add_option(settings, "synth.midi-bank-select", "mma");
   
+  fluid_settings_register_str(settings, "synth.volenv", "emu", 0, NULL, NULL);
 }
 
 /**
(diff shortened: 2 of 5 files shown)


derselbst:
@ensonic Could you please read through the doc of `synth.volenv`. Is it clear enough?

doc/fluidsynth.1
```diff
@@ -221,6 +221,9 @@ Must always to be true for usage by fluidsynth executable.
 .TP
 .B synth.verbose            BOOL  [def=False]
 Print received MIDI events to stdout.
+.TP
+.B synth.volenv             STR   [def='emu' vals:'compliant', 'emu']
```
ensonic: Would it make sense to have the previous behaviour as the default (compliant)?

src/synth/fluid_synth.c
```diff
@@ -554,8 +557,29 @@ new_fluid_synth(fluid_settings_t *settings)
...
+    {
+            fluid_conversion_set_atten_power(FLUID_ATTEN_POWER_DEFAULT_COMPLIANT);
+    }
+    else if (fluid_settings_str_equal (settings, "synth.volenv", "emu") == 1)
+    {
+            fluid_conversion_set_atten_power(FLUID_ATTEN_POWER_DEFAULT_EMU);
+    }
+    else
+    {
+        if (fluid_settings_copystr(settings, "synth.volenv", buf, 64) == 1)
```
ensonic: if (fluid_settings_copystr(settings, "synth.volenv", buf, sizeof(buf)) == 1)

src/utils/fluid_conv.c
```diff
@@ -21,6 +21,12 @@
 #include "fluid_conv.h"
 
 
+/* EMU 8k/10k don't follow spec in regards to volume attenuation.
```
ensonic: doesn't follow

src/utils/fluid_conv.c
```diff
@@ -21,6 +21,12 @@
...
+ * This factor is used in the equation pow (10.0, cb / FLUID_ATTEN_POWER_FACTOR).
+ * By the standard this should be -200.0. */
+/* 07/11/2008 modified by S. Christian Collins for increased velocity sensitivity.  Now it equals the response of EMU10K1 programming.*/
```
ensonic: break long line?

doc/fluidsynth.1
derselbst: Well... I'm not sure. [Element suggested "emu" to be the default](https://github.com/FluidSynth/fluidsynth/issues/70#issue-238323938) and I'm also not feeling comfortable changing this again to the previous behaviour. So I think we should leave it as it is.

src/synth/fluid_synth.c
derselbst: you're right

src/utils/fluid_conv.c
derselbst: I consider 8k/10k to be plural, sry :)

src/utils/fluid_conv.c
derselbst: clang-format (#206) should take care of that some when in the future.

PR #212, opened 2017-09-23
Consider channel pressure, key pressure and pitch wheel for lower attenuation boundary calculation

### LINKED ISSUE #163: Calculation of lower boundary for attenuation should also consider channel pressure, key pressure and pitch wheel

mawe42:
The current lower attenuation boundary calculation only considers CCs that target the initial attenuation generator. But channel pressure, key pressure and pitch wheel controllers can also be mapped to attenuation.

Fixes #163

### DIFF
--- a/src/synth/fluid_voice.c
+++ b/src/synth/fluid_voice.c
@@ -1568,7 +1568,14 @@ fluid_voice_get_lower_boundary_for_attenuation(fluid_voice_t* voice)
 
         /* Modulator has attenuation as target and can change over time? */
         if ((mod->dest == GEN_ATTENUATION)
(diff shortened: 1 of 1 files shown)


derselbst:
Looks good. Thanks.

PR #220, opened 2017-10-02
move fluidsettings doc to xml

### LINKED ISSUE #211: Document settings at one central place
Currently documentation of fluidsynths settings is spread over three places:
* [the devdocs](http://www.fluidsynth.org/api/index.html#CreatingSettings) (pretty up-to-date)
* [the man page](https://github.com/FluidSynth/fluidsynth/blob/master/doc/fluidsynth.1) (also quite up-to-date)
* [the wiki](https://github.com/FluidSynth/fluidsynth/wiki/FluidSettings) (out-of-date)

Merge them at one place together, ~favorably in the wiki.~

derselbst:
allows the settings to be documented at one central place. also makes separation of data and layout possible (XML and XSL).

ToDo:
- [x] remove settings documentation from API docs
- [x] remove settings doc from fluidsynths man page

Instead provide a link to that new single xml file documenting all the available fluid settings.

fixes #211


derselbst:
@mawe42 Could you checkout the [fluidsettings-doc branch](https://github.com/FluidSynth/fluidsynth/tree/fluidsettings-doc) locally, have a look at `doc/fluidsettings.xml` and tell me your opinion about:

1. the basic idea of documenting the settings in xml
1. the layout and design used to do so.

mawe42:
I like the idea of having one central place where we document settings and I understand the rationale behind it. But I must admit... I don't like hand-written XML :-)

If I understand this right, then you want to generate the HTML documentation and also the man page from the XML file, correct? Or do you also want to generate C code for the help texts in fluid_cmd.c?

derselbst:
> If I understand this right, then you want to generate the HTML documentation and also the man page from the XML file, correct? Or do you also want to generate C code for the help texts in fluid_cmd.c?

I want to provide this single xml file to the user. The API docs will link to it, the man page will link to it, as well as the webpage (Main Menu > Documentation > Fluidsynth Settings)

mawe42:
If the file is supposed to be read directly, then I don't quite understand why you went for XML. What is the benefit over something like plain text or markdown?

derselbst:
I saw the separation of data (XML) and layout (XSL) as advantage. Both combined generate HTML which has much more powerful abilities than just plain text or markdown files (esp. thinking of colors). All what can be achieved using markdown is something like [this](https://github.com/FluidSynth/fluidsynth/wiki/FluidSettings), which is not very readable IMO.

But I'm also not a very creative guy, so I'm open to suggestions.

mawe42:
What about using an external service to offload some of the work, like readthedocs.io? They support markdown and reStructuredText: https://docs.readthedocs.io/en/latest/getting_started.html
They integrate very well with GitHub, allow automatic builds, PDF generation, etc. And I really like how their documentation is presented.

derselbst:
But it also only uses markdown. And if we find a nice way of presenting the settings in markdown, we could also document it in the github wiki.

I dont think it's a good idea to use even more services/media (webpage, wiki and API-doc is more than enough IMO). And I wasnt intending to question or rework fluidsynth's whole documentation or move it somewhere else. Only trying to find a less redundant and potentially nicer way of documenting fluidsettings.

derselbst:
...simple things like marking a setting as deprecated and display it in a different color should aid the user and as a solution XML came to my mind.

mawe42:
Sorry, I misunderstood the scope of this change. If it's just about the settings, then moving to something like readthedocs is way too invasive, I agree.

Still, as you asked for my opinion :-), let me repeat my dislike of XML for anything that is meant to be read by humans. I would rather go with a simpler, maybe boring and less colourful but more readable markdown, reStructuredText or even plain text file. But ultimately, this is just bike-shedding... and an up-to-date XML file with all settings and good descriptions is much better than the current situation.

mawe42:
Ok, another idea... reStructuredText is imho much better than markdown, more expressive and standardized. And if we want to generate html anyway, then we could use sphinx or a simpler RST parser to generate it. That way we get nice colours in the output, and still have a readable text file that you can read if you only have access to the sources and no RST parser. I'll draft an example so show what I mean.

derselbst:
> let me repeat my dislike of XML for anything that is meant to be read by humans

My intention was that a user will (usually) open the fluidsettings.xml in a webbrowser, since we are shipping doxygen generated HTML API doc anyway. Except those very few console-only-hardliner-users.


> and still have a readable text file that you can read if you only have access to the sources and no RST parser.

That's another problem. I think most people dont have an RST parser, never heard of it, are too lazy or simply dont know to install one in order to generate that nicer doc. So they would probably keep looking at that raw rst file with their favorite text editor. Ofc we as upstream could do the rst parsing and roll out both files, but again we would have duplicate documentation. Whereas they could simply double click the xml file to view it in the webbrowser.

mawe42:
> My intention was that a user will (usually) open the fluidsettings.xml in a webbrowser, since we are shipping doxygen generated HTML API doc anyway. Except those very few console-only-hardliner-users.

Ah yes, and Firefox and probably every browser other than Chrome/Chromium will evaluate the XSL that way. I tried it with Chromium first and had to start a local webserver to get it to read the XSL. But like I said, that's Chromiums problem. But the comparison with the API doc is a little misleading, I think. The settings documentation is interesting to both users and developers. The API doc is only for developers, so they can be expected to install doxygen and generate docs themselves.

> So they would probably keep looking at that raw rst file with their favorite text editor. 

Yes, and that is where a format like reStructuredText, Markdown or even plain text show their strength: they are very easily readable in source format. XML is not.

But what about this: lets cut this short and just go with your XML/XSL idea. That means we'll have well structured settings information in a single place. And if we later want to have it in a different format, it's just an new XSL away.

mawe42:
> Will look into that chrome issue.

Don't waste your time on this, Chrome developers are convinced that allowing XML to load local XSL is a security risk. Has been wontfixed many times: https://bugs.chromium.org/p/chromium/issues/detail?id=70088

derselbst:
Yeah, only added a comment to point this out

PR #226, opened 2017-10-04
Activate LADSPA subsystem only if "synth.ladspa.active" is true 

mawe42:
When LADSPA support is enabled at compile time, the LADSPA machinery is *always* active. The "synth.ladspa.active" setting is not being used anywhere in the code.

This pull requests adds code to check synth.ladspa.active when a new synth is being created. It also adds checks in all ladspa command handlers to ensure that FluidSynth doesn't crash if any of those commands are used with an inactive LADSPA (i.e. without a valid synth->LADSPA_FxUnit).

### DIFF
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -40,9 +40,13 @@
 #define L(x);
 
 fluid_LADSPA_FxUnit_t* new_fluid_LADSPA_FxUnit(fluid_synth_t* synth){
+  if(synth == NULL)
+    return NULL;
+
   fluid_LADSPA_FxUnit_t* FxUnit=FLUID_NEW(fluid_LADSPA_FxUnit_t);
-  assert(FxUnit);
-  assert(synth);
+  if(FxUnit == NULL)
+    return NULL;
+
   /* The default state is 'bypassed'. The Fx unit has to be turned on explicitly by the user. */
   /* Those settings have to be done in order to allow fluid_LADSPA_clean. */
   FxUnit->Bypass=fluid_LADSPA_Bypassed;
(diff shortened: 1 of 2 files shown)


derselbst:
Oh indeed, thanks!

derselbst:
Just made a change so that the synth is properly cleaned up if LADSPA alloc fails. Do you agree?

mawe42:
Yes, that looks much nicer now!

I stumbled across this because I was thinking of moving code that I added to the FluidSynth reverb into a LADSPA plugin, removing the need to patch FluidSynth for my special reverb. But going through the LADSPA codebase... it looks quite a mess, to be honest.

I've partly started to bring the code into shape. First of all giving it a reformat (whitespace, indentation), but also renaming variables and identifiers to be more in line with the rest of the codebase. Would you take a pull request doing this sort of cleanup first and adding functional changes and bugfixes later?

derselbst:
> Would you take a pull request doing this sort of cleanup first and adding functional changes and bugfixes later?

I would prefer the other way: bug fixes first, functional additions + refactorings later.

Regarding indentation and other formatting stuff: dont invest too much time in that. clang-format #206 will take care of that. But before actually applying it there are some older patches / commits to be merged in, so it doesnt end in a complete mess. But I want to have reformatted code prior to the next major release.

mawe42:
Ok, I'll do the fixes first, then optimizations and wait for the reformat of the codebase. #227 contains the changes from this pull request and the additional fixes.

derselbst:
@mawe42 Because I dont see the reformat to happen soon, I merged #206. So if you have time right now, feel free to do your ladspa refactorings and apply `clang-format` to `fluid_ladspa.c` if you want.

mawe42:
@derselbst Will do, thanks!

PR #227, opened 2017-10-04
LADSPA fixes

mawe42:
This pull request contains some fixes (and one optimization) to the LADSPA subsystem that make it usable again. After these fixes, I could load a simple LADSPA plugin and hear the effects. There are quite a few additional optimizations that could (and should) be done to the code, but those are for another pull request :-)

Note that the "constant value node" char has changed from '#' to '$', as the command parser treats '#' as the start of a comment. So to load the simple delay.so plugin from the LADSPA sources and connect it to an output port, you could use a command like the following:

`ladspa_add /usr/lib/ladspa/delay.so delay_5s Input <- in1_L Output -> out1_L Output -> out1_R Delay <- $1.0 Dry <- $0.5`
`ladspa_start`

### DIFF
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -40,9 +40,13 @@
 #define L(x);
 
 fluid_LADSPA_FxUnit_t* new_fluid_LADSPA_FxUnit(fluid_synth_t* synth){
+  if(synth == NULL)
+    return NULL;
+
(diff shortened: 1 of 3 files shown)


derselbst:
wont have this in 1.1.8, too many conflicts

mawe42:
Ah, so master is for the next major release, 1.1.8 will be cut from the 1.1.x branch, correct?

derselbst:
Yes, with 1.1.x only receiving the most important bug fixes or fixes that I messed up in 1.1.7.

PR #233, opened 2017-10-15
Correct use of FLUID_INLINE in all sources.

carlo-bramini:
Some sources are still using "inline" keyword, while others are using "FLUID_INLINE" macro.
This patch puts all sources at the same level.


carlo-bramini:
Thank you very much for your support.
Please excuse me, it's just the second time that I used github...

PR #235, opened 2017-10-19
LADSPA effect unit refactor

mawe42:
This pull request contains a major refactor of the LADSPA subsystem and the associated shell commands. In it's current form it is more of an RFC than a proper pull request. I'd really like to get your feedback on the size and "mergeability" of this change.

I've basically completely rewritten src/bindings/fluid_ladspa.c. Also, the shell commands have been streamlined, removing the need for extremely long command lines. So something like this:

    ladspa_add ./aw.so alienwah_stereo Input1 <- in1_L Input2 <- in1_R Output1 -> out1_L Output2 -> out2_R Frequency <- $0.6 Initial <- $0 Feedback <- $0.5 Delay <- $20

becomes

    ladspa_plugin ./aw.so alienwah_stereo
    ladspa_port Input1 < in1_L
    ladspa_port Input2 < in1_R
    ladspa_port Output1 > out1_L
    ladspa_port Output2 > out1_R
    ladspa_port Frequency = 0.6
    ladspa_port Initial = 0
    ladspa_port Feedback = 0.5
    ladspa_port Delay = 2.0

I don't think we need to be careful of backwards compatibility here, as the LADSPA system was broken for a very long time before the fix that changed the "number prefix" from # to $.

There are more changes I would like to make, like supporting default parameters for plugin ports,
avoiding the constant buffer copying if FluidSynth is compiled with float instead of double, etc. But before I do that, I wanted to know show the current state first and ask for feedback.

### DIFF
--- a/doc/ladspa.md
+++ b/doc/ladspa.md
@@ -1,0 +1,64 @@
+# FluidSynth LADSPA Interface
+
+The [LADSPA](http://ladspa.org/) (Linux Audio Developer's Simple Plugin API)
+binding can be used to route the FluidSynth audio output through any number
+of LADSPA plugins. As the name implies, it is only available on Linux.
+
+## Configuration
+
+To configure and compile FluidSynth with LADSPA support, make sure you have
+the LADSPA SDK (basically the ladspa.h header file) installed. Then enable
+LADSPA when calling cmake:
+
+    cmake path/to/source/ -Denable-ladspa=1
+
+You should see `LADSPA support: yes` in the cmake output.
+
+To enable the LADSPA engine, use the `synth.ladspa.active` setting when
+starting FluidSynth:
+
+    fluidsynth -o synth.ladspa.active=1 ...
+
+
+# Signal Flow
+
+The LADSPA effects unit runs immediately after the internal reverb and chorus
+effects have been processed. When no plugins have been configured, the
+effects unit is dormant and uses no additional system resources.
+
+When at least one plugin is configured and the engine is activated, the
+rendered audio is passed into the LADSPA effects unit, each plugin is
+run in the order that they were created and the resulting audio is
+passed back into FluidSynth (and from there to the sound card or other
+output).
+
+
+# Loading and Connecting Plugins
+
+Currently the only way to configure the effects unit is via the FluidSynth
+shell or via a config file.
+
+## Example Setups
+
+All examples assume that your `LADSPA_PATH` environment variable points
+to the directory containing the plugin libraries (e.g. /usr/lib/ladspa).
+
+### Single Plugin
+
+The following loads the delay.so plugin library from the LADSPA SDK and
+instantiates the `delay_5s` plugin from that library. It connects the
+main left channel output from FluidSynth with the plugin input, the
+main left channel input to FluidSynth with the plugin output. It also
+sets the two control ports of the plugin to example values and starts
+the engine.
+
+    ladspa_plugin delay.so delay_5s
+    ladspa_port Input < in1_L
+    ladspa_port Output > out1_L
+    ladspa_port Delay = 1.0
+    ladspa_port Dry/Wet = 0.5
+
+    ladspa_start
+
+The audible effect should be an untouch right channel and a slightly
+lower volume on the left with a delay effect of 1 second on top.
--- a/doc/ladspa.md
+++ b/doc/ladspa.md
@@ -10,7 +10,7 @@ To configure and compile FluidSynth with LADSPA support, make sure you have
 the LADSPA SDK (basically the ladspa.h header file) installed. Then enable
 LADSPA when calling cmake:
 
-    cmake path/to/source/ -Denable-ladspa=1
+    cmake -Denable-ladspa=1 <path-to-source>
 
 You should see `LADSPA support: yes` in the cmake output.
 
@@ -60,5 +60,5 @@ the engine.
 
     ladspa_start
 
-The audible effect should be an untouch right channel and a slightly
+The audible effect should be an untouched right channel and a slightly
 lower volume on the left with a delay effect of 1 second on top.
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -97,6 +97,20 @@ fluid_ladspa_fx_t *new_fluid_ladspa_fx(fluid_real_t sample_rate, int audio_group
     fx->effects_channels = effects_channels;
     fx->audio_channels = audio_channels;
 
+    /* Setup mutex and cond used to signal that fluid_ladspa_run has finished */
+     fx->run_finished_mutex = new_fluid_cond_mutex();
+     if (fx->run_finished_mutex == NULL)
+     {
+         delete_fluid_ladspa_fx(fx);
+         return NULL;
+     }
+     fx->run_finished_cond = new_fluid_cond();
+     if (fx->run_finished_cond == NULL)
+     {
+         delete_fluid_ladspa_fx(fx);
+         return NULL;
+     }
+
     /* Finally, create the nodes that carry audio in and out of the LADSPA unit.
      * They will always be the first in the fx->nodes array and not removed on
      * fluid_ladspa_reset but only when this LADSPA fx instance is deleted. */
@@ -132,6 +146,16 @@ void delete_fluid_ladspa_fx(fluid_ladspa_fx_t *fx)
         delete_fluid_ladspa_node(fx->nodes[i]);
     };
 
+    if (fx->run_finished_cond != NULL)
+    {
+        delete_fluid_cond(fx->run_finished_cond);
+    }
+
+    if (fx->run_finished_mutex != NULL)
+    {
+        delete_fluid_cond_mutex(fx->run_finished_mutex);
+    }
+
     fluid_rec_mutex_destroy(fx->api_mutex);
 
     FLUID_FREE(fx);
@@ -141,8 +165,7 @@ void delete_fluid_ladspa_fx(fluid_ladspa_fx_t *fx)
  * Set the sample rate of the LADSPA effects.
  *
  * Resets the LADSPA effects if the sample rate is different from the
- * previous sample rate. If the LADSPA effects is active, it tries to 
- * deactivate and reset it once, without timeout.
+ * previous sample rate.
  *
  * @param fx LADSPA fx instance
  * @param sample_rate new sample rate
@@ -164,16 +187,11 @@ int fluid_ladspa_set_sample_rate(fluid_ladspa_fx_t *fx, fluid_real_t sample_rate
 
     if (fluid_ladspa_is_active(fx))
     {
-        /* Deactivate without timeout */
-        if (fluid_ladspa_deactivate(fx, 0) != FLUID_OK)
+        if (fluid_ladspa_reset(fx) != FLUID_OK)
         {
-            FLUID_LOG(FLUID_ERR, "Failed to deactivate LADSPA, unable to change sample rate");
+            FLUID_LOG(FLUID_ERR, "Failed to reset LADSPA, unable to change sample rate");
             LADSPA_API_RETURN(fx, FLUID_FAILED);
         }
-
-        /* A new sample rate means the plugin parameters need to be updateded anyway, so
-         * no point in keeping the old setup */
-        clear_ladspa(fx);
     }
 
     fx->sample_rate = new_sample_rate;
@@ -242,13 +260,11 @@ int fluid_ladspa_activate(fluid_ladspa_fx_t *fx)
  * @note This function may sleep.
  *
  * @param fx LADSPA fx instance
- * @param timeout_ms number of milliseconds to wait for deactivation. Set to 0 to return immediately on failiure.
  * @return FLUID_OK if deactivation succeeded, otherwise FLUID_FAILED
  */
-int fluid_ladspa_deactivate(fluid_ladspa_fx_t *fx, int timeout_ms)
+int fluid_ladspa_deactivate(fluid_ladspa_fx_t *fx)
 {
     int i;
-    int retries;
 
     LADSPA_API_ENTER(fx);
 
@@ -258,32 +274,16 @@ int fluid_ladspa_deactivate(fluid_ladspa_fx_t *fx, int timeout_ms)
         LADSPA_API_RETURN(fx, FLUID_OK);
     }
 
-    /* Try to set the state to inactive. If this fails, then fluid_ladspa_run is currently
-     * processing plugins, so wait a little and try again. Set pending_deactivation to inform
-     * fluid_ladspa_run to bail out early, otherwise deactivation might clash with it
-     * again on the next retry. If it's still not possible to deactivate after the configured timeout,
-     * then log an error and report failure to the caller. */
+    /* Notify fluid_ladspa_run that we would like to deactivate and that it should
+     * send us a signal when its done if it is currently running */
     fx->pending_deactivation = 1;
 
-    for (retries = 0; retries <= timeout_ms; retries++)
+    fluid_cond_mutex_lock(fx->run_finished_mutex);
+    while (!fluid_atomic_int_compare_and_exchange(&fx->state, FLUID_LADSPA_ACTIVE, FLUID_LADSPA_INACTIVE))
     {
-        if (fluid_atomic_int_compare_and_exchange(&fx->state, FLUID_LADSPA_ACTIVE, FLUID_LADSPA_INACTIVE))
-        {
-            break;
-        }
-        if (timeout_ms)
-        {
-            /* FIXME: this should probably be a FLUID_* style macro */
-            g_usleep(1000); /* wait one millisecond */
-        }
+        fluid_cond_wait(fx->run_finished_cond, fx->run_finished_mutex);
     }
-
-    if (fluid_atomic_int_get(&fx->state) != FLUID_LADSPA_INACTIVE)
-    {
-        fx->pending_deactivation = 0;
-        FLUID_LOG(FLUID_ERR, "Unable to deactivate LADSPA after %dms!", timeout_ms);
-        LADSPA_API_RETURN(fx, FLUID_FAILED);
-    }
+    fluid_cond_mutex_unlock(fx->run_finished_mutex);
 
     /* Now that we're inactive, deactivate all plugins and return success */
     for (i = 0; i < fx->num_plugins; i++)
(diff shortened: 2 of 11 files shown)


mawe42:
@derselbst Would it be possible to get another Travis build with -Denable-ladspa=1?

src/bindings/fluid_ladspa.h
```diff
@@ -21,220 +21,129 @@
...
+    fluid_ladspa_plugin_t *plugins[FLUID_LADSPA_MAX_PLUGINS];
+    int num_plugins;
+
+    /* used to generate the unique plugin ids */
+    int next_plugin_id;
+
+    fluid_rec_mutex_t api_mutex;
 
-#endif  /* _FLUID_LADSPA_H */
+    volatile int state;
```
derselbst: why volatile?

src/rvoice/fluid_rvoice_mixer.c
```diff
@@ -642,10 +642,20 @@ void delete_fluid_rvoice_mixer(fluid_rvoice_mixer_t* mixer)
...
+  mixer->ladspa_fx = ladspa_fx;
+}
+
+void fluid_rvoice_mixer_deactivate_ladspa(fluid_rvoice_mixer_t *mixer)
+{
+    if (mixer->ladspa_fx == NULL) return;
+    mixer->ladspa_fx->state = FLUID_LADSPA_INACTIVE;
+
+    /* Signal LADSPA engine that deactivation has been processed */
+    fluid_cond_mutex_lock(mixer->ladspa_fx->state_mutex);
```
derselbst: mutex lock in `rvoice_mixer`. Is that a good idea?

src/synth/fluid_synth.c
```diff
@@ -5269,3 +5270,24 @@ int fluid_synth_set_channel_type(fluid_synth_t* synth, int chan, int type)
...
 }
 
+#ifdef LADSPA
+/**
+ * Deactivate the LADSPA effects
+ *
+ * @param synth FluidSynth instance
+ * @return FLUID_OK on success, FLUID_FAILED otherwise
+ */
+int fluid_synth_deactivate_ladspa(fluid_synth_t *synth)
```
derselbst: So the only purpose is to deactivate ladspa if the user changes his mind? Shall it be part of public API (FLUIDSYNTH_API)?

src/bindings/fluid_ladspa.h
```diff
@@ -21,220 +21,129 @@
...
-int fluid_LADSPA_handle_start(fluid_synth_t* synth, int ac, char** av, fluid_ostream_t out);
+    const LADSPA_Descriptor *desc;
+    LADSPA_Handle *handle;
 
-/* Purpose:
- * Adds one plugin into the list of the LADSPA Fx unit.
- * This is only allowed, while the Fx block is in 'bypass' state (after clear).
- */
-int fluid_LADSPA_handle_add(fluid_synth_t* synth, int ac, char** av, fluid_ostream_t out);
+    volatile int active;
```
derselbst: why volatile?

src/bindings/fluid_ladspa.c
```diff
@@ -22,1107 +22,1288 @@
...
+    fx->state_mutex = new_fluid_cond_mutex();
+    if (fx->state_mutex == NULL)
+    {
+        FLUID_FREE(fx);
+        return NULL;
+    }
+    fx->state_cond = new_fluid_cond();
+    if (fx->state_cond == NULL)
+    {
+        FLUID_FREE(fx);
```
derselbst: missing `delete_fluid_cond_mutex(fx->state_mutex);`

src/bindings/fluid_ladspa.c
```diff
@@ -22,1107 +22,1288 @@
...
+
+    fx = FLUID_NEW(fluid_ladspa_fx_t);
+    if (fx == NULL)
+    {
+        return NULL;
+    }
+
+    fx->state = FLUID_LADSPA_INACTIVE;
+
+    fx->sample_rate = synth->sample_rate;
```
derselbst: I'm not a fan of shadowing variables. And the synth may actually change the samplerate. I guess this would cause trouble. Couldnt you store the synth instance and use the synth API getters instead?

src/bindings/fluid_ladspa.c
```diff
@@ -22,1107 +22,1288 @@
...
+static int create_input_output_nodes(fluid_ladspa_fx_t *fx)
+{
+    char name[99];
+    int i;
+
+    /* These nodes transport the main audio generated by FluidSynth into the
+     * LADSPA effects unit. Create left and right input nodes for each audio group. */
+    for (i = 0; i < fx->audio_groups; i++)
+    {
+        FLUID_SPRINTF(name, "in%i_L", (i + 1));
```
derselbst: use FLUID_S**N**PRINTF pls

`FLUID_SNPRINTF(name, sizeof(name), "in%i_L", (i + 1));`

src/bindings/fluid_ladspa.c
```diff
@@ -22,1107 +22,1288 @@
...
+ * Does a case insensitive 'fuzzy' prefix match.
+ *
+ * Checks if haystack starts with needle, ignoring case differences and
+ * treating spaces and underscores as equal.
+ *
+ * @param haystack string to search in
+ * @param needle string prefix to search for
+ * @return TRUE if match was found, otherwise FALSE
+ */
+static int fuzzy_prefix_match(const char *haystack, const char *needle)
```
derselbst: Couldnt you achieve easier that with `FLUID_STRCASECMP` ?

src/bindings/fluid_ladspa.c
```diff
@@ -22,1107 +22,1288 @@
...
-    printf("***Error032***\n"
-	     "Too many plugins.\nChange FLUID_LADSPA_MaxPlugins!");
-    fluid_LADSPA_clear(FxUnit);
-    return(PrintErrorMessage);
-  };
-
-  /* CommandLine (token sequence) is terminated with NULL.
-   * Add two more NULLs, so that a chunk of three tokens can be checked later without risk.*/
-
-  CommandLine=FLUID_ARRAY(char*, (ac+3));assert(CommandL
```
derselbst: I would prefer just handing it to `delete_fluid_ladspa_plugin()` than doing incremental cleanup.

derselbst: same here: just call `delete_fluid_ladspa_node()`

derselbst:
will look into travis

src/bindings/fluid_ladspa.c
mawe42: The LADSPA engine can't handle sample rate changes yet, so I thought it would be best to make this more or less obvious by copying that setting on instantiation of the LADSPA fx. But this is only temporary, I will add another commit that introduces proper handling of sample rate changes and will address this comment then. Would that be ok for you?

mawe42: Just to be clear... the old LADSPA engine couldn't handle sample rate changes either. Even worse... the sample rate was hardcoded to 44100.

src/bindings/fluid_ladspa.c
mawe42: Ack, fixed.

src/bindings/fluid_ladspa.c
mawe42: Ack, fixed

src/bindings/fluid_ladspa.c
mawe42: Yes, the case-insensitive prefix match would be much easier iwth strcasecmp. But there is still the spaces-to-underscores translation, so I would have to copy the string and iterate over it anyway. Might still be cleaner... the current implementation is simply a cleaned up version of the old code.

src/bindings/fluid_ladspa.c
mawe42: Sorry, I don't understand what you are referring to here. The marked code is all deleted.

mawe42: Same here... :-) What are you referring to?

src/bindings/fluid_ladspa.h
mawe42: To ensure this variable is guaranteed to be updated before the API mutex is released. Otherwise there could be a race condition, leading to a plugin being activated or deactivated more than once (which is forbidden in the LADSPA specs).

src/bindings/fluid_ladspa.h
mawe42: To ensure that the state variable is guaranteed to be written before the API mutex is released. Otherwise there could be a race condition, leading to two command handlers activating or deactivating the effects engine more than once.

src/bindings/fluid_ladspa.c
derselbst: Oh, github messed this up or I made some wrong click.

I referred to your "failed alloc" handling in `new_fluid_ladspa_plugin()`. I would prefer just giving that to the destructor function instead of doing custom cleanups.

derselbst: again failed alloc handling but for `new_fluid_ladspa_node()`

src/rvoice/fluid_rvoice_mixer.c
mawe42: I wondered as well, but the old code did exactly that in fluid_LADSPA_run. I don't think it's a problem, because the mutex is only used for the cond signalling. So it's going to be a wait-free lock in almost 100% of cases. And that code is not called very often, currently only if a user enters ladsap_reset or ladspa_stop into the shell.

src/synth/fluid_synth.c
mawe42: No need for a public API here, this should really be called via fluid_ladspa_deactivate. I think some of the fluid_ladspa_* functions could be made public, thought. But I suggest leaving that for later.

doc/ladspa.md
```diff
@@ -0,0 +1,64 @@
...
+
+    ladspa_plugin delay.so delay_5s
+    ladspa_port Input < in1_L
+    ladspa_port Output > out1_L
+    ladspa_port Delay = 1.0
+    ladspa_port Dry/Wet = 0.5
+
+    ladspa_start
+
+The audible effect should be an untouch right channel and a slightly
```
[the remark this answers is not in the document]
mawe42: Thanks, fixed.

doc/ladspa.md
```diff
@@ -0,0 +1,64 @@
...
+binding can be used to route the FluidSynth audio output through any number
+of LADSPA plugins. As the name implies, it is only available on Linux.
+
+## Configuration
+
+To configure and compile FluidSynth with LADSPA support, make sure you have
+the LADSPA SDK (basically the ladspa.h header file) installed. Then enable
+LADSPA when calling cmake:
+
+    cmake path/to/source/ -Denable-ladspa=1
```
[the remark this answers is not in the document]
mawe42: Thanks, fixed.

src/bindings/fluid_ladspa.c
derselbst: I'm a bit two minded about this. On the one hand I would accept it, on the other "Nothing lasts longer than a temporary..."

src/rvoice/fluid_rvoice_mixer.c
derselbst: Agreed, ok then.

src/bindings/fluid_ladspa.h
derselbst: You're always accessing it via a structure. I think it should be updated properly. However if you want guarantees you should use atomic setters to be sure.

src/bindings/fluid_ladspa.c
mawe42: You are so right :-) Let me rephrase: I will address this before signing off on this pull request.

src/bindings/fluid_ladspa.c
mawe42: So initialize all pointers that are going to be malloc'ed to NULL just after creation, change the destructor to check for NULL for those pointers and call destructor on each malloc fail?

derselbst: Yes. `memset(this, 0, sizeof(this));` at the beginning of `new_*()`. And then you can safely call `delete_*()`, see as it has been done in the synth: https://github.com/FluidSynth/fluidsynth/blob/bf3b5a9419080054a9f0fb08f90b4cd7de2f911a/src/synth/fluid_synth.c#L736

mawe42: Ok, fixed

src/bindings/fluid_ladspa.c
mawe42: ce79b5f adds sample rate change handling. I still shadow the sample rate in the ladspa structure, because the plugins want it in unsigned long and not float. And also because I think it's better to work with a stable value for the sample rate and only change it in fluid_ladspa_set_sample_rate. What do you think?

src/bindings/fluid_ladspa.h
mawe42: Sorry, wrong idea and wrong explanation. What I really want the volatile to protect the code against is weird compiler optimizations. The optimizer might try to store it in a local register and never re-read it again if it's not declared volatile. That might happen on the while loop in fluid_ladspa_deactivate. And as far as I know, accessing a struct member as opposed to a "normal" variable makes no difference to the compiler... but I might be mistaken?

On the plugin->active I think you are right, it is unnecessary and I will remove it.

src/bindings/fluid_ladspa.c
```diff
@@ -160,6 +162,43 @@ void delete_fluid_ladspa_fx(fluid_ladspa_fx_t *fx)
...
+ *
+ * Resets the LADSPA effects if the sample rate is different from the
+ * previous sample rate.
+ *
+ * @param fx LADSPA fx instance
+ * @param sample_rate new sample rate
+ */
+void fluid_ladspa_set_sample_rate(fluid_ladspa_fx_t *fx, fluid_synth_t *synth)
+{
+    unsigned long new_sample_rate = (unsigned long)synth->sample_rate;
```
derselbst: `synth->sample_rate` is mutex protected, potentially causing a dirty read here. Prefer handing `new_sample_rate` directly into `fluid_ladspa_set_sample_rate()` as extra param.

Also consider doing this:

`(unsigned long)(synth->sample_rate + 0.5) /* to minimize overall casting error */`

src/synth/fluid_synth.c
```diff
@@ -2383,6 +2383,16 @@ fluid_synth_set_sample_rate(fluid_synth_t* synth, float sample_rate)
   fluid_synth_update_mixer(synth, fluid_rvoice_mixer_set_samplerate, 
 			   0, sample_rate);
   fluid_synth_api_exit(synth);
```
derselbst: I'm afraid we are giving up our lock too early here. Consider this thread gets suspended right after  `fluid_synth_api_exit()` and the next call would be `delete_fluid_synth()` deleting `synth->ladspa_fx`.

On the other hand I'm not sure if we might experience a deadlock waiting to enter LADSPA_API.

derselbst: And ofc if you're postponing `fluid_synth_api_exit()` so `ladspa_set_sample_rate` is protected you can ignore my first comment.

mawe42: Yes, it deadlocks if the fluid API is not released before calling fluid_ladspa_set_sample_rate. Hm...

src/bindings/fluid_ladspa.h
derselbst: Ok, then just add a small comment to `state`, pointing out that register optimization thingy.

src/synth/fluid_synth.c
mawe42: Ah... we could handle this via the mixer event queue as well. Then there would be no deadlock. I'll draft the idea in a commit.

src/bindings/fluid_ladspa.h
derselbst: I still think `state` is not optimized to a register, because it's properly accessed via struct and the compiler cannot be sure that this memory isnt aliased, because it doesnt live in local scope.

But since it's performance irrelevant it probably doesnt hurt to be volatile.

src/bindings/fluid_ladspa.c
derselbst: We can also ignore that for now, so you can concentrate on the samplerate change.

mawe42:
Ok, I've reworked a lot of the LADSPA engine. It is now completely decoupled from FluidSynth. Everything it needs to function (sample rate, number  of input and output channels) is passed in on creation. Beside the obvious benefit of easier maintenance of FluidSynth and the LADSPA engine due to the loose coupling, it also has an additional benefit in that users can create their own (even additional) effect engines for further or other processing. After we expose a public API for LADSPA, of course.

Synchronization is done only with atomic operations now. And the sample rate change trigger is also greatly simplified now, only a very small hook in FluidSynths fluid_rvoice_mixer_set_samplerate.

I'm quite happy with it now. The only snag is the g_usleep call, which isn't available as a FLUID_* macro yet. But that can easily be added. Also, while explicit sleep calls like this might seem dirty, it is only used as a fallback. During my tests of starting and stopping LADSPA while FluidSynth was rendering sound, I never managed to actually trigger this sleep.

Looking forward to your feedback! Thanks for taking the time to get this change polished!

derselbst:
That timeout strikes me. Apart from that it's absolutely fine.

derselbst:
Anyway, I think I'll just merge it and look at it later again.

src/bindings/fluid_ladspa.c
mawe42: I've removed the weird fuzzy_prefix_match function. It's not needed anymore anyway, as the command line parser knows how to deal with quoted names like "Dry / Wet Mix". They are passed in as one string. Needed to add FLUID_STRNCASECMP to fluid_priv.h though.

derselbst:
cecf115a38f35e1c25be5b0f2eb9ef67827616de: Oh ok, nice. Waiting for you being ready.

mawe42:
> That timeout strikes me. Apart from that it's absolutely fine.

Ok, then dont' merge yet, please. It could probably also be solved with a cond and mutex...

derselbst:
> cond and mutex...

Yes, that would be my favourite.

mawe42:
Done, and much nicer than the timeout. Thanks for insisting :-)

derselbst:
Yeah, that looks good. Thanks for fixing, really nice work! Will merge it once travis done.

mawe42:
Added one last cleanup commit, removing unused includes. I will leave this branch alone now.

PR #238, opened 2017-10-21
Fluidsynth on mcu 1

carlo-bramini:
This is the first set of fixes that I made for running Fluidsynth on my embedded platform.

bindings/fluidmax/fluidmax.c
src/synth/fluid_gen.c

- The array fluid_gen_info[] is declared as const, by looking the location where it is used, it seems good to be a read-only object.

src/bindings/fluid_cmd.c

- The array of commands could be a const object. Strings inside this array could be constant too. In my opinion, it would be a good idea to declare a fluid_cmd_int_t typedef, which is similar to the fluid_cmd_t but without the 'data' member, to avoid useless waste of space and it also simplify the next changes.
- I remove the sentinel with all NULL pointers. The number of elements in the list is known, thanks to the FLUID_N_ELEMENTS() macro.
- To avoid the warning of signed/unsigned comparison mismatch, the integer variables used with the FLUID_N_ELEMENTS() macro should be an unsigned type.
- In line #1628, in the code for handling the help, there is a copy of the current fluid_commands[i] item in RAM. It seems to me that there is not an explanation for doing this, at least with the current code, so I replaced that into the if() statement and the strcmp() function with direct access to the structure.
- In functions new_fluid_cmd_handler(), instead of setting the 'data' member and then resetting it back to NULL, it just makes a fluid_cmd_t structure, filled with the values inside fluid_commands[i]. Required because fluid_commands[i] is now constant and declared with reduced, space-saving structure and now also thread-safe.

src/bindings/fluid_filerenderer.c

- Actually, objects like format_names[] are made of 2 parts, the array of pointers and the strings to be addressed. For having a fully constant object, both the array and the strings should be declared constant, by writing "const char * const" in the declaration. They have been also declared as static since they are used only locally.
- I removed the sentinel used of detecting the end of the lists type_names[], format_names[] and endian_names[]. Their length is known by using FLUID_N_ELEMENTS() macro.

src/drivers/fluid_adriver.c

- 'name' field has been declared as const into struct fluid_audriver_definition_t.
- fluid_audio_drivers[] is declared as static const object.
- The sentinel used of detecting the end of fluid_audio_drivers[] has been removed. Its length is known by using FLUID_N_ELEMENTS() macro.
- By using FLUID_N_ELEMENTS() macro, the integer variable for parsing fluid_audio_drivers[] has been declared as unsigned type, to avoid warnings of signed/unsigned comparison mismatches.

src/drivers/fluid_adriver.h

- "fluid_sys.h" is included instead of "fluidsynth_priv.h", required for using FLUID_N_ELEMENTS() macro.
- 'name' field has been declared as const into struct fluid_audio_driver_t.

src/drivers/fluid_mdriver.c

- 'name' field has been declared as const into struct fluid_mdriver_definition_t.
- fluid_midi_drivers[] is declared as static const object.
- The sentinel used of detecting the end of fluid_midi_drivers[] has been removed. Its length is known by using FLUID_N_ELEMENTS() macro.
- By using FLUID_N_ELEMENTS() macro, the integer variable for parsing fluid_midi_drivers[] has been declared as unsigned type, to avoid warnings of signed/unsigned comparison mismatches.

src/drivers/fluid_mdriver.h

- "fluid_sys.h" is included instead of "fluidsynth_priv.h", required for using FLUID_N_ELEMENTS() macro.
- 'name' field has been declared as const into struct fluid_midi_driver_t.

src/synth/fluid_synth.c
src/synth/fluid_synth.h

- The revmodel_preset[] has been declared as const and the sentinel at the end has been removed.
- The fluid_synth_set_reverb_preset() function can be implemented much simpler, by using FLUID_N_ELEMENTS() macro to test the effective number of elements. For this reasons, the 'num' parameterhas been changed to unsigned type.
- 'name' field has been declared as pointer to const strings.

src/utils/fluid_sys.c

- fluid_libname[] can be a constant string, rather than a RAM pointer to RAM string.

src/rvoice/fluid_rev.h

- The 'name' field of structure fluid_revmodel_presets_t has been declared as const.

src/sfloader/fluid_defsfont.c

- The arrays idlist[], badgen[] and badpgen[] can be static const objects.

src/sfloader/fluid_defsfont.h

- Since the arrays idlist[], badgen[] and badpgen[] has been redeclared as static, I removed their declaration as 'extern'.

src/utils/fluid_conv.c
- To simplify calculations, log(10.0) and log(2.0) can be replaced by M_LN10 and M_LN2 macros.

src/utils/fluidsynth_priv.h
- If M_LN10 and M_LN2 macros do not exists, declare them, like it had been already done for M_PI.

### DIFF
--- a/bindings/fluidmax/fluidmax.c
+++ b/bindings/fluidmax/fluidmax.c
@@ -1065,7 +1065,7 @@ fluidmax_version(t_object *o)
   post("  Max/MSP integration by Norbert Schnell IMTR IRCAM - Centre Pompidou");
 }
 
-extern fluid_gen_info_t fluid_gen_info[];
+extern const fluid_gen_info_t fluid_gen_info[];
 
 static void 
 fluidmax_print(t_object *o, Symbol *s, short ac, Atom *at)
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -59,6 +59,17 @@ struct _fluid_shell_t {
   fluid_ostream_t out;
 };
 
+/**
+ * Reduced command information structure for constant data.
+ * For internal use only.
+ */
+typedef struct {
+  const char *name;             /**< The name of the command, as typed in the shell */
+  const char *topic;            /**< The help topic group of this command */ 
+  fluid_cmd_func_t handler;     /**< Pointer to the handler for this command */
+  const char *help;             /**< A help string */
+} fluid_cmd_int_t;
+
 static int fluid_shell_run(fluid_shell_t* shell);
 static void fluid_shell_init(fluid_shell_t* shell,
                              fluid_settings_t* settings, fluid_cmd_handler_t* handler,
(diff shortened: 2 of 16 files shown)


src/bindings/fluid_cmd.c
```diff
@@ -59,6 +59,17 @@ struct _fluid_shell_t {
...
+/**
+ * Reduced command information structure for constant data.
+ * For internal use only.
+ */
+typedef struct {
+  const char *name;             /**< The name of the command, as typed in the shell */
+  const char *topic;            /**< The help topic group of this command */ 
+  fluid_cmd_func_t handler;     /**< Pointer to the handler for this command */
+  const char *help;             /**< A help string */
+} fluid_cmd_int_t;
```
derselbst: You want to optimize in size, that's ok. But I dont like that code duplication. I would rather prefer to somehow have the public struct re-using the internal struct. However anonymous structs are a non-standard extension. So I currently cant offer a really better alternative, expect than some fancy preprocessor-macro-hack.

I'm even thinking about removing `fluid_cmd_handler_register()` from the public API. Cant think of any use-case nor of any downstream-application using it.

src/synth/fluid_synth.c
```diff
@@ -3897,19 +3895,15 @@ fluid_synth_set_reverb_on(fluid_synth_t* synth, int on)
...
-    if (i == num) {
-      fluid_synth_set_reverb (synth, revmodel_preset[i].roomsize,
-                              revmodel_preset[i].damp, revmodel_preset[i].width,
-                              revmodel_preset[i].level);
-      return FLUID_OK;
-    }
-    i++;
-  }
-  return FLUID_FAILED;
+fluid_synth_set_reverb_preset(fluid_synth_t* synth, unsigned int num)
```
derselbst: I would leave `num` signed and check for `<0`. Else you would have to adjust it's usage in `fluid_handle_reverbpreset()`.

src/synth/fluid_synth.c
```diff
@@ -3897,19 +3895,15 @@ fluid_synth_set_reverb_on(fluid_synth_t* synth, int on)
...
+{
+  if (num >= FLUID_N_ELEMENTS(revmodel_preset))
+    return FLUID_FAILED;
```
derselbst: Please use `fluid_return_val_if_fail()` for consistency.

src/bindings/fluid_cmd.c
carlo-bramini: In my opinion, it looks a code duplication, but just apparently.
Besides the deletion of the 'data' field, at the end the real purpose is the declaration of a container for const objects and this have to be necessarily done in this way, while the public fluid_cmd_t has not this strict requirement, at least this is what I have understood after reading the sources, so the global structure cannot be used for this purpose. Actually, I do not know many other ways to do that.

src/synth/fluid_synth.c
carlo-bramini: Done.

src/synth/fluid_synth.c
carlo-bramini: Done, but if I could say my opinion, this was not a problem because the test is safe even if you pass a negative number at upper level (I had seen the call to atoi() before doing this change, so I did it intentionally).

bindings/fluidmax/fluidmax.c
```diff
@@ -1065,7 +1065,7 @@ fluidmax_version(t_object *o)
   post("  Max/MSP integration by Norbert Schnell IMTR IRCAM - Centre Pompidou");
 }
 
-extern fluid_gen_info_t fluid_gen_info[];
+extern const fluid_gen_info_t fluid_gen_info[];
```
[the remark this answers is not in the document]
carlo-bramini: I see.
If I could say my opinion, perhaps it would be worth to this change to 'static' when you will decide to officially discard this fluidmax thing (I do not know what it is exactly, actually) by doing both actions together... what do you think?

src/synth/fluid_synth.c
derselbst: Dont see it. Forgot to push?

src/synth/fluid_synth.c
derselbst: Given 32 bit integers: If the user enters -4294967295 it would select preset 2, which I dont quite consider to be correct ;)

bindings/fluidmax/fluidmax.c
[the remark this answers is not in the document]
derselbst: I dont think this would mean to discard it. It would just become a bit more "unmaintained". But this decision seems to have been made years ago.

src/bindings/fluid_cmd.c
derselbst: Ok, just leave it for now. Will perhaps address that later.

bindings/fluidmax/fluidmax.c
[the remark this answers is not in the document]
derselbst: Ok, so let me staticallyzise it.

src/synth/fluid_synth.c
carlo-bramini: 32 bit signed integers range from -2147483648 to 2147483647.
2147483647 is 0x7FFFFFFF
-2147483648 is 0x80000000, that is 2147483648 if tested as unsigned.
So, for the properties of the boolean arithmetic, the smallest negative signed integer is always greater than biggest positive signed value. That's why it is safe to do this assumption and why the compilers too sometimes converts things like "if (a<0 || a>max)" to "if ((unsigned )a>max)".

i=atoi("-4294967295") always returns 1, because that value overflows the limits of the type, so it does not matter too much to have the test positive or not... :)

src/synth/fluid_synth.c
carlo-bramini: Excuse, I was waiting before doing the final push, for reading your comments...

src/synth/fluid_synth.c
derselbst: Oh yeah, completely confused, sry. Should go to bed.

src/synth/fluid_synth.c
```diff
@@ -3899,17 +3897,15 @@ fluid_synth_set_reverb_on(fluid_synth_t* synth, int on)
...
-      fluid_synth_set_reverb (synth, revmodel_preset[i].roomsize,
-                              revmodel_preset[i].damp, revmodel_preset[i].width,
-                              revmodel_preset[i].level);
-      return FLUID_OK;
-    }
-    i++;
-  }
-  return FLUID_FAILED;
+  fluid_return_val_if_fail (
+    num < 0 || num >= (int)FLUID_N_ELEMENTS(revmodel_preset),
```
derselbst: Should be `num>=0 && num < FLUID_N_ELEMENTS()` because if this fails (evaluates to false) it returns. But you could also change num back to unsigned, sorry for the confusion.

carlo-bramini: Ouch, you are right, I misunderstood the meaning of the macro, sorry.

PR #243, opened 2017-10-22
Fixed nasty loop test

rmattes:
There's a nasty bug introduced in fluid_cmd.c that lead to a segfault when the shell command ```help all``` is used.  ```FLUID_N_ELEMENTS(fluid_commands)``` is a int >= 0 so will always be true, hence the loop over all commands never terminates.


derselbst:
Thanks!

PR #245, opened 2017-10-22
Misc fixes 1

carlo-bramini:


include/fluidsynth.h
```diff
@@ -48,7 +48,7 @@ extern "C" {
 
 #if defined(__GNUC__) || defined(__clang__)
 #    define FLUID_DEPRECATED __attribute__((deprecated))
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && _MSC_VER > 1400
```
derselbst: Perhaps `>= 1500` ?

src/synth/fluid_voice.c
```diff
@@ -488,22 +488,7 @@ fluid_voice_calculate_gen_pitch(fluid_voice_t* voice)
...
- * conversion will be done in the DSP function. This is the case, for
- * example, for the pitch since it is modulated by the controllers in
- * cents. */
-static int
-fluid_voice_calculate_runtime_synthesis_parameters(fluid_voice_t* voice)
-{
-  int i;
-
-  int list_of_generators_to_initialize[35] = {
+static int const list_of_generators_to_initialize[] = {
```
derselbst: Really move this to global scope? It only used by `fluid_voice_calculate_runtime_synthesis_parameters()` so IMO it should only be visible in that function.

src/synth/fluid_voice.c
```diff
@@ -1113,7 +1113,7 @@ fluid_voice_update_param(fluid_voice_t* voice, int gen)
  */
 int fluid_voice_modulate(fluid_voice_t* voice, int cc, int ctrl)
 {
-  int i, k;
+  unsigned int i, k;
```
kmatheussen: IMHO, there's no reason to use unsigned integers here, except perhaps readability. But unsigned integers often causes unnecessary bugs [1], and at least when I see one, I check that the usage of it is not dangerous. By not using unsigned integers you both avoid these bugs and the need to check if the use of unsigned can cause bugs. IMO, unsigned integers should only be used when doing bit-operations, for instance to store flags, plus in rare situations where you need that extra bit and using a larger integer type is not a good option.

[1] Search the internet. For instance, Qt does not use unsigned integers, and the google style guide recommends not using them (at least that's what I've read, but I would be surprised if it didn't.).

src/synth/fluid_voice.h
```diff
@@ -68,7 +68,7 @@ struct _fluid_voice_t
 	fluid_channel_t* channel;
 	fluid_gen_t gen[GEN_LAST];
 	fluid_mod_t mod[FLUID_NUM_MOD];
-	int mod_count;
+	unsigned int mod_count;
```
kmatheussen: This one is dangerous. If mod_count somewhere in the code could be negative (for instance as a temporary state), it would become a very large number.

src/synth/fluid_voice.c
kmatheussen: Here's what the google style guide says about integers: https://google.github.io/styleguide/cppguide.html#Integer_Types

src/synth/fluid_voice.h
derselbst: At least for the current implementation it should never become negative. Given the fact that it will never become greater `FLUID_NUM_MOD` (==64), we could change it's type to char, as we're having one padding byte over in `fluid_voice_t` anyway.

kmatheussen: Maybe. My point was that this use of unsigned is more dangerous since the type is not immediately visible from where "mod_count" is actually used, since the declaration is in a different file. In the future, this could theoretically cause a bug.

I've been bitten by unnecessary bugs caused by use of unsigned so many times that it's become a habit of mine to point this out in other peoples code, and never use unsigned in my own code.

mawe42: FWIW: I always disliked using anything but int for "normal" scale numbers, but that was just a gut feeling and sometimes I used unsigned anyway. The case made in the google style guide against unsigned ints sounds very convincing to me. Thanks for the pointer!

src/bindings/fluid_cmd.c
```diff
@@ -397,7 +398,11 @@ fluid_source(fluid_cmd_handler_t* handler, const char *filename)
     return file;
   }
   fluid_shell_init(&shell, NULL, handler, file, fluid_get_stdout());
-  return fluid_shell_run(&shell);
+  result = fluid_shell_run(&shell);
+
+  close(file);
```
derselbst: Shouldnt that be `_close()` for WIN32?

src/synth/fluid_voice.h
derselbst: Ok then, lets define it as `signed char mod_count;` and use an `int` when iterating over it. That way the `mod_count` would be promoted to `int` in any loop.

include/fluidsynth.h
carlo-bramini: Unfortunately, I cannot verify it.
MSVC surely does not support it.

src/bindings/fluid_cmd.c
carlo-bramini: Actually, it compiles anyways with oldnames activated, but you are right, probably it's better to use directly the name of the dllimport. I will change it.

include/fluidsynth.h
carlo-bramini: After some research, I think that MSVC6 was really the last version to not support deprecated declspec, so I modified the test on _MSC_VER to be more strict.

src/synth/fluid_voice.h
carlo-bramini: Well, the reason because I proposed to change "mod_count" to unsigned type was not because I really wanted to optimize it, but because I got a complain of sign mismatch with FLUID_N_ELEMENTS(), so I got 2 choices: writing "(int)FLUID_N_ELEMENTS(blablabla)" or adjusting mod_count. I tried to suggest the second choice because it looked more correct to me, after I verified that the fix was safe in all parts of the current source code.

Unsigned types are dangerous if you do not use them correctly. That's true. Is this particular change dangerous? In my opinion, it should not, if mod_count becomes negative (or 0xffffffff if read as unsigned), Fluidsynth will hang in both cases. Actually, Google coding style guide just shows you some dangers could happen *if* the coder does not take the right care, that's true. Even the pointers can dangerous if you do pay attention... 

However, instead of changing to "signed char", I really prefer to revert the change and restore "int".
Byte access is another thing that should be used with some care because it can have several speed penalties on modern CPU.
Since this is not an array and there is not the reason for saving space, I would recommend to use the previous code if this change is not acceptable. If it's ok for you too, I can revert everything and I will cast FLUID_N_ELEMENTS() to (int) type.

derselbst: Byte access might not be the best, however it will only happen once. I dont see a speed penalty here. I would rather prefer to decrease `sizeof(fluid_voice_t)` as there may be many instances of voices alive. But sure, that's not your problem right now. Just change it back to int for now. However, casting `FLUID_N_ELEMENTS` to `int` definitely feels wrong to me.

carlo-bramini:
Thank you very much.

Issue #145, opened 2015-10-04
Patch to remove dependency on glib?


derselbst:
Would this be accepted? I've got the code to a point where it compiles with stubbed functions - the reason for doing this is because compiling on windows is quite painful. I was thinking of swapping which .c file gets compiled with cmake on windows with one that uses win32 api directly.

Reported by: paulsapps

Original Ticket: [fluidsynth/tickets/147](https://sourceforge.net/p/fluidsynth/tickets/147)

derselbst:
looking for a pull request

derselbst:
related discussion: http://lists.nongnu.org/archive/html/fluid-dev/2016-06/msg00007.html

derselbst:
glib is currently THAT part of fluidsynth which makes it available for multiple OSs. I would agree to a solution that gives windows an own implementation of mutexes, atomics, etc. For the question whether keeping glib at all, it would be interesting to know whether fluidsynth is used on any non windows and non posix OS (OS/2, BeOS??).

One might think about dropping support for non posix and non windows OS on a potential 1.2 / 2.0 release. That would allow also dropping glib and having Win32 API and pthreads taking care of all these synchronization tasks.

derselbst:
I dont want to get back to something like that: https://github.com/FluidSynth/fluidsynth/commit/5fcedecdfd6736e9d58c46767dd1bff5b0f04558#diff-df29015513acb3992896d306ff5dadc2L135

derselbst:
Conclusion from the mailing list: The only answer saying goodbye to glib can be moving to C11. As of today (2017) Windows doesnt support this standard at all and probably never will given the fact it still doesnt support C99. Also glibc still has no support for C11 threads. As of today, glib is the best option we have.

Moving back to custom defines as done before glib is not an option. Alternatively porting to C++11 might be an unrealistic overkill.

Closing as wontfix.

PR #247, opened 2017-10-22
Fixed dumptuning

rmattes:
The dumptuning currently returns uninitialized data instead of reporting an  error (or returning the standard equal temprament tuning). This patch implements minimal error reporting.

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -45,7 +45,7 @@ struct _fluid_cmd_handler_t {
   fluid_synth_t* synth;
   fluid_midi_router_t* router;
   fluid_cmd_hash_t* commands;
-  
+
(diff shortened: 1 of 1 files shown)


derselbst:
Thanks. Consider associating the email used for the commits with your github account, so you get recognized as contributor.

derselbst:
Oh, and make sure to declare variables always at the beginning of a scope. Though it's ugly, that's the only way C89 supports it.

rmattes:
Oh, sorry, I was (wrongly?) assuming that the project moves to C11. Wasn't that the plan to get glib-free threading support?

derselbst:
Wont happen too soon, see [#145](https://github.com/FluidSynth/fluidsynth/issues/145#issuecomment-337562513).

PR #248, opened 2017-10-22
Add in-place rendering to LADSPA

mawe42:
This set of commits make some performance optimizations to the LADSPA system. It adds an in-place rendering mode, which is only available if FluidSynth has been compiled with `-Denable-floats=1`. To enable it, a new setting `synth.ladspa.in-place` has been introduced. In-place rendering means that the LADSPA plugins operate directly on the mixer buffers when reading or writing to the system nodes. This eliminates two memcpy operations per input and output channel.

This pull request also changes the "normal" (not in-place) rendering entry point fluid_ladspa_run, removing the need for VLAs, removing lots of for-loops in rvoice mixer and reducing the number of executed atomic compare and exchange operations. fluid_ladspa_run and the new fluid_ladspa_run_in_place methods process all blocks to be rendered as a single large block, calling each plugin only once.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -82,6 +82,13 @@ https://stackoverflow.com/a/6251757
                 When set to "yes" the LADSPA subsystem will be enabled. This subsystem allows to load and interconnect LADSPA plug-ins. The output of the synthesizer is processed by the LADSPA subsystem. Note that the synthesizer has to be compiled with LADSPA support. More information about the LADSPA subsystem later.</desc>
         </setting>
         <setting>
+            <name>ladspa.in-place</name>
+            <type>bool</type>
+            <def>0 (FALSE)</def>
+            <desc>
+                (Only supported if FluidSynth has been compiled to use floats instead of doubles). When set to "yes" the LADSPA subsystem will enable plugins to render audio in-place, without the overhead of copying input and output buffers.</desc>
+        </setting>
+        <setting>
             <name>midi-channels</name>
             <type>int</type>
             <def>16</def>
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -502,7 +502,7 @@ static FLUID_INLINE int ladspa_run_start(fluid_ladspa_fx_t *fx)
     /* Somebody wants to deactivate the engine, so let's give them a chance to do that.
      * And check that there is at least one plugin loaded, to avoid the overhead of the
      * atomic compare and exchange on an unconfigured LADSPA engine. */
-    if (fluid_atomic_int_get(&fx->pending_deactivation) || fx->num_plugins == 0)
+    if (fx->pending_deactivation || fx->num_plugins == 0)
     {
         return FLUID_FAILED;
     }
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -430,18 +430,23 @@ static int create_input_output_nodes(fluid_ladspa_fx_t *fx)
 {
     char name[99];
     int i;
+    int buf_size;
+
+    /* If this LADSPA fx is set to work on the buffers passed into fluid_ladspa_run in-place,
+     * then don't allocate space on the nodes. */
+    buf_size = (fx->in_place) ?  0 : FLUID_BUFSIZE;
 
     /* Create left and right input nodes for each audio group. */
     for (i = 0; i < fx->audio_groups; i++)
     {
         FLUID_SNPRINTF(name, sizeof(name), "in%i_L", (i + 1));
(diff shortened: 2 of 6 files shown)


derselbst:
Is introducing that setting really necessary? Why not always rendering
in_place in case `WITH_FLOAT`? Would there be any benefit of *not* having
this enabled that I'm missing? The buffers of rvoice_mixer would be
changed/polluted anyway.

mawe42:
The reason I introduced that setting is to have identical behaviour, regardless of the compile time switch. When using in-place rendering, the input and output buffer nodes (e.g. "in1_L" and "out1_L") point to the same buffer. When copying the buffers, they are two separate arrays. I wanted to make sure that this change in behaviour is actively requested by the user. 

It's especially relevant as some LADSPA plugins can't cope with shared buffers for input and output. LADSPA even defines an INPLACE_BROKEN hint that plugins can set to inform the host of that fact. (I will add a check for that flag in a later pull request)

mawe42:
Hm... but you got me thinking. Maybe we can simplify the interface again by removing the ladsap_run_in_place function. So that the standard LADSPA usage would be like this:

in fluid_synth.c:
    fx = new_fluid_ladspa_fx(sample_rate, audio_groups, effects_channels, audio_channels)

in fluid_rvoice_mixer_set_ladspa:
    fluid_ladsa_set_host_buffers(fx, left_buf, right_buf, fx_left_buf, fx_right_buf, in_place)

in fluid_rvoice_process_fx:
    fluid_ladspa_run(fx, block_count, block_size);

derselbst:
> I wanted to make sure that this change in behaviour is actively requested by the user.

I was just thinking that if there is no observable change in behaviour to the user, we wouldnt need the setting. I could imagine that only a very few users would be consciously activating it. To me it seems this extra buffering only had the purpose to convert from fluid_real_t to LADSPA_FLOAT. So leaving that out and always doing in_place rendering if possible seems like a good idea one could only benefit from.

How about leaving the decision whether in_place rendering or not completely to the LADSPA subsystem? If compiled with double or one plugin sets INPLACE_BROKEN: render as usual. Else render in_place.

mawe42:
> I was just thinking that if there is no observable change in behaviour to the user, we wouldnt need the setting.

That's the thing: there is an observable effect to the user. An example: Lets say a user attaches a noise generator to out1_L, then another plugin to read from in1_L and output to out2_L:
```
ladspa_plugin noise.so white_noise
ladspa_port Output > out1_L

ladspa_plugin reverb.so reverb
ladspa_port Input < in1_L
ladspa_port Output > out2_L
```
Without in-place rendering, this works just fine: out1_L contains noise, out2_L contains in1_L with reverb added. 

With in-place rendering, the first plugin writes to out1_L and therefore also to in1_L (because it's the same buffer). Then the second plugin runs, taking the generated noise in in1_L as input, and not the original contents of that buffer.

So IMHO it definitely needs to be a conscious decision to enable in-place rendering.

derselbst:
I see (sorry, never used ladspa myself).

I dont think that these strange side effects are clear to a user at least from the current description of this setting. But even if you had pointed out this side effect I still think that even less users would actually use it. Think of it:

1. WITH_FLOAT is 0 by default (i.e. double, and most linux distros compile and ship it like that)
1. This requires the user to have a custom build where he is actually using `-Denable-float=1`
1. The user must be familiar with LADSPA
1. The user must use plugins that dont specify INPLACE_BROKEN
1. The user must manually activate in_place rendering

Then and only then will inplace rendering acutally happen and give you a slight performance boost. Is that worth it?

mawe42:
Granted... I'm scratching my own itch here. I want to use LADSPA for a custom reverb, and the device I'm going to use it on has limited resources and is designed to run on battery. So I'm trying to save CPU cycles and memory bandwidth wherever I can.

But you are probably right. The performance gains are only very minimal, even on my 1 GHz ARMv7 CPU. 

Ok, what about this: when FluidSynth is compiled with floats, then the input nodes (main and effect sends) are always passed on (point to the mixer.buffers), the output nodes are always copied. That way the behaviour is the same, we don't need the in-place setting anymore and there is still a small performance gain. Especially because the number of inputs is usually higher than the number of outputs.

derselbst:
Yes, that would be nice :)

mawe42:
You know... you are completely right. The performance gains are so minimal, it's not worth complicating the code and the interface for it. It might make a difference it if somebody is using many audio-groups for multi-channel output. But then they would need to use Jack, which can be used to apply effects very efficiently anyway. In the standard case with one audio group, it only gets rid of a maximum of 4 memcpy's. So let's call it what it is: premature optimization. And scrap the whole idea. Thanks for keeping an eye on these things!

So let me close this and prepare another pull request with the other fixes.

PR #249, opened 2017-10-23
More LADSPA cleanup and refactor

mawe42:
- Contains the render cleanup from the previous (but closed) pull request
- Adds check for in-place broken plugins
- Restructures fluid_ladspa_check function into smaller units

### DIFF
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -854,10 +854,11 @@ int fluid_ladspa_control_defaults(fluid_ladspa_fx_t *fx)
 int fluid_ladspa_check(fluid_ladspa_fx_t *fx, char *err, int err_size)
 {
     int i;
-    unsigned int k;
+    unsigned int j, k;
     int has_connections;
     int num_system_nodes;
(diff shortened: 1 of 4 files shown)


derselbst:
So because of a1ff8d0e55a588e893ca22bc6cd74afa41000f77 from now on the effect of any ladspa output node will be heard when pulling audio from the synth via `fluid_synth_(n)write_float()`? I'm fine with that. Just asking because I cant comprehend why it hasnt been done like that before.

mawe42:
Sorry, I dont' understand your question... All that a1ff8d0 functionally changes is to reduce the number of times fluid_ladspa_run (and the run method of all plugins) is called when rendering more than one block at a time.

derselbst:
Oh no, you're right. We are having buffers of buffers ofc. Argh.

mawe42:
The old code with the VLAs duplicated *the pointers* into the mixer buffers. Just to be able to advance the pointers by adding FLUID_BLOCK to each one after the fluid_ladspa_run call.

mawe42:
Great, thanks for the review!

PR #251, opened 2017-10-25
Some optimizations

derselbst:
as suggested by @jjceresa


jjceresa:
Both commits 

- avoid redundant zero checking
- use fluid_cb2amp() for lfo calculation
 
are Ok for me:

Issue #197, opened 2017-09-01
refactor rvoice_mixer


derselbst:
`rvoice_mixer` needs a bit refactoring after the workaround to fix #192. Think about:

- [x] remove `fluid_rvoice_eventhandler_t::is_threadsafe`
- [x] remove "synth.parallel-render"
- [x] remove fluid_voice_t::debug
- [x] refactor chrous and reverb types
  - fluid_revmodel_set takes floats.
  - fluid_revmodel_presets_t uses fluid_real_t
  - fluid_revmodel_t uses fluid_real_t
  - fluid_rvoice_mixer_set_chorus_params takes doubles
  - fluid_rvoice_mixer_set_reverb_params take doubles
  - fluid_rvoice_event_t::realparams is fluid_real_t.
  - public synth API takes doubles
  - etc.
- [x] ~each `rvoice` gets its own mono resampled `buffer[FLUID_MIXER_MAX_BUFFERS_DEFAULT*FLUID_BUFSIZE]`?~
- [x] ~openMP parallelization?~
- [x] get rid of that hug macro magic switch case in rvoice_event_handler
- [x] variable naming not very intuitive (buf_blocks, buf_count, current_blockcount ???)
- [x] loop that initializes rvoice buffer in `fluid_rvoice_buffers_check_bufnum()` is superfluous
- [x] efficiently render "silence"
- [x] rethink fluid_synth_check_finished_voices()
- [x] ~fluid_rvoice_set_output_rate() is called twice (by fluid_voice_set_output_rate() and fluid_rvoice_mixer_set_samplerate())~
- [x] process effects units in parallel

jjceresa:
About the mixer.
> variable naming not very intuitive (buf_blocks, buf_count, current_blockcount ???)

Some infos follows:
![mixer-fluid_rvoice_mixer_t](https://user-images.githubusercontent.com/32781294/36636041-9e2f6704-19be-11e8-934d-7a47346772da.JPG)

- `buf_count`: is the number of buffers .
This number  depends of the number of `audio output`. This number is specified by `mixer->buffers.buf_count`. (Probably a better naming should be related to the audio output number).

- Internally each buffer have the same maximum size (`buf_blocks`). (not a good name too).
- The size of a buffer (buf_blocks) is specified in blocks. Each block have `FLUID_BUFSIZE `samples

- `current_blockcount`: is the current  index (0 to buf_blocks) of the block been mixed.
***
[...]

PR #252, opened 2017-10-25
allow system override (to get rid of glib when not available)

loki666:
- use system includes for fluid_sys.h and fluidsynth_priv.h to allow overriding of sys implementation
- use fluid_atomic function to read and set atomics variables
- add typedef for atomics

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -20,12 +20,12 @@
 
 #include <glib.h>
 
-#include "fluidsynth_priv.h"
+#include <fluidsynth_priv.h>
 #include "fluid_cmd.h"
 #include "fluid_synth.h"
 #include "fluid_settings.h"
 #include "fluid_hash.h"
-#include "fluid_sys.h"
+#include <fluid_sys.h>
 #include "fluid_midi_router.h"
 #include "fluid_sfont.h"
 #include "fluid_chan.h"
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -20,12 +20,12 @@
 
 #include <glib.h>
 
-#include <fluidsynth_priv.h>
+#include "fluidsynth_priv.h"
 #include "fluid_cmd.h"
 #include "fluid_synth.h"
 #include "fluid_settings.h"
 #include "fluid_hash.h"
-#include <fluid_sys.h>
+#include "fluid_sys.h"
 #include "fluid_midi_router.h"
 #include "fluid_sfont.h"
 #include "fluid_chan.h"
--- a/src/rvoice/fluid_rvoice_event.c
+++ b/src/rvoice/fluid_rvoice_event.c
@@ -229,7 +229,7 @@ new_fluid_rvoice_eventhandler(int is_threadsafe, int queuesize,
    * that too many events are dispatched too early, causing incorrectly timed audio
    */
   eventhandler->is_threadsafe = TRUE;
-  eventhandler->queue_stored = 0;
+  fluid_atomic_int_set(&eventhandler->queue_stored, 0);
   
   eventhandler->finished_voices = new_fluid_ringbuffer(finished_voices_size,
                                                        sizeof(fluid_rvoice_t*));
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -736,7 +736,7 @@ fluid_mixer_get_mt_rvoice(fluid_rvoice_mixer_t* mixer)
 #define THREAD_BUF_TERMINATE 3
 
 /* Core thread function (processes voices in parallel to primary synthesis thread) */
-static FLUID_THREAD_RETURN_TYPE
+static fluid_thread_return_t
 fluid_mixer_thread_func (void* data)
 {
   fluid_mixer_buffers_t* buffers = data;  
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -45,7 +45,7 @@ struct _fluid_mixer_buffers_t {
   fluid_rvoice_t** finished_voices; /* List of voices who have finished */
   int finished_voice_count;
 
-  fluid_atomic_int ready;             /**< Atomic: buffers are ready for mixing */
+  fluid_atomic_int_t ready;             /**< Atomic: buffers are ready for mixing */
 
   int buf_blocks;             /**< Number of blocks allocated in the buffers */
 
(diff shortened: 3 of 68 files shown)


src/bindings/fluid_cmd.c
```diff
@@ -20,12 +20,12 @@
 
 #include <glib.h>
 
-#include "fluidsynth_priv.h"
+#include <fluidsynth_priv.h>
```
mawe42: Is this really necessary to be able to override fluidsynth_priv.h with your own header? Couldn't you solve this with a custom include path for the quote-headers, like the -iquote option for gcc?

src/rvoice/fluid_rvoice_event.c
```diff
@@ -229,7 +229,7 @@ new_fluid_rvoice_eventhandler(int is_threadsafe, int queuesize,
    * that too many events are dispatched too early, causing incorrectly timed audio
    */
   eventhandler->is_threadsafe = TRUE;
-  eventhandler->queue_stored = 0;
+  fluid_atomic_int_set(&eventhandler->queue_stored, 0);
```
derselbst: We are in a constructor function. I dont see why this is necessary.

src/rvoice/fluid_rvoice_mixer.c
```diff
@@ -736,7 +736,7 @@ fluid_mixer_get_mt_rvoice(fluid_rvoice_mixer_t* mixer)
 #define THREAD_BUF_TERMINATE 3
 
 /* Core thread function (processes voices in parallel to primary synthesis thread) */
-static void
+static FLUID_THREAD_RETURN_TYPE
```
derselbst: Why?

src/bindings/fluid_cmd.c
loki666: Nope quote includes always looks first one the directory of the file that make the include

src/synth/fluid_synth.c
```diff
@@ -439,18 +441,12 @@ fluid_synth_init(void)
 
 static FLUID_INLINE unsigned int fluid_synth_get_ticks(fluid_synth_t* synth)
 {
-  if (synth->eventhandler->is_threadsafe)
-    return fluid_atomic_int_get(&synth->ticks_since_start);
-  else
-    return synth->ticks_since_start;
```
derselbst: Too early. Will be addressed in #197.

src/synth/fluid_synth.c
```diff
@@ -595,8 +593,10 @@ new_fluid_synth(fluid_settings_t *settings)
   
   synth->settings = settings;
 
-  fluid_settings_getint(settings, "synth.reverb.active", &synth->with_reverb);
-  fluid_settings_getint(settings, "synth.chorus.active", &synth->with_chorus);
+  fluid_settings_getint(settings, "synth.reverb.active", &with_reverb);
+  fluid_atomic_int_set(&synth->with_reverb, with_reverb);  
+  fluid_settings_getint(settings, "synth.chorus.active", &with_chorus);
+  fluid_atomic_int_set(&synth->with_chorus, with_chorus);
```
derselbst: Again: Constructor: dont see necessity.

src/sfloader/fluid_defsfont.c
```diff
@@ -2030,24 +2034,40 @@ fluid_sample_import_sfont(fluid_sample_t* sample, SFSample* sfsample, fluid_defs
...
+#define INT32_FROM_LE(x) x
+#define INT16_FROM_LE(x) x
+#endif
+
+
 #define READCHUNK(var,fd)	do {		\
 	if (!safe_fread(var, 8, fd))			\
 		return(FAIL);				\
-	((SFChunk *)(var))->size = GUINT32_FROM_LE(((SFChunk *)(var))->size);  \
+	((SFChunk *)(var))->size = UINT32_FROM_LE(((SFChunk *)(var))->size);  \
```
derselbst: I'm not feeling comfortable to start duplicating functionality that is provided by glib.

src/bindings/fluid_cmd.c
derselbst: What keeps you from hard-replacing these files with your own header?

loki666: Ok I'll revert these changes, an replace these files in my own branch

src/rvoice/fluid_rvoice_event.c
loki666: atomic types needs to be opaque types for one to use a different implementation

src/rvoice/fluid_rvoice_mixer.c
loki666: the thread function return type might be different (for example with pthread it's void*)

src/synth/fluid_synth.c
loki666: again, atomics type need to be opaque, i didnt want to get rid of is_threadsafe, but I need a fuction to access the atomic

src/sfloader/fluid_defsfont.c
loki666: the goal is to try to allow one to get rid of glib.
I can add a define in fluid_sys.h for that then ?

src/synth/fluid_synth.c
loki666: atomic type need to be opaque

mawe42: I actually quite like this change and also the fluid_atomic_* typedefs. They make it immediately obvious that these values are meant to be accessed atomically throughout the codebase.

src/rvoice/fluid_rvoice_mixer.c
derselbst: Then you should have setup a `typedef fluid_thread_return_t` and consistently applied that everywhere we use `new_fluid_thread` and also to `fluid_thread_func_t`.

src/synth/fluid_synth.c
mawe42: They might even help us to get rid of these annoying gcc warnings:
```
/usr/include/glib-2.0/glib/gatomic.h:117:5: warning: cast from function call of type ‘_Bool’ to non-matching type ‘int’ [-Wbad-function-cast]
```

src/sfloader/fluid_defsfont.c
derselbst: Not sure. @mawe42 What do you think of duplicating that byteswap logic?

src/synth/fluid_synth.c
derselbst: Maybe. However it should be `typedef fluid_atomic_int_t`. And it should also cleanly be done in one commit only addressing this issue and not something else.

src/rvoice/fluid_rvoice_mixer.c
loki666: on it

src/synth/fluid_synth.c
loki666: so you want me to remake splited pull requests ?

loki666:
ok let me split this it's a mess now

src/sfloader/fluid_defsfont.c
mawe42: Couldn't @loki666 just define GUINT32_FROM_LE in his own glib-free fluid_sys.h?

loki666: yes but fluid_defsfont.h still include glib.h

mawe42: Ok... but wouldn't you need to remove that as well, to get rid of the glib dependency?

loki666: well... I d'like to avoid to replace the fluid_defsfont.c/h just because of the deps on simple GLIB marcos for this... but if you don't want to change that I'll do it in my branch.

I wanted to have a minimum changes in my branch so I can merge from your repository without having too much conflicts

derselbst: Then please `#define FLUID_LE32TOH` in `fluidsynth_priv.h` which replace `G*_FROM_LE`.

loki666: Ok I'll do that in another PR

derselbst: Nevermind. I'll draft an example.

PR #253, opened 2017-10-25
add typedef for atomics

loki666:
access atomics via fluid_atomic_X_set/get

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -116,7 +116,7 @@ static void fluid_synth_stop_LOCAL (fluid_synth_t *synth, unsigned int id);
  */
 
 /* has the synth module been initialized? */
-static int fluid_synth_initialized = 0;
+static fluid_atomic_int_t fluid_synth_initialized = 0;
 static void fluid_synth_init(void);
 static void init_dither(void);
 
@@ -272,8 +272,6 @@ fluid_version_str (void)
 static void
 fluid_synth_init(void)
 {
-  fluid_synth_initialized++;
-
 #ifdef TRAP_ON_FPE
   /* Turn on floating point exception traps */
   feenableexcept (FE_DIVBYZERO | FE_UNDERFLOW | FE_OVERFLOW | FE_INVALID);
@@ -553,7 +551,7 @@ new_fluid_synth(fluid_settings_t *settings)
   int with_chorus = 0;
   
   /* initialize all the conversion tables and other stuff */
-  if (fluid_synth_initialized == 0)
+  if (fluid_atomic_int_compare_and_exchange(&fluid_synth_initialized, 0, 1))
   {
     char buf[64];
     if (fluid_settings_str_equal (settings, "synth.volenv", "compliant"))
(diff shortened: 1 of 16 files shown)


src/synth/fluid_synth.c
```diff
@@ -555,6 +549,8 @@ new_fluid_synth(fluid_settings_t *settings)
   double gain;
   int i, nbuf;
   int with_ladspa = 0;
+  int with_reverb;
+  int with_chorus;
```
mawe42: I would initialize with_reverb and with_chorus explicity. Otherwise compilers might complain about a possible unitialized use of those variables...

src/synth/fluid_synth.h
```diff
@@ -142,22 +142,22 @@ struct _fluid_synth_t
...
-  int chorus_nr;                     /**< Shadow of chorus number */
-  float chorus_level;                /**< Shadow of chorus level */
-  float chorus_speed;                /**< Shadow of chorus speed */
-  float chorus_depth;                /**< Shadow of chorus depth */
-  int chorus_type;                   /**< Shadow of chorus type */
+  fluid_atomic_int_t chorus_nr;                     /**< Shadow of chorus number */
+  fluid_atomic_float_t chorus_level;                /**< Shadow of chorus level */
+  fluid_atomic_float_t chorus_speed;                /**< Shadow of chorus speed */
+  fluid_atomic_float_t chorus_depth;                /**< Shadow of chorus depth */
+  fluid_atomic_int_t chorus_type;                   /**< Shadow of chorus type */
```
mawe42: I think the chorus_* and reverb_* members of fluid_synth_t can be simple ints, no need for atomics here. Access to those values are protected via the fluid API lock anyway. They are passed by value to the rvoice eventhandler, so the other threads don't access them either.

loki666: should I make chorus_level, chorus_speed, chorus_depth double type (since they are passed as double in the functions fluid_synth_set_chorus_level, fluid_synth_set_chorus_speed, fluid_synth_set_chorus_depth ?

derselbst: Yes double might be better.

loki666: dont

mawe42: The `fluid_rvoice_eventhandler_push*` functions take `fluid_real_t` values, not float or double. If FluidSynth is compiled with `-Denable-floats=1`, wouldn't the precision loss cause a warning?
I would suggest defining them as `fluid_real_t` instead.

loki666: then the API should also change
fluid_synth_set_reverb_full is using double, and the args are directly passed to fluid_rvoice_eventhandler_push5

derselbst: `fluid_rvoice_mixer_set_chorus_params` and `fluid_rvoice_mixer_set_reverb_params` take doubles, while `fluid_rvoice_event_t::realparams` is `fluid_real_t`. I'll take care of that afterwards.

mawe42: There's more:
- fluid_revmodel_set takes floats.
- fluid_revmodel_presets_t uses fluid_real_t
- fluid_revmodel_t uses fluid_real_t

It's all a little... inconsistent :-)

derselbst: Ok, too much for now. Lets postpone this until the new fx_api struct thing. Noted in  #197.

loki666:
should be good now

PR #254, opened 2017-10-25
thread function return type

loki666:
- add typedef for thread starting routine return type
- add #define for thread return value

### DIFF
--- a/src/utils/fluid_sys.h
+++ b/src/utils/fluid_sys.h
@@ -288,8 +288,11 @@ fluid_atomic_float_get(volatile float *fptr)
 
 /* Threads */
 
+typedef fluid_thread_return_t void
+#define FLUID_THREAD_RETURN_VALUE
+
 typedef GThread fluid_thread_t;
-typedef void (*fluid_thread_func_t)(void* data);
+typedef fluid_thread_return_t (*fluid_thread_func_t)(void* data);
 
 #define FLUID_THREAD_ID_NULL            NULL                    /* A NULL "ID" value */
 #define fluid_thread_id_t               GThread *               /* Data type for a thread ID */
--- a/src/utils/fluid_sys.h
+++ b/src/utils/fluid_sys.h
@@ -288,8 +288,10 @@ fluid_atomic_float_get(volatile float *fptr)
 
 /* Threads */
 
-typedef void* fluid_thread_return_t;        /* other system thread implementation might changes this for their needs */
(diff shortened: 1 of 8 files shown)


loki666:
for fluid_cmd the issue is that this function is used both as a thread starting routing and directly and that it returns the error code.

Should I move the error code in the fluid_shell_t struct ?

derselbst:
Sry, fluid_alsa.c is still incomplete. And ea48bf9a47f80cf75ac5d8c97281b3ba65ba3a60 is a bit mistaken.

> Should I move the error code in the fluid_shell_t struct ?

Good point. Moving it to the struct is a bit far away. How about using a call-by-ref variable instead?

`fluid_shell_run(fluid_shell_t* shell, int* error_out)`

Where `error_out` may be NULL if the result is not needed.

src/utils/fluid_sys.h
```diff
@@ -288,8 +288,11 @@ fluid_atomic_float_get(volatile float *fptr)
 
 /* Threads */
 
+typedef void fluid_thread_return_t;
+#define FLUID_THREAD_RETURN_VALUE
```
derselbst: And consider adding a comment here pointing out why this is needed to avoid that another maintainer optimizes that out in the future.

loki666:
@derselbst thread starting functions only takes one argument

loki666:
sorry for messing up fluid_alsa.c, hope you can merge commits

derselbst:
> thread starting functions only takes one argument

Oops, wishful thinking. Ok then let me think about it again. In the meantime: what about `fluid_alsa_seq_run()` and `fluid_alsa_midi_run()`?

derselbst:
How about:

```
typedef void* fluid_thread_return_t;
#define FLUID_THREAD_RETURN_VALUE NULL
```

Equivalent to what GThreadFunc is typedefed.

For `fluid_shell_run()` we could make some ugly custom int cast then. Which I still consider to be better than placing a return value of that function in the struct, because it is needed exactly once inside `fluid_source()`.


You missed `fluid_server_socket_run()` btw.

loki666:
Should I merge back master in my branch to resolve this?

derselbst:
No, I'm on it.

derselbst:
Hm, strange, had to merge twice before conflict got resolved. Anyway, I think it looks good now. Thinking about squashing this rather than merging.

PR #255, opened 2017-10-25
Autoconnect ALSA MIDI inputs.

### LINKED ISSUE #156: Auto-connect ALSA MIDI inputs
While using QSynth+fluidsynth mostly as a hobbist, I came across this need for
reconnecting the ALSA MIDI input to QSynth/fluidsynth everytime I restart the engine.

So I worked out the attached patch that adds a bool setting for auto-connecting ALSA
MIDI inputs. This is the dual of the "Auto-connect JACK outputs" already there in
fluidsynth/QSynth.

If this makes sense, I'd follow up with a corresponding patch to QSynth to set this
property when instantiating the fluidsynth engine.

Reported by: tcucinotta

Original Ticket: [fluidsynth/tickets/158](https://sourceforge.net/p/fluidsynth/tickets/158)

derselbst:
fixes #156

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -240,14 +240,6 @@ https://stackoverflow.com/a/6251757
             </desc>
         </setting>
         <setting>
-            <name>alsa.autoconnect</name>
-            <type>bool</type>
-            <def>0 (FALSE)</def>
-            <desc>
-                If 1 (TRUE), automatically connects FluidSynth to available ALSA MIDI input ports.
-            </desc>
-        </setting>
-        <setting>
             <name>alsa.device</name>
             <type>str</type>
             <def>default</def>
@@ -412,6 +404,14 @@ https://stackoverflow.com/a/6251757
     <midi>
         <setting>
             <isFirst>MIDI driver settings</isFirst>
+            <name>autoconnect</name>
+            <type>bool</type>
+            <def>0 (FALSE)</def>
+            <desc>
+                If 1 (TRUE), automatically connects FluidSynth to available MIDI input ports. alsa_seq is currently the only driver making use of this.
+            </desc>
+        </setting>
+        <setting>
             <name>driver</name>
             <type>str</type>
             <def>alsa_seq (Linux),<br />
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -240,6 +240,14 @@ https://stackoverflow.com/a/6251757
             </desc>
         </setting>
         <setting>
+            <name>alsa.autoconnect</name>
+            <type>bool</type>
+            <def>0 (FALSE)</def>
+            <desc>
+                If 1 (TRUE), automatically connects FluidSynth to available ALSA MIDI input ports.
+            </desc>
+        </setting>
+        <setting>
             <name>alsa.device</name>
             <type>str</type>
             <def>default</def>
--- a/src/drivers/fluid_alsa.c
+++ b/src/drivers/fluid_alsa.c
@@ -179,14 +179,14 @@ static void fluid_alsa_autoconnect_inputs(fluid_alsa_seq_driver_t* dev, const sn
       snd_seq_port_subscribe_set_dest(subs, dest);
 
       if (snd_seq_get_port_subscription(seq, subs) == 0) {
-	fprintf(stderr, _("Connection %s is already subscribed\n"), pname);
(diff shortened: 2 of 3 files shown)


derselbst:
@tomcucinotta FYI

tomcucinotta:
hi there, thanks for the notification, can I do anything to help out here ? my old patch was from ~1year ago, was it rebased ?

doc/fluidsettings.xml
```diff
@@ -240,6 +240,14 @@ https://stackoverflow.com/a/6251757
             </desc>
         </setting>
         <setting>
+            <name>alsa.autoconnect</name>
+            <type>bool</type>
+            <def>0 (FALSE)</def>
+            <desc>
+                If 1 (TRUE), automatically connects FluidSynth to available ALSA MIDI input ports.
```
derselbst: @tomcucinotta Haven't tested it yet. You could let me know if this documentation is sufficient.

tomcucinotta: docs seem ok, thanks!

tomcucinotta: Hi, I also just recompiled this new alsa-autocon branch, and tried the feature, plugging my USB MIDI adapter, then launching:

  ./src/fluidsynth -o audio.alsa.autoconnect=1 -a alsa -i /usr/share/sounds/sf2/FluidR3_GM.sf2 -s

I've got the USB MIDI adapter, as well as the MIDI through, readily connected to FLUID Synth :-)!

derselbst: Confirmed.

mawe42:
> But sending data to those ports doesn't work, of course, because they are not sequencer but raw ports.
Forget that. Sending data does work, just not with aplaymidi. If I send raw midi events to the port, FluidSynth reacts just fine.

So the only remaining valid comment is the suggestion to rename `audio.alsa.autoconnect` to `midi.alsa.autoconnect`.

derselbst:
Ok, but should it be `midi.alsa.autoconnect` or `midi.alsa_seq.autoconnect`? It applies for both. I would also vote for the first. Or we have to make it work only with real alsa_seq ports, not with raw ones.

tomcucinotta:
anyone willing to explain what a raw MIDI port is useful for? I'm just reading from alsa docs that it doesn't add timestamps, but I confess I don't really know what it's talking about ... sorry for the newbie Q, thanks.

derselbst:
Never figured that out myself. All I know: "There are seq ports and raw ports. Always use seq ports!" Sry.

mawe42:
A raw MIDI port is basically what your USB keyboard creates when it attaches to your computer. It's a port that outputs (or takes in) raw MIDI events. So no MIDI files, with headers etc, just the plain MIDI protocol data in binary form. Attaching FluidSynth with alsa_raw and midi.alsa.device=hw:1,0 (or similar) to your MIDI device is the best way for very low latency MIDI input. Going through alsa_seq always adds a little more latency, as I understand it.

mawe42:
Also, alsa_raw ports are usually opened in exclusive mode, so only one application can use the port. The sequencer ports can be subscribed to by many clients at the same time. But like I said, this flexibility comes at the cost of higher resource consumption and latency.

mawe42:
And with regard to the timestamps in alsa_seq, I think it works as follows: a raw midi event (like note on, channel 0, note 60, velocity 127) get sent by the MIDI device (for example a keyboard). It comes in via a raw midi device and gets processed by the alsa sequencer engine. The sequencer adds a timestamp to it saying "this is when I received the event". Then the sequencer passes on the event with the timestamp to all connected clients.

For real time control, those timestamps would not be necessary. But the alsa_seq interface can also take events with a timestamp in the future. And that is basically what a MIDI file is.

derselbst:
Aha, so does it even make sense then connecting to raw ports?

mawe42:
> Aha, so does it even make sense then connecting to raw ports?

Yes, very much so. Especially if you want to control FluidSynth in real time with a MIDI instrument or other MIDI controller.

derselbst:
Ok, so `midi.alsa.autoconnect` or `midi.alsa_seq.autoconnect`?

mawe42:
With the current implementation, it should probably be `midi.alsa_seq.autoconnect`, because it uses the ALSA sequencer api and doesn't have any effect for `-m alsa_raw`.

Or we could go for `midi.autoconnect`, assuming other midi drivers could also benefit from an auto-connect feature. But I don't know anything about coremidi or Windows MIDI...

PR #256, opened 2017-10-26
remove glib.h import from fluid_defsfont.h

loki666:
not needed anymore since the glib macro is now wrapped in a custom FLUID macro


derselbst:
Ok

PR #257, opened 2017-10-26
dont register synth.default-soundfont if not defined

loki666:
on some system a DEFAULT_SOUNDFONT might not exists


derselbst:
Should always be defined as long as you build with cmake. But doesnt hurt either.

PR #260, opened 2017-10-28
add contribution guidelines

### LINKED ISSUE #250: Coding Style
Somewhat related to #179, but a little larger: I would love it if we could decide on a coding style that we expect in new contributions. The FluidSynth codebase has grown over many years and with many different coding styles. The differences are not just in the white-space, but also with things like naming of functions and variables, naming of macros, whether to use braces for single line statements after if, etc...

I think it would help a lot towards a more readable and maintainable codebase to set something, anything, as soon as possible, publish it and apply those rules to new code.

Even just a few rules would go a long way. I don't have a strong preference, but personally I like:

- Macros always UPPER_CASE_AND_SEPARATED_BY_UNDERSCORES
- Function names always lower case, words separated by underscores
- Variable names always lower case, words separated by underscores
- Every block after an if, else, while or for should be enclosed in braces (but single `continue` or `break` statements are allowed)
- Every function, even static ones, should have a short comment explaining it's purpose
- Every non-static function should be documented with purpose, params and return value
- (From the discussion about unsigned int) Use plain `int` for numbers, even if that number should only ever be positive.

What do you think?

derselbst:
\+ style guide, fixes #250 

comments, suggestions, improvements welcome

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,27 @@
+# Contributing
+
+Thanks for considering to contribute to FluidSynth. Before implementing
+any huge new feature, consider bringing up your ideas on our mailing list:
+https://lists.nongnu.org/mailman/listinfo/fluid-dev
+
+Contributing can be done by submitting pull requests on Github:
+https://help.github.com/articles/proposing-changes-to-your-work-with-pull-requests/
+
+Or by submitting patches to the mailing list. Patches should be created with
+
+```
+git format-patch
+```
+
+So in every case you should be familiar with the basics of git.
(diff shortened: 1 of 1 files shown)


derselbst:
> I don't think promising a response time of three days (or even one day) is necessary.

I actually just copied it from somewhere :D but I guess you're right.

> do you have a preference if you would rather have pull requests or patches?

As long as it's a `git format-patch` I dont really care, both are easy to apply.

jjceresa:
Good proposal as far it makes easier for any reviewers.

PR #261, opened 2017-10-30
add seek to player

loki666:


loki666:
I'm postponing the seek until `fluid_player_callback()` for threat safeness.
Playing with the position while the callback is running is probably not a good idea.

`fluid_track_send_events()` send out all events, but I want to skip note on / off

derselbst:
> I'm postponing the seek until fluid_player_callback() for threat safeness.

Ok, sure, why not.

> fluid_track_send_events() send out all events, but I want to skip note on / off

And why not calling this function with a parameter `is_seeking` that if `true` would skip notes?

loki666:
No really, I don't see how I can use `fluid_track_send_events()`.
`fluid_track_send_events()` is processing a single track, but `fluid_player_seek_local()` is processing all the tracks together, like for seeking backward...It needs to reset all tracks

both functions do indeed process the tracks events, but their intent is very different. I don't understand why you insist on using `fluid_track_send_events()`
I mean, my understanding of clean code was to have small dedicated functions instead of one big function that does everything (and a boolean flag to change the behavior of a functions in generally not clean code)

derselbst:
When I was looking at `fluid_player_seek_local()` I thought that this `while(1)` loop basically does the same job as `fluid_track_send_events()`. The only difference is that the seek indeed processes all tracks together, but I dont see a reason why this is necessary. Doing it one by one should work as well. And prior to that `fluid_player_seek_local()` could still reset tracks if seeking backwards.

So unless I'm missing something, I'd rather prefer introducing a boolean flag to some existing function, instead of duplicating code in order to do basically the same thing but differently.

loki666:
this should works

src/midi/fluid_midi.c
```diff
@@ -1716,6 +1734,26 @@ fluid_player_get_status(fluid_player_t *player)
 }
 
 /**
+ * Seek in the current playing file.
```
derselbst: Should be currently, it's an adverb.

src/midi/fluid_midi.c
```diff
@@ -1716,6 +1734,26 @@ fluid_player_get_status(fluid_player_t *player)
...
+ *   #FLUID_OK otherwise
+ * @since 1.1.8
+ * 
+ * The actual seek is performed during the player_callback.
+ */
+int fluid_player_seek(fluid_player_t *player, int ticks)
+{
+    if (ticks > fluid_player_get_total_ticks(player)) {
+        return FLUID_FAILED;
+    }
```
derselbst: Calling it with a negative tick count should also return `FLUID_FAILED`.

src/midi/fluid_midi.c
```diff
@@ -1716,6 +1734,26 @@ fluid_player_get_status(fluid_player_t *player)
 }
 
 /**
+ * Seek in the current playing file.
+ * @param player MIDI player instance
+ * @param ticks the position to seek to in the current file
+ * @return #FLUID_FAILED if ticks is after the latest tick of the file,
+ *   #FLUID_OK otherwise
+ * @since 1.1.8
```
derselbst: since 1.1.9

src/midi/fluid_midi.c
```diff
@@ -1716,6 +1734,26 @@ fluid_player_get_status(fluid_player_t *player)
 }
 
 /**
+ * Seek in the currently playing file.
+ * @param player MIDI player instance
+ * @param ticks the position to seek to in the current file
+ * @return #FLUID_FAILED if ticks is negative of after the latest tick of the file,
```
derselbst: > if ticks is negative of after the

find the typo...

src/midi/fluid_midi.c
```diff
@@ -1202,6 +1202,13 @@ fluid_track_send_events(fluid_track_t *track,
 {
     int status = FLUID_OK;
     fluid_midi_event_t *event;
+    int seeking = player->seek_ticks >= 0;
+
+    if (seeking) {
+        ticks = player->seek_ticks; /* update target ticks */
+        if (track->ticks > ticks)
+            fluid_track_reset (track); /* reset track if seeking backwards */
```
derselbst: Seems to work. But you should call `fluid_synth_all_sounds_off()` in this if clause to avoid hanging notes.

derselbst:
Looks good now. Last question: do you really need this for 1.1.9? I was actually planning to have the 1.1.x releases as maintenance / bug fixing releases only and to keep new features on master (=2.0). So if you're having an "urgent" use case for this I would ship it with 1.1.9, but if you're going to use recent git anyway I'll leave it on master only.

loki666:
yes it's the latest pull request I need for a MIDI player, so I don't need to use my fluidsynth fork which I can't keep up to date with this repository anymore.
too much conflict on merge

It's not urgent but I'd really like to get rid of my fork

derselbst:
Sounds like you assume that your previous PRs (atomics, glib, thread return...) will be part of 1.1.x. This wont be the case as 1.1.x will stay mostly untouched only receiving bug fixes and not all these code cleanups that might cause trouble for someone else and I would be infinitely busy fixing an EOL release. So if you dont rely on ABI stability your best option would be to branch master. Else I'm afraight you'll have to wait until 2.0, sry.

loki666:
oh branching master is fine for me

willem179:
@loki666 Thanks for implementing my patch!
You clearly used the source of fluid_player_seek_local() from my webpage (where it is since 2014-11-11): https://wim.vree.org/svgParse/testplayer.html

PR #264, opened 2017-10-31
support for registering audio drivers based on actual needs

### LINKED ISSUE #218: Initialize only adrivers based on actual needs
As reported by @ReinholdH:

Buggy soundcard drivers have caused portaudio to segfault. If fluidsynth is compiled with support for portaudio it will always attempt to `Pa_Initialize()` within `new_fluid_settings()` (even if the user doesnt want to use portaudio) and thus causing fluidsynth to crash as well as any application using fluidsynth.

Thus introduce the ability to register only certain audio drivers based on acutal needs. Achieve this similar to other projects (like FFMpeg) by introducing 

```
int fluid_register_adriver(...)
```

which can be called once after fluidsynths startup to register audio driver that shall be usable throughout fluidsynths lifetime. If this function is not being called, all available audio drivers are registered by default.

If possible, have this functionality for the next maintenance release, else for the next major one.

derselbst:
fixes #218


derselbst:
@ReinholdH Please review the API doc of `fluid_audio_driver_register()` and give me feedback.
https://github.com/FluidSynth/fluidsynth/pull/264/files#diff-b5264a77b9838a6ccbd55c0efab35053R452

jjceresa:
>  @warning This function may only be called if no thread is residing in fluidsynth's API and no instances of any kind are alive (e.g. as it would be the case right after fluidsynth's inital creation). Else the behaviour is undefined.

Probably you mean that ` fluid_audio_driver_register(const char** adrivers)` ****must be call** before** any call to `new_fluid_settings()` like in this example ?:

```
void init() 
{
    fluid_settings_t* settings;
    fluid_synth_t* synth;
    fluid_audio_driver_t* adriver;
    /* 
        Here the place to call fluid_audio_driver_register().
    */
    /* Set the synthesizer settings, if necessary */
    fluid_settings_setstr(settings, "audio.driver", "jack");
    adriver = new_fluid_audio_driver(settings, synth);
}
```
Does this example is the correct practice to obtain the intended behavior ?

derselbst:
> Probably you mean that fluid_audio_driver_register(const char** adrivers) must be call before any call to new_fluid_settings()

No. You can call this function if needed, you dont have to. If you dont call it, everything works as before, as this note points out:

> By default all audio drivers fluidsynth has been compiled with are registered.

However if you need to call this function, the example given by you would be the correct/intended use case.

jjceresa:
Thanks for this useful precision.

src/drivers/fluid_adriver.c
```diff
@@ -460,18 +460,16 @@ delete_fluid_audio_driver(fluid_audio_driver_t* driver)
...
  * The \c names are the same as being used for the \c audio.driver setting.
  * 
- * By default all audio drivers fluidsynth has been compiled with are registered.
+ * By default all audio drivers fluidsynth has been compiled with are registered, so calling this function is optional.
  * 
- * Any attempt of using audio drivers that have not been registered is undefined behaviour!
+ * @warning This function may only be called if no thread is residing in fluidsynth's API and no instances of any kind
+ * are alive (e.g. as it would be the case right after fluidsynth's inital creation). Else the behaviour is undefined.
+ * Furtermore any attempt of using audio drivers that have not been registered is undefined behaviour!
  *
```
jjceresa: I'm regularly doing the sames typos errors ("behavior") line 466 and 467 :)

derselbst: `Behaviour` is British English, `Behavior` American. British is the correct English ;P

jjceresa: Good to know.

PR #267, opened 2017-11-04
Stdint fixes

carlo-bramini:


derselbst:
The problem with this change is that it will fail to compile on platforms that dont have `stdint.h`. So instead of emulating it with the MSVC types, you should use the glib types instead.

derselbst:
I was actually meaning to replace MSVC types with glib's int types. Because as long as we have glib as OSAL, additionally specifying MSVC types feels redundant.

For completeness you should also `typedef gint64  int64_t;`

carlo-bramini:
I removed custom type for MSVC as you suggested. Actually, I can obtain the same result (compiling fluidsynth without GLIB dependencies) by installing msinttypes package.

PR #269, opened 2017-11-06
FluidSynth for window: Window needed by dsound is now owned by dsound audio driver

jjceresa:
This PR is related to http://lists.nongnu.org/archive/html/fluid-dev/2017-11/msg00006.html

`Window creation handle` is now owned by `dsound_audio_driver`.

- A very simple enhancement allows the application to instruct `dsound audio driver` to use the `application window handle` (see `fluid_win32_set_window()` in fluid_dll.c).
That means that if the application is`windowed`, it can benefit of the `cooperative window behavior` intended by dsound.

- If  the application doesn't instruct `dsound audio driver`, everything works as before (i.e the driver creates/destroys it own window).

The `DllMain()` function  was cleaned and is no longer in charge of window creation handle.


derselbst:
@carlo-bramini FYI

carlo-bramini:
Thank you very much.
Perhaps, after reading the code, together with this change, it would be also the chance to fix some things:

I suggest to make a const string rather than a const pointer to a RAM string.
Strings should be also declared as TCHAR if we want to use "generic" functions (otherwise we must take care if we are compiling for UNICODE or basic ASCII).

```
/***
  Window creation functions needed for dsound.
***/
static const TCHAR fluid_class_name[] = _T("FluidSynth");
```

Use correct declaration of WNDPROC

```
/* window procedure */
static LRESULT CALLBACK
fluid_wndproc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
{
    return DefWindowProc(hWnd, message, wParam, lParam);
}

```
Declare a global class only once for the current process.

```
/* Class creation.
 * @creates the class for supporting window(s)
 */
static int
fluid_win32_create_class(HINSTANCE hInstance)
{
  WNDCLASS wc;

  /* Check previous creation of the global class */
  if (GetClassInfo(hInstance, fluid_class_name, &wc))
    return 1;

  ZeroMemory(&wc, sizeof(wc));

  wc.lpszClassName = fluid_class_name;
  wc.style         = CS_GLOBALCLASS;
  wc.lpfnWndProc   = fluid_wndproc;
  wc.hInstance     = hInstance;

  return RegisterClass(&myClass);
}
```

Just use TCHAR strings, as said before.

```
/* window creation 
 * @return handle of window if success NULL otherwise
 */
static HWND
fluid_win32_create_window(void)
{
  HINSTANCE hInstance = fluid_get_hinstance();

  if (fluid_win32_create_class(hInstance))
  {
    /* now create the window */
    return
    CreateWindow(fluid_class_name,
                 _T("FluidSynth"),
                 WS_OVERLAPPEDWINDOW,
                 CW_USEDEFAULT,
                 CW_USEDEFAULT,
                 CW_USEDEFAULT,
                 CW_USEDEFAULT,
                 NULL,
                 NULL,
                 hInstance,
                 NULL);
  } else
    return NULL;
}
```

derselbst:
> Strings should be also declared as TCHAR if we want to use "generic" functions (otherwise we must take care if we are compiling for UNICODE or basic ASCII).

Is it even possible to compile fluidsynth for UNICODE? We're using plain old `char*` throughout the whole lib. Anyway, I dont mind going with TCHAR here.

> Use correct declaration of WNDPROC

Do we even need to wrap `DefWindowProc()` inside `fluid_wndproc()` couldnt we just use it directly?

jjceresa:
@carlo-bramini 
>  suggest to make a const string rather than a const pointer to a RAM string.

> Use  correct declaration of WNDPROC

Good suggestions, thanks.

> Declare a global class only once for the current process.

The class `fluid_window_class_name` here is intended only for a `default  window` required by dsound driver only:
-  for this reason, the couple (`class, window`) are both `created `at `driver creation` time and both `deleted ` at `driver deletion` time.
- The `intended behavior `of this window (defined  inside the body of fluid_win32_wndproc()) is `empty` (which means :` do nothing`).
Effectively, inside fluid_win32_wndproc() there is no processing of the `message `parameter (apart  DefWindowProc(); more on this later). In others words this window is unable to dialog with the user.
Even the style of the window (chosen by dsound driver at creation time) is an `invisible `window. Effectively at driver creation time any user don't expect to see a window that only able to do `nothing`.

Declaring a global class `fluid_window_name` only once for the current process means that the library FluidSynth intends to give the  application the possibility to create new window with the class `fluid_window_class_name`. 
From the application developer view  what could be the purpose of creating a window that do nothing (i.e a window unable to dialog with the user) ?

> Strings should be also declared as TCHAR if we want to use "generic" functions (otherwise we must take care if we are compiling for UNICODE or basic ASCII).

This `static `string `fluid_window_class_name` only expected by API RegisterClass() and CreateWindow() are type`LPCTSTR ` (i.e Pointer to a constant null-terminated string of 8-bit (ANSI). So declaring it `static const char` seems sufficient and have been compiled without problem.

As Tom said , `"We're using plain old char* throughout the whole lib. Anyway, I dont mind going with TCHAR here"`.

Did you get any issue when compiling fluid_dsound.c ? 
***
@derselbst 
> Do we even need to wrap DefWindowProc() inside fluid_wndproc() couldnt we just use it directly?
```
static LRESULT CALLBACK fluid_win32_wndproc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
{
    return DefWindowProc(hWnd, message, wParam, lParam);
}
```
`fluid_win32_wndproc()` registered in Windows OS (by `API RegisterClass()`) are callback function called by Windows OS at appropriate time. The `fluid_win32_wndproc() `function must return an expected `result value` that depend of the `message ` parameter. When `fluid_win32_wndproc() `process a specific `message` it must return` 0`. For the others `message ` values  that he doesn't process the developer have no knowledge of the value to return, only `DefWindowProc()` function has this knowledge, so `DefWindowProc()` must be called.

A usual thanks for reading so far :)

jjceresa:
@derselbst
> Do we even need to wrap DefWindowProc() inside fluid_wndproc() couldnt we just use it directly?

Please, ignore my previous comment on this point..

As we don't need to process `message `inside `fluid_win32_wndproc()`, wee don't need `fluid_win32_wndproc()`, so `DefWindowProc()` could be registered by  API `RegisterClass()`. This should work fine. Thanks for pointed this.
***
@carlo-bramini 
> Declare a global class only once for the current process.

Please ignore my previous comment on this point.
In fact, i think you mean that the intend is to allow  the application(i.e the process) the possibility of creation of more than one `dsound driver instance` (associated with more that one synth instance). Effectively in this case, the application need to take care calling  `RegisterClass()` only one one time (otherwise subsequent call will fail). In this case `we don't must call UnregisterClass() `at deletion driver time (as in my example).
```
  if(dev->hwnd ) 
  {
	  DestroyWindow(dev->hwnd);
//	  UnregisterClass(fluid_window_class_name,fluid_get_hinstance());
  }

```
Your fix is one possible solution implementation:
```
/* Class creation.
 * @creates the class for supporting window(s)
 */
static int
fluid_win32_create_class(HINSTANCE hInstance)
{
  WNDCLASS wc;

  /* Check previous creation of the global class */
  if (GetClassInfo(hInstance, fluid_class_name, &wc))
    return 1;

  ZeroMemory(&wc, sizeof(wc));

  wc.lpszClassName = fluid_class_name;
  wc.style         = CS_GLOBALCLASS;
  wc.lpfnWndProc   = fluid_wndproc;
  wc.hInstance     = hInstance;

  return RegisterClass(&myClass);
}
```
- An other solution could be to call RegisterClass(), UnregisterClass() in DllMain() at dll initialization time.
```
BOOL WINAPI DllMain(HANDLE hModule, DWORD ul_reason_for_call, LPVOID lpReserved)
{
    FLUID_LOG(FLUID_DBG, "DllMain");
    switch (ul_reason_for_call) 
    {
        case DLL_PROCESS_ATTACH:
        /* register the window class */
        {
             WNDCLASS myClass;
	     ZeroMemory(&myClass, sizeof(myClass));
		  myClass.lpszClassName = (LPSTR) fluid_class_name;
		  myClass.hInstance = hModule;
		  myClass.style = CS_GLOBALCLASS;
		  myClass.lpfnWndProc = DefWindowProc;
		  RegisterClass(&myClass);
	}
        break;
        case DLL_PROCESS_DETACH:
	    UnregisterClass(fluid_window_class_name,hModule);
        break;
     }
     return TRUE;
}
```
Your opinion ?

jjceresa:
Now dsound driver uses the desktop window  [5644ec3](https://github.com/FluidSynth/fluidsynth/commit/5644ec31194c06bb17b8f50f016fa07f83617723), So

- No need to register a new class window.
- No need to create a local window.
- No need of DllMain(). No need of fluid_dll.c.

carlo-bramini:
Hello,

1- well, I was curious to make that test, so I tried to compile FluidSynth as UNICODE library. Everything was fine except the device name used into dsound and midiin drivers. Not a big trouble, if someone will really have the need to do so in the future, we should just add a single line of code for converting from wide-char to multibyte utf8 before reading/writing the settings, but until we won't need such feature, I agree to leave it as it is.

2- no, placing RegisterClass() and other stuff from USER32 or GDI32 was wrong.

3- using desktop window handle is a typical hack that it is also used into other  programs. Usually, it works fine.

jjceresa:
> no, placing RegisterClass() and other stuff from USER32 or GDI32 into DllMain() was wrong.

I'am sorry but It is not true for RegisterClass(). The documentation API about WNDCLASS explains that when the class style is intended to be CS_GLOBALCLASS the class could be registered in DllMain().

> using desktop window handle is a typical hack that it is also used into other programs. Usually, it works fine.

Anyway, here it seems we are in a FluidSynth library that is not dedicated to a particular OS , and that library actually need no class at all. 

- Assuming the application make use of dsound driver (this driver can use  a default window like `desktop ` window).
Actually this works on Window XP. An there no reason it  should not work (imho).
If for any reason the application doesn't like this window it can use another window (using `fluid_win32_set_window()` in fluid_sound.c)
- Assuming the Windows application choose another driver (i.e asio,..) (and no use of dsound driver) , actually no others components  in FluidSynth library needs any window handle (imho).

Consequently actually ,there is no reason to create a class.

In the future, in the case of the FluidSynth library will became a windowed library that will need some initialization in `DllMain()` it should be appropriate to redefine the use of` fluid_dll.c `at appropriate time.
Guess, that i will not still alive when this time will occur :).

derselbst:
@carlo-bramini Is there anything left to do here or can I review/merge it?

jjceresa:
@carlo-bramini 

> Actually this works on Window XP. An there no reason it should not work (imho).

a)By the way, i have no other Window OS (i.e Win7,W10) available here. It should be nice if Carlo can do
some possible checks using the last commit.

> 1- well, I was curious to make that test, so I tried to compile FluidSynth as UNICODE library. Everything was fine....

>3- using desktop window handle is a typical hack that it is also used into other programs. Usually, it works fine.

Did you mean you have already done some check  as described in  point a ?
Many thanks for you help.

carlo-bramini:
I had seen that the desktop window handle is also used by other softwares like VLC, MPlayer and others and until now nobody complained about this choice, so I do not look bad sides on that. It is not clear to me how to call fluid_win32_set_window() since it seems to me that it is not exported, however it looks a bit thread-unsafe that thing. In my opinion, it would be better to just use desktop window and forget the custom handle. Basically, it is the same thing it has been done until now, but with a public handle.

OT:
*IF* I remember well, the creation of window classes into DllMain is subjected to some conditions, which includes the declaration of the DLL in a branch of the registry and other few things to be done for safety and if I'm not mistaken, this behavior is visible for the first time with the introduction of Windows XP and its theming support. In the past, you had to register the classes of the common controls by calling InitCommonControls() until version 5, while starting from version 6 those functions are just empty stubs included for source compatibility. Those controls were not registered directly in DllMain not because the MS engineers were not smart enough, but because that was the right way to do it. In my opinion, the new behavior had been a design choice that it had been made because, for not touching things into user32, commctrl also provides traditional controls with their new features and they had to be initialized in this way. So, I think that it would be better to avoid all this kind of stuff if it is not needed.

jjceresa:
> It is not clear to me how to call fluid_win32_set_window() since it seems to me that it is not exported, however it looks a bit thread-unsafe that thing.

Ok, i will remove the fluid_win32_set_window(). So in summary we will obtain a more clear commit that before the ancestor of this branch as this:
A commit that will contain the minimum necessary and sufficient stuff:
-  Dsound driver is now making of an already existing  window handle : the desktop window.
-  Removing of fluid_dll.c because it has becoming useless.

Please, did you do a simple check on Win 7 ?

carlo-bramini:
Here, on Windows 7 Home Premium, it works fine.

jjceresa:
> Here, on Windows 7 Home Premium, it works fine.

Thanks very much. I guess it should be true also on Win 10.

derselbst:
So we eliminated `fluid_dll.c`, nice. I think it would be good to have this in 1.1.9 due to the potential deadlock it could cause. Could it break any application using fluidsynth, i.e. could some app rely on that HINSTANCE? I guess not since it's private, but in case I'm missing something...

carlo-bramini:
Actually, I have no idea about that... since in the past it had been included to the exported functions, I think it should not be a big trouble to keep DllMain just for retrieving the hInstance.

jjceresa:
The `HINSTANCE `represent the module of the dll loaded in memory. It is for example useful when this module contains things  called ressources (bitmap, cursor , strings,...) intended to be used by the code inside the dll (or by the code of the application if the variable was declared public).
Effectively, since this variable was declared private, the application cannot use it. 

Actually the dll contains only PUBLIC fluidsynth API (and no ressources).

Another example. Any application that wants to unload a ddl need the hinstance of that dll, but usually any application that intends to to that rely to the load/unload API and does't need DllMain() to get the HINSTANCE.

Actually there is no code inside the dll  that reference this variable. Anyway we can keep the dllmain() in case of future Windows code that will need of this variable.
```
HINSTANCE fluid_hinstance = NULL;

BOOL WINAPI DllMain(HANDLE hModule, DWORD ul_reason_for_call, LPVOID lpReserved)
{
 fluid_hinstance = hModule; 
 return TRUE;
}
```

derselbst:
> Anyway we can keep the dllmain() in case of future Windows code that will need of this variable.

If we can avoid that "unused" code I would highly appreciate to do so. Writing code for any potential future usecase is usually not a good idea. So if you guys say (as I understand it) that everything should work as before, I'm happy to merge this to 1.1.9

jjceresa:
> If we can avoid that "unused" code I would highly appreciate to do so. Writing code for any potential future usecase is usually not a good idea.

I agree. Above i have explained why this code is now becoming "unused". 
In summary, with the last commit of this branch (i.e without  fluid_dll.c):
- everything should work as before.
- with a Windows code more clear to read, that actually works on my Win XP and on Carlo 'Windows 7 Home Premium.

So, please merge this to 1.1.9.

carlo-bramini:
Ouch, I arrived too late...
You forgot to remove the dllexport of fluid_get_hinstance() and fluid_set_hinstance() from include/misc.h.
Since you removed fluid_dll.c, you should remove them too.

jjceresa:
I forgot these because usually i use decoration in my own program in the signature of the function. When theses decorations are present, it helps to identify that the function is an API.
I am sorry Tom!

derselbst:
OH, I completely missed that as well. It's actually the reason that we cannot take this for 1.1.9, as symbols must not be removed from public API without a major version bump. Will restore the branch and revert it. Additionally I've reached out to ReinholdH to see if they are using `fluid_get|set_hinstance()` in their commercial product. Will let you know how to continue with this.

jjceresa:
> Additionally I've reached out to ReinholdH to see if they are using fluid_get|set_hinstance() 

If they do that, please ask him the reason why they do that !.

PR #271, opened 2017-11-10
fix library file names when building with MinGW and CMake

fabiangreffrath:
We have applied this patch to the mingw-w64-fluidsynth package [1] in
MSYS2 to recieve the same library file names when building with CMake
as we got when building with Autotools.

[1] https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-fluidsynth

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -268,6 +268,15 @@ elseif ( OS2 )
         VERSION ${LIB_VERSION_INFO}
         SOVERSION ${LIB_VERSION_CURRENT}
     )
+elseif ( WIN32 )
+  set_target_properties ( libfluidsynth
+    PROPERTIES
+      ARCHIVE_OUTPUT_NAME "fluidsynth"
+      PREFIX "lib"
+      OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
+      VERSION ${LIB_VERSION_INFO}
+      SOVERSION ${LIB_VERSION_CURRENT}
+    )
 else ( MACOSX_FRAMEWORK )
   set_target_properties ( libfluidsynth
     PROPERTIES


src/CMakeLists.txt
```diff
@@ -268,6 +268,15 @@ elseif ( OS2 )
         VERSION ${LIB_VERSION_INFO}
         SOVERSION ${LIB_VERSION_CURRENT}
     )
+elseif ( WIN32 )
+  set_target_properties ( libfluidsynth
+    PROPERTIES
+      ARCHIVE_OUTPUT_NAME "fluidsynth"
+      PREFIX "lib"
+      OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
```
derselbst: Do we even need to specify `OUTPUT_NAME`, wouldnt `ARCHIVE_OUTPUT_NAME` override it anyway?

fabiangreffrath: I think `ARCHIVE_OUTPUT_NAME` applies to both static and shared libraries and then for the shared library file name it is overridden by `OUTPUT_NAME` to add `LIB_VERSION_CURRENT`.

derselbst:
Ok, then lets take it. Thanks!

src/CMakeLists.txt
fabiangreffrath: If a static build should end up with a `fluidsynth.a` library despite the `PREFIX "lib"` stanza, it might be necessary to change `ARCHIVE_OUTPUT_NAME` to `"fluidsynth"`. Sorry, I am not sure about this anymore since the static library built is disabled by default...

fabiangreffrath: > Sorry, I am not sure about this anymore since the static library built is disabled by default...

Nope, just tested and IMHO everything looks good!

This is what gets installed with `-DBUILD_SHARED_LIBS=ON`:
```
lib/libfluidsynth.dll.a
bin/libfluidsynth-1.dll
```
and this is what gets installed with `-DBUILD_SHARED_LIBS=OFF`:
```
lib/libfluidsynth.a
```

PR #274, opened 2017-11-11
 introduce separate type for client IDs

derselbst:
typedef the unique client ID returned by `fluid_sequencer_register_client()` and `fluid_sequencer_register_fluidsynth()`

### DIFF
--- a/include/fluidsynth/event.h
+++ b/include/fluidsynth/event.h
@@ -67,8 +67,8 @@ FLUIDSYNTH_API fluid_event_t* new_fluid_event(void);
 FLUIDSYNTH_API void delete_fluid_event(fluid_event_t* evt);
 
 /* Initializing events */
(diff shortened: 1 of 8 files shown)


derselbst:
@mawe42 Please have a look over it, esp. the API doc of fluid_event.c

derselbst:
My motivation was to fix those DOCMEs of `fluid_event_set_dest()` and friends. So I needed to explain what that client id is and how/where to obtain it from. It's returned by two functions, but when I first used this part of the API (several years ago) only looking for a short that could be a client ID was hard for me. Introducing a dedicated type is what I considered to be a good way of clarifying this.

Also note that `fluid_sequencer_client_is_dest()` and `fluid_sequencer_get_client_name()` were incorrectly using `int` as client IDs which again points out the confusion of this mess.

PR #277, opened 2017-11-13
return FLUID_FAIL when sampletype is OGG but not compiled with LIBSND…

loki666:
…FILE_SUPPORT


derselbst:
Faster than I, thanks.

Issue #279, opened 2017-11-15
Reverb roomsize issue and level/width dependency


jjceresa:
Hello,
1) I have noticed one minor issue on upper limit `roomsize` setting for value above 1.
Actually the command handler propose an `upper limit` value `1.2`.
For value `above 1.07`, the output amplitude grow exponentially. This entering the user in high stress :).
An upper limit of `1.0 `seems sufficient as it produce a long `reverb time` (an can even ovoid  load speakers damages).
***
2) `Width `is a nice setting for value above 0 (`[0..3]`)  as we benefit of the `internal stereo effect `of the unit.
However for value in the upper range `[1..100]`, the` output amplitude` augment (rather linearly versus `width`).
Actually for `width value > 50`, the output amplitude can be clipped by the audio card . This make also unhappy musician.
Of course the user can compensate by changing the `level `setting to lower value .This is not easy as he must try multiples times before finding the right level.

We can also compensate automatically the output amplitude (i.e `rev->wet`) like this:

```
//#define scalelevelwidth 0.16
#define scalelevelwidth 0.2
		rev->wet = (rev->level ) /   (1 + rev->width * scalelevelwidth);
```

Actually, with a  `scalelevelwidth `of 0.2, for a given level setting (i.e `rev->level` ), the output amplitude `rev->wet` seems rather independent of  `width` setting (i.e`rev->width`).
***
I can easily  do a Pull Request if you want.

derselbst:
Regarding 1. I'm absolutely with you, I also ran into those kind of problems. Regarding 2. we'll have to see if this has a too big impact on how the reverb response to user settings. So yeah a PR would be nice.

jjceresa:
> Regarding 2. we'll have to see if this has a too big impact on how the reverb response to user settings. 

I was waiting for this question. 
Applying this compensation doesn't impact the `width `effect response but only the `scale `of the  width. 

- In others words for some user already `existing configurations settings` (particularly those with `width > 3`), it is possible that the user will need to augment the `Width `value and re-augment the `level `settings

- For future `configurations width` (those width > 3), it should be easier to build a `new width configuration` from a `previous width configuration`. Because the user should not be obliged to compensate  drastically and empirically on the level setting.

- For others old configurations (with width < 3), i had heard a negligible impact.

Anyway i will create a branch `reverb-settings-adjust` and 2 commits

- One commit with the roomsize limit.
- The next commit will include the `level-width compensation`.

[...]

PR #280, opened 2017-11-16
Reverb settings adjustements

jjceresa:
The PR corresponding to issue [Reverb roomsize issue and level/width dependency #279](https://github.com/FluidSynth/fluidsynth/issues/279)

### DIFF
--- a/src/rvoice/fluid_rev.c
+++ b/src/rvoice/fluid_rev.c
@@ -239,13 +239,13 @@ fluid_comb_getfeedback(fluid_comb* comb)
 #define numcombs 8
 #define numallpasses 4
 #define	fixedgain 0.015f
-/* scale_wet_width is a compensation weigth factor to get an output
-   amplitude (wet) rather  independent of width setting.
-    0: the output amplitude is fully dependant of width setting.
-    >0: the output amplitude is less dependant of width setting.
+/* scale_wet_width is a compensation weight factor to get an output
+   amplitude (wet) rather independent of the width setting.
+    0: the output amplitude is fully dependant on the width setting.
+   >0: the output amplitude is less dependant on the width setting.
    With a scale_wet_width of 0.2 the output amplitude is rather 
-   independent of width setting.(see fluid_revmodel_update()).
-*/
+   independent of width setting (see fluid_revmodel_update()).
+ */
 #define scale_wet_width 0.2f 
 #define scalewet 3.0f
 #define scaledamp 1.0f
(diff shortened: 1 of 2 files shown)


src/rvoice/fluid_rev.c
```diff
@@ -481,11 +481,31 @@ fluid_revmodel_processmix(fluid_revmodel_t* rev, fluid_real_t *in,
...
-
+  
+  /* The stereo amplitude equation  (wet1 and wet1 below) have a 
+  tendency to produce high amplitude with high width value ( 1 < width < 100).
+  This result in a unwanted output clipped by the audio card.
+  To avoid this dependency, we divide by (1 + rev->width * scalelevelwidth)
+  Actually, with a scalelevelwidthof 0.2, (regardless of level setting), 
+  the output amplitude (wet) seems rather independent of width setting */	
+  rev->wet = (rev->level * scalewet) / 
+             (1.0f + rev->width * scalelevelwidth);
```
derselbst: Looks like `rev->wet` is only use in the local scope here. It should be removed from `struct _fluid_revmodel_t` then.

src/rvoice/fluid_rev.c
```diff
@@ -481,11 +481,31 @@ fluid_revmodel_processmix(fluid_revmodel_t* rev, fluid_real_t *in,
...
 
 static void
+/* scalelevelwidth is compensation weigth factor to get output
+   amplitude rather  independent of width setting.
+   0: the output amplitude is fully dependant of width setting.
+   >0: the output amplitude is less dependant of width setting.
+   With a scalelevelwidthof of 0.2 the output amplitude is 
+   rather independent of width setting.
+*/
+#define scalelevelwidth 0.2f
```
derselbst: This is a pretty unsuitable location to place a define. Please place it some lines above (around 242) where the other `scale*` defines are located.

src/rvoice/fluid_rev.c
jjceresa: Yes. Since the arrival of rev->level,   rev->wet need to be removed.

derselbst:
I had a listen. I actually like it, it avoids that the reverb gets too noisy. However now it seems that when setting `width==10` it sounds pretty much the same like when setting `width==100` (while `level==1`). Not sure if a user actually relies on the effect of this range.

jjceresa:
> it avoids that the reverb gets too noisy

Yes, without the compensation factor (i.e `scale_wet_width = 0`), the `output amplitude` dependency of `Width` value was making the reverb very noisy (particularly for high value of Width).

> However now it seems that when setting width==10 it sounds pretty much the same like when setting width==100 (while level==1).

Yes , in general regardless of the compensation factor presence  , the perceived effect of whole Width range [0..100] could be roughly cut  into 2 ranges (ie `wRange1 [0..10]` and` wRange2 [10..100]`). `wRange1` values brings a more prominent perceived  `Width effect` than `wRange2` values (probably due to the logarithmic response of our ear). This remark is more obvious with the presence of compensation (i.e  `scale_wet_width > 0`). 

> Not sure if a user actually relies on the effect of this range.

Not sure too. The upper limit of the range feels excessive.

derselbst:
Ok, I'll merge this for master only.

jjceresa:
Many thanks for fixing grammar.

Issue #232, opened 2017-10-13
Compiling Fluidsynth  1.1.8 with MinGW: undefined reference toDirectSoundEnumerateA@8


YesterPlay80:
### FluidSynth version 
1.1.8

### I'm submitting a ... 
[x] bug report
[ ] feature request

### Current behavior
After generating the code with CMAKE and running mingw32-make, the following error occurs:

[ 94%] Linking C shared library libfluidsynth.dll
CMakeFiles\libfluidsynth.dir/objects.a(fluid_dsound.c.obj):fluid_dsound.c:(.text+0x35c): undefined reference to `DirectSoundEnumerateA@8'
CMakeFiles\libfluidsynth.dir/objects.a(fluid_dsound.c.obj):fluid_dsound.c:(.text+0x820): undefined reference to `DirectSoundEnumerateA@8'
collect2.exe: error: ld returned 1 exit status
src\CMakeFiles\libfluidsynth.dir\build.make:1041: recipe for target 'src/libfluidsynth.dll' failed
mingw32-make[2]: *** [src/libfluidsynth.dll] Error 1
CMakeFiles\Makefile2:86: recipe for target 'src/CMakeFiles/libfluidsynth.dir/all' failed
mingw32-make[1]: *** [src/CMakeFiles/libfluidsynth.dir/all] Error 2
Makefile:150: recipe for target 'all' failed
mingw32-make: *** [all] Error 2

### Related code

```
if available, insert short code snippets here
```

### Other information

```
MINGW with GCC 6.3.0
```

derselbst:
Hm, bad timing. Please attach cmakes full configure output as well as the full build log when building with

```
make VERBOSE=1
```

YesterPlay80:
Here's the configure output of CMAKE GUI before changing any options:
```
The C compiler identification is GNU 6.3.0
Check for working C compiler: D:/Entwicklung/MINGW/bin/gcc.exe
Check for working C compiler: D:/Entwicklung/MINGW/bin/gcc.exe -- works
Detecting C compiler ABI info
Detecting C compiler ABI info - done
Detecting C compile features
Detecting C compile features - done
Checking whether system has ANSI C header files
Looking for 8 include files dlfcn.h, ..., float.h
Looking for 8 include files dlfcn.h, ..., float.h - found
Performing Test memchrExists
Performing Test memchrExists - Success
Performing Test freeExists
Performing Test freeExists - Success
ANSI C header files - found
[...]

PR #281, opened 2017-11-17
do not set the library directory suffix when building with MinGW

fabiangreffrath:
This is another patch from MSYS2: Omit the library directory suffix
when building with MinGW for Windows.


derselbst:
Ok, seems useful. Just for interest: Is it meant to fix any "undefined reference to dsound" (#232) or the common "cannot find -lintl" error?

fabiangreffrath:
I don't think so. To be honest, #232 sounds more like a build environment issue to me. Fluidsynth builds and links perfectly fine with MinGW32 in MSYS2.

derselbst:
> Fluidsynth builds and links perfectly fine with MinGW32 in MSYS2.

Good to hear.

PR #282, opened 2017-11-19
Use vcpkg for Windows CI builds

mawe42:
Microsoft have published a nice tool for fetching and compiling dependencies:
https://github.com/Microsoft/vcpkg

These two commits switch the CI build to using vcpkg and remove the need to manually download very old binary builds of the glib libraries. Initial build takes much longer (+15 Minutes for each platform), but the dependencies are cached, so the next build will be just as fast as before. The cache is invalidated if the .appveyor.yml file changes.

To use the vcpkg provided libraries with cmake, you have to disable pkg-config and pass in the cmake toolchain file that vcpkg automatically generates. For cmake to find the libs, the NO_DEFAULT_PATH argument had to be removed in CMakeLists.txt. To be honest, I'm not sure why it was there in the first place. @derselbst do you know more about that? Will there be any bad side-effects if we also search in the default path if pkg-config is disabled?


derselbst:
Building without pkgconfig was actually only an attempt (a pre alpha feature if want to), that I manually cherry-picked from a contributor here on github (without thinking too much about NO_DEFAULT_PATH, dont see any bad side effects either). That was before I've seen how simple it is to download pkgconfig + glib and extract them all to build that on windows. The problem without pkgconfig is that it's only able to specify glib's location, not that of the other build dependencies.

I was thinking about compiling the windows build against libsndfile and portaudio. This would require pkgconfig. I however dont want to introduce additional cmake variables (more redundancy) to also being able to build those (and potentially all other) dependencies without using pkgconfig. Which is why I was thinking about completely removing that "build without pkgconfig mode" again.

On the other hand vcpkg is a really nice tool, I have to admit. Currently, I dont know...

mawe42:
> That was before I've seen how simple it is to download pkgconfig + glib and extract them all to build that on windows

It's only simple if you find pre-build binaries. The ones from ftp.gnome.org are *very* old. They work, but it feels just wrong to use a seven year old build of glib. Or do you know of an up-to-date binary build for glib and the other deps for Win32 and Win64?

derselbst:
> It's only simple if you find pre-build binaries.

Agreed.

> The ones from ftp.gnome.org are very old. They work, but it feels just wrong to use a seven year old build of glib.

Yes, and after releasing 1.1.7 I was surprised to get three bug reports within a very short time pointing out that it does not build against glib<2.30, and I was like :O

So we need a CI build against old glib, and this seems to be most useful on windows.

> Or do you know of an up-to-date binary build for glib and the other deps for Win32 and Win64?

libsndfile provides upstream .exe files, that however cannot be easily unpacked with 7z. AFAIK portaudio doesnt provide upstream binaries.

I think the core question is: _Do we want to have a "build-without-pkgconfig-mode"? And if yes, how much of it (only glib, a few more dependencies, or all)?_
Where my answer would be: Preferably not; but in case we do: have as less as possible (i.e. only glib) and build CI windows builds with both: with vcpkg (newest glib and without pkgconfig) and by manually downloading zip files (old glib, using pkgconfig and possibly more dependencies).

mawe42:
Hm... we could also use pure cmake and ditch pkg-config altogether. There's not too many dependencies and glib is probably the most complicated. Here's a module for glib: https://github.com/nemequ/gnome-cmake/blob/master/modules/FindGLib.cmake

And for pulseaudio:
https://github.com/tdfischer/pulseaudio-qt/blob/master/cmake/modules/FindPulseAudio.cmake

I'm sure we could write or adapt modules for the other deps as well. But I must admit I don't really know much about cmake, so I'm probably missing something here.

derselbst:
Phew, that would be a tough job if done properly: Introducing cmake modules that work on all platforms, using consistent naming conventions for cmake varialbes (lib_dirs, include_dirs, search_paths) so these can be set from outside when calling cmake. Might be a little too much for now.

Also: Consider building fluidsynth with a statically linked libsndfile: It would require to not only link against -lsndfile but also against -lvorbis and -lFLAC depending on how sndfile was built. Not sure if this is a realistic usecase, but it should be pretty easily possible with pkgconfig, not sure if/how this could be achieved using solely cmake.

mawe42:
Yes, after making that suggestion I read a few cmake modules from various projects. It's looks like a lot of work... and like quite a mess, actually. Lot's of `if ( WIN32 )` stuff there, and most use pkg-config when `(not WIN32)` anyway. So please forget my suggestions of ditching pkg-config.

And pkg-config is available on Windows anyway, it just doesn't work for vcpkg. But those .pc files look very simple, we could generate our own for Appveyor. Or just go with msys2 for the builds. Or both...

derselbst:
Never used MSYS2 myself, only know from fabiangreffrath that it seems to work quite well. If you can make it work in appveyor and it works with pkgconfig, I'd vote for that.

mawe42:
Ok, will try to switch to msys2 and create another pull request.

PR #283, opened 2017-11-19
LADSPA subsystem

mawe42:
I'm happy with the way the subsystem works now and I view it as feature complete (for now). It has been tested on Windows and Linux by multiple people (both using the command line and the API). 

gmodule is now only required if LADSPA is enabled. I also removed the libdl support, as that was only used by ladspa (which now uses gmodule).

A little more documentation could be added, an example program using the ladspa API functions would be nice. But those can all be added later, I think.


src/bindings/fluid_ladspa.c
```diff
@@ -25,93 +25,103 @@
...
+} fluid_ladspa_effect_t;
 
 struct _fluid_ladspa_fx_t
 {
     unsigned long sample_rate;
 
-    int audio_groups;
-    int effects_channels;
-    int audio_channels;
+    /* The buffer size for all audio buffers */
```
derselbst: "size for all audio buffers" is too unspecific, ie the unit is missing (bytes, audio frames?). Avoids tracing back it's usages to understand.

src/bindings/fluid_ladspa.c
```diff
@@ -265,6 +260,55 @@ void delete_fluid_ladspa_fx(fluid_ladspa_fx_t *fx)
...
+ * at least as large as the buffer size given to new_fluid_ladspa_fx.
+ *
+ * @param fx LADSPA fx instance
+ * @param prefix common name prefix for the created nodes
+ * @param num_buffers number of of buffers buffer array
+ * @param buffers array of pointers to buffers
+ * @return FLUID_OK on success, otherwise FLUID_FAILED
+ */
+int fluid_ladspa_add_host_ports(fluid_ladspa_fx_t *fx, const char *prefix,
+        int buffer_count, fluid_real_t *buffers[])
```
derselbst: `buffer_count` should be called `num_buffers`, or vice versa.

src/bindings/fluid_ladspa.c
```diff
@@ -265,6 +260,55 @@ void delete_fluid_ladspa_fx(fluid_ladspa_fx_t *fx)
...
+
+        if (new_fluid_ladspa_node(fx, name,
+                    FLUID_LADSPA_NODE_AUDIO | FLUID_LADSPA_NODE_HOST,
+                    buffers[i]) == NULL)
+        {
+            return FLUID_FAILED;
+        }
+    }
+
+    return FLUID_OK;
```
derselbst: Direct return without `LADSPA_API_RETURN()` intended?

src/bindings/fluid_ladspa.c
```diff
@@ -517,195 +554,200 @@ void fluid_ladspa_run(fluid_ladspa_fx_t *fx, fluid_real_t *left_buf[], fluid_rea
...
+    LADSPA_API_RETURN(fx, can_mix);
+}
+
+/**
+ * Set if the effect should replace everything in the output buffers (mix = 0, default)
+ * or add to the buffers with a fixed gain (mix = 1).
  *
  * @param fx LADSPA fx instance
+ * @param name the name of the effect
+ * @param mix (boolen) if to enable mix mode
```
derselbst: "boole**a**n whether to..."

src/bindings/fluid_ladspa.c
```diff
@@ -951,46 +956,40 @@ int fluid_ladspa_check(fluid_ladspa_fx_t *fx, char *err, int err_size)
     int i;
```
derselbst: This is a comment on the line above "@ return FLUID_OK on success, otherwise FLUID_FAILED" which I cannot comment on:

It it not clear that FLUID_FAILED is returned when a "problem" is detected.

src/bindings/fluid_ladspa.c
```diff
@@ -1000,87 +999,30 @@ int fluid_ladspa_check(fluid_ladspa_fx_t *fx, char *err, int err_size)
...
-
-static FLUID_INLINE void buffer_to_node(fluid_real_t *buffer, fluid_ladspa_node_t *node, int num_samples)
-{
-    int i;
-
-    /* If the node is not used by any plugin, then we don't need to fill it */
-    if (node->num_outputs > 0)
-    {
-        for (i = 0; i < num_samples; i++)
+        effect->active = 1;
```
derselbst: Better use TRUE.

src/bindings/fluid_ladspa.c
```diff
@@ -1000,87 +999,30 @@ int fluid_ladspa_check(fluid_ladspa_fx_t *fx, char *err, int err_size)
...
+static void deactivate_effect(fluid_ladspa_effect_t *effect)
 {
-    int i;
-
-    /* If the node has no inputs, then we don't need to copy it to the node */
-    if (node->num_inputs > 0)
+    if (effect->active)
     {
-        for (i = 0; i < num_samples; i++)
+        effect->active = 0;
```
derselbst: Better use FALSE ;)

src/bindings/fluid_ladspa.c
mawe42: Definitely not, well spotted. Seems like I haven't tested simultaneous access to the API from multiple threads, otherwise this should have tripped me up.

src/utils/fluid_sys.h
```diff
@@ -37,6 +37,7 @@
 #define _FLUID_SYS_H
 
 #include <glib.h>
+#include <gmodule.h>
```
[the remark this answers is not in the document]
loki666: I wouldn't bother, for system where glib is not available, fluid_sys.h is overridden anyway.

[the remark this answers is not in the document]
derselbst: > fluid_sys.h is overridden anyway.

Irrelevant. The assumption that `gmodule.h` is available when `glib.h` exists, is incorrect. Although usually shipped together, those are two different packages/libraries. That makes the #ifdef check mandatory.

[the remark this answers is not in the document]
mawe42: Need to check it anyway, because cmake only checks for gmodule if LADSPA is enabled.

src/bindings/fluid_ladspa.c
mawe42: Áck, fixed

src/bindings/fluid_ladspa.c
```diff
@@ -955,6 +995,13 @@ int fluid_ladspa_check(fluid_ladspa_fx_t *fx, char *err, int err_size)
     const char *str2;
     fluid_ladspa_effect_t *effect;
 
+    fluid_return_val_if_fail(fx != NULL, FLUID_FAILED);
+    fluid_return_val_if_fail(err_size >= 0, FLUID_FAILED);
+    if (err_size > 0 && err == NULL)
+    {
+        return FLUID_FAILED;
+    }
```
derselbst: I support a usecase for allowing `err` to be NULL. But perhaps better consider something like:

```
if (err != NULL)
{
    fluid_return_val_if_fail(size>0, FLUID_FAILED);
}
else;// err msgs not wanted, ignore err_size
```

Better make sure to guard snprintf calls with `err!=NULL`. And consider updating the doc to point out that `err` may be NULL ("if applicable" may also be understood as "err to be filled if a problem detected and unchanged if not").

mawe42: Done.

mawe42:
I think all requested changes have been addressed now. Any other comments on this?

derselbst:
> Any other comments on this?

Not yet. I wanted to actually test it myself before merging. Hopefully I
can do this at the weekend.

mawe42:
> Not yet. I wanted to actually test it myself before merging. Hopefully I
can do this at the weekend.

Ok. And no rush, I just wanted to make sure you've seen the last changes. Take your time.

derselbst:
Tested and works really nice. Great!

mawe42:
Thanks a lot for reviewing and testing!

PR #286, opened 2017-11-20
make dsound use the desktop window handle

derselbst:
basically #269 but without breaking API `fluid_get|set_hinstance()`


jjceresa:
Just a complement that shows that `fluid_set_hinstance()` API is particularly not useful from any application code. This will help to feel more confident (imho).
https://github.com/FluidSynth/fluidsynth/blob/c0cd1887db17b5d31aa07fe27814480850ae89ac/src/fluid_dll.c#L25
https://github.com/FluidSynth/fluidsynth/blob/c0cd1887db17b5d31aa07fe27814480850ae89ac/src/fluid_dll.c#L62-L68

The function stores the parameter `hinstance `in the HINSTANCE `fluid_hinstance` variable.
This works only when the function is called `the first time` (i.e subsequent calls will do nothing).

In the original code , the function is called the `first time` in `case DLL_PROCESS_ATTACH `(i.e when the application process is started (and the dll is loaded)). 

```
BOOL WINAPI DllMain(HANDLE hModule, DWORD ul_reason_for_call, LPVOID lpReserved)
{
   case DLL_PROCESS_ATTACH:
    ... 
    fluid_set_hinstance((void*) hModule);
    ...
}
```
So, if  later an application call `fluid_set_hinstance()` , the function does nothing !. Not a useful function.

All this stuff in` fluid_dll.c `(present since v1.0.9 (2004)  or before) probably haven't never maintained nor reviewed.

derselbst:
ReinholdH does not use HInstance. I agree on the uselessness of `fluid_set_hinstance()`. I will merge this PR as it is into master. Additionally I'll deprecate `fluid_*hinstance()` for 1.1.x and implement the minimalist get/set @jjceresa  suggested previously.

Issue #167, opened 2017-06-26
API cleanup


derselbst:
### issue collection
upon 2.0 release bump libfluidsynth's soversion and cleanup api:

- [x] remove [misspelled macro](https://github.com/FluidSynth/fluidsynth/blob/0edbe06621e8c9cc10837a754af14c645bf6ed60/fluidsynth/include/fluidsynth/event.h#L64)
- [x] remove deprecated LADCCA support
- [x] move `FLUID_INTERP_DEFAULT` and `FLUID_INTERP_HIGHEST` to `enum fluid_interp`
- [x] remove unused "synth.dump" setting?
- [x] make all the `fluid_settings_*()` return FLUID_FAILED or FLUID_OK
- [x] ~introduce a new `fluid_seq_event_type` that allows to use `fluid_synth_start()` and `fluid_synth_stop()` (to enable user of handling overlapping notes individually)~ too low level, the user can use callback events to achieve that
- [x] ~either provide a way of reading doubles from the synth or remove `WITH_FLOAT` and default to float?~
- [x] think about removing public function from the synth that manipulate synth->settings; we have the settings API for doing that (like `fluid_synth_setstr()` and friends)
- [x] ~remove `fluid_synth_get_program()` in favour of `fluid_synth_get_channel_info()`~
- [x] without deprecated `fluid_synth_get_channel_preset()` there is no way of retrieving the channel preset for `fluid_synth_start()`
- [x] `struct _fluid_mod_t` should be made private
- [x] remove `fluid_gen_set_default_values()`?
- [x] split `fluid_synth_set_reverb|chorus()` into atomic setters like done in 5f8c3d22e9a3a1f130613a71af98c09b538200dc
- [x] fluid_ladspa_fx_t can be made opaque
- [x] ~potentially misused fluid_voice_off() [here](https://github.com/FluidSynth/fluidsynth/blob/3babbed01e6fb3a264bb61d6e0be0d67e45cd853/src/synth/fluid_synth.c#L1910) and elsewhere (1.1.9!)~
- [x] ~remove unused "synth.effects-channels" setting?~
- [x] ladspa shell cmd handle functions signature were not updated after cb036d20233982567c3c6100999c3150f501b8f1
- [x] refactor fluid_handle_*() to return `FLUID_OK` or `FAILED`
- [x] ~check return value of delete_fluid_voice()~
[...]

PR #288, opened 2017-11-23
Some smaller fixes

mawe42:
Mainly removal of unused code, cleanup of comments. But also a missing memory allocation fix.


mawe42:
There are more unused functions that we could remove, but I'm a little unsure if we should do it:
```
fluid_event_get_time
fluid_hashtable_get_keys
fluid_hashtable_get_values
fluid_hashtable_iter_get_hash_table
fluid_hashtable_ref
fluid_hashtable_size
fluid_hashtable_steal_all
fluid_hz2ct
fluid_is_empty
fluid_player_count_tracks
fluid_player_get_track
fluid_settings_remove_option
fluid_track_count_events
fluid_track_first_event
fluid_track_get_name
fluid_tuning_set_key
fluid_tuning_set_name
fluid_voice_gen_value
```

derselbst:
fluid_event_get_time() could (should?) be used in
_fluid_seq_queue_insert_entry()

fluid_tuning_set_key() is the same as fluid_tuning_set_pitch(), so yes
remove it

perhaps look around where fluid_tuning_set_name() could be used suitably,
else remove that as well

fluid_player*() and fluid_track*() can be removed IMO, only using direct
member access here

fluid_is_empty() can be removed

fluid_voice_gen_val() is yet to be reviewed (#167)

The other functions may turn out to be useful, so let's better keep them
for now.

mawe42:
> fluid_event_get_time() could (should?) be used in _fluid_seq_queue_insert_entry()

Not sure. That function is not exposed via public API. Is that because nobody needs the time anyway?

> fluid_tuning_set_key() is the same as fluid_tuning_set_pitch(), so yes remove it

Done

> perhaps look around where fluid_tuning_set_name() could be used suitably, else remove that as well

fluid_tuning_set_name is now used in multiple places (and returns an error on memory allocation failure)

> fluid_player*() and fluid_track*() can be removed IMO, only using direct member access here

Done

> fluid_is_empty() can be removed

Done

> fluid_voice_gen_val() is yet to be reviewed (#167)

Ok, leaving it in that case.

> The other functions may turn out to be useful, so let's better keep them for now.

Yes, sounds good.

I've made two additional changes:
- fix the two "not guarded by if" warnings of gcc by correcting the indentation
- glib includes are now marked as system includes, so that gcc does not report those annoying bad function case warnings. They come from glib.h, there's nothing we can do about problems in that code. So no need for warnings here. With those two changes, FluidSynth compiles on my gcc without any warnings, which helps in spotting *real* problems.

derselbst:
> glib includes are now marked as system includes

Oh nice! Good idea.

mawe42:
About that system includes change for glib: it seems like clang doesn't honour it properly, it still throws those warnings. clang seems to handle macro expansion and system includes differently. Another way to get rid of those warnings would be a wrapper header for glib.h with some `#pragma GCC system_header` and `#pragma clang system_header`.

derselbst:
Indeed. So in case we are not missing any further flags, to me this looks like a bug in clang (tested with version 4.0.1, which is yours?).

> Another way to get rid of those warnings would be a wrapper header for glib.h with some #pragma GCC system_header and #pragma clang system_header.

I wouldnt go that far for now, hoping that this will be fixed some when in either glib or clang. Perhaps I'll try some different versions of clang and google around.

derselbst:
Looking into `fluid_voice_gen_value()` again, I think it was once thought to replace that `_GEN()` macro in `fluid_voice.h`. However I have no clue why Peter didnt get rid of that `_GEN` macro then. Introduced in ab066eb7450cc6214291c9be4bca8bb555c8b08d he put so much work into that commit to support absolute NRPN values but ultimately did not make use of this new function and without it I dont see how absolute NRPN should have ever worked correctly.

So I really think that we could just complete that by replacing that `_GEN` macro with that function. I dont see any change in behaviour (...except that now also absolute generator values via `fluid_synth_set_gen()` are supported).

mawe42:
> Indeed. So in case we are not missing any further flags, to me this looks like a bug in clang (tested with version 4.0.1, which is yours?).

Sorry, missed that comment. I've also tested with clang-4.0. But it seems like it affects previous versions as well. And it's not a bug, but a (missing?) feature: http://clang-developers.42468.n3.nabble.com/Warnings-from-macro-expansions-vs-system-headers-td4028109.html

mawe42:
> Looking into fluid_voice_gen_value() again, I think it was once thought to replace that _GEN() macro in fluid_voice.h. However I have no clue why Peter didnt get rid of that _GEN macro then. Introduced in ab066eb he put so much work into that commit to support absolute NRPN values but ultimately did not make use of this new function and without it I dont see how absolute NRPN should have ever worked correctly.

I think you are right... Looking through the code and the relevant commits, I think that feature has never worked. Maybe ask on the mailing list to see if anybody can remember anything? Otherwise I'd say get rid of all the absolute generator value handling code.

PR #290, opened 2017-11-25
Bsd socket fixes

carlo-bramini:


src/utils/fluid_sys.c
```diff
@@ -1132,13 +1001,18 @@ static fluid_thread_return_t fluid_server_socket_run (void *data)
...
+
 #ifdef HAVE_INETNTOP
+#ifdef IPV6_SUPPORT
+  char straddr[INET6_ADDRSTRLEN];
+#else
   char straddr[INET_ADDRSTRLEN];
-#endif
-#endif
+#endif /* HAVE_INETNTOP */
+#endif /* IPV6_SUPPORT */
```
derselbst:
```
#endif /* IPV6_SUPPORT */
#endif /* HAVE_INETNTOP */
```

carlo-bramini:
Argh... you are right, comments adjusted ;)

I tested it in CYGWIN for the POSIX part and in MINGW (32bit) for the Windows part.
Connecting with Teraterm was successful, shell visible at port 9800.
I have to say that, unlike CYGWIN, when cross compiling Windows version IPV6 is not detected by CMake, perhaps because it does not force _WIN32_WINNT to at least Windows Vista and inet_ntop() is not found. Unfortunately I don't know much of this build system. This happened also before this fix. At the moment, this is better because it does not break compatibility with older Windows, but fixing this limitation for making it working everywhere should not be difficult, if someone will want in the future.

derselbst:
Ok, also works fine under linux, will take it now. Regarding cross compilation and IPv6 I have no clue, sry.

PR #292, opened 2017-11-30
Refactor and cleanup of settings

mawe42:
I was going to propose a small addition to the overflow handling and wanted to add a new setting. While working on it, I noticed how badly the settings could do with a little refactor and cleanup, so I did that first.

First of all, I removed the callback and data parameters from all settings_register_* functions. Those parameters were only used in a few function calls in new_fluid_synth, never anywhere else. But each settings_register_* call had to set those to `NULL, NULL`. So I added a dedicated function to set a callback for a setting and removed the params from all other calls.

Then I reworked the callback handling, getting rid of the distinction between int, num and str callbacks. I thinks it's all a bit cleaner now, more understandable.

The last commit fixed a bug that was there before the refactor: when setint and setnum failed due to the numbers being out of range, the settings mutex was left in a locked state.

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -80,8 +80,8 @@ static int fluid_handle_voice_count (void* data, int ac, char **av,
 
 void fluid_shell_settings(fluid_settings_t* settings)
 {
-  fluid_settings_register_str(settings, "shell.prompt", "", 0, NULL, NULL);
-  fluid_settings_register_int(settings, "shell.port", 9800, 1, 65535, 0, NULL, NULL);
+  fluid_settings_register_str(settings, "shell.prompt", "", 0);
+  fluid_settings_register_int(settings, "shell.port", 9800, 1, 65535, 0);
 }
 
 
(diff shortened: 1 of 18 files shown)


derselbst:
While looking at it, I think that `/* Call under lock */`-mentality is actually dangerous, consider:

- Thread `A` wants to do a settings callback and has locked the fluid_settings API
- Within the callback function `A` attempts to lock the fluid_synth API
- But the synth API is already locked by Thread `B`, which for some reason wants to enter the settings API to do some read or whatever

--> deadlock.

This problem seemed to exist before this refactor and could only occur when calling `fluid_synth_update_sample_rate()` and `fluid_synth_set_sample_rate()` concurrently. But I also dont want to bother with that when doing some change to the synth in the future. So instead of pulling the new value from the settings, that new value should be given as param to the callback function and the callback should not be guarded by the lock. One could think about

`typedef void (*fluid_settings_callback_t)(fluid_settings_t *settings, const char *name, void *userdata, void* new_value);`


where `new_value` would be a pointer to that `val` variable allocated on the callers stack. The callee will know how to correctly cast it. Not nice due to even more error hiding casts. If there is another change to that, the old implementation (i.e. distinction between int, num and str callbacks) would not compile (or at least with warnings). So all in all I'm not sure if it's a good idea to place `fluid_settings_callback_t callback` into `fluid_setting_node_t`.

mawe42:
Yes, very good point about the deadlock. I noticed that as well in the original implementation, but thought it could be dealt with later on. But of course, the unified callback and pulling the values from the settings makes it actually impossible to remove that lock.

I will change it back to separate callbacks for each type.

PR #293, opened 2017-12-01
Settings cleanup and refactor

mawe42:
Second attempt at cleaning up the settings, especially the callback handling.

As in the previous PR I have added dedicated functions to register setting callbacks. This enables the removal of the callback and data parameters to all settings_register_* functions.

The callback function signatures have been cleaned up, returning void now, as the callbacks return value wasn't used anyway. And the error-hiding casts have been removed from the callback registering, making the cast from void* to fluid_synth_t* explicit in the callback handlers.

The handlers have been renamed to clarify their purpose and distinguish them from the other _update* functions in fluid_synth.h.

The settings mutex is now released before calling the callbacks.

A little unrelated, but still touches settings code: with_reverb and with_chorus on fluid_settings_t don't need to be atomics. Setting those via fluid_synth_reverb_on and fluid_synth_chorus_on is (or should be) protected via the api mutex anyway. fluid_synth_reverb_on wasn't protected but is now. It's calling fluid_synth_update_mixer, and that definitely needs to be protected by the API lock.

Please also note two backwards-incompatible changes:
- fluid_settings_setstr was allowed to be called for integer settings, evaluating "yes" and "no" to TRUE and FALSE. This hack has been removed. The fluidsynth binary and the fluid shell still accept "yes", "true", "t" (all case-insensitive) and anything else but 0 as TRUE, as they handle it themselves. A programmer using the FluidSynth API should be able to use _setint for an int setting and not rely on "yes" and "no".
- All fluid_settings_set* functions silently registered a new setting if the specified setting didn't exist. This hack has also been removed. The fluidsynth binary and the fluid shell check if a setting exists and return an error if it does not. The API should be just as strict and not hide errors like that.

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -80,8 +80,8 @@ static int fluid_handle_voice_count (void* data, int ac, char **av,
 
 void fluid_shell_settings(fluid_settings_t* settings)
 {
-  fluid_settings_register_str(settings, "shell.prompt", "", 0, NULL, NULL);
-  fluid_settings_register_int(settings, "shell.port", 9800, 1, 65535, 0, NULL, NULL);
+  fluid_settings_register_str(settings, "shell.prompt", "", 0);
+  fluid_settings_register_int(settings, "shell.port", 9800, 1, 65535, 0);
 }
 
 
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -1363,9 +1363,9 @@ fluid_handle_set(void* data, int ac, char** av, fluid_ostream_t out)
       if (fluid_settings_get_hints (handler->synth->settings, av[0], &hints) == FLUID_OK
           && hints & FLUID_HINT_TOGGLED)
       {
-          if (FLUID_STRCMP (av[1], "yes") == 0 || FLUID_STRCMP (av[1], "True") == 0
-              || FLUID_STRCMP (av[1], "TRUE") == 0 || FLUID_STRCMP (av[1], "true") == 0
-              || FLUID_STRCMP (av[1], "T") == 0)
+          if (FLUID_STRCASECMP (av[1], "yes") == 0
+              || FLUID_STRCASECMP (av[1], "true") == 0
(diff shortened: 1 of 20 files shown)


derselbst:
Looks good, thanks!

mawe42:
Thanks for the review! I'm now moving on to my real target: adding a feature to the overflow handling. But before I do that, there's one more thing that is bugging me...

The fluid_overflow_prio_t struct defines the priorities as `fluid_real_t`. And the code that uses those values also works with that type. I'm just wondering why it has been done like that. I can see the need to decide if audio data and all associated calculations use float/double for performance or compatibility reasons. But here I see no reason to use double, nor do I see a reason why it should be a compile-time decision.

So I'm thinking about changing those to use float. And related to that: what do you think about adding type specific `fluid_settings_getnum` functions to avoid the annoying "create double var, read setting, assign to float or fluid_real_t var` dance? I would propose:
```
fluid_settings_getfloat(fluid_settings_t *settings, const char *name, float *val);
fluid_settings_getreal(fluid_settings_t *settings, const char *name, fluid_real_t *val);
```
Those functions would call fluid_settings_getnum and cast the value to float or fluid_real_t before returning.

And maybe also:
```
fluid_settings_getdouble(fluid_settings_t *settings, const char *name, double *val);
```
as an alias to `fluid_settings_getnum`, to make it explicit that getnum is double.

mawe42:
Thinking about it more, maybe it would be clearer to just define the following:
```
fluid_settings_getnum_float(fluid_settings_t *settings, const char *name, float *val);
fluid_settings_getnum_real(fluid_settings_t *settings, const char *name, fluid_real_t *val);
```

derselbst:
Regarding `fluid_overflow_prio_t` I at least dont see a reason to use `fluid_real_t` here, `float` should be absolutely sufficient.

`fluid_settings_getnum_float()` and `fluid_settings_getnum_real()` seem like a good way to go. For now I would however refrain from introducing `fluid_settings_getnum_double()`.

mawe42:
Ok, thanks. I'll incorporate those changes into the PR with the new overflow feature.

PR #294, opened 2017-12-02
New feature for overflow calculation: important channels

mawe42:
The FluidSynth overflow priority calculation determines which voice to kill in an overflow situation, i.e. when the polyphony limit has been exceeded and FluidSynth tries to kill an active voice in order to service a note-on event.

This PR adds a new feature to the overflow priority calculation, making it possible to mark a number of MIDI channels as "important" and add a user-defined value to their score. The use-case for this is to protect long running notes on separate MIDI channels from being killed by the overflow handling.

In my particular use-case there are a number of MIDI channels which play "drone" notes (i.e. notes that get started once and always sound white the instrument is playing) and another MIDI channels that plays a melody. I want those drone notes protected from killing, as it is very noticeable if one of them goes away.

Please consider this as an RFC. It works very well for my use-case, but maybe there is better approach for this?

The other three commits are just cleanup/refactor and documentation.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -128,6 +128,30 @@ https://stackoverflow.com/a/6251757
             </desc>
         </setting>
         <setting>
+            <name>overflow.important</name>
+            <type>num</type>
+            <def>5000</def>
+            <min>-50000</min>
+            <max>50000</max>
+            <desc>
+                This score is added to voices on channels marked with the
+                synth.overflow.important-channels setting.
+            </desc>
+        </setting>
+        <setting>
+            <name>overflow.important-channels</name>
+            <type>str</type>
+            <def>""</def>
+            <desc>
+                This setting is a comma-separated list of MIDI channel numbers that should
+                be treated as "important" by the overflow calculation, adding the score
+                set by synth.overflow.important to each voice on those channels. It can
+                be used to make voices on particular MIDI channels
+                less likely (synth.overflow.important &gt; 0) or more likely
+                (synth.overflow.important &lt; 0) to be killed in an overflow situation.
+            </desc>
+        </setting>
+        <setting>
             <name>overflow.percussion</name>
             <type>num</type>
             <def>4000</def>
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -115,6 +115,71 @@ https://stackoverflow.com/a/6251757
                 Sets the minimum note duration in milliseconds. This ensures that really short duration note events, such as percussion notes, have a better chance of sounding as intended. Set to 0 to disable this feature.</desc>
         </setting>
         <setting>
+            <name>overflow.age</name>
+            <type>num</type>
+            <def>1000</def>
+            <min>-10000</min>
+            <max>10000</max>
+            <desc>
+                This score is divided by the number of seconds this voice has been
+                active and is added to the overflow priority. It is usually a positive
+                value and gives voices which have just been started a higher priority,
+                making them less likely to be killed in an overflow situation.
+            </desc>
+        </setting>
+        <setting>
+            <name>overflow.percussion</name>
+            <type>num</type>
+            <def>4000</def>
+            <min>-10000</min>
+            <max>10000</max>
+            <desc>
+                Sets the overflow priority score added to voices on a percussion
+                channel. This is usually a positive score, to give percussion voices
+                a higher priority and less chance of being killed in an overflow
+                situation.
+            </desc>
+        </setting>
+        <setting>
+            <name>overflow.released</name>
+            <type>num</type>
+            <def>-2000</def>
+            <min>-10000</min>
+            <max>10000</max>
+            <desc>
+                Sets the overflow priority score added to voices that have already
+                received a note-off event. This is usually a negative score, to give released
+                voices a lower priority so that they are killed first in an overflow
+                situation.
+            </desc>
+        </setting>
+        <setting>
+            <name>overflow.sustained</name>
+            <type>num</type>
+            <def>-1000</def>
+            <min>-10000</min>
+            <max>10000</max>
+            <desc>
+                Sets the overflow priority score added to voices that are currently
+                sustained. With the default value, sustained voices are considered less
+                important and are more likely to be killed in an overflow situation.
+            </desc>
+        </setting>
+        <setting>
+            <name>overflow.volume</name>
+            <type>num</type>
+            <def>500</def>
+            <min>-10000</min>
+            <max>10000</max>
+            <desc>
+                Sets the overflow priority score added to voices based on their current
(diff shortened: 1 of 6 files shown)


mawe42:
> Please consider this as an RFC. It works very well for my use-case, but maybe there is better approach for this?

A possible different implementation would be to specify the channel priority directly. So instead of simply marking channels as important and adding a fixed value, give the user the ability to set priorities for each channel in a comma separated list:
```
fluidsynth -o synth.overflow.channels=0,0,0,-5000,30000,20000,10000
```
This would give channels 1-3 unchanged priority, channel 4 a penalty of -5000 (making it more likely to be killed), channels 5,6 and 7 an individual priority boost.

Thinking about it, this is much more flexible and seems preferable over the proposed implementation. What do you think? And should I take this to the mailing-list for further discussion?

derselbst:
Honestly speaking I dont like any attempt of overflow / priority voice killing algorithm stuff. If one runs out of polyphony, one simply cannot play a multi voice MIDI. End of the story.

But as we already have such a functionality, let me try to be more constructive:


> So instead of simply marking channels as important and adding a fixed value, give the user the ability to set priorities for each channel in a comma separated list:

Yes it is more flexible, but what if the user wants to have channel 16 the only important channel?

`-o synth.overflow.channels=0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,30000`

is not very readable and prone to errors. So, if I had to choose between your first comment and your second, I'd go with the first one.

But please consider: I think that this kind of overflow handling feature is too low-level, i.e. too fluidsynth specific. [The soft synth of the Nintendo64 SDK](https://level42.ca/projects/ultra64/Documentation/man/pro-man/pro17/index17.4.html) handled those channel priorities via General Purpose MIDI CC 0x10. Putting that into the context of fluidsynth it should actually be very straight forward to implement that priority handling via the modulator functionality. By using the new API to manip def mods a user application is not even bound to using any specific MIDI CC. And even the scale can be adjusted to the users needs. Because in the end, the MIDI file being played should now best which channels carry the main melody and which ones are less important.

mawe42:
> Honestly speaking I dont like any attempt of overflow / priority voice killing algorithm stuff. If one runs out of polyphony, one simply cannot play a multi voice MIDI. End of the story.

You forget that FluidSynth is not only used for MIDI playback, but also (mainly?) as a real-time software synth. An upper bound on polyphony and voice killing is a must-have feature for any such system. For example, the instrument that I'm building uses FluidSynth as a real-time synth. I'm running on a small ARM core and can only support a limited amount of polyphony. If a player presses a lot of keys in quick succession and has a sound with a long release time loaded, voice killing prevents the system from going to it's knees. And I know I'm not the only one using FS in this manner.

> By using the new API to manip def mods a user application is not even bound to using any specific MIDI CC. And even the scale can be adjusted to the users needs. Because in the end, the MIDI file being played should know best which channels carry the main melody and which ones are less important.

I wasn't thinking about MIDI file playing when implementing this feature. I think for that use-case, the available settings are working quite well. The "important channels" feature is for real-time use, like explained above. So there's no MIDI file "knowing best"... more like a "instrument developer knowing best" :-)

derselbst:
> You forget that FluidSynth is not only used for MIDI playback, but also (mainly?) as a real-time software synth.

Granted.

Anyway I think that modulator approach is also viable for real-time usage. An "instrument developer" could just start fluidsynth with a config that executes `cc 14 16 127` to make channel 14 a very important channel right from the beginning (ofc. requires somehow a priority modulator being set as def mod). Because if you need to change the prio on-the-fly you would have to fiddle with the shell anyway. Or am I missing something?

mawe42:
> Anyway I think that modulator approach is also viable for real-time usage.

Yes, definitely. The question is how usable this modulator approach is in real-life. Not for my use-case... I don't care if I need to create a custom modulator and then send CC messages, or if I set a setting. Just a few more lines of code for the modulator approach. The setting is real-time capable, btw. So no need to fiddle with the shell, I can simply call `fluid_settings_setstr` again during runtime if the channel setup changes.

But if the motivation for the modulator approach is to have the setting more accessible, I'm not sure it fits the purpose. After all, this modulator would be non-standard, so there won't be any SoundFont editors that can create such a modulator. And the CC this modulator reacts to would also be non-standard. So how could a MIDI file creator know which CC to use to affect channel priority? I don't think there is an official MIDI spec for this kind of thing, is there?

mawe42:
More more comment: this feature (and all the other synth.overflow.* settings) is definitely a "power-user" feature. People who use FluidSynth to play MIDI files on a standard desktop computer or Notebook will never have to touch or even know about this feature. It really is only useful for people on very resource constrained systems or with very special real-time needs (or both... like me :-).

derselbst:
> The question is how usable this modulator approach is in real-life.

You're right, it would be a complete non-standard custom approach, that would probably require some hybrid of a composer and a developer to actually make use of that. And even then it would not be of much use for MIDI files.

...I just thought it would fit in that modulator paradigm and would be esp. useful if you need different priorities for different channels. But if you only need a single score for important channels your suggestion to go with a setting also seems viable.

mawe42:
I think a modulator approach would be great if the overflow priorities need to be changed frequently and in response to real-time midi events. But I don't think that is the case. They really are *settings*, which are set up at synth start, maybe changed occasionally but very infrequently. And most probably not during playback. So I think using the settings fit quite well.

src/synth/fluid_synth.c
```diff
@@ -5289,3 +5308,79 @@ fluid_ladspa_fx_t *fluid_synth_get_ladspa_fx(fluid_synth_t *synth)
...
+            FLUID_LOG(FLUID_ERR, "Out of memory");
+            goto exit;
+        }
+        scores->num_important_channels = synth->midi_channels;
+    }
+
+    for (i = 0; i < scores->num_important_channels; i++)
+    {
+        scores->important_channels[i] = FALSE;
+    }
```
derselbst: Perhaps replace this loop by FLUID_MEMSET?

doc/fluidsettings.xml
```diff
@@ -114,6 +114,95 @@ https://stackoverflow.com/a/6251757
...
+            <name>overflow.important-channels</name>
+            <type>str</type>
+            <def>""</def>
+            <desc>
+                This setting is a comma-separated list of MIDI channel numbers that should
+                be treated as "important" by the overflow calculation, adding the score
+                set by synth.overflow.important to each voice on those channels. It can
+                be used to make voices on particular MIDI channels
+                less likely (synth.overflow.important &gt; 0) or more likely
+                (synth.overflow.important &lt; 0) to be killed in an overflow situation.
```
derselbst: You could mention here that channel numbers are one-based.

mawe42:
I've added a comment about channels being 1-based and changed the loop to memset.

PR #295, opened 2017-12-02
Allow overwriting the compile-time flags for gcc/clang.

mawe42:
As discussed in http://lists.nongnu.org/archive/html/fluid-dev/2017-11/msg00077.html, this PR adds the ability to override the C flags.


derselbst:
Ok, do you need this for 1.1.9?

mawe42:
I'm working on master with my own patches added, so no rush from my side.

PR #297, opened 2017-12-03
YAC

### LINKED ISSUE #186: make FLUID_SEQ_LASTEVENT private
upon next major release, at least add a comment (or invent some hacky preprocessor macro) that makes clear that `FLUID_SEQ_LASTEVENT` can change at any time. else we are not able to add new event types to the synth without breaking ABI.

...same for all other public enums

derselbst:
Yet Another Cleanup:

* make sure doxygen generates ladspa API
* remove deprecated tuning functions
* remove `fluid_synth_get_channel_info()` (opinions?)
* general cleanup of public API
* use `fluid_voice_gen_value()` rather than _GEN macro
* bump SOVERSION

I actually think absolute generator values could turn out to be useful, so I want to give that a try.

Please also give me your opinion about `GEN_LAST`, `LAST_LOG_LEVEL`, `FLUID_SEQ_LASTEVENT` and `FLUID_MIDI_ROUTER_RULE_COUNT`: I want them to be removed from the public API. Is it enough to hide them in the doxygen docs or should they also be removed from the public header itself to avoid any downstream lib from using them?

Fixes #186.


src/synth/fluid_voice.c
```diff
@@ -663,22 +676,24 @@ calculate_hold_decay_buffers(fluid_voice_t* voice, int gen_base,
...
 {
-  double q_dB;
-  fluid_real_t x;
-  fluid_real_t y;
-  unsigned int count, z;
   // Alternate attenuation scale used by EMU10K1 cards when setting the attenuation at the preset or instrument level within the SoundFont bank.
   static const float ALT_ATTENUATION_SCALE = 0.4f;
+  unsigned int count, z;
+  fluid_real_t q_dB;
+  fluid_real_t x = fluid_voice_gen_value(voice, gen);
```
mawe42: I understand why you want to have this outside of the switch, but it does mean that on some generators this the gen value is calculated twice (on coarse/fine generators). Probably doesn't have a huge performance impact... but still, is it really worth it just to save a line of code in each `case`?

mawe42:
> Please also give me your opinion about GEN_LAST, LAST_LOG_LEVEL, FLUID_SEQ_LASTEVENT and FLUID_MIDI_ROUTER_RULE_COUNT: I want them to be removed from the public API. Is it enough to hide them in the doxygen docs or should they also be removed from the public header itself to avoid any downstream lib from using them?

Well, as it is documented that they are not part of the API/ABI stability guarantee, I think it's ok to leave them in the public header.

src/synth/fluid_voice.c
derselbst: > but it does mean that on some generators the gen value is calculated twice (on coarse/fine generators).

Yes. The generators in question should not be triggered too often and if they do, the required generator values are most likely to reside in L1 cache. So I thought that way it would be more readable as the majority of the generators use `x`. But if you insist I change it back.

mawe42: I don't (insist). It's probably fine.

mawe42:
You could also remove the comment lines about fluid_synth_set_gen2 and document the absolute NRPN values.

derselbst:
> You could also remove the comment lines about fluid_synth_set_gen2

I'm not yet quite sure whether this one was an unnecessary API change, i.e. we could also leave `fluid_synth_set_gen()` calling `fluid_synth_set_gen2()` with the last two params set to FALSE... idk.

> document the absolute NRPN values

Not quite sure what you mean here.

mawe42:
>> document the absolute NRPN values

> Not quite sure what you mean here.

I meant documenting why those non-standard absolute NRPN generator values are there. What they could be used for. I must admit I don't really know a use-case for that feature.

derselbst:
The only use-case I can currently think of is if one wants to manually implement the modulator functionality. This could be useful if the default value of a generator doesnt fit your needs, or if you want to use a custom math-function for scaling MIDI CCs to dest generator values. But this is also more like a hack.

I'm really torn between removing that functionality and keeping it. I would ask on the mailing list, but since this feature never worked and we never got a bug request, I dont think this will lead to any enlightenment.

derselbst:
I will merge this now. Discussion about the future of `fluid_synth_set_gen2()` has been triggered and will be tracked in #167. I still think this feature is useful and lightweight enough to keep it's implementation. As a compromise (in case no one comes up with a use-case for this function) I suggest marking it deprecated in 1.1.9 so that we can safely remove it from the public API next major release. Once we removed this API dependency, eventually doing any actual cleanup of this functionality can be done trivially in the future.

mawe42:
Ok, sounds good!

PR #298, opened 2017-12-03
Simpler implementation of WINMIDI driver.

carlo-bramini:
Since this is a driver that it will run just and only on Windows, there is a much simpler way to do it, without using synchronization objects and atomic variables.

### DIFF
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -26,8 +26,7 @@
  * NOTE: Unfortunately midiInAddBuffer(), for SYSEX data, should not be called
  * from within the MIDI input callback, despite many examples contrary to that
  * on the Internet.  Some MIDI devices will deadlock.  Therefore we add MIDIHDR
- * pointers to a queue and re-add them in a separate thread, using a mutex and
- * conditional to wake up the thread.  Lame-o API! :(
+ * pointers to a queue and re-add them in a separate thread.  Lame-o API! :(
  */
 
 #include "fluidsynth_priv.h"
@@ -71,8 +70,7 @@ void delete_fluid_winmidi_driver(fluid_midi_driver_t* p);
 
 void CALLBACK fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
 				    DWORD_PTR msg, DWORD_PTR extra);
-static char* fluid_winmidi_input_error(int no);
-int fluid_winmidi_driver_status(fluid_midi_driver_t* p);
+static char* fluid_winmidi_input_error(MMRESULT no);
 
 
 void fluid_winmidi_midi_driver_settings(fluid_settings_t* settings)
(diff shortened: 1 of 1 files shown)


derselbst:
I'm not familiar with WinMIDI. Please test it and if everything works as before I'll take it.

@jjceresa Could you review it?

jjceresa:
> Since this is a driver that it will run just and only on Windows, there is a much simpler way to do it, without using synchronization objects and atomic variables.

1)Yes, it is a better way that it is only dependent of Window API. And It works has before when sysex message are received..

2)Note, i use a simpler way that doesn't need of a thread to add a buffer in the driver, like this:
```
  case MIM_LONGDATA:    /* SYSEX data */
     pMidiHdr = (LPMIDIHDR)dwParam1;
    data = (unsigned char *)(pMidiHdr->lpData);

    /* We only process complete SYSEX messages (discard those that are too small or too large) */
    if (pMidiHdr->dwBytesRecorded > 2 && data[0] == 0xF0
        && data[pMidiHdr->dwBytesRecorded - 1] == 0xF7)
    {
      fluid_midi_event_set_sysex (&event, pMidiHdr->lpData + 1,
                                  pMidiHdr->dwBytesRecorded - 2, FALSE);
      (*dev->driver.handler)(dev->driver.data, &event);
    }
    /* Now the synth doesn't use pMidiHdr->lpData anymore. So we can safely add the buffer
      in the driver */
    midiInAddBuffer(dev->hmidiin, pMidiHdr, sizeof(MIDIHDR));
    break;
```
This works fine also as far the buffer  `pMidiHdr->lpData` is no longer in use by the synth instance when calling midiInAddBuffer().  Calling `dev->driver.handler` is synchronous, which means that   `pMidiHdr->lpData `is no more in use by the synth instance. I use this since many years without any issues even when Sysex messages are received.

There is no reason not to use midiInAddBuffer() in the callback even at interrupt time. I disagree with the NOTE at the top of the file:`NOTE: Unfortunately midiInAddBuffer(), for SYSEX data, should not....`
****
So as Carlo' solution  is better than before and works as before it can be merged. 

some minor cleanup to do:
1) make use of variable `res `only and  removing variable `err`
2) removing unused `fluid_winmidi_driver_status() `function.
3) Update the NOTE comments (removing `using a mutex and....`)

carlo-bramini:
I had read that sentence before doing the change and I had no problems to put that call into MIM_LONGDATA but, evidently, the original developer also tried to put midiInAddBuffer() in the callback, but it did not work (perhaps I have an idea).
So, I left the same behaviour of previous code because no detailed information were given. If I can say my opinion, until a bit more details are given (test hardware, OS version, driver version, external connections, etc), it would be better to leave it as it is.

derselbst:
> If I can say my opinion, until a bit more details are given, it would be better to leave it as it is.

It's 8 years since this change, I'm not sure if Element can even remember all the details of that. I dont want to bother him with that either, the `PostThreadMessage` solution is probably fine and I would merge it if that's all.

jjceresa:
> the PostThreadMessage solution is probably fine and I would merge it if that's all.

Yes , it works so it can be merged.

PR #299, opened 2017-12-06
add support for lyrics and text midi events

### LINKED ISSUE #111: Add hooks for Karaoke
Karaoke files contain Lyric or Text meta-events in addition to the other MIDI events. Playing such files in a Karaoke system requires that all of the lyrics are available to the karaoke application at the start of the song and the Lyric/Text events need to be captured at the time they are "played". I suggest two changes to make this possible:  

a) Add an event handler that is called on completion of file loading:  

> typedef int (*handle_onload_func_t)(void* data, fluid_player_t* player);  

> int fluid_player_set_onload_callback(fluid_player_t* player,   

>
>> handle_onload_func_t handler, void* handler_data);  

This will expose the player _after_ it has loaded a file so that its parsed MIDI file data can be examined (see example below).  

b) Add code to fluid_midi.c to handle Text/Lyric events and send them from the sequencer to the MIDI synthesizer. NOTE: the default synthesizer fluid_synth_handle_midi_event() is NOT changed so will continue to ignore such events. Only if a custom event handler is installed will these events be seen (see example below).  

I have working code for this (including appropriate garbage collection) and can send patch files, against 1.1.6.  

An example showing how this could work follows. It is based on file_player.c from the API documentation  

```
#include <fluidsynth.h>  
#include <fluid_midi.h>


/**  
  * This MIDI event callback filters out the TEXT and LYRIC events 
  * and passes the rest to the default event handler. 
  * Here we just print the text of the event, more 
  * complex handling can be done */ 
int event_callback(void *data, fluid_midi_event_t *event) {  
fluid_synth_t* synth = (fluid_synth_t*) data;  
int type = fluid_midi_event_get_type(event);  
int chan = fluid_midi_event_get_channel(event);  

switch(type) {  
case MIDI_TEXT:
printf("Callback: Playing text event %s (length %d)\n", event-&gt;paramptr, event-&gt;param1);  
return FLUID_OK;  

case MIDI_LYRIC:  
printf("Callback: Playing lyric event %d %s\n", event-&gt;param1, event-&gt;paramptr);  
return FLUID_OK;  
}  
return fluid_synth_handle_midi_event( data, event);  
}  


/**  
  * This is called whenever new data is loaded, such as a new file. 
  * Here we extract the TEXT and LYRIC events and just print them 
  * to stdout. They could e.g. be saved and displayed in a GUI 
  * as the events are received by the event callback. */ 
int onload_callback(void *data, fluid_player_t *player) {  
printf("Load callback, tracks %d \n", player-&gt;ntracks);  
int n;  
for (n = 0; n &lt; player-&gt;ntracks; n++) {  

fluid_track_t *track = player-&gt;track[n]; fluid_settings_setint(settings, "synth.polyphony", 64);  
printf("Track %d\n", n);  
fluid_midi_event_t *event = fluid_track_first_event(track);  
while (event != NULL) {  

switch (event-&gt;type) {  
case MIDI_TEXT:  
case MIDI_LYRIC:  
printf("Loaded event %s\n", event-&gt;paramptr);  


}  
event = fluid_track_next_event(track);  

 }  
 }  
return FLUID_OK;  
}  

int main(int argc, char** argv)  
{  
int i;  
fluid_settings_t* settings;  
fluid_synth_t* synth;  
fluid_player_t* player;  
fluid_audio_driver_t* adriver;  
fluid_settings_setstr(settings, "audio.driver", "alsa");  
player = new_fluid_player(synth);  
/* Set the MIDI event callback to our own function rather than the system default */  
fluid_player_set_playback_callback(player, event_callback, synth);  

/* Add an onload callback so we can get information from new data before it plays */  
fluid_player_set_onload_callback(player, onload_callback, NULL);

adriver = new_fluid_audio_driver(settings, synth);  
/* process command line arguments */  
for (i = 1; i &lt; argc; i++) {  

if (fluid_is_soundfont(argv[i])) {  

fluid_synth_sfload(synth, argv[1], 1);  

 } else {  
 fluid_player_add(player, argv[i]);  
 }  
}  
/* play the midi files, if any */  
fluid_player_play(player);  
/* wait for playback termination */  
fluid_player_join(player);  
/* cleanup */  
delete_fluid_audio_driver(adriver);  
delete_fluid_player(player);  
delete_fluid_synth(synth);  
delete_fluid_settings(settings);  
return 0;  
}  
```

Reported by: jannewmarch

Original Ticket: [fluidsynth/tickets/112](https://sourceforge.net/p/fluidsynth/tickets/112)

derselbst:
* partly apply patch of #111 
* allow MIDI_SET_TEMPO to be passed to `_fluid_player_t::playback_callback`

closes #111


mawe42:
Have you tried if this works as intended? As I understand the MIDI spec, the distinction between TEXT and LYRIC is important for Karaoke. It seems to me that TEXT contains the whole text line, LYRIC then contains only the syllables that are meant to be sung for the currently sounding note.

derselbst:
I have not tested it as I really dont have any use-case or example app I
could test this, but I'm convinced it will work. The thread opener of #111
argued that "that all of the lyrics are available to the karaoke
application at the start of the song and the Lyric/Text events need to be
captured at the time they are played". The way he/she did that was by
introducing yet another onload_callback, exposing struct fluid_player_t to
public API and using a hacky way to iterate over all midi events of all
fluid_track_t. This is no option for us. If the application really needs to
know all text and lyrics events one could use fluid_player_seek(). That way
the app would receive all events via the playback_callback.

derselbst:
I would merge it then... well squash it. The commit history is a mess.

PR #300, opened 2017-12-06
add cmake toolchain files

derselbst:
...as provided by Carlo in #275. Since there was no input from the mailing list yet, I propose to create a subdir `/cmake_admin/cross/` where we place any toolchain files. As naming convention I suggest using the target system this file is dedicated for (i.e. as indicated by the `TOOLCHAIN_PREFIX`). Although I'm not sure if the host system matters for these files. So this can be seen as RFC.

@carlo-bramini @jjceresa @mawe42 @ensonic Any opinions?

### DIFF
--- a/cmake_admin/cross/i686-w64-mingw32.cmake
+++ b/cmake_admin/cross/i686-w64-mingw32.cmake
@@ -1,0 +1,23 @@
+# Cross compile toolchain configuration based on:
+# http://www.cmake.org/Wiki/CMake_Cross_Compiling
+
+# the name of the target operating system
+SET(CMAKE_SYSTEM_NAME Windows)
+
+# 32 bit mingw-w64
+SET(TOOLCHAIN_PREFIX "i686-w64-mingw32")
+
+# compilers to use for C and C++
+FIND_PROGRAM(CMAKE_C_COMPILER NAMES ${TOOLCHAIN_PREFIX}-gcc)
+FIND_PROGRAM(CMAKE_CXX_COMPILER NAMES ${TOOLCHAIN_PREFIX}-g++)
+FIND_PROGRAM(CMAKE_RC_COMPILER NAMES ${TOOLCHAIN_PREFIX}-windres)
+FIND_PROGRAM(PKG_CONFIG_EXECUTABLE ${TOOLCHAIN_PREFIX}-pkg-config)
+
+# path to the target environment
+SET(CMAKE_FIND_ROOT_PATH /usr/${TOOLCHAIN_PREFIX}/sys-root/mingw)
+
+# search for programs in the build host directories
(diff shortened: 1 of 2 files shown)


jjceresa:
Unfortunately, i have no knowledge about cmake scripting. Also i don't usually use cross compilation tools.
The only chains i have used to build Fluidsynth are the followings:
1) cmake  + make on  Debian.
2) cmake + make on Raspbian (for Raspberry pi2 hardware).
3) cmake + visual studio 6.0. (actually for Window XP target).

Actually i use (3) often.
So, i am not competent to give an objective opinion.

mawe42:
I'm not sure. I do cross-compile regularly to ARM on my Linux box, but I use a custom buildroot package for that (so I have no need for cmake toolchain files). Also I have never used mingw to cross-compile to Windows.

But I do wonder how portable those toolchain files are. At least the following looks very host specific to me:
```
SET(CMAKE_FIND_ROOT_PATH /usr/${TOOLCHAIN_PREFIX}/sys-root/mingw)
```

I have x86_64-w64-mingw32 toolchain and also an iconv-lib for mingw-w64 installed on my Ubuntu 17.10 and there is no sys-root/mingw in the toolchain dir. The iconv.dll file got installed into `/usr/x86_64-w64-mingw32/bin` and there is a winpthread dll in ` /usr/x86_64-w64-mingw32/lib`, so I guess that I would have to modify that path to point to `/usr/x86_64-w64-mingw32/`?

carlo-bramini:
yes, that's right. That root path is expected to be used for CYGWIN.
I do not know if it is possible to make something conditional with cmake and other hosts, unfortunately I know it just very little.

mawe42:
Ah, ok. In that case, maybe it would be better to create a wiki page instead of adding those files to the source tree.

derselbst:
`/usr/${TOOLCHAIN_PREFIX}/sys-root/mingw` is also used on my Suse machine, so this seems to be common. Anyway a wiki page might really be a better idea.

jjceresa:
The idea of wiki page is interesting. This way fluidsynth users can not say that things are not documented.
****
Talking about wiki editing. Did you know if any documentation exist about formatting , editing, table of  contents ?. Sorry for this question out of the subject :)

derselbst:
It's done by using Markdown. Just search for it, you'll find plenty tutorials on the internet.

derselbst:
I added information about cross compilation: https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake#cross-compilation

It was surprisingly simple once I figured out how to convince pkgconfig where to search for the target platform libraries. Once you guys are fine with that I'd close this PR.

jjceresa:
Many thanks.

PR #304, opened 2017-12-13
Fix crash to TCP server (WIN32).

carlo-bramini:
This is the fix for the bug that I signaled in the mailing list.
I have not removed the join into fluidsynth.c because, as I have written in the mailing list, I have not understood what it is its purpose.


derselbst:
Ok

PR #306, opened 2017-12-16
Polymono

### LINKED ISSUE #158: Patch-PolyMono-functionality
This patch add Poly Mono functionality to Fluidsynth v 1.1.6

Part 1
This part describes Poly/Mono mode:
- This patch handle the MIDI specifications concept of channel basic
- MIDI modes messages in Fluidsynth:Omni On/Off, Poly OnOff.
- New API for channel basic and mode change.
- New shell commands for basic channel and mode change.

Part 2 
This part describes mainly the monophonic behavior:
- Type of MIDI input controler (monophonic/polyphonic).
- MIDI CC messages handled
  CC legato(68d) On/Off
  CC portamento(65d) On/Off
  CC portamento time (msb,lsb) (5, 37d)
- Use of sustain/sostenuto pedal in monophonic mode.
- Use of CC Breath.
  New API, shell commands to handle default breath to Attenuation modulator.
- Legato mode playing.
  New API, shell commands to handle legato mode.

Please see PatchFluidPolyMono-0001.pdf

Reported by: jjceresa

Original Ticket: [fluidsynth/tickets/160](https://sourceforge.net/p/fluidsynth/tickets/160)

derselbst:
This PR adds the polymono functionality based on #158 to fluidsynth. Its main purpose is to keep track of the TODOs and aid polishing this patch until it's ready to be merged. I will therefore periodically review it.

Closes #158.


mawe42:
@jjceresa have you tried your profiling patches with the polymono changes? It be great to know how the polymono patch affects the performance of FluidSynth.

derselbst:
What currently strikes me are the legato modes. I had a listen with your provided legato_demo.sf2 and additionally with a soundfont that has only one instrument zone. No matter what instrument I try, to my ears legato modes 0,1,4 dont sound legato at all. Modes 2 and 3 sound a little legato, though I cant tell the difference between those.

As the effect of legato playing is even more limited the more instruments zones exist, the question is: Do we really need five legato modes?

Not only that a difference between the modes is hardly audible, the underlying code is IMO also hardly maintainable. If there is any change to the codebase probably no one (execpt you jjc) will be able to re-test this functionality and verify that it still works as before.

Thus I would vote for only one, at most two legato modes. Though I have to admit that I only see this from my little programming world, as I'm totally missing a musicians view here.

jjceresa:
@mawe42 

> have you tried your profiling patches with the polymono changes?

The performance are rather the same than before. This is normal because the dsp chain (voices processing + [reberb] + [chorus] is rather not impacted by poly/mono patch. To be exact the only part that is impacted is the `phase calculation` in `voice processing` to get a `portamento effect`. Here is the code:

https://github.com/FluidSynth/fluidsynth/blob/d0adfe31f4961d034483a95705ac8b86839cf11d/src/rvoice/fluid_rvoice.c#L328-L346

When portamento is enabled (i.e when voice->dsp.pitchinc ! = 0) there is an addition. So the impact is the test of `voice->dsp.pitchinc`  and one addition with real variable (`voice->dsp.pitchoffset += voice->dsp.pitchinc`).

Later `voice->dsp.pitchoffset ` is used in phase calculation (One more addition). Here.
 https://github.com/FluidSynth/fluidsynth/blob/d0adfe31f4961d034483a95705ac8b86839cf11d/src/rvoice/fluid_rvoice.c#L321-L322
So the minimum impact is `One test + 1 addition` (when portamento is disabled).
     the maximum impact is `One test + 2 additions` (when portamento is enabled).

jjceresa:
> What currently strikes me are the legato modes. I had another listen with your provided legato_demo.sf2 and additionally with a soundfont that has only one instrument zone. No matter what instrument I try, to my ears legato modes 0,1,4 dont sound legato at all.

The perceptive effect of the chosen legato mode is strongly dependent of:

- volume ADHDSR shape.
- keyRange, velRange switching preset/instruments zones.

In others word these soundfont data have often precedence on the current legato mode. So, i have to admit that the legato_demo.sf2 demo and tutorial (3.1.4) are incomplete because of the difficulty to demonstrate these dependency.

>As the effect of legato playing is even more limited the more instruments zones exist, the question is: Do we really need five legato modes?

We can live with less legato modes, most musicians should be happy with two modes only.

>Thus I would vote for only one, at most two legato modes. Though I have to admit that I only see this from my little programming world, as I'm totally missing a musicians view here.

Modes 1 and 4 should be sufficient. When playing n1,n2 (n2 being played legato with n1):
Mode 1: On n2On,
   - voices of n1 are forced in release segment.
   - the new note n2  starts with new voices.

Mode 4:  On n2On,
- if n2  falls inside keyrange/velrange zone of n1, the new note n2 makes used of voices of the previous note n1.
- But if n2  falls outside the keyrange zone of n1 , new note n2  starts with new voices (i.e Mode 4 will react the same as Mode 1).

Using the convenient soundfont, Mode 4 behaves like acoustic monophonic instrument.
This mode should be preferred by musician playing on wind controller.

Hopefully the underlying code remains (imo) maintainable.

I'm ready to do these changes. Any opinions ?.

derselbst:
Interesting that you've chosen 1 and 4. I would have voted for mode 2.

Isnt mode 1 pretty much the same that would happen without any legato playing?

jjceresa:
>Interesting that you've chosen 1 and 4. I would have voted for mode 2.

If we vote for only one mode,  we could choose beetwen and 2 or 4, but  after more thinking`mode 2` is more  interesting than mode 4.

If we vote for two modes: we should choose 1 and 2 (rather than 1 and 4 as indicated in my previous comment)

>Isnt mode 1 pretty much the same that would happen without any legato playing?

In fact the `"legato playing gesture"` is NOT ONLY used in the intend to obtain
slurs (note merged with the previous one) (as mode 2 allow). For example when the musician wants to play fast trills (i.e 2 notes fast repeated like C-E-C-E-C-E...). In this case the notes need to be played as this: 
1) with a `very regular and short transition` between note (the more as possible).
   To aquire this, it is more easy using the same gesture than `"legato playing"`.
2) Each note must be perceived `individually` (i.e not merged with the previous one). In others words a very short "valley" must exist between notes. (This type of articulation is a bit similar to "détaché").
   To aquire this, only Mode 1 is appropriate.
 
Note that when a musician wants to play fast trills on an acoustic monophonic instrument he uses also a "legato like" gesture and to obtain individual perceived note he blows on each note. 
Also we note than attempting to obtain fast trill without using a "legato like" gesture is unnatural and very difficult to do.
( I agree with the fact that in this case the use of "legato mode", "legato like" terminology is inappropriate and doesn't help to understand the real sonic intend). The most important is to keep in mind the sonic intend, regardless of the mechanical gesture used to get this result. 

This is why i have chosen Mode 1 in addition to Mode 2.

derselbst:
Ok, then let's use mode 2 as default mode (as being the most "interesting" one). And if you are convinced that playing these "fast trills" is best done in legato mode and cannot be better done with appropriate instrument ADSR envelopes or whatever, then we could also have mode 1.

jjceresa:
Ok, note that numbering 1 and 2 will be changed to 0 and 1 respectively (to get it based to 0).

derselbst:
Sure, just make sure that this "most interesting" mode 2 is the default ;)

jjceresa:
Sure, i had noticed that.

mawe42:
> The performance are rather the same than before.

@jjceresa  That's great. I've tested it on my small ARM machine and didn't see any noticable performance increases. Excellent.

Thanks for these great features, jjc. I am going to make use of them in my instrument once this PR has been cleaned up, especially the portamento and basic channels part.

While reading up on the channel modes, a thought occured to me: in your PDF you explain the channel modes, basic channels etc in great detail. But that feature is already specified quite thoroghly in the official MIDI spec. Maybe it would be easier (and less documentation work) to just point to the MIDI spec? And maybe explain where FluidSynth's implementation differs (if at all?).

src/synth/fluid_voice.c
```diff
@@ -1168,6 +1197,74 @@ int fluid_voice_modulate_all(fluid_voice_t* voice)
...
+{
+	fluid_channel_t* channel= voice->channel;
+	fluid_real_t PitchBeg = fluid_voice_calculate_pitch(voice,fromkey);
+	fluid_real_t PitchEnd = fluid_voice_calculate_pitch(voice,tokey);
+	fluid_real_t pitchoffset = PitchBeg - PitchEnd;
+	/* Increment number is function of PortamentoTime (ms)*/
+	unsigned int countinc = (int)(((fluid_real_t)voice->output_rate * 
+					0.001f *
+			        (fluid_real_t)fluid_channel_portamentotime(channel))  /
+					(fluid_real_t)FLUID_BUFSIZE  +0.5);
```
derselbst: Why are you casting this to `int` when assigning it to an `unsigned int`?

src/synth/fluid_voice.c
```diff
@@ -1168,6 +1197,74 @@ int fluid_voice_modulate_all(fluid_voice_t* voice)
...
+	UPDATE_RVOICE2(fluid_rvoice_set_portamento, countinc, pitchoffset); 
+}
+
+/*---------------------------------------------------------------*/
+extern fluid_gen_info_t fluid_gen_info[];
```
derselbst: This `extern` seems to be redundant and `fluid_gen_info` can be made static again.

src/synth/fluid_voice.c
```diff
@@ -586,6 +604,17 @@ fluid_voice_calculate_runtime_synthesis_parameters(fluid_voice_t* voice)
     fluid_voice_update_param(voice, list_of_generators_to_initialize[n]);
   }
 
+  /* Start portamento if enabled */
+  {	/* fromkey note comes from "GetFromKeyPortamentoLegato()" detector.
+	When fromkey is set to ValidNote , portamento is started */
+	  /* Return fromkey portamento */
+	  int fromkey = voice->channel->synth->fromkey_portamento;
```
derselbst: Please think of a better solution to retrieve `fromkey_portamento` here and elsewhere. The synth and therefore also `fromkey_portamento` is mutex protected, which you are ignoring. Why are you not storing this key inside `fluid_channel_t`?

src/synth/fluid_synth.h
```diff
@@ -204,7 +218,25 @@ void fluid_synth_process_event_queue(fluid_synth_t* synth);
...
+int fluid_synth_noteoff_monopoly(fluid_synth_t* synth, int chan, int key, char Mono);
+
+fluid_voice_t*
+fluid_synth_alloc_voice_LOCAL(fluid_synth_t* synth, fluid_sample_t* sample, int chan, int key, int vel, fluid_zone_range_t* zone_range);
+
+void fluid_channel_set_onenote_monolist(fluid_channel_t* chan, unsigned char key, unsigned char vel);
+void fluid_channel_clear_monolist(fluid_channel_t* chan);
+void fluid_channel_invalid_prev_note_staccato(fluid_channel_t* chan);
+void fluid_channel_cc_legato(fluid_channel_t* chan, int value);
+void fluid_channel_cc_breath_note_on_off(fluid_channel_t* chan, int value);
```
derselbst: All `fluid_channel_*()` functions should be moved to fluid_chan.c

derselbst:
@jjceresa Ok thanks, looks better. However I again highly recommend that you start using the command line tools. Having numerous commits with the same message created via the web interface is not nice.

jjceresa:
@derselbst 
>However I again highly recommend that you start using the command line tools.....
Yes, i am aware of this, this is somewhere on top to my todo list ;)

include/fluidsynth/synth.h
```diff
@@ -46,6 +46,97 @@ extern "C" {
...
+ API: Poly mono mode 
+ */
+
+/* Macros interface to poly/mono mode variables */
+enum fluid_basic_channel_mode_flags
+{
+    MONO = 0x01, /* bit 0, 0: poly on , 1: mono on */
+    OMNI = 0x02, /* bit 1, 0: omni on, 1:omni off */
+    BASIC_CHANNEL = 0x04, /* bit 2, 1: channel is basic channel */
+    ENABLED = 0x08,         /* bit 3, 1: channel is listened */
```
mawe42: I think it's a bit misleading that MONO and OMNI are switched on if the bit is 0, BASIC_CHANNEL and ENABLED are switched on if the bit is 1. I would vote to make them all behave in the same way: enabled if bit is set, otherwise disabled.

Also a small typo here "Channel is listened" doesn't really make sense. Maybe "channel is enabled"?

mawe42: > I think it's a bit misleading that MONO and OMNI are switched on if the bit is 0, 

Sorry, that only applies to OMNI, of course. MONO is already on if the bit is set.

derselbst: If I understood correctly, this was made because by default fluidsynth has one basic channel
on MIDI channel 0 with PolyOn and OmniOn, so that the synth plays/responds to all MIDI data received on all 16 channels.

derselbst: Though I agree it would be clearer to have this bit flipped.

jjceresa: MIDI spec `channel mode (1 to 4)` are in Fluidsynth zero based (0 to 3) respectively.
This value is coded internally on 2 bits (bits b1,b0) and simply represent the four MIDI spec values (0 based):
- OMNION_POLY  : mode 0
- OMNION_MONO: mode 1
- OMNIOFF_POLY: mode 2
- OMNIOFF_MONO: mode 3 

Consequently, MONO On is represented by bit 0 when 1, and OMNI ON is represented by bit b2 when 0. 

>by default fluidsynth has one basic channel on MIDI channel 0 with PolyOn and OmniOn, so that the synth plays/responds to all MIDI data received on all 16 channels.

Yes, mode 0 on channel 0 is the preferable default mode(and usually recommended by MIDI spec).

>Also a small typo here "Channel is listened" doesn't really make sense. Maybe "channel is enabled"?

Yes, it could be done

src/synth/fluid_voice.c
jjceresa: Since the last change this extern now need to be removed

src/synth/fluid_synth.h
jjceresa: This will be moved.

src/synth/fluid_voice.c
jjceresa: This a writing error and this should be cast  `(unsigned int)`.

src/synth/fluid_voice.c
jjceresa: >The synth and therefore also fromkey_portamento is mutex protected, which you are ignoring.

I'm aware of the mutex, please which issue did you see on this subject ?.

>Why are you not storing this key inside fluid_channel_t ?

Simply because on each `noteOn` (regardless of the channel), this `key` is always updated by `get_fromkey_portamento_legato()`  https://github.com/FluidSynth/fluidsynth/blob/67ff72659f4e83a0e0e24912e919071ae692f590/src/synth/fluid_synth_mono.c#L475

and then this `key` is read in the context of this `noteOn` regardless of the channel concerned, so to avoid waste of memory i have chosen `fluid_synt_t` rather than  `fluid_channel_t`.

include/fluidsynth/synth.h
mawe42: > Consequently, MONO On is represented by bit 0 when 1, and OMNI ON is represented by bit b2 when 0.

Ok, if it is important to have the same numbering of modes inside of FluidSynth as used in the spec, then maybe we could rename the enums to better reflect what they express. Maybe POLY_OFF and OMNI_OFF.

The current state is really quite confusing, with OMNI being the only one where the logic is reversed. What tripped me up was the following code:

```
if (chan->mode &  BASIC_CHANNEL)
{
    int new_mode, new_val;
--> if(!(chan->mode & OMNI)) /* channel is actually Omni On */
    {new_mode = OMNION_MONO; new_val=0;} /* set channel in mode 1 */
    else /* channel is actually Omni Off */
    {new_mode = OMNIOFF_MONO; new_val= value;} /* set channel in mode 3 */
    return fluid_synth_set_basic_channel_LOCAL(synth,channum, 
                                        new_mode, new_val);
}
```

So we have ```(mode & BASIC_CHANNEL)``` to mean "is basic channel" and ```(mode & OMNI)``` to mean "is NOT omni". This is really confusing and a great source for errors. If we would rename MONO and OMNI to POLY_OFF and OMNI_OFF, then those bit tests suddenly become meaningful and consistent again.

jjceresa: >Ok, if it is important to have the same numbering of modes inside of FluidSynth as used in the spec, then maybe we could rename the enums to better reflect what they express. Maybe POLY_OFF and OMNI_OFF.


Yes, the naming should be changed `POLY_OFF and OMNI_OFF` or `MONO_ON and OMNI_OFF` ?  which did  you prefer ?.

src/synth/fluid_voice.c
derselbst: I just dont like accessing anything via this indirection `voice->channel->synth`, yes it's done like that at several places in `fluid_voice.c`... However looking at it again I think when accessing `fromkey_portamento` that way the synth API should be locked anyway, so never mind. It's probably fine.

mawe42: I must admit I don't understand why you store fromkey_portamento on the synth instance in the first place. Couldn't you just return it in a return parameter of the get_fromkey_portamento_legato function?

include/fluidsynth/synth.h
derselbst: `POLY_OFF and OMNI_OFF`

mawe42: I personally would prefer POLY_OFF and OMNI_OFF, makes it more consistent.

src/synth/fluid_voice.c
mawe42: Ok, having read more of the code now, I think I understand: you need it somewhere that fluid_voice_start can reach. And passing it directly would mean changing a lot of public function signatures. Still, it seems very strange to have get_fromkey_portamento_legato return the legato fromkey directly and the portamento fromkey indirectly in such a far away place. But I don't have a good suggestion on how to do it otherwise...

include/fluidsynth/synth.h
jjceresa: So, `MONO , OMNI` will be changed  to `POLY_OFF`, `OMNI_OFF` respectively.

src/synth/fluid_voice.c
jjceresa: >derselbst : I think when accessing fromkey_portamento that way the synth API should be locked anyway, so never mind. It's probably fine.

On fluid_synth_noteon() , between the last update of `fromkey_portamento`  (made by get_fromkey_portamento_legato()) to the use of this `fromkey_portamento` (in  fluid_voice_start()), effectively there is a call to `FLUID_API_ENTRY_CHAN` call (in  fluid_synth_alloc_voice_LOCAL()) but the mutex involved is recursive, so it doesn't lock the current task (but should lock another task calling fluid_synth_noteon() API). In short , On noteOn, PolyMono calls fluid_preset_noteon() like on noteOn without PolyMono. I don't see a potential deadlock that wasn't already there before.

>mawe42 : And passing it directly would mean changing a lot of public function signatures.

Yes, unfortunately.

>Still, it seems very strange to have get_fromkey_portamento_legato return the legato fromkey directly and the portamento fromkey indirectly in such a far away place. But I don't have a good suggestion on how to do it otherwise..

Me too, i regret to be forced in this situation and don't have a good  suggestion on how to do it otherwise.

For information,note the same regrettable situation (not related to polymono) for `storeid` which is also a "far away parameter" for fluid_voice_init().
https://github.com/FluidSynth/fluidsynth/blob/67ff72659f4e83a0e0e24912e919071ae692f590/src/synth/fluid_synth.c#L4541-L4542.

jjceresa:
> mawe42 : But that feature is already specified quite thoroghly in the official MIDI spec.

i have used the MIDI 1.0 version 4.2, have you read others Specs ?

>Maybe it would be easier (and less documentation work) to just point to the MIDI spec?

Sure, it could be done. Talking about Basic channel with only one Basic channel, the specs are pretty clear, but when there are more then one basic channel i haven't not found the Specs very explicit this is why i have added chapter 2.5.
Could you please point the chapters numbers (apart chapter 1) that seems too redundant ?.

>And maybe explain where FluidSynth's implementation differs (if at all?).

Apart the channels mode numbering 1 to 4 that is zero based  (0 to 3) i don't see any others difference between MIDI spec and this PR.
Of course, things like legato mode , breath mode, portamento mode are out of the scope of MIDI specs.
Please don't hesitate to point things that seems unclear or not understandable.

src/synth/fluid_voice.c
derselbst: > I don't see a potential deadlock that wasn't already there before.

No no, there is no deadlock. I just wanted to be sure this way of accessing `fromkey_portamento` is threadsafe. And since there is a call to `FLUID_API_ENTRY_CHAN` it's fine.

include/fluidsynth/synth.h
mawe42: Then I would suggest also renaming the other member of that enum to add ```_BIT```, e.g. BASIC_CHANNEL_BIT, ENABLED_BIT

My preference would be to prefix everything with CHANNEL_:
```
CHANNEL_POLY_OFF
CHANNEL_OMNI_OFF
CHANNEL_BASIC
CHANNEL_ENABLED
```

Or ```CHAN_POLY_OFF```, if you think CHANNEL_ is too long.

mawe42:
>  have used the MIDI 1.0 version 4.2, have you read others Specs ?

I'm reading MIDI 1.0 version 96.1 3rd edition, as downloaded from https://www.midi.org/specifications.

> Please don't hesitate to point things that seems unclear or not understandable.

mawe42:
> i have used the MIDI 1.0 version 4.2, have you read others Specs ?

I'm using MIDI 1.0 version 96.1 3rd edition, as downloaded from https://www.midi.org/specifications

> Could you please point the chapters numbers (apart chapter 1) that seems too redundant ?

Well, maybe I spoke too soon (again). It's probably a good thing to have it all explained in a single text and not rely on the MIDI spec. As you say, the MIDI spec leaves many things open, especially multiple basic channels in a single instrument.

Do you think it would be possible to have the polymono docs in an editable format? I'm just thinking about maintaining that doc if anything about the polymono or basic channels feature changes...

include/fluidsynth/synth.h
derselbst: > Or CHAN_POLY_OFF, if you think CHANNEL_ is too long.

Because it's in a public header, we should also prefix `FLUID_`. I would also like to put the `BREATH_*` flags into that enum:

```
enum fluid_basic_channel_mode_flags
{
    FLUID_CHANNEL_POLY_OFF = 0x01, /* bit 0, 0: poly on , 1: mono on */
    FLUID_CHANNEL_OMNI_OFF = 0x02, /* bit 1, 0: omni on, 1:omni off */
    FLUID_CHANNEL_BASIC = 0x04, /* bit 2, 1: channel is basic channel */
    FLUID_CHANNEL_ENABLED = 0x08,         /* bit 3, 1: channel is listened */

    FLUID_CHANNEL_BREATH_POLY = 0x10,
    FLUID_CHANNEL_BREATH_MONO = 0x20,
    FLUID_CHANNEL_BREATH_SYNC = 0x40,

    FLUID_CHANNEL_MODE_MASK,
    FLUID_CHANNEL_MODE_OMNION_POLY,
    FLUID_CHANNEL_MODE_OMNION_MONO,
    ...
    FLUID_CHANNEL_MODE_LAST // NOT PART OF PUBLIC API/ABI stability guarantee
};
```

If you agree I would take care of these changes.

jjceresa: >Because it's in a public header, we should also prefix FLUID_. I would also like to put the BREATH_* flags into that enum:

Yes, these  missing flags need to be added.

>and also fix logic of POLY_OFF bit then.

Keeping the same numbering of `mode value` inside of FluidSynth as used in the `spec` simplify the setting/getting of this `mode value` in API `fluid_synth_set_basic_channel(), fluid_synth_get_basic_channels(), and fluid_synth_get_channel_mode()`. 
Changing the logic of bits `(b1,b0)` will complicate the construction of  `MIDI spec. mode` from the `FluidSynth mode b1,b0`  inside `fluid_synth_get_basic_channels(), and fluid_synth_get_channel_mode()`. This also will complicate the construction of `FluidSynth mode b1,b1` from the `MIDI spec. mode` inside `fluid_synth_set_basic_channel()`. So what could be chosen ?

jjceresa:
>Do you think it would be possible to have the polymono docs in an editable format? I'm just thinking about maintaining that doc if anything about the polymono or basic channels feature changes...

Sure it is possible. The original docs is a .doc (Microsoft Word), it should be editable using Libre Office.

include/fluidsynth/synth.h
derselbst: >> and also fix logic of POLY_OFF bit then.

Sorry, I was confused. The logic will be fine ofc, once the flags have been renamed. So if you agree with the re-naming I'd take care of this then.

jjceresa: >So if you agree with the re-naming I'd take care of this then.
Yes ofc, these missing flags BREATH_* need to be added and also the re-naming is necessary.

jjceresa:
Todo list:
-------------
1- 

> I am going to make use of them in my instrument once this PR has been cleaned up, especially the portamento and basic channels part.

Please could you point anythings about `portamento ` and `basic channels` ?
****

2- /doc/polymono directory isnt yet up to date (see examples files *.txt).
This is a memory stick for me, ignore it
****

3.1- Note (not related to polymono).

- (a) `fluid_synth_alloc_voice_LOCAL()` is called in the context of  `fluid_synth_noteon()` API which is  is threadsafe (i.e there is `FLUID_API_ENTRY_CHAN(), FLUID_API_RETURN`inside fluid_synth_noteon()).
- (b) `fluid_synth_alloc_voice_LOCAL()` now doen't need  to be PUBLIC API, so `parameters checking` and  `FLUID_API_ENTRY_CHAN(), FLUID_API_RETURN` should be moved in the right `fluid_synth_alloc_voice()` API.

3.2- Note that the same observation for `fluid_synth_start_voice()` API which could call its non API companion `fluid_synth_start_voice_LOCAL()` (called also by fluid_defpreset_noteon(), fluid_rampreset_noteon() ).

The benefice is to avoid a `fluid_synth_noteon()` API (thread safe) making call to others API (thread safe). 
Sure this practice works fine, but it is a bit difficult to understand for reviewer;  also calling multiple time the same mutex (which need to be recursive) is useless, with a slight lost of performance).

Please,any opinion ?
****

derselbst:
Since we already have `fluid_synth_alloc_voice_LOCAL()`, we could move `param checking`,  `FLUID_API_ENTRY_CHAN()`, `FLUID_API_RETURN` as you suggested.

On the other hand (for now) I dont want to start introducing more `_LOCAL()` functions. I would like to concentrate polish polymono itself. So far I have only reviewed the headers. On occasion I'll continue reviewing.

src/synth/fluid_synth.c
```diff
@@ -1181,20 +1250,51 @@ fluid_synth_remove_default_mod(fluid_synth_t* synth, const fluid_mod_t* mod)
...
+	  fluid_channel_set_cc (channel, num, val);
+	  result = fluid_synth_cc_LOCAL (synth, chan, num);
+  }
+  else /* chan is disabled so it is a candidate for global channel */
+  {	/* looks for next basic channel */
+	int n_chan = synth->midi_channels; /* MIDI Channels number */
+	int basicchan ;
+	if (chan < n_chan-1) basicchan = chan + 1; /* next channel */
+	else  basicchan = 0; /* wrap to 0 */
+	channel = synth->channel[basicchan];
```
derselbst: 1. Do we need this global channel thing at all?
1. Is it necessary to make this global channel check so complicated? Couldnt we just

```
 if (synth->channel[chan]->mode & FLUID_CHANNEL_BASIC && FLUID_CHANNEL_MODE_OMNIOFF_MONO)
  then
    /* send cc to all channels in this basic channel */
```

jjceresa: >1 . Do we need this global channel thing at all?

Yes we need it, Global Channel is a MIDI specification for channel basic in mode 3 (Omni off - Mono) .

>Is it necessary to make this global channel check so complicated?

The electric guitar is an example of instrument that could be put in mode 3 . We need 6 MIDI channels each for one string. When playing the musician could do two actions at the same time: 
1) use the `vibrato lever` to produce a `global vibrato` on the six strings . (This could be done using CC pitch bend on `global channel`)
2) use a bending on string 1 only. (This could done using  CC pitch bend on the `individual string MIDI channel`) . 
We get string 1 with a different bend then the other strings.

That mean that `global channel` and ` individual channel` need to be independent each to others.

To differentiate `global channel` the implementation logic follows the MIDI specs (please see pdf chapter 3.4.12). Of course it is a bit complicated for fluidsynth, but it is simpler for the application making use of instrument in mode 3.

include/fluidsynth/synth.h
```diff
@@ -317,6 +312,89 @@ FLUIDSYNTH_API int fluid_synth_handle_midi_event(void* data, fluid_midi_event_t*
...
+    FLUID_CHANNEL_BASIC = 0x04,    /**< if flag set the corresponding midi channel is a basic channel */
+    FLUID_CHANNEL_ENABLED = 0x08,  /**< if flag set the corresponding midi channel is enabled, else disabled, i.e. channel ignores any MIDI messages */
+};
+
+/* breath mode bits infos */
+enum fluid_channel_breath_flags
+{
+    FLUID_CHANNEL_BREATH_POLY = 0x10,  /**< bit 4, 1: default breath poly On */
+    FLUID_CHANNEL_BREATH_MONO = 0x20,  /**< bit 5, 1: default breath mono On */
+    FLUID_CHANNEL_BREATH_SYNC = 0x40,  /**< bit 6, 1: BreathSyn On */
```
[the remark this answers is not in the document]
jjceresa: Yes, it is necessary, and will be done.

src/synth/fluid_synth.c
```diff
@@ -168,6 +173,15 @@ void fluid_synth_settings(fluid_settings_t* settings)
   fluid_settings_register_str(settings, "synth.default-soundfont", DEFAULT_SOUNDFONT, 0);
 #endif
 
+  /* basic channel settings for poly/mono mode */
+  /* synth.basic-channel: default: 0, min:0, max:255, no hint,no callback function */
+  fluid_settings_register_int(settings, "synth.basic-channel", 0, 0, 255, 0);
+  /* synth.basic-channel.mode: default: 0, min:0, max:3, no hint,no callback */
+  fluid_settings_register_int(settings, "synth.basic-channel-mode", 0, 0, 3, 0);
+  /* synth.basic-channel.modeval: default: 0, min:0, max:255, no hint,no callback*/
+  fluid_settings_register_int(settings, "synth.basic-channel-modeval", 0, 0, 255, 0);
```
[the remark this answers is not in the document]
jjceresa: Initially these `synth.basic-channel*` settings was intended for the fluidsynth application console (using the -o option). But now,since the presence of commands `resetbasicchannels and setbasicchannels` theses commands can be put in a config file (using -f option).

Similarly since the presence of `fluid_synth_set_basic_channel() and fluid_synth_set_basic_channel()` API, an application can do what it need after synth creation. 
i think these settings could be removed. It will be done.

src/synth/fluid_synth.c
derselbst: > the implementation logic follows the MIDI specs (please see pdf chapter 3.4.12)

Ok.

> for channel basic in mode 3 (Omni off - Mono) .

In this case I'll correct the API comment.

jjceresa: Ok, later, i will do the change stated above about `breath mode` comment and removing `basic channel settings`.

src/synth/fluid_synth_polymono.c
```diff
@@ -0,0 +1,536 @@
...
+ * 
+ * @return 
+ * - FLUID_OK on success.
+ * - FLUID_FAILED
+ *   - synth is NULL.
+ *   - chan or val is outside MIDI channel count.
+ *   - mode is invalid.
+ *   - val has a number of channels overlapping the next basic channel.
+ */
+int fluid_synth_set_basic_channel(fluid_synth_t* synth, int basicchan, int mode, int val)
```
derselbst: Why doesnt `fluid_synth_set_basic_channel()` receive a `fluid_basic_channel_infos_t*`?

src/synth/fluid_synth_polymono.c
```diff
@@ -0,0 +1,536 @@
...
+ * @param val Number of monophonic channels (for mode 3 only) (0 to MIDI channel count).
+ * 
+ * @return 
+ * - FLUID_OK on success.
+ * - FLUID_FAILED
+ *   - basicchan is outside MIDI channel count.
+ *   - val has a number of channels overlapping the next basic channel.
+ */
+char * warning_msg1 = "Basic channel %d has been narrowed to %d channels.";
+char * warning_msg2 = "Basic channel %d have number of channels that overlaps.\n\
```
derselbst: `static const` and inside function scope, please.

src/synth/fluid_synth_polymono.c
```diff
@@ -0,0 +1,536 @@
...
+ *   - modeInfos is NULL.
+ */
+int fluid_synth_get_channel_mode(fluid_synth_t* synth, int chan,
+				fluid_basic_channel_infos_t  *modeInfos)
+{
+	/* check parameters first */
+	fluid_return_val_if_fail (modeInfos!= NULL, FLUID_FAILED);
+	FLUID_API_ENTRY_CHAN(FLUID_FAILED);
+	/**/
+	modeInfos->basicchan= chan;
```
derselbst: Why are you setting `chan` as basic channel? Wouldnt you need to search to what basic channel `chan` belongs to and return this instead?

src/bindings/fluid_cmd.c
```diff
@@ -1879,6 +1925,587 @@ int fluid_handle_router_par2(void* data, int ac, char** av, fluid_ostream_t out)
...
+	"Basic channel:  3, poly omni off(2), nbr:  1"
+	"Basic channel:  8, mono omni off(3), nbr:  2"
+	"Basic channel: 13, mono omni on (1), nbr:  3"
+*/
+char * mode_name[]={"poly omni on (0)","mono omni on (1)",
+					"poly omni off(2)","mono omni off(3)"};
+int fluid_handle_basicchannels (void* data, int ac, char** av, 
+								fluid_ostream_t out)
+{
+    fluid_cmd_handler_t* cmd_hdlr = (fluid_cmd_handler_t*)data;
```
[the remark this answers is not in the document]
jjceresa: Ok,

src/bindings/fluid_cmd.c
```diff
@@ -1879,6 +1925,587 @@ int fluid_handle_router_par2(void* data, int ac, char** av, fluid_ostream_t out)
...
+        free(bci); 
+	}
+	/* n is 1 or more basic channel number (never 0)*/
+	else if(n < 0) return -1; /* error */
+	if (n == 0) fluid_ostream_printf(out,"no basic channels\n");
+	return 0;
+}
+
+char *invalid_arg_msg =" invalid argument\n";
+char *too_few_arg_msg = " too few argument, chan mode val [chan mode val]...\n";
```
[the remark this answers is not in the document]
jjceresa: Ok.

src/bindings/fluid_cmd.c
```diff
@@ -1879,6 +1925,587 @@ int fluid_handle_router_par2(void* data, int ac, char** av, fluid_ostream_t out)
...
+		int chan = ac ? atoi(av[i]): i;
+		result = fluid_synth_get_channel_mode(synth, chan, &bci);
+		if (result == FLUID_OK)
+		{
+			if(bci.mode &  FLUID_CHANNEL_ENABLED)
+			{	/* This channel is enabled */
+				char * basic_channel="basic channel"; /* field basic channel */
+				char * bc_msg; /* field basic channel */
+				char * poly_msg ="poly"; /* field mode */
+				char * mono_msg ="mono"; /* field mode */
```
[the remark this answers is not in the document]
jjceresa: Ok

src/bindings/fluid_cmd.c
```diff
@@ -1879,6 +1925,587 @@ int fluid_handle_router_par2(void* data, int ac, char** av, fluid_ostream_t out)
...
+				char * p_mode; /* field mode */
+				char * blank="--";
+				char nbr[10]; /* field Nbr */
+				char *p_nbr; /* field Nbr */
+				int mode = bci.mode &  FLUID_CHANNEL_MODE_MASK;
+				if (bci.mode &  FLUID_CHANNEL_BASIC)
+				{	/* This channel is a basic channel */
+					bc_msg = basic_channel;
+					sprintf(nbr,"nbr:%3d",bci.val);
```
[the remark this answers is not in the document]
jjceresa: Ok

src/synth/fluid_synth_polymono.c
jjceresa: It could done, if we prefer that the caller build a fluid_basic_channel_infos_t ?

src/synth/fluid_synth_polymono.c
jjceresa: Ok

src/synth/fluid_synth_polymono.c
jjceresa: >Wouldnt you need to search to what basic channel chan belongs to and return this instead?.

No, we need only informations on this particular channel (regardless of the basic channel it belongs). I admit that the used return structure name `fluid_basic_channel_infos_t` is improper and confusing. 
Should we define a new named structure for this API ?  or simply could we better document the API ?

src/synth/fluid_synth_polymono.c
derselbst: I'm not a fan of introducing these helper structs. However in this case I currently cannot think of a better way. And since we already have this `fluid_basic_channel_infos_t`, we should also consistently use it.

src/synth/fluid_synth_polymono.c
derselbst: No not introducing yet another struct.

I still dont understand why you dont return the basic channel `chan` belongs to. There is no gain in information returning `chan` for this member. In fact I find it more confusing. The caller knows anyway with what `chan` the function was called.

jjceresa: >I still dont understand why you dont return the basic channel chan belongs to.

Ok, it will be done.

src/bindings/fluid_cmd.c
```diff
@@ -2133,27 +2135,25 @@ int fluid_handle_channelsmode (void* data, int ac, char** av,
...
 				char *p_nbr; /* field Nbr */
 				int mode = bci.mode &  FLUID_CHANNEL_MODE_MASK;
 				if (bci.mode &  FLUID_CHANNEL_BASIC)
 				{	/* This channel is a basic channel */
-					bc_msg = basic_channel;
-					sprintf(nbr,"nbr:%3d",bci.val);
+					char nbr[10]; /* field Nbr */
+					bc_msg = "basic channel";
+					nbr[sizeof(nbr)-1] = 0;
+					FLUID_SNPRINTF(nbr,sizeof(nbr)-1,"nbr:%3d",bci.val);
```
derselbst: Manual null termination is not necessary for c99 snprintf, you can use the full buffer.

jjceresa: Sorry, when using  this example:
```
 char nbr[4]; /* field Nbr */
 nbr[sizeof(nbr)-1] = 0;					
 snprintf(nbr,sizeof(nbr),"%d",9999); 
 if(nbr[sizeof(nbr)-1] != 0) 
 {        
  printf "not null termined"; 
 }
```
The message "not null termined" is displayed.  What could we do ?

mawe42: Are you compiling this with VC? It seems like VC returns an error code if the buffer is too small to fit the formatted string + null byte: https://github.com/magnumripper/JohnTheRipper/issues/851
Maybe we need a similar (if  not the same) shim for old VC versions?

mawe42:
>> I am going to make use of them in my instrument once this PR has been cleaned up, especially the portamento and basic channels part.

> Please could you point anythings about portamento and basic channels ?

@jjceresa sorry, my comment was probably unclear: I didn't mean that portamento and basic channels need special clean-up. I wanted to say that I'm particularly interested in those two features.

src/bindings/fluid_cmd.c
derselbst: Seems like @jjceresa trapped into microsofts non standard, non compliant, shitty extension `_snprintf()`. That's why I did this yesterday to master: 272b8073833b1b1867caf71759924617592e1b2f

...perhaps it's work cherry-picking it to `1.1.x`.

jjceresa:
> I wanted to say that I'm particularly interested in those two features.

Basic channel is interesting for large setup with a lot of instruments (mono and poly). We usually don't need it daily.
Portamento should be used with taste because not all instruments are favorable to be played with this type of articulations.
Probably, thee most interesting is the aptitude to play monophonic instrument regardless the MIDI device ( MIDI keyboard or wind controller).

mawe42:
When playing around with the shell commands, I had to constantly go back to the documentation and remind myself what those mode numbers actually mean. And it's not helped by the fact that MIDI mode numbers are 1-based, where FluidSynth modes are 0-based.

What do you think about changing the user-interface so that the mode numbers are removed completely? Let the user specify omni and poly/mono directly. After all... all possible combinations are valid modes. So maybe something like this:

```setbasicchannel <channel> <length> <omni|single> <poly|mono>```

```
setbasicchannel 5 3 omni poly
setbasicchannel 11 2 single mono
```

**Comment by derselbst** [comment]
> What do you think about changing the user-interface so that the mode numbers are removed completely?

For the shell it's definitely a viable change, I agree.

File: src/bindings/fluid_cmd.c (hunk 16 above)
Comment (mawe42) [reply src/bindings/fluid_cmd.c]: >  That's why I did this yesterday to master: 272b807

Ah, that's much better. I must say I really like having glib around :-)

Comment (derselbst) [reply src/bindings/fluid_cmd.c]: > I must say I really like having glib around

That's so true. It is still very sad that microsoft needed until VS2015 for a c99 compliant library implementation.

**Comment by jjceresa** [comment]
>What do you think about changing the user-interface so that the mode numbers are removed completely?
>Let the user specify omni and poly/mono directly. 

Yes of course the idea is very seducing. Basically all shell commands (not only polymono) are rustic but  `very efficient`. Probably the most important is to stay with consistent commands. If we begin to add `names `for parameters, we need to do that for all commands and not only for polymono commands.

With the actual commands one probably need to get used  with the following `poly = 0 or 2` and `mono = (1 or 3) `regardless of (`OmniOn = 0 or 1` and `OmniOff = 2 or 3`). If you think one function at a time this become easier. (Note that names `Poly`, `Mono`, `OmniOn`, `OmniOff `are consistent with MIDI spec wording).

Please don't forget that `resetbasicchannels/rsetbasicchannels` allow to set `any number` of `basic channel` in only one command efficiently and mainly useful for large setup. The same is true for others `set*` command . Typically theses commands should be placed in  `config `file (`option -f`) or in a file invoked by the `source `command (preferably documented by the `echo `command).

If we are on small setup (only one or 2 instruments) we usually don't need to use `setbasicchannel `or `resetbasicchannel `to put a channel in `mono`  (regardless of the actual mode). For example if we want to put channel 0 in `mono `mode use this very fast command:

- CC 0 68 127  (`CC legato on`) Channel 0 is now `mono`
- CC 0 68 0      (`CC legato off`) Channel 0 returns to `poly`.

**Comment by mawe42** [comment]
I like your last change that introduces named modes. That addresses it equally well, I think.

File: src/bindings/fluid_cmd.c [hunk 17]
```diff
@@ -1969,14 +1969,45 @@ int fluid_handle_basicchannels (void* data, int ac, char** av,
...
   Sets the list of MIDI basic channels with mode
+
   This list replaces any previous basic channels list.
 
   With no parameters the function set one channel basic at
   basicchan 0 mode 0 (Omni On Poly) (i.e all the MIDI channel are polyphonic).
+
+  Mode can be: 
+      numeric: 0 to 3 or
+	  name: poly_omnion , mono_omnion, poly_omni_off, mono_omnioff
```
mawe42: Small typo here: poly_omnioff

jjceresa: >Small typo here: poly_omnioff.

Thanks for reporting this and probably any others grammar errors.

src/rvoice/fluid_rvoice.c
```diff
@@ -319,11 +319,32 @@ fluid_rvoice_write (fluid_rvoice_t* voice, fluid_real_t *dsp_buf)
...
+	if (voice->dsp.pitchoffset > 0.0f) 
+		voice->dsp.pitchoffset = voice->dsp.pitchinc = 0.0f;
+  }
+  else if (voice->dsp.pitchinc < 0.0f)
+  {	/* portamento is enabled, so update pitchoffset */
+	voice->dsp.pitchoffset += voice->dsp.pitchinc;
+	/* when pitchoffset reaches 0.0f, portamento is disabled */
+	if (voice->dsp.pitchoffset < 0.0f) 
+		voice->dsp.pitchoffset = voice->dsp.pitchinc = 0.0f;
+  }
```
derselbst: Couldn't this whole portamento pitch calc. be simplified to:

```
voice->dsp.pitchoffset += voice->dsp.pitchinc;

if ((voice->dsp.pitchinc > 0.0f && voice->dsp.pitchoffset > 0.0f) ||
    (voice->dsp.pitchinc < 0.0f && voice->dsp.pitchoffset < 0.0f))
  voice->dsp.pitchoffset = voice->dsp.pitchinc = 0.0f;
```

jjceresa: yes it could, it will be done.

src/synth/fluid_chan.c
```diff
@@ -294,3 +314,360 @@ fluid_channel_get_sfont_bank_prog(fluid_channel_t* chan, int *sfont,
...
+ *         |                        |
+ *      i_first                   i_last
+ * 
+ * @param chan  fluid_channel_t.
+ * @param key MIDI note number (0-127) to search.
+ * @return index of the note if find, INVALID_NOTE otherwise.
+ * 
+ */
+unsigned short
+fluid_channel_search_monolist(fluid_channel_t* chan, unsigned char key)
```
derselbst: I would prefer if this function returns an `int`, i.e. either `FLUID_FAILED` or index `i`.

src/synth/fluid_chan.c
```diff
@@ -294,3 +314,360 @@ fluid_channel_get_sfont_bank_prog(fluid_channel_t* chan, int *sfont,
...
+{
+	short n = chan->n_notes; /* number of notes in monophonic list */
+	short i= chan->i_first; /* searching starts from i_first included */
+	while(n) 
+	{
+		if(chan->monolist[i].note == key) break; /* found */
+		i = chan->monolist[i].next; /* next element */
+		n--;
+	}
```
derselbst: Could you transform this to a for loop and return directly from it?

src/synth/fluid_chan.c
```diff
@@ -294,3 +314,360 @@ fluid_channel_get_sfont_bank_prog(fluid_channel_t* chan, int *sfont,
...
+ * Note: the following variables in Channel keeps trace of the situation.
+ *       - i_last index keeps a trace of the most recent note played even if
+ *       the list is empty.
+ *       - prev_note keeps a trace of the note removed if it is i_last.
+ *       - FLUID_CHANNEL_LEGATO_PLAYING bit keeps a trace of legato/staccato playing state.
+ * 
+ * More informations in FluidPolyMono-0003.pdf chapter 4 (Appendices).
+ */
+unsigned char
+fluid_channel_remove_monolist(fluid_channel_t* chan, short i)
```
derselbst: Same here. If this function returns an index I would prefer returning an `int` or `FLUID_FAILED`.

src/synth/fluid_chan.c
```diff
@@ -294,3 +314,360 @@ fluid_channel_get_sfont_bank_prog(fluid_channel_t* chan, int *sfont,
...
+ * if portamento pedal is Off.
+ *
+ * The function is intended to be called when the following event occurs:
+ * - On noteOff (in poly or mono mode), to mark prev_note invalid.
+ * - On Portamento Off(in poly or mono mode), to mark prev_note invalid.
+ * @param chan  fluid_channel_t.
+ */
+void fluid_channel_invalid_prev_note_staccato(fluid_channel_t* chan)
+{
+	if(!(chan->mode  & FLUID_CHANNEL_LEGATO_PLAYING)) /* the monophonic list is empty */
```
derselbst: If you want to check whether the list is empty, couldnt you explicitly check `chan->n_notes == 0`?

src/bindings/fluid_cmd.c
```diff
@@ -1000,6 +1024,28 @@ fluid_handle_echo(void* data, int ac, char** av, fluid_ostream_t out)
...
+{
+  if (ac < 1) {
+    fluid_ostream_printf(out, "sleep: too few arguments.\n");
+    return -1;
+  }
+  if (!fluid_is_number(av[0])) {
+    fluid_ostream_printf(out, "sleep: argument should be a number in ms.\n");
+    return -1;
+  }
+  g_usleep(atoi(av[0]) * 1000);	/* delay in micro second.*/
```
derselbst: This should now be `fluid_msleep(atoi(av[0]))`

jjceresa: Yes,it will be done.

src/synth/fluid_chan.c
jjceresa: Yes,it will be done.

src/synth/fluid_chan.c
jjceresa: Yes of course. It will be done.

src/synth/fluid_chan.c
jjceresa: Ok.

src/synth/fluid_chan.c
jjceresa: Here the comment  `/* the monophonic list is empty */` is in incorrect and should be `/* check if the playing is staccato */`. This correction will be done.

derselbst:
So I thought about API design a little bit, because currently `struct fluid_basic_channel_infos_t` strikes me most, as being an internal data type that should not really be exposed to the public. I suggest changing the basic channel API in the following way:

```
// disables all channels, deletes any basic channel assignment
int fluid_synth_reset_basic_channels(fluid_synth_t* synth);


// makes chan a basic channel of mode "mode", assignes "val" subsequent channels
// to this basic channel as well
// "val" must always be given no matter of the mode to use, however it can be set to -1
// to indicate synth->midi_channels
//
// in case "chan" is already a basic channel or belongs to a channel group, or
// channels overlap, this function should fail. in other words, all channels that are 
// changed by this function should be disabled
//
// we could also think about some fluid_synth_remove_basic_channel() that removes
// a whole basic channel group, so the user wouldnt always need to call
// fluid_synth_reset_basic_channels() to make a minor change. not sure how often
// this could be a use case though...
int fluid_synth_set_basic_channel(fluid_synth_t* synth, int chan, int mode, int val);


// gets the basic channel "chan" belongs to
int fluid_synth_get_basic_channel(fluid_synth_t* synth,
                                 int chan,
                                 int *basic_chan_out);


// gets the mode "chan" is set to
int fluid_synth_get_channel_mode(fluid_synth_t* synth,
                                 int chan,
                                 int *mode_out);
```

`fluid_synth_get_basic_channels()` should be removed from the public API, as long as there is no urgent need for it. The user should know about all basic channels and in case he doesnt, he can use `fluid_synth_get_basic_channel()` and `fluid_synth_get_channel_mode()`.

jjceresa:
> "val" must always be given no matter of the mode to use,..

For mode `poly omnion` and `mono omnion` val value can be effectively the number of MIDI channels intended to belong to the group but for mode `poly omnioff`  the number of channel of this group must be 1 regardless of val. This is required by the MIDI specs.

> "val" must always be given ......however it can be set to -1 to indicate synth->midi_channels

or to indicate all MIDI channels available (i.e disabled) until the next basic channel.
We could use the 0 value instead of -1 to stay coherent with the MIDI specs convention for val.

> // we could also think about some fluid_synth_remove_basic_channel() that removes
> // a whole basic channel group, so the user wouldnt always need to call
> // fluid_synth_reset_basic_channels() to make a minor change. not sure how often
> // this could be a use case though...

This is necessary, to change the mode of an existing basic channel as does `CC omni on`,` CC omni off`, `CC poly on` and `CC mono on` and also the shell commands. We don't need a new API and could use of :
```
// disables all channels belonging to basic channel chan. 
// If chan is -1 deletes any basic channel assignment (i.e disable all channels).
int fluid_synth_reset_basic_channels(fluid_synth_t* synth, int chan);
```

>fluid_synth_get_basic_channels() should be removed from the public API, as long as there is no urgent need for it.

It could be removed, but as the returned `val `is lost  `fluid_synth_get_channel_mode` should be replaced by a new signature to return `val`:
```
int fluid_synth_get_channel_mode_val (fluid_synth_t* synth,
                                 int chan,
                                 int *mode_out,
                                 int *val_out
);
```
Now  in the intend to save the basic channels infos in permanent storage, the application must  itterates using int fluid_synth_get_channel_mode_val (). This is not too difficult to do.

> int fluid_synth_get_basic_channel(fluid_synth_t* synth,
>                                  int chan,
>                                  int *basic_chan_out);

I don't see a use case for this function as `fluid_synth_get_channel_mode_val()` returns all the bits informations in `mode`.

derselbst:
> but for mode poly omnioff the number of channel of this group must be 1 regardless of val. This is required by the MIDI specs.

Sure, if that's what the spec says, the API should also check for this.

> or to indicate all MIDI channels available (i.e disabled) until the next basic channel.

That would be inverse logic to what the function is actually supposed to do. Why would that be useful?

> We could use the 0 value instead of -1 to stay coherent with the MIDI specs convention for val.

In the context of fluidsynth -1 means all channels. I would like to stay consistent here.

> It could be removed, but as the returned val is lost

Why would an application need `val`?

> fluid_synth_get_basic_channel() I don't see a use case for this function

Dont you think that it could be useful for an application to find out to what basic-channel some midi channel belongs to?

jjceresa:
> Why would an application need val?

- An application can load a private configuration file for basic channels to set inside a `synth `using the new function API `fluid_synth_set_basic_channel(synth, basic_chan,mode,val)` with `basic_chan`,`mode`, and `val `coming from this file among others things.

- Then as the application is interactive using a fluidsynth shell, it allows the end user to change the current basic channels setting using the `setbasichannels chan_x  mode_x val_x` commands.

- Then the application allows to save in its private configuration file the actual basic channels configuration coming from the `synth `using for example an iteration through `fluid_synth_get_channel_mode_val ( synth, chan_x,*mode_out, *val_out) `to get `mode_x` and `val_x `for each basic channel `chan_x `.

Also the command `channelsmode `needs this val value.
***
Note also that with your new proposal , the command `basicchannels `doen't not need any more of the 
`fluid_synth_get_basic_channels()` nor  `fluid_basic_channel_infos_t`, instead it could use `fluid_synth_get_channel_mode_val()` as `channelsmode` command does.
***
> Dont you think that it could be useful for an application to find out to what basic-channel some midi channel belongs to?

Actually i don't see that it could be useful. But if the case occurs,it could be easy for the the application to iterate through the MIDI channels count using `fluid_settings_getint(settings, "synth.midi-channels",..)` and `fluid_synth_get_channel_mode_val (chan)` to find this basic channel.
***
i find your API proposal for `fluid_synth_reset_basic_channels()`, `fluid_synth_set_basic_channel()` more straightforward for the user. The result is a more simpler API interface and for the application it is easy to replace the lost of `fluid_synth_get_basic_channels()`. Good compromise.

derselbst:
I must admit that I find your use-case of val a little constructed... but I understand your point. However since `val` refers to the number of midi channels in a basic channel group, I would vote for putting this parameter to `fluid_synth_get_basic_channel()`, i.e.:

```
int fluid_synth_get_basic_channel(fluid_synth_t* synth,
                                 int chan,
                                 int *basic_chan_out
                                 int *val);
```

Semantically this seems to be the better way to go, imo.

jjceresa:
The command `channelsmode `needs both `mode ` and `val ` also.
In fact an API returning only `mode ` value is not usable and 
an API returning only the` basic channel` that chan belongs to without returning both `mode `and `val `is not usable also.

The only API that is useful is the one that return the `mode` of the group that `chan `belongs to and also the `val` value of the basic channel of this group. Eventually this API can return also the `basic channel` chan belongs to.

So if we want to keep the name `fluid_synth_get_basic_channel` it  must return the `mode `of chan also 
```
int fluid_synth_get_basic_channel (fluid_synth_t* synth, 
           int chan,
           int *basic_chan_out, // the returned basic channel chan belongs to (if any)
           int *mode_chan_out,// returned mode of chan
           int *basic_val_out);  // returned val of basic channel
```

This function API is sufficient we don't need `fluid_synth_get_channel_mode()` any more.
So the set API could be reduced to:
- int fluid_synth_reset_basic_channels(synth, chan)
- int fluid_synth_set_basic_channel(synth, chan,mode,val)
- int fluid_synth_get_basic_channel (synth, chan, *basic_chan_out, *mode_chan_out, * basic_val_out)

derselbst:
Ok, then let's do it that way.

Note: The `*_out` params of `fluid_synth_get_basic_channel()` should allowed to be NULL, in case some info is not needed.

jjceresa:
>The *_out should allowed to be NULL.

Ok.

derselbst:
I cleaned it up a little, [pls have a look](https://github.com/FluidSynth/fluidsynth/compare/15a52f6...0689257). As it seems `fluid_synth_set_basic_channel()` can also be (ab)used to change the `val` of a basic chan group (not only the `mode`). Is this intended? Is it a use-case?

jjceresa:
> As it seems fluid_synth_set_basic_channel() can also be (ab)used to change the val of a basic chan group (not only the mode). Is this intended? Is it a use-case?

Before my last changes, the API will be able to determine internally the `number of channel` depending of the requested `mode`. For example in `omnion `(poly or mono) `the number of channel `determined was all channels `from the basic channel `to the `next basic channel -1`(if any) otherwise to the `MIDI channel count  -1` . It seems i have stripped this functionality and it should be restored.
I will review this and your change too.

jjceresa:
In summary, now from the last changes (yours 15a52f6  to 0689257 and  mine 13a8c6c  ) we have a `basic channel API` that look clear for the caller:
- (1) `fluid_synth_reset_basic_channel(), fluid_synth_set_basic_channel()`.
- (2) `flluid_synth_get_basic_channel()`.

The couple `fluid_synth_reset_basic_channel(), fluid_synth_set_basic_channel() `allows the application to manage the layout of `full MIDI channels set` in basic channel groups in a flexible way fully compliant with the MIDI specs.
That is:
- (1) `fluid_synth_set_basic_channel()` not only allows to set a `new basic channel group`, but it allows to changes an `existing basic channel group`. Changing an existing basic channel group is flexible (as it `clears` automatically the basic channel before doing the change), and coherent with the behaviour of MIDI `CCs` `omnioff`, `omnion`, `mono`,`poly`.
- Each time `fluid_synth_set_basic_channel()` specify a basic channel that overlaps any `neighbour existing basic channel group` the function fails . This occurs when the `basic channel parameter` overlaps an `existing basic channel group below`, or when `val parameter` overlaps an `existing basic channel group above`. Theses overlapping cases can be corrected using `fluid_synth_reset_basic_channel()` that allows to clear (i.e removing) any `existing basic channel group`.

-(2)  `fluid_synth_get_basic_channel()` allows to get information about any `MIDI channels` regardless it is a `basic channel` or `not`. These information can be used for general purpose cases like `displaying` on `human interface screen` or `saving` basic channel group in `permanent storage`.

derselbst:
I still dont quite like `fluid_synth_set_basic_channel()`. It is too complex for my taste, specifically that it allows to change existing basic channels by silently clearing them out. That's the purpose of `fluid_synth_reset_basic_channel()` and the user application should be compelled to reset this basic channel itself, because if we did it silently this hides functionality and makes it hard to understand what's going on from a user perspective.

So IMO `fluid_synth_set_basic_channel()` should only **set a new basic channel group**, i.e. it should only succeed if all the channels that are requested to be changed are disabled. Else it should fail. I think this is the most transparent and best maintainable approach.

The current solution is pretty hard to understand, maintain and test. Imagine:

```c
// init some plain new basic channel groups
fluid_synth_set_basic_channel(synth, 0, 0, 4);
fluid_synth_set_basic_channel(synth, 4, 0, 2);

// oh, actually I wanted the first group to have 5 channels
ret = fluid_synth_set_basic_channel(synth, 0, 0, 5);

// ret == FLUID_FAILED
// oops, there is an overlapp ofc, damn...
// ... well nevermind, I can live with 4 channels as well
// now let's play some notes
ret = fluid_synth_noteon(synth, 0, 60, 127);

// oh no, something went terribly wrong:
// ret == FLUID_FAILED
// well the last call to fluid_synth_set_basic_channel() silently
// disabled the whole group instead of leaving it untouched
```

Those kinds of bugs I fear most when API functions are doing too many magic things under the hood.

jjceresa:
>the user application should be compelled to reset this basic channel itself, because if we did it silently this hides functionality and makes it hard to understand what's going on from a user perspective.

Ok , i understand your point .

>So IMO fluid_synth_set_basic_channel() should only set a new basic channel group, i.e. it should only succeed if all the channels that are requested to be changed are disabled. Else it should fail. I think this is the most transparent and best maintainable approach.

Ok, i understand this will done.

derselbst:
Sry, I dont like the `enable_change` flag either. We should consistently use our own API internally, as proof-of-concept. I mean it could turn out, that my proposal of this API is completely useless, which we would never recognize, if we do our own custom things internally.

jjceresa:
> I mean it could turn out, that my proposal of this API is completely useless, which we would never recognize, if we do our own custom things internally.

I am not sure to understand what you prefer?.

The MIDI spec for  `MIDI CCs omnioff, omnion, mono,poly `requires  `that fluid_synth_cc()` are able to change an existing basic channel (actually via `fluid_synth_set_basic_channel_LOCAL ()`) . But now we have an API  `fluid_synth_set_basic_channel() `that denies this possibility. Effectively this is not consistent.

Do you think that it is better for `fluid_synth_set_basic_channel_LOCAL () `to have the same limitation that  `fluid_synth_set_basic_channel()` ?. Consequently in this case, `On CCs omnioff, omnion, mono,poly` we should be forced to `call fluid_reset_basic_channel()` before calling  `fluid_synth_set_basic_channel_LOCAL ()` ? .

derselbst:
> Do you think that it is better for fluid_synth_set_basic_channel_LOCAL ()to have the same limitation that fluid_synth_set_basic_channel() ?.

Yes.

> we should be forced to call fluid_reset_basic_channel() before calling fluid_synth_set_basic_channel_LOCAL ()

Right.

> but we could not do this as we don't know if fluid_synth_set_basic_channel_LOCAL () will fail ?

Why should it fail? The current basic channel group has been cleared out, there should be enough room to set a new group. And in case there isnt, the current implementation would fail as well, wouldnt it?

jjceresa:
> Why should it fail? The current basic channel group has been cleared out, there should be enough room to set a new group.And in case there isnt, the current implementation would fail as well, wouldnt it?

Example: let  2 basichannels:  `0 mono_ominoff  3 channels`    and    `3 mono_omioff  6 channels`

- (1)Now, we send `CC  mono 4`  to `channel basic  0 `using `fluid_synth_cc()`. The new val value is 4 . Current `fluid_synth_set_basic_channel_LOCAL ()` will fail as there is an overlap with channel basic 3, but basic channel 0 is  left untouched. `fluid_synth_cc()` return FLUID_FAILED

- (2)Now, we send `CC  mono 2` to `channel basic  0` using `fluid_synth_cc()`. The new val is 2.
Current `fluid_synth_set_basic_channel_LOCAL ()` will accept the change as there is no overlap.

I we call `fluid_reset_basic_channel() `before calling `fluid_synth_set_basic_channel_LOCAL()`, in case (2),  `fluid_synth_set_basic_channel_LOCAL ()` will fail. But now basic channel 0 is cleared. This is a bug as` CC mono` is only intended to change an existing basic channel not to clear this channel.

derselbst:
Couldnt we check for case (1)?

```c
case POLY_OFF: /* Mono On */
    int basic_chan, old_mode, old_val;
    if (fluid_synth_get_basic_channel_LOCAL(synth, channum, &basic_chan, &old_mode, &old_val) == FLUID_OK &&
        baisc_chan == channum && /* allowed only if channum is a basic channel */
        value <= old_val) /* and if there is enough space for a new group */
    {
        int new_mode;
        if(!(old_mode & FLUID_CHANNEL_OMNI_OFF)) /* channel is actually Omni On */
        {	/* sets channel channum in mode mono omnion (1) */
            /* MIDI specs: value is 0 that means all channels to MIDI channel count -1.
                However, if value is > 0 (i.e 4) , the group of channels will be be
                limited to 4. */
            new_mode = FLUID_CHANNEL_MODE_OMNION_MONO; 
        } 
        else /* channel is actually Omni Off */
        {	/* sets channel channum in mode mono omnioff (3) */
            /* MIDI specs: value is the number of channels for mode mono omnioff */
            new_mode = FLUID_CHANNEL_MODE_OMNIOFF_MONO;
        }

        /* Changes an existing basic channel group by resetting the current and setting a new one */
        fluid_synth_reset_basic_channel_LOCAL(synth, channum);
        if(fluid_synth_set_basic_channel_LOCAL(synth, channum, new_mode, value) == FLUID_FAILED)
        {
            FLUID_LOG(FLUID_ERR, "THIS SHOULD NEVER HAPPEN: fluid_synth_set_basic_channel_LOCAL failed :(");
            /* now, we could try to enforce the old setting of this group */
            fluid_synth_set_basic_channel_LOCAL(synth, channum, old_mode, old_val);
            return FLUID_FAILED;
        }
        return FLUID_OK;
    }
    else  
    {
        return FLUID_FAILED;
    }
    break;
```

derselbst:
Oh, ofc the check is not as trivial as I initially though, so thanks for getting it right. Now the only nitpick I have is that `fluid_synth_check_next_basic_channel()` is called twice (once in `fluid_synth_cc_LOCAL()` and again in `fluid_synth_set_basic_channel_LOCAL()`).
How about moving this if clause:
https://github.com/FluidSynth/fluidsynth/blob/7976be599a4940ec0dc4420954ef8cfe310dd386/src/synth/fluid_synth_polymono.c#L208-L215
 to  `fluid_synth_set_basic_channel()` then? As far as I see internally we can always call `fluid_synth_set_basic_channel_LOCAL()` with a specific  non zero `val`.

jjceresa:
Yes, moving `fluid_synth_check_next_basic_channel() `to `fluid_synth_set_basic_channel()` is functionally equivalent so it will to be done.

derselbst:
Ok, I cleaned it up a little, please have a look.

src/synth/fluid_synth_mono.c
```diff
@@ -0,0 +1,683 @@
...
+					case FLUID_CHANNEL_LEGATO_MODE_MULTI_RETRIGGER: /* mode 1 */
+						/* Skip in attack section */
+						fluid_voice_update_multi_retrigger_attack(voice,tokey,vel);
+					break;
+					default: /* Invalid mode */
+					FLUID_LOG(FLUID_WARN, 
+						"Failed to execute legato mode: %d",legatomode);
+					return FLUID_FAILED;
+				}
+				if (legatomode >= FLUID_CHANNEL_LEGATO_MODE_MULTI_RETRIGGER)
```
derselbst: I think this if clause should now be moved to the previous switch-case.

src/synth/fluid_synth_mono.c
```diff
@@ -0,0 +1,683 @@
...
+                            char Mono)
+{
+	int status = FLUID_FAILED;
+	fluid_voice_t* voice;
+	int i;
+	fluid_channel_t* channel = synth->channel[chan];
+	/* Key_sustained is prepared to return no note sustained (-1) */
+	if (Mono)
+	{
+		channel->key_mono_sustained = -1; /* no mono note sustained */
```
derselbst: Why does `key_mono_sustained` use -1 as invalid not and not `INVALID_NOTE`?

jjceresa: This is inconsistent , key_mono_sustained should use INVALID_NOTE. This will be corrected.

src/synth/fluid_synth_mono.c
jjceresa: Right. It will be done.

derselbst:
Generally I'm fine with this now. In the next days / weeks I'll test the new features I just merged into master. Then I'll come back to polymono.

jjceresa:
> In the next days / weeks I'll test the new features.

For infos, here the new basic channel API` (fluid_synth_set_basic_channel(), fluid_synth_reset_basic_channel(), fluid_synth_get_basic_channel())` have been tested and works as intended. Idem for the corresponding shell commands.

jjceresa:
Now as the `fluid_synth_set_basic_channel()` is intended to `set a new` basic channel group but `not to change an existing` basic channel,  the actual name has become inappropriate . I would prefer `fluid_synth_add_basic_channel()` that seems a bit more significant than using `set` verb. Any thought ?

derselbst:
I'm open to renaming this. But if you do I would also recommend to rename
fluid_synth_reset_basic_channel() to fluid_synth_remove_basic_channel().

jjceresa:
Ok for the renaming of `set/reset` to `add/remove`. Of course this should be applied also to the corresponding shell commands `setbasicchannels, resetbasicchannels` renamed `addbasicchannels removebasicchanels`.

jjceresa:
Since the most recent master merge https://github.com/FluidSynth/fluidsynth/pull/306/commits/49c1ee39d2ef525ac0af22ab22a4d8f8d49ab0e9 Travis build find now a lot of warnings like this. 

```
/home/travis/build/FluidSynth/fluidsynth/src/rvoice/fluid_adsr_env.h:32:15: warning: padding struct to align ‘coeff’ [-Wpadded]
  fluid_real_t coeff;
```
I don't think this issue is related to this merge !?. Any opinion ?

derselbst:
You can safely ignore it for now. I want to start refactoring rvoice_mixer
after having polymono merged and also looking into padding of those struct
members.

jjceresa:
I abandon the renaming of fluid_synth_set_basic_channel(), fluid_synth_reset_basic_channel(). This seems fine and well documented.

jjceresa:
> Do you think there is anything left to be done?

In the actual state i don't see anything that need to be done. The pdf history resumes the pertinent changes compared to the initial patch. So i think that squashing can be safely done. ( i am curious about squashing strategy and commands invoked, so many thanks for logging this).

jjceresa:
>How did you test (extensive portamento use or none) ?.

none portamento.
>I guess you used the VC++6 compiler with optimizations on?

Yes, optimization was on and when `speed optimization` is on the lost of performance is -21.8%
Same results using VC10.

The measurement has be done using the profiling patch (for easy measurement). (A patch for v2.0 will be ready sooner). The measurement is based on `relative voice duration` in percent.
To get  pertinent compare it was necessary to run a `Release `or `RelMinSize` version but not a `Debug build` that introduces `higher voice duration` masking any possible small duration measurement.

The compare (between 2 algorithms a1,a2) was done without portamento (i.e when dsp.pitchinc is 0) in 3 steps like this. (Build RelMinSize):
0) A reference measurement `without` the portion of portamento code.
1) Measurement `with the former algo. "slower"`. The lost of performance is -12% relative to (0) (without speed optimization) , - 21.8 (with speed optimization).
2) Measurement `with the recent algo. "faster"`. The lost of performance is -0.5% relative  to (0).

>They are nearly the same. However the former, slower version seems to have higher register pressure, >as it uses up to four floating point regs, 
```
// (1.a) when pitchinc is 0, E1, E2, E4 are always evaluated (slower than 2.a).
// (1.b) when pitchinc is > 0, E1, E2, E3 , E4 are always evaluated until portamento is finished (slower than 2.b).
// (1.c) when pitchinc is > 0 and portamento is finished, E1, E2, E3, E6 are evaluated (same as  2.c).
// (1.d) when pitchinc is < 0, E1, E2, E4 are always evaluated until portamento is finished (same as 2.d).
// (1.e) when pitchinc is < 0 and portamento is finished  E1, E2, E4, E5, E6 are evaluated (same as 2.e).
voice->dsp.pitchoffset += voice->dsp.pitchinc; /* E1 */
if ((voice->dsp.pitchinc > 0.0f /* E2 */ && voice->dsp.pitchoffset > 0.0f) /* E3 */ ||
    (voice->dsp.pitchinc < 0.0f /* E4 */ && voice->dsp.pitchoffset < 0.0f /* E5 */))
    {
	voice->dsp.pitchoffset = voice->dsp.pitchinc = 0.0f; /* E6 */
    }

```
Please note 1.a /2.a  the  useless expressions E1 (when portamento is disabled )   and  1.b/2.b E4 always evaluated when portamento is running and not yet finished (i.e E3 false).

>while your version is satisfied with three. Might this be the problem? 

It seems you point right.
```
// (2.a) when pitchinc == 0, E2,E4 are always evaluated (faster than 1.a).
// (2.b) when pitchinc  > 0, E2,E1,E3 are always evaluated until portamento is finished (faster than 1.b).
// (2.c) when pitchinc >0 and portamento is finished, E2,E1,E3, E6 are evaluated (same as  1.c).
// (2.d) when pitchinc is < 0, E2, E4, E1 are always evaluated until portamento is finished (same as 1.d).
// (2.e) when pitchinc is < 0 and portamento is finished  E2, E4, E1, E5, E6 are evaluated (same as 1.e).
if (voice->dsp.pitchinc > 0.0f) /* E2 */
{	/* portamento is enabled, so update pitchoffset */
	voice->dsp.pitchoffset += voice->dsp.pitchinc; /* E1 */
	/* when pitchoffset reaches 0.0f, portamento is disabled */
	if (voice->dsp.pitchoffset > 0.0f) /* E3 */
		voice->dsp.pitchoffset = voice->dsp.pitchinc = 0.0f; /* E6 */
}
else if (voice->dsp.pitchinc < 0.0f) /* E4 */
{	/* portamento is enabled, so update pitchoffset */
	voice->dsp.pitchoffset += voice->dsp.pitchinc; /* E1 */
	/* when pitchoffset reaches 0.0f, portamento is disabled */
	if (voice->dsp.pitchoffset < 0.0f) /* E5 */ 
		voice->dsp.pitchoffset = voice->dsp.pitchinc = 0.0f; /* E6 */
}
```

Following the disasm, find the two versions attached.
[fluid_rvoice_asm_slow.txt](https://github.com/FluidSynth/fluidsynth/files/1773416/fluid_rvoice_asm_slow.txt)
[fluid_rvoice_asm_fast.txt](https://github.com/FluidSynth/fluidsynth/files/1773417/fluid_rvoice_asm_fast.txt)

derselbst:
I see. And unlike to my machine, yours seems to always push the floats back to the stack and re-read it from there which in combination with the addition of 0.0 gets expensive. Lesson learned, thx.

derselbst:
FYI: I did the manual squash as follows:

```bash
git checkout master
git pull
git checkout polymono
git pull
git merge master # ensure polymono is up to date with master
git checkout master
git merge --squash polymono # all changes from polymono branch are now put into staging
git commit --author="jjceresa <jjc_fluid@orange.fr>"
```

Not sure why github doesnt recognize this custom squash though...

mawe42:
@derselbst , @jjceresa Thank you both very much for this great addition to FluidSynth and your long and hard word getting the patch into shape! Great stuff!

PR #312, opened 2017-12-19
Last explicit cast seen.

jjceresa:
Explicit casts to avoid narrowing conversion warnings.


derselbst:
I dont like it. It makes this macro magic even more complicated. But I'll take it and will probably take care of that in #197, by using unions and the same function signature for every `rvoice_event_handler` function. That will hopefully allow us to get rid of this huge switch case and call the functions directly.

PR #314, opened 2017-12-23
Thread-safe driver selection.

carlo-bramini:
In my opinion, there is no need to waste so much RAM with 2 copies of the list of available drivers.
Actually, you need just one bit for each driver to know if it is activated or not.
While I was on the topic, I took the chance to make the driver selection thread-safe.

new_fluid_audio_driver() and new_fluid_audio_driver2() are also almost identical, so I introduced a find_fluid_audio_driver() that searches the selected driver, so the above 2 functions are reduced to few lines.

### DIFF
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -203,11 +203,10 @@ static const fluid_audriver_definition_t fluid_audio_drivers[] =
 #endif
 };
 
-#define ENABLE_AUDIO_DRIVER(_drv, _idx)     _drv[(_idx) / 32] &= ~(1 << ((_idx) % 32))
-#define IS_AUDIO_DRIVER_ENABLED(_drv, _idx) (!(_drv[(_idx) / 32] & (1 << ((_idx) % 32))))
+#define ENABLE_AUDIO_DRIVER(_drv, _idx)     _drv[(_idx) / 8] &= ~(1 << ((_idx) % 8))
+#define IS_AUDIO_DRIVER_ENABLED(_drv, _idx) (!(_drv[(_idx) / 8] & (1 << ((_idx) % 8))))
 
-static fluid_mutex_t fluid_adriver_mutex = FLUID_MUTEX_INIT;
-static uint32_t      fluid_adriver_enabled[FLUID_N_ELEMENTS(fluid_audio_drivers)] = {0};
+static uint8_t fluid_adriver_disable_mask[(FLUID_N_ELEMENTS(fluid_audio_drivers)+7)/8] = {0};
 
 void fluid_audio_driver_settings(fluid_settings_t* settings)
 {
@@ -288,16 +287,12 @@ void fluid_audio_driver_settings(fluid_settings_t* settings)
   fluid_settings_add_option(settings, "audio.driver", "file");
 #endif
 
-  fluid_mutex_lock(fluid_adriver_mutex);
-
   for (i = 0; i < FLUID_N_ELEMENTS(fluid_audio_drivers); i++) {
     if (fluid_audio_drivers[i].settings != NULL &&
-        IS_AUDIO_DRIVER_ENABLED(fluid_adriver_enabled, i)) {
+        IS_AUDIO_DRIVER_ENABLED(fluid_adriver_disable_mask, i)) {
       fluid_audio_drivers[i].settings(settings);
     }
   }
-
-  fluid_mutex_unlock(fluid_adriver_mutex);
 }
 
 static const fluid_audriver_definition_t*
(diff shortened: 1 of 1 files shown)


derselbst:
Yes, I have to admit that I did not really care about memory when implementing this, so basically you're right.

However your change actually breaks the functionality introduced by #264: It was intended to call the `_fluid_audriver_definition_t::settings()` functions of only those audio drivers previously registered with `fluid_audio_driver_register()`. However you're initializing all available drivers in `fluid_audio_driver_settings()` ignoring to what `fluid_adriver_enabled` is set to. You should instead:

```
if (fluid_audio_drivers[i].settings != NULL
    && IS_AUDIO_DRIVER_ENABLED(fluid_adriver_enabled, i)) {
   fluid_audio_drivers[i].settings(settings);
}
```

And since calling `fluid_audio_driver_register()` is optional, you must ensure that all available drivers are initialized by default, i.e. at compile-time. Thus perhaps only use a single global

`static uint32_t fluid_adriver_enabled = 0xFFFFFFFFU;`

i.e. not an array. Having bits for 32 audio drivers should be more than enough.

carlo-bramini:
Thank you, actually I misunderstood the calling order of those functions.
I reversed the logic of the fluid_adriver_enabled[], so if the bit is zero, then the driver is activated. In this way, since I forced the array filled with zero at compile time, any number of drivers will be allowed.

derselbst:
> I reversed the logic of the fluid_adriver_enabled[], so if the bit is zero, then the driver is activated.

Talking about bitmasks this is IMO a bit obfuscated. We should call it `fluid_adriver_disabled_mask` then. And speaking about memory: when fluidsynth is built with 4 adrivers

`static uint32_t fluid_adriver_enabled[FLUID_N_ELEMENTS(fluid_audio_drivers)]`

would have 4 integers (==16 bytes), although only the first integer would ever be accessed. If you really want this generic approach, why not:

`static uint8_t fluid_adriver_disabled_mask[(FLUID_N_ELEMENTS(fluid_audio_drivers) / (sizeof(uint8_t) * 8u + 1u)) + 1u]`

Why do we need `fluid_adriver_mutex`?

* I dont think it's a use-case to create `settings` objects concurrently (and therefore cause race conditions in `fluid_audio_driver_settings()`).
* When calling `new_fluid_audio_driver*()`, the threads operate on their own `settings` struct.
* And again I dont think it's a use-case to call `new_fluid_audio_driver*()` concurrently with the same `settings` instance.
* It is also not allowed to modify `fluid_adriver_disabled_mask` when already working with fluidsynth's API.

Or am I missing something?

carlo-bramini:
Yes, you are right, I forgotten to rescale the size of the array to bit size.
I will also rename it to fluid_adriver_disabled_mask[] if you think it is a more meaningful.
I added a mutex because I had the suspect that things may not work if the user tries to use fluid_audio_driver_register() concurrently with the new_fluid_audio_driver(). But since you wrote that it not allowed to modify fluid_adriver_disabled_mask[] in that way, I think I can remove that mutex.

src/drivers/fluid_adriver.c
```diff
@@ -120,149 +120,93 @@ void delete_fluid_file_audio_driver(fluid_audio_driver_t* p);
...
-    AVAILABLE_AUDRIVERS
+#endif
 };
 
-static fluid_audriver_definition_t fluid_audio_drivers[] =
-{
-    AVAILABLE_AUDRIVERS
-};
+#define ENABLE_AUDIO_DRIVER(_drv, _idx)     _drv[(_idx) / 8] &= ~(1 << ((_idx) % 8))
+#define IS_AUDIO_DRIVER_ENABLED(_drv, _idx) (!(_drv[(_idx) / 8] & (1 << ((_idx) % 8))))
```
derselbst: I dont like hard coding these special values like `8`. I'd prefer

```
_drv[(_idx) / (sizeof(*_drv)*8)] &= ~(1 << ((_idx) % (sizeof(*_drv)*8)))
```

src/drivers/fluid_adriver.c
```diff
@@ -476,49 +434,42 @@ delete_fluid_audio_driver(fluid_audio_driver_t* driver)
  */
 int fluid_audio_driver_register(const char** adrivers)
 {
-    unsigned int i=0, add=0;
-    int res = FLUID_FAILED;
+    unsigned int i;
+    uint8_t      disable_mask[(FLUID_N_ELEMENTS(fluid_audio_drivers)+7)/8];
```
derselbst: `uint8_t disable_mask[(FLUID_N_ELEMENTS(fluid_adriver_disable_mask)];`

src/drivers/fluid_adriver.c
```diff
@@ -476,49 +434,42 @@ delete_fluid_audio_driver(fluid_audio_driver_t* driver)
...
     
-    if(adrivers == NULL)
-    {
-        res = FLUID_OK;
-        goto cleanup;
-    }
-    else
+    if (adrivers == NULL)
+      return FLUID_OK;
```
derselbst: From the doc:

`@param adrivers [...] Pass NULL to register all available drivers.`

src/drivers/fluid_adriver.c
```diff
@@ -476,49 +437,46 @@ delete_fluid_audio_driver(fluid_audio_driver_t* driver)
...
-                    break;
-                }
-            }
-            
-            if(j >= FLUID_N_ELEMENTS(fluid_audio_drivers_template))
+            if (FLUID_STRCMP(adrivers[i], fluid_audio_drivers[j].name) == 0)
             {
-                /* requested driver not found, failure */
-                goto cleanup;
+                ENABLE_AUDIO_DRIVER(disable_mask, i);
```
derselbst: `ENABLE_AUDIO_DRIVER(disable_mask, j);` otherwise wont work.

derselbst:
Ok, now it works. And it's the better solution, thanks!

PR #315, opened 2017-12-24
Misc fixes 3

carlo-bramini:
I could test "Avoid zero-size array if MIDI input unavailable." but not the others, unfortunately.
But I think that they are quite easy and simple, so I providing them anyways.

### DIFF
--- a/src/drivers/fluid_jack.c
+++ b/src/drivers/fluid_jack.c
@@ -28,7 +28,6 @@
  *
  */
 
-#include "fluid_sys.h"
 #include "fluid_synth.h"
 #include "fluid_adriver.h"
 #include "fluid_mdriver.h"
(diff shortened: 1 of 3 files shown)


derselbst:
So this is only to avoid compilation error if no MIDI? What do you mean by "the others" then?

carlo-bramini:
> So this is only to avoid compilation error if no MIDI?

20b7143  "Avoid zero-size array if MIDI input unavailable." does it, it's tested.

> What do you mean by "the others" then?

771fa81 "Remove redundant include."
806b3d6 "Check for JACK_SUPPORT"
74122f2 "Remove redundant include."
c6a2b0c "Use FLUID_MUTEX_INIT instead of G_STATIC_MUTEX_INIT"

I could not test them on windows. Tested only partially on CYGWIN. But I think that they are fine, so I provided them for review.

derselbst:
> I could not test them on windows.

Oh ok. I misunderstood it as "test during compilation" which confused me. So yes, "the others" are fine I think, but I'll only cherry-pick 20b7143ba962024ada82367626406a1f0d14d087 to 1.1.9. The others only for `master`.

PR #316, opened 2017-12-27
Introduce fluid_msleep function.

carlo-bramini:


src/utils/fluid_sys.c
```diff
@@ -344,6 +344,19 @@ fluid_is_soundfont(const char *filename)
...
 }
 
+/**
+ * Suspend the execution of the current thread for the specified amount of time.
+ * @param milliseconds to wait.
+ */
+void fluid_msleep(unsigned int msecs)
+{
+#ifdef _WIN32
+  Sleep(msecs);
```
derselbst: I dont see a necessity of having this special treatment for WIN32.

carlo-bramini: You are right, it is not a strict necessity... since Windows Sleep() handles directly milliseconds, I introduced it and I avoided x1000, but if it looks unnecessary, we can fall down to g_usleep().

derselbst: Yes please. glib is doing all that platform dependent handling. We only move it to fluid_sys.c to ease getting rid of it for downstream if needed.

PR #318, opened 2018-01-06
Attenuation modes

mawe42:
Implements the changes discussed in #70 and 29f6ac1. This is explicity an RFC pull-request... still not sure if this functionality actually adds anything of value to Fluidsynth.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -259,12 +259,12 @@ https://stackoverflow.com/a/6251757
             </desc>
         </setting>
         <setting>
-            <name>volenv</name>
+            <name>attenuation-mode</name>
             <type>str</type>
             <def>emu</def>
-            <vals>compliant, emu</vals>
+            <vals>compliant, emu, timidity</vals>
             <desc>
-                Specifies the kind of volume envelope processing. This esp. influences the way fluidsynth responses to noteon velocity. The default setting 'emu' causes the envelope to be highly dynamic (i.e. compatible with the EMU10K1). Alternatively this may be set to 'compliant' for a less dynamic envelope, as it was done before fluidsynth 1.0.9. Note that this setting can only be changed until the first synth has been created. Changing it afterwards will have no effect for the rest of fluidsynths lifetime.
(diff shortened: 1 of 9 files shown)


derselbst:
Ok, so far I accomplished to get the timidity curve by solely manipulating default modulators. Ofc this requires modifying all default modulators that have the GEN_ATTENUATION destination generator, namely velocity, CC7 and CC11. Simply setting an amount of `960 * 0.4` is sufficient to have the same volume response like timidity. Here my thoughts:

### The manip. def. mod. attempt

#### PROs
* great freedom of designing custom volume responses
* one doesnt need to patch every single soundfont file applying the same modulator(s) to every single instrument
* it would allow us to get rid of that `if(num==GEN_ATTENUATION) { do_work_depending_on_atten_setting() }` stuff (that 0.4 factor could simply be hard-coded when initially reading in soundfont files and processing it's initial attenuation, i.e. no need to do that every time when calculation voice attenuation)

#### CONs
* only skilled users will be able to do that
* ergo it cannot be quickly applied to see (hear) the effects
* it does not work for user-defined modulators in soundfont files, they would remain untouched (or is that a pro?)

### The `synth.attenuation-mode` setting attempt

#### PROs
* very user friendly, doesnt require deep knowledge of fluidsynth
* allows to quickly switch between modes  and potentially test which mode works best for certain midi files 
* this attempt would also affect user-defined modulators in soundfont files (or is that a con?)

#### CONs
* very limited freedom
* could encourage people to create midi files with a strange non standard volume response
* does any "normal" user acutally need this??

mawe42:
> - could encourage people to create midi files with a strange non standard volume response

Well, there are so many different volume responses in software synths out there, so that ship has already sailed, I think.

> - does any "normal" user acutally need this??

This is my main gripe with this attenuation-mode setting. There hasn't been a request for this switch for over 7 years. And EMU behaviour is the defacto standard, it seems.

We could just remove this and only address it if and when somebody comes forward with a new request. For now simply clean up the handling of the factor (apply it on soundfont load, as you say), remove atten2amp, increase cb2amp range to 1440 and be done with it.

derselbst:
> For now simply clean up the handling of the factor (apply it on soundfont load, as you say), remove atten2amp, increase cb2amp range to 1440 and be done with it.

Alright, let's do it that way. Could you propose another PR pls? :)

mawe42:
> Alright, let's do it that way. Could you propose another PR pls? :)

Will do, but probably not before next week.

mawe42:
Ok, the #324 is the new PR. Closing this one.

PR #321, opened 2018-01-06
Simplified code for modulation waveforms

carlo-bramini:
When I finally run FluidSynth on an embedded ARM Cortex M3, I already expected not so good performance since it has not an hardware floating point unit. Surprisingly, it required more than 20 seconds to reaching the prompt on the console, too much if I think that, afterall, it is clocked at 120MHz.
After a bit of debugging, I discovered that the bottleneck is the calculation of the modulation waveform.
I simplified the code by extracting the calculations that never changed during the loop, this has improved the speed results a lot, by decreasing that time to 8 seconds. Evidently, the calculation of the sin() value is the most expensive one and it cannot be avoided. I applied the same rewrite also for the triangular waveform. Since the "depth" parameter cannot be bigger than MAX_SAMPLES, it could be possible to calculate fluid_chorus_triangle() with an integer implementation, but unlike the the fluid_chorus_sine(), it is calculate in a blink of eye, even with soft-float, the prompt appears immediately after poweron, so I think there are no problems to leave it as is.

### DIFF
--- a/src/rvoice/fluid_chorus.c
+++ b/src/rvoice/fluid_chorus.c
@@ -479,13 +479,22 @@ static void
 fluid_chorus_sine(int *buf, int len, int depth)
 {
   int i;
-  double val;
+  double angle, incr, mult;
 
+  /* Pre-calculate increment between angles. */
+  incr = (2. * M_PI) / (double)len;
+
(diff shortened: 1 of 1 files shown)


derselbst:
Looks like they are equivalent. Ok thanks.

PR #324, opened 2018-01-08
Cleanup initial attenuation generator value handling

mawe42:
The EMU8k/10k hardware applies a 0.4 factor to all initially set
attenuation generator values. Do this once on soundfont load.

Remove the invalid -531.509 power factor that was computed from Timidiy's
non-standard conversion table. And remove the synth.volenv switch again,
as the "compliant" setting wasn't compliant at all.

This allows us to get rid of the atten2amp function and use cb2amp
everywhere, with a range of 0 1440 centibels.

Related to #70 and #318.


derselbst:
Ok, great. Should be fine now.

PR #328, opened 2018-01-19
No need to do extra addition at every cycle.

carlo-bramini:


derselbst:
Ok. Just let me nit-pick that the triangle now starts at -786431, whereas before it was -786432. Doesnt matter, never mind.

carlo-bramini:
Actually, it is the rounding to be wrong.
It should add 0.5 when the number is positive, and subtract 0.5 when it is negative.

derselbst:
Yes, if you want you can change it. Need to wait for the macos CI build anyway.

carlo-bramini:
However, it seems to me that the values will be always negative.
In function fluid_chorus_set(), the value of "modulation_depth_samples" cannot be bigger than MAX_SAMPLES. Actually, that is the "depth" parameter passed to fluid_chorus_triangle() and fluid_chorus_sine(). Just doing a subtraction (instead of current addition) with 0.5 rather than testing the sign of the number seems to be enough.

PR #329, opened 2018-01-19
implement 24 bit sample support for fluid_voice_optimize_sample()

derselbst:
In addition to #301, `fluid_voice_optimize_sample()` should calculate the noise floor using 24 bit samples as well. Otherwise very quiet voices could be turned off too early.

@mawe42 Do you agree with `FLUID_NOISE_FLOOR`?

### DIFF
--- a/src/rvoice/fluid_rvoice.h
+++ b/src/rvoice/fluid_rvoice.h
@@ -36,9 +36,10 @@ typedef struct _fluid_rvoice_t fluid_rvoice_t;
 
 /* Smallest amplitude that can be perceived (full scale is +/- 0.5)
  * 16 bits => 96+4=100 dB dynamic range => 0.00001
- * 0.00001 * 2 is approximately 0.00003 :)
+ * 24 bits => 144-4 = 140 dB dynamic range => 1.e-7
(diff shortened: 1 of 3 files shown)


src/rvoice/fluid_rvoice.h
```diff
@@ -36,9 +36,10 @@ typedef struct _fluid_rvoice_t fluid_rvoice_t;
 
 /* Smallest amplitude that can be perceived (full scale is +/- 0.5)
  * 16 bits => 96+4=100 dB dynamic range => 0.00001
- * 0.00001 * 2 is approximately 0.00003 :)
+ * 24 bits => 144-4 = 140 dB dynamic range => 1.e-7
+ * 1.e-7 * 2 == 2.e-7 :)
  */
-#define FLUID_NOISE_FLOOR 0.00003
+#define FLUID_NOISE_FLOOR 2.e-7
```
jjceresa: On 24 bits the noise is -(144 - 6) = -138 dB. So noise floor amplitude =  is 10^(-138/20) = 1.26e-7. Saying that we could keep 2.e-7.  :).

derselbst: Ok, thanks :)

PR #331, opened 2018-01-20
Additional general-purpose IIR filter

derselbst:
This is an RFC PR for adding and exposing an additional general-purpose IIR filter.

#### Motivation

1. @mawe42 Needs a custom high-pass filter conforming to the SF2 spec.
1. I need a custom low-pass filter not compliant with SF2 to reproduce certain MIDI files I have around.

#### Solution

Introduce a general-purpose IIR filter, that the user can configure with `fluid_synth_set_custom_filter()` and control with default modulators, see this minimal usage example:

```
// set a custom low pass filter that uses a linear filter Q, switches the
// filter off if Q==0 and does no filter_gain correction depending on Q
fluid_synth_custom_filter(this->synth,
                          FLUID_IIR_LOWPASS,
                          FLUID_IIR_NO_GAIN_AMP | FLUID_IIR_Q_LINEAR);

fluid_mod_t* my_mod = new_fluid_mod();

// add a custom CC34 to custom lowpass filter cutoff freq.
{
    fluid_mod_set_source1(my_mod, 34,
                FLUID_MOD_CC
                | FLUID_MOD_SIN
                | FLUID_MOD_UNIPOLAR
                | FLUID_MOD_POSITIVE
                );
    fluid_mod_set_source2(my_mod, 0, 0);
    fluid_mod_set_dest(my_mod, GEN_CUSTOM_FILTERFC);
    fluid_mod_set_amount(my_mod, 10500); // filter cutoff freq. form 0 to 10500
    fluid_synth_add_default_mod(this->synth, my_mod, FLUID_SYNTH_OVERWRITE);
}

// add a custom CC33 to custom lowpass filter Q def mod.
{
    fluid_mod_set_source1(my_mod, 33,
                FLUID_MOD_CC
                | FLUID_MOD_CONCAVE
                | FLUID_MOD_UNIPOLAR
                | FLUID_MOD_POSITIVE
                );
    fluid_mod_set_source2(my_mod, 0, 0);
    fluid_mod_set_dest(my_mod, GEN_CUSTOM_FILTERQ);
    fluid_mod_set_amount(my_mod, 20);
    fluid_synth_add_default_mod(this->synth, my_mod, FLUID_SYNTH_OVERWRITE);
}

delete_fluid_mod(my_mod);
```

This basically initializes the custom filter and adds two default modulators to control the filters cutoff freq. and Q.

The range of the destination generator `GEN_CUSTOM_FILTERFC` is `[0; 22050]` with a default of 0 which makes it easier to calculate the desired cutoff freq using modulators. `GEN_CUSTOM_FILTERQ` can be used as dB as given by the SF2 spec or it can be given a linear range (my use-case), in which case the filter will be temporarily turned off if `Q==0`.

I think it's worth to add a small example of how to use this filter to the API docs, once this is generally accepted.

P.S.: Pls. dont look at the commits, it's a complete history of try and error. This one will be squashed.


jjceresa:
The filter will be created at synth instance level, then it will be controlled by CCs (via default modulators).
Is it a `general purpose sound filter` for each MIDI channels (i.e only one filter by MIDI channel ) ? or 
is it a `new filter added `to the existing filter for each voices belonging to each MIDI channels ?. Sorry i don't understand the expected intend ?.

derselbst:
> or is it a new filter added to the existing filter for each voices belonging to each MIDI channels ?

Yes, that's it, the filter is created at voice level:

https://github.com/FluidSynth/fluidsynth/blob/aa978955f377351fee8997374c00ab698832c050/src/rvoice/fluid_rvoice.h#L148-L155

The intention is to have a separate user-adjustable filter controlled by custom MIDI CCs for any custom use-case without interfering with the default low-pass filter required by the SF2 spec.

jjceresa:
>The intention is to have a separate user-adjustable filter controlled by custom MIDI CCs for any custom use-case without interfering with the default low-pass filter required by the SF2 spec.

In this case only one filter by MIDI Channel is sufficient and there is no need of multiple filters for each voices which is more CPU intensive. Saying that, having a general purpose filter for each voice is a good orientation if later the dsp engine will be intended to be a sfz player. But this is another story.

May be, if this new filter allow `pass band`  and `low pass` it will be possible to consider for an application using MIDI CC 74d dedicated to "Brigthness" ?. Another dedicated CC is 71d "timbre/harmonic intensity" both CC are in the family of "timbre controller". have not yet looked the possibility of these new filter.

derselbst:
> In this case only one filter by MIDI Channel is sufficient and there is no need of multiple filters for each voices which is more CPU intensive.

I dont see how this could fit in the current architecture. Modulators are applied on voice level. Starting to implement some strange hack to make some modulators apply on channel level would be premature optimization. By default this custom filter is off and doesnt cause any additional overhead. And if an application turns it on, it must be aware that there is some sort of overhead.

> May be, if this new filter allow pass band and low pass it will be possible to consider for an application using MIDI CC 74d dedicated to "Brigthness" ?.

The IIR filter currently doesnt support a band-pass filter. This would mean tweaking its implementation by seaparating b0 and b2 filter coefficients, which are currently stored in one variable because they are always the same for high-pass and low-pass.

Any application is free to use whatever MIDI CC it needs by using fluidsynth's API for manipulating default modulators (i.e. `fluid_synth_add_default_mod()`). I dont want to hardcode even more non-standard default modulators in fluidsynth.

jjceresa:
>I dont see how this could fit in the current architecture. 

Yes of course with the current voice list architecture, there is no way that to put that filter in a channel structure.

>The IIR filter currently doesnt support a band-pass filter.

So actually this filter can be a low pass or a high pass filter.

>I dont want to hardcode even more non-standard default modulators in fluidsynth.

sure , i understand.

jjceresa:
> @mawe42

 
Did you find something in the SF2 specs related to the use of a high-pass filter ?

src/synth/fluid_synth.c
```diff
@@ -196,6 +198,9 @@ void fluid_synth_settings(fluid_settings_t* settings)
   fluid_settings_add_option(settings, "synth.midi-bank-select", "gs");
   fluid_settings_add_option(settings, "synth.midi-bank-select", "xg");
   fluid_settings_add_option(settings, "synth.midi-bank-select", "mma");
+
+  fluid_settings_register_int(settings, "synth.high-pass-filter", 0, 0, 1,
```
mawe42: This can be removed

src/synth/fluid_synth.c
```diff
@@ -139,6 +139,8 @@ static fluid_mod_t default_reverb_mod;         /* SF2.01 section 8.4.8  */
 static fluid_mod_t default_chorus_mod;         /* SF2.01 section 8.4.9  */
 static fluid_mod_t default_pitch_bend_mod;     /* SF2.01 section 8.4.10 */
 
+static fluid_mod_t custom_cc2hpfilterfc_mod;
```
mawe42: This can be removed

src/synth/fluid_synth.c
```diff
@@ -3289,6 +3296,8 @@ fluid_synth_alloc_voice(fluid_synth_t* synth, fluid_sample_t* sample, int chan,
     default_mod = default_mod->next;
   }
 
+  fluid_voice_add_mod(voice, &custom_cc2hpfilterfc_mod, FLUID_VOICE_DEFAULT);
```
mawe42: This can be removed

src/rvoice/fluid_rvoice.h
```diff
@@ -150,6 +150,7 @@ struct _fluid_rvoice_t
 	fluid_rvoice_envlfo_t envlfo;
 	fluid_rvoice_dsp_t dsp; 
 	fluid_iir_filter_t resonant_filter; /* IIR resonant dsp filter */
+	fluid_iir_filter_t resonant_custom_filter; /* IIR resonant dsp high-pass filter */
```
mawe42: maybe use "custom" instead of "high-pass"? Or also mention low-pass...

src/rvoice/fluid_iir_filter.h
```diff
@@ -26,28 +26,32 @@
...
 void fluid_iir_filter_calc(fluid_iir_filter_t* iir_filter, 
                            fluid_real_t output_rate, 
-                           fluid_real_t fres_mod); 
+                           fluid_real_t fres_mod);
 
 /* We can't do information hiding here, as fluid_voice_t includes the struct
    without a pointer. */
 struct _fluid_iir_filter_t
 {
+	enum fluid_iir_filter_type type; /* specifies the type of this filter (highpass, lowpass or bandpass) */
```
mawe42: "bandpass" is invalid

src/rvoice/fluid_rvoice.c
```diff
@@ -361,12 +361,17 @@ fluid_rvoice_write (fluid_rvoice_t* voice, fluid_real_t *dsp_buf)
...
   fluid_iir_filter_calc(&voice->resonant_filter, voice->dsp.output_rate,
-  		        fluid_lfo_get_val(&voice->envlfo.modlfo) * voice->envlfo.modlfo_to_fc +
- 		        fluid_adsr_env_get_val(&voice->envlfo.modenv) * voice->envlfo.modenv_to_fc);
+                    fluid_lfo_get_val(&voice->envlfo.modlfo) * voice->envlfo.modlfo_to_fc +
+                    fluid_adsr_env_get_val(&voice->envlfo.modenv) * voice->envlfo.modenv_to_fc);
 
   fluid_iir_filter_apply(&voice->resonant_filter, dsp_buf, count);
-
+  
+  /* additonal high-pass filter - only uses the fixed modulator, no lfos... */
```
mawe42: "custom filter" may be better. And "the fixed modulator" is a bit misleading, it sounds like there is a default modulator for that custom filter in the source somewhere.

src/rvoice/fluid_iir_filter.c
```diff
@@ -47,6 +50,12 @@ void
 fluid_iir_filter_apply(fluid_iir_filter_t* iir_filter,
                        fluid_real_t *dsp_buf, int count)
 {
+  if(iir_filter->type == FLUID_IIR_DISABLED || iir_filter->q_lin == 0)
```
mawe42: Why does setting q_lin to 0 disable the filter?

And it seems like this would not only apply to the custom filter, but also to the standard low-pass. Is it according to SD2 spec that the filter switches off if q_lin = 0?

derselbst:
Thanks for reviewing, will fix all those leftovers.

> Just out of curiosity: what can the custom low-pass filter do that the standard filter can't?

Short story: Basically it can do the same in a more non standard way by setting flags like `FLUID_IIR_Q_LINEAR` or `FLUID_IIR_NO_GAIN_AMP`.

Long story: I'm using fluidsynth to re-create the soundtracks of some Nintendo64 games. Nearly all of them use sequenced audio but synthesize only at 22kHz. Extracting the MIDI and soundbank allows me to listen to those tunes in 44kHz, multi-channel (each MIDI chan to a separate stereo chan) and builtin loops. A couple of those N64 games (e.g. [CBFD](https://en.wikipedia.org/wiki/Conker%27s_Bad_Fur_Day) and [JFG](https://en.wikipedia.org/wiki/Jet_Force_Gemini)) additionally use a biquad LP IIR filter for some of their musical themes. Since my reverse engineering skills are very limited, I played around with fluidsynths IIR filter to make it behave similar to the N64 one. It appears that the N64 IIR specifies the filter cutoff freq. in a sinusoidal way. I'm not yet quite sure whether the filter Q really scales linearily meaning if I need `FLUID_IIR_Q_LINEAR`. However I am sure that if the MIDI CC in charge of controlling the filter Q is set to zero, the filter is "off". And `FLUID_IIR_NO_GAIN_AMP` because the N64 IIR does not seem to have this kind of gain reduction. At least 3 non-standard customizations here.

I tried to implement the N64 IIR with soundfont modulators, but could not get it to work, there are just too many custom issues here. GEN_FILTERFC defaults to 13500Hz, which is incompatible to what the N64 IIR expects (messing around with negative unipolar mods didnt really help). When overriding FC with 0Hz (for every instrument bag), the scale matches again, but some instruments are missing because completely filtered away because Q never gets zero thus the IIR is never switched off. And the gain reduction makes it unnecessarily silent.

After all I think it's better to have this as a pure custom optional feature, rather than to start overengineering a soundfont with a bunch of strange modulators that wouldnt work anyway due to Q never gets zero, filter always on.

Note that `q_lin == 0` issue should not affect the default SF2 filter, because `pow(10.0f, q_dB / 20.0f)` is the very last step when converting Q from dB to linear, which never gets zero.

jjceresa:
After a quick try of `iir2 branch` current commit (7a60cdc) using the fluidsynth application console under the following classic use:
- soundfont : GUGSv1_47.sf2
- preset 0 or 1  (first pianos).
- playing some notes at low and high velocity.
The result is not the same as `on master  branch` (or others). We hear an obvious filtering cutoff frequency depending of velocity. The test is easy to reproduce.

It seems that it isn't the default expected behaviour  ?.

derselbst:
Confirmed, thanks for noticing. Seems to happen for velocity `<=65`? Need to investigate.

derselbst:
Should be fixed now... I must have been very drunk -_-

jjceresa:
> Should be fixed now... 

Yes, now the piano sound as expected.

I haven't yet look in depth. Does this new filter "integrated" in the existing SF2 filter or is it an independent new filter (in serial with the SF2 filter) ?. Ofc i guess the later.

derselbst:
Works fine now, will merge it.

Boscop:
Is it correct that fluidsynth doesn't have a Lowpass filter out of the box, like Bassmidi has (that can be controlled by CC 74 for cutoff and CC 71 for resonance, as defined in the GM standard)?

Boscop:
Ah thanks for clearing that up. 
Do you know since which version or which date fluidsynth supports the GM filter?
I've [read](https://lists.gnu.org/archive/html/fluid-dev/2016-03/msg00013.html) that it doesn't support it..

I was asking because with the fluidsynth.dll that I used to create a VST on Windows, the filter CCs didn't seem to work.

jjceresa:
>I was asking because with the fluidsynth.dll that I used to create a VST on Windows, the filter CCs didn't seem to work.

As explained above, if you want **CC 74 and CC 71** acting on  **cutoff** and **resonance** you need to:
1) add the corresponding `modulators `in the used soundfont using a soundfont editor. This works for any fluidsynth version.
2) alternativelly if you dont want to change the soundfont you can use the new API  `fluid_synth_add_default_mod()` http://www.fluidsynth.org/api-unstable/index.html#NewIn2_0_0 in your application. This is bring by recent version described here [2.0 beta fluidsynth version](http://lists.nongnu.org/archive/html/fluid-dev/2018-05/msg00019.html).
If you need  more help, it would be preferable to continue this on a new thread to the mailing list.

Boscop:
Thanks.
The same soundfont file was working with the filter CCs using bassmidi, so it should also work with fluidsynth with no modifications, right? 
I think maybe my version of the fluidsynth.dll was too old (?) Which version is the first one that supports the filter CCs? :)

jjceresa:
>The same soundfont file was working with the filter CCs using bassmidi, so it should also work with fluidsynth with no modifications, right?

No, because the modulators needed are not part of the default modulators described by SF2 specification. This is why you need to add those yourself.

Boscop:
Does that mean all GM soundfonts that I want to work with fluidsynth have to be manually modified to add those modulators? Why can't fluidsynth be changed so that it works like bassmidi? I want to have a drop-in replacement VST for the bassmidi-based SF player, where I can load any soundfont into it, and others too. I don't want to require the user to have to manually add modulators to every Soundfont that they want to load..

By GM soundfont I mean e.g. GeneralUser etc. Shouldn't fluidsynth work out of the box with GM soundfonts, including the filter CCs? :)

derselbst:
It's a requirement by the SF2, compelling **you** to add those modulators to the soundfont files you use to make your use-case work. If bassmidi works out-of-the box, they are using non-standard custom default modulators for satisfying lazy users. fluidsynth is very cautious in adding those kinds of non-standard compliant modulators and as of 2.0 it only provides one for CC8.

> Shouldn't fluidsynth work out of the box with GM soundfonts, including the filter CCs?

No.

If you are a programmer, check out `fluid_synth_add_default_mod()` available as of fluidsynth 2.0.

Boscop:
Thanks. 
Would this be the correct way to do it?
```C
// link CC 74 to filter cutoff

fluid_mod_t mod_filter_fc;

fluid_mod_set_source1(&mod_filter_fc, 74,              /* index=74 */
               FLUID_MOD_CC                            /* CC=1 */
               | FLUID_MOD_LINEAR                      /* type=0 */
               | FLUID_MOD_UNIPOLAR                    /* P=0 */
               | FLUID_MOD_POSITIVE                    /* D=0 */
               );
fluid_mod_set_source2(&mod_filter_fc, 0, 0);           /* No second source */
fluid_mod_set_dest(&mod_filter_fc, GEN_FILTERFC);      /* Target: filter cutoff */
fluid_mod_set_amount(&mod_filter_fc, 1000);            /* Amount: 1000 ('tenths of a percent') */

fluid_voice_add_mod(synth, &mod_filter_fc, FLUID_VOICE_DEFAULT);

// link CC 71 to filter resonance

fluid_mod_t mod_filter_q;

fluid_mod_set_source1(&mod_filter_q, 71,               /* index=71 */
               FLUID_MOD_CC                            /* CC=1 */
               | FLUID_MOD_LINEAR                      /* type=0 */
               | FLUID_MOD_UNIPOLAR                    /* P=0 */
               | FLUID_MOD_POSITIVE                    /* D=0 */
               );
fluid_mod_set_source2(&mod_filter_q, 0, 0);            /* No second source */
fluid_mod_set_dest(&mod_filter_q, GEN_FILTERQ);        /* Target: filter resonance */
fluid_mod_set_amount(&mod_filter_q, 1000);             /* Amount: 1000 ('tenths of a percent') */

fluid_voice_add_mod(synth, &mod_filter_q, FLUID_VOICE_DEFAULT);
```

---

Btw, does fluidsynth now support controlling the portamento settings via the CCs 5, 65, 84?

---

What are the necessary steps to build fluidsynth 2.0 as a win32 dll? Has anyone done that already (and uploaded it)? :)

---

Btw, I didn't mean that this should work out of the box because of the SF2 standard but because of the GM standard. The CCs 74 and 71 are listed in the GM standard in the list of CCs that should affect the filter by default (if a filter is available). Not only bassmidi, also the cakewalk SFZ player (and SFZ+) do this, btw! (So far, all the soundfont players I've tested behave this way, except fluidsynth.)

Bassmidi supports all the CCs that the GM standard requires:
https://github.com/kode54/BASSMIDI-Driver/blob/master/www/archive/BASSMIDI_Readme.txt#L41-L67

So why was the decision made for fluidsynth to not behave this way? 
Would any user be mad if the filter is affected by CC 74 and 71 when they are playing a midi file/stream through fluidsynth? Wouldn't it make more sense to support the expected GM use case by default, so that midi files sound the same in fluidsynth as in other soundfont players?
(If I don't want certain CCs to have an effect, I put a CC filter/blocker before the synth I'm using, for those specific CCs..)

jjceresa:
> Btw, does fluidsynth now support controlling the portamento settings via the CCs 5, 65, 84?

Yes, this is in recent version described here [2.0 beta fluidsynth version](http://lists.nongnu.org/archive/html/fluid-dev/2018-05/msg00019.html). Please again read this mail.

>What are the necessary steps to build fluidsynth 2.0 as a win32 dll?

The steps are described on the Wiki.https://github.com/FluidSynth/fluidsynth/wiki/BuildingFluidSynth

>Has anyone done that already

A lot of guys on the mailing list could probably help. 

>Btw, I didn't mean that this should work out of the box because of the SF2 standard but because of the GM standard. The CCs 74 and 71 are listed in the GM standard in the list of CCs that should affect the filter by default (if a filter is available).

Yes, those CC are called TIMBRE CONTROLLERS (MIDI 1.0 Detailed Specification 4.2 p 15)
**"Both of these controllers are intended as a performance controllers – not as a sound parameter
editing controllers (in other words, these messages do not change the memorized data of a preset)."**

That means those are real-time performance CCs. On fluidsynth those CCs are to be implemented via SF2 modulators . SF2 standard is not against MIDI standard but at the service of MIDI standard. 
Again, It's a requirement by the SF2, compelling you to add those modulators.
It is important to understand that each time a  CC (standard or not) will become necessary for a particular use case to modulate something in the sound engine, it should be preferable that the decision to add this modulator should be done by the application via API (or by the soundfont designer) .

Again, because of the nature of the questions and because a lot of persons are aware of MIDI, SF2 and building fluidsynth, If you need more help, it would be preferable to continue this on a new thread to the mailing list. It is really the right place.

derselbst:
> So why was the decision made for fluidsynth to not behave this way?

Forget about GM. fluidsynth is a **pure SF2** synth. Thus we solely obey to the SF2 spec when it comes to sound synthesization.

> Bassmidi supports all the CCs that the GM standard requires:

I've crawled all the way through the complete midi 1.0 spec. There is not a single useful word on how to implement or make this filter behave exactly. No words about scales, ranges, shapes of this filter. Does it behave linear, convex, concave and what is the maximum cutoff frequency or filter Q?  Why should **we** add a non-standard default modulator of some kind  and break use-cases of all other users that might be using CC74 / 71 completely differently? It is your job to do it, either via soundfont files or by manually adding default modulators via API.

> Would this be the correct way to do it?

Not quite. Bring this to the mailing list and I'll advise.

Issue #241, opened 2017-10-21
Feature request: read raw SF2 data from memory


sykhro:
Hello,
while cleaning up a Linux port of a certain tool, I hit a particular use case: I have to load SF2 data from memory. I can read MIDI from a buffer just fine, but when it comes to loading a soundfont there's no direct way. The problem was somewhat fixed by implementing [this](https://github.com/vgmtrans/vgmtrans/tree/master/lib/fluidsynth/mem_sfloader/mem_sfloader.c), but it doesn't feel like a great solution because the code itself is very old and it's more of a hack around the api. 
I understand that this will probably be turned down as it won't probably be used by a lot of people, but I'm not sure if it's worth re-implementing an entire loader for a relatively minor addition

derselbst:
Implementing another sfont loader wont happen. However one might think about creating a user supplied struct of function pointers, that replicate the behaviour of fopen(), fread(), etc. Something like this: 

https://gitlab.kode54.net/kode54/psflib/blob/7da887b3a41ece72b2461cbf8876af6dda2738fc/psflib.h#L36-55

So instead that fluidsynth calls `fread()` directly, it would call the user supplied function pointer.

There are currently other open issues around here, so I wont have time implementing that. Feel free to draft a PR though.

sykhro:
That seems like a good solution; I'll look into it whenever I can

sykhro:
finally had time to try to do something about this, but there are a few problems. first off, I'm unsure to what the callbacks struct should be passed to (I was thinking of adding it to the loader struct, and then referencing it when needed. I wanted to avoid changing `new_fluid_defsfloader`'s prototype though. there's a way around it with variatic functions, but it feel overcomplicated). secondly, the filename remains a problem, because it's used to set the sfont filename. this ends up in the sfont filename always being "RIFFE", because it reads from the file header, and it stops at null bytes.

derselbst:
You could add the callback struct as a member to `struct _fluid_sfloader_t`. `new_fluid_defsfloader()` would create the `_fluid_sfloader_t` and init the callbacks to the standard C file api functions. After exposing `new_fluid_defsfloader()` to public API, the user could retrieve this sturct, manipulate the callbacks and then add this loader via `fluid_synth_add_sfloader()`. Internally you will only have to pass the callback struct to `fluid_defsfont_load()` and every other function that would need to call fopen(), fread() and friends.

When the user then wants to load a sf2 from memory he could:

```
[...]

PR #334, opened 2018-01-26
Refactor Soundfont loader API

### LINKED ISSUE #309: refactor soundfont loader API
evaluate possibilities on refactoring the soundfont loader API using an interface with less exposed internals (at least `_fluid_sfloader_t`, perhaps also `_fluid_sfont_t` should be private)

... as suggested by https://github.com/FluidSynth/fluidsynth/blob/master/TODO

derselbst:
This refactors the sfloader API:

* makes all the public structs private
* provides dedicated setter + getter functions to manipulate those data types
* in the hope this part of the API will become clearer and easier to use and to maintain.

Tested by porting SWAMI to this new API (the only known downstream app actually using the sfloader API).

closes #309
addresses #241 

[0001-swami-port-fluidsynth-plugin-to-2.0-API.txt](https://github.com/FluidSynth/fluidsynth/files/1669020/0001-port-fluidsynth-plugin-to-2.0-API.txt)


derselbst:
I think it's good now.

@mawe42 Could you have a look at it? Esp. at the functions that are now exposed to the public (see `sfont.h`). I tried to make `_fluid_sfont_t::free` return `void` (as every destructor/cleanup function should do), but it seems with the current architecture this is not possible, as there is no way to know during soundfont unloading when all rvoices have finished, i.e. are not processed by rvoice_mixer anymore. Which means we have to keep spawning a timer in `fluid_synth_sfunload()` if there are still samples in use by (r)voices.

mawe42:
Sorry for the late reply! From a quick glance this looks very good. Just a thought about the unloading: do you think it would be possible to make the unload optionally blocking? So that if you unload a font which has no refs anymore but where samples are still in use, fire up the timer internally but use a mutex wait until unloading has succeeded?

derselbst:
> do you think it would be possible to make the unload optionally blocking?

Unfortunately not. First we would have to wait until somebody calls `fluid_synth_[n]write*()` so that pending rvoice_events are dispatched and the finished rvoices are collected in `_fluid_rvoice_eventhandler_t::finished_voices`. Secondly someone would have to call `fluid_synth_check_finished_voices()` (so that voices + rvoices release the samples by decrementing the sample refcount in `fluid_voice_stop()`) which however is only called on `fluid_synth_api_enter()`. And since the current thread would reside in the synth API (not necessarily by keeping the mutex locked) we would still need a third thread that comes along tries to enter the API in order to unblock ourself :(

P.S.: I also just got in touch with Element, so no need for a detailed review in case you're lacking time.

Issue #216, opened 2017-09-27
Windows: libfluidsynth.dll should show version info in explorer file detail view


harborsiem:
This is an example german detail view from libfluidsynth.dll:

![properties](https://user-images.githubusercontent.com/28147549/30904161-7151f17a-a371-11e7-8a93-0170a16a5f19.jpg)

You can do this with the following resource file inside the zip:

[libfluidsynth_dll.zip](https://github.com/FluidSynth/fluidsynth/files/1336385/libfluidsynth_dll.zip)

In the cmake generated libfluidsynth.vcxproj file you have to add at the end of file:

<Project ...
...

  <ItemGroup>
    <ResourceCompile Include="libfluidsynth_dll.rc" />
  </ItemGroup>
  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
  <ImportGroup Label="ExtensionTargets">
  </ImportGroup>
  <ProjectExtensions>
    <VisualStudio>
      <UserProperties RESOURCE_FILE="libfluidsynth_dll.rc" />
    </VisualStudio>
  </ProjectExtensions>
</Project>

harborsiem:
ups, the last lines are lost
 
 `<ItemGroup>
    <ResourceCompile Include="libfluidsynth_dll.rc" />
  </ItemGroup>
  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
  <ImportGroup Label="ExtensionTargets">
  </ImportGroup>
  <ProjectExtensions>
    <VisualStudio>
      <UserProperties RESOURCE_FILE="libfluidsynth_dll.rc" />
    </VisualStudio>
  </ProjectExtensions>
</Project>`

derselbst:
The screenshot shows the way it should be?

Anyway, I dont quite know what you want me to do. Could you please elaborate?

harborsiem:
It would be a nice feature to see the version of a *.dll or *.exe with the file explorer. One can also look with an other program to these informations via winapi functions.

> The screenshot shows the way it should be?

Yes, Sir.

> Could you please elaborate?

I give it a try.
First of all take a look to the file "libfluidsynth_dll.rc" (inside the above .zip file) with a text editor.
[...]

PR #336, opened 2018-01-28
Windows version info for dll

harborsiem:
This is a solution for issue #216
 
I'm not shure if it works for MINGW. So I set

`if ( WIN32 AND NOT MINGW )  
include(generate_product_version)  
generate_product_version(  
#...etc... `

Without the `include(...)` I get an error.


derselbst:
Looks good, thanks.

> Without the include(...) I get an error.

Yep, my mistake, sry.

Issue #333, opened 2018-01-23
VCPKG package?


stekyne:
I noticed you were using VCPKG to get some dependencies. Any thoughts on adding fluidsynth to VCPKG altogether? It would be handy for all Windows users I'd imagine.

Thanks

derselbst:
> Any thoughts on adding fluidsynth to VCPKG altogether?

I'm open to add a minimalist build of fluidsynth (i.e. no libsndfile, no portaudio, no readline, etc.) to vcpkg. However as Linux user I have no means to test it. Could you draft a `fluidsynth.cmake`?

stekyne:
By fluidsynth.cmake, do you mean the config file for VCPKG or something else?

derselbst:
The vcpkg config file.

stekyne:
Nearly finished this.  Just wondering if it's possible to build Fluidsynth statically (easily) as that's one of the build options in VCPKG?

derselbst:
Great! Do you mean something like

```
cmake -DBUILD_SHARED_LIBS=0 ..
```

stekyne:
That is being set by VCPKG actually. The problem was something else. There is just an issue of the libs being put in a folder named "lib64" instead of "lib". Once that's solved it will be good to go.

derselbst:
There is a `LIB_SUFFIX` var in fluidsynth's CMake file, might this be responsible for it?

https://github.com/FluidSynth/fluidsynth/blob/9cd870ea7a06ae049590b9bb539ef3864e057f34/CMakeLists.txt#L94-L103

This has already been disabled for mingw in #281, if it helps I wouldnt mind disabling this also for MSVC or WIN32 resp.

stekyne:
Closed with #339

ollpu:
The wiki says that the package is available via VCPKG now, but I can't find it (at least using Microsoft's repository). Is someone going to add it upstream?

derselbst:
The wiki refers to 1.1.10, which has not been released yet. Please be patient.

ollpu:
Ah, I see. Thanks!

stekyne:
https://github.com/Microsoft/vcpkg/pull/3016 

I've opened the PR here. Sorry for the delay.

stekyne:
This has now been pulled into VCPKG now so should be available.

PR #339, opened 2018-02-07
No longer prefix the lib folder for MSVC builds

stekyne:
As discussed in #333, I've remove the lib prefix for MSVC builds. 

Should resolve the issue with the VCPKG.


derselbst:
"It should resolve the issue" sounds like a vague guess. Have you tested it with 1.1.9 whether it works?

stekyne:
Tested both x64 and x64 static builds now. Works well. I've tested with a sample cmake project and header/libs are being found.

stekyne:
I've used this FindFluidSynth.cmake.

# - Find fluidsynth
# Find the native fluidsynth includes and library
#
#  FLUIDSYNTH_INCLUDE_DIR - where to find fluidsynth.h
#  FLUIDSYNTH_LIBRARIES   - List of libraries when using fluidsynth.
#  FLUIDSYNTH_FOUND       - True if fluidsynth found.

if (FLUIDSYNTH_INCLUDE_DIR AND FLUIDSYNTH_LIBRARIES)
  # Already in cache, be silent
  set (FluidSynth_FIND_QUIETLY TRUE)
endif ()

find_path (FLUIDSYNTH_INCLUDE_DIR fluidsynth.h)

find_library (FLUIDSYNTH_LIBRARIES NAMES fluidsynth libfluidsynth)
mark_as_advanced (FLUIDSYNTH_LIBRARIES FLUIDSYNTH_INCLUDE_DIR)

# handle the QUIETLY and REQUIRED arguments and set FLUIDSYNTH_FOUND to TRUE if 
# all listed variables are TRUE
include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(FluidSynth DEFAULT_MSG FLUIDSYNTH_LIBRARIES FLUIDSYNTH_INCLUDE_DIR)

derselbst:
Ok, nice. I'd like to look into the vcpkg config file for fluidsynth. Could you upload it as well?

stekyne:
https://gist.github.com/stekyne/91ec20d9aeaa01b175b83fe4013fbda6

stekyne:
Obviously the 1.1.9 zip won't work without these PR changes so I guess we'd have to switch it to whatever the next version is. Or else download the latest git version in the portfile instead.

derselbst:
1.1.10 will be next (end of February I guess). Please **DO NOT** use buggy, unstable, recent git versions, there is not API/ABI stability guarantee if you do so.

Else this looks good. Will merge it tomorrow and add a few notes to the [wiki](https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake#building-with-visual-studio-on-windows). Thanks!

stekyne:
Just on this. I finally integrated into csound as a static x64 build (in VCPKG). There are a few things though. You have to define FLUIDSYNTH_NOT_A_DLL for the static builds which I wasn't aware of. You'll also have to separately link in glib-2.0 and whatever else is missing. 

Perhaps we should set the "not a dll" flag in the port file seeing as we'll know the target triple.

derselbst:
>  You have to define FLUIDSYNTH_NOT_A_DLL for the static builds

The only reason I see so far is that you probably dont want to have a
DllMain() function in your static lib? Does compilation or linkage
break if you do otherwise?

Besides: having thought a little more about this "static lib" idea, I
think it's a bad one. Two main reasons come to my mind:

1. Static libs dont provide symbol visibility, which we are enforcing
since 1.1.7, to avoid the emerging or support of broken downstream
apps using private functions of fluidsynth.
2. Fluidsynths license is LGPL. Any software statically linking in
fluidsynth must be (L)GPL as well. This is probably not the case for
windows apps, which are mostly proprietary.

People are lazy and due to the fact that vcpkg is a very comfortable
solution, static libs would probably encourage downstream to violate
fluidsynth's license when using it in their proprietary app, I could
imagine. Is there no way of providing a dll via vcpkg?

stekyne:
There were build errors without that flag set. It was complaining about missing fluidsynth functions. 
https://github.com/FluidSynth/fluidsynth/blob/master/include/fluidsynth.h#L36
The FLUIDSYNTH_API was being set to __declspec(dllimport) without it. 

Yeah, you can build shared DLLS in VCPKG. That's mostly the default way of doing it but they provide support for static builds also. There is also the possibility of building in the CRT statically or dynamically if you wanted. https://github.com/Microsoft/vcpkg/blob/master/docs/users/triplets.md 

That's what I meant about target triplet earlier. You have to set how you want to install each package e.g. "vcpkg install fluidsynth:x64-windows-static fluidsynth:x64-windows". Which is the static 64bit and shared 64bit respectively. So in the portfile we could add any CMake flags to build it correctly for each type.

I was building it statically because that's the way we've been integrating 3rd party libs in Csound recently. Just by changing the target triplet in our appveyor file, we can link them in shared or statically which can be handy.

The license for Fluidsynth is copied with the VCPKG by their design so it should be very clear how it is to be used. I don't think there is any difference with someone who would download the source and build with CMake the same way.

derselbst:
Ah, so since vcpkg supports shared and static version (and apparently
defaults to shared) I'm fine with this. However the build failure of static
windows builds is a bug in fluidsynths build system. I could think of some
fix that checks for if BUILD_SHARED_LIBS==0 then define not a dll flag.
Additionally I'll try to setup an appveyor CI build that tests static
builds for windows. I wont have time for this in the next days, but I'll do
so before 1.1.10. In the meantime, please open a new issue a.la. "Static
build fails on windows".

PR #345, opened 2018-03-03
Add profiling command interface

derselbst:
This patch adds "profiling" commands interface for performance measurement.

Overall this looks quite well. It's an internal (developer only) feature, so I dont really care that it only works with one synth instance. If need be this can be changed any time later on.

@jjceresa Some general notes:

* Please do not `fluid_atomic_int_get (&synth->active_voice_count)`; `active_voice_count` is not of type `fluid_atomic_int_t`, it is not set atomically, this would probably cause undef. behaviour. Please call `fluid_synth_get_active_voice_count()` instead.
* Please avoid assignments in conditions. I've seen it several times in `fluid_cmd.c`, compilers might warn, humans might misread it.
* For my taste there are too many whitespace changes in `fluid_synth.c`. I dont want to nit-pick too much, but it unnecessarily changes history, like: "What? The preset fallback mechanism was changed by the profiling patch? ... Ah, no, only whitespace changes...". So, perhaps consider restoring the `fluid_synth.c` from master and adjust those 4 `fluid_profile` lines in hindsight.


jjceresa:
> It's an internal (developer only) feature, so I dont really care that it only works with one synth instance. If need be this can be changed any time later on.

Working with only `one instance` is required as it is very important that during measurement of part of code of one synth instance (using `prof_start`) this will never preempted by another instance. This is ensured by a simple `lock flag` inside the `prof_start `command.

However, for the same reason it is important that when using profiling on a machine, we don't forget to paid attention to stop any background OS tasks (no running ftp services, no web service nor web client making a download,....).

jjceresa:
> So, perhaps consider restoring the fluid_synth.c from master and adjust those 4 fluid_profile lines in hindsight.

Sorry, It seems that in https://github.com/FluidSynth/fluidsynth/pull/345/commits/2aca9482dfa8349756e4d6a6f55fa9c2df974f67 i have restored fluid_synth.c from profile branch by mistake rather using that of master branch. Perhaps it will be preferable to restart the profile branch from the actual master, next i will do the patching taking in account yours initial general notes ?

derselbst:
> Working with only one instance is required as it is very important that during measurement of part of code of one synth instance

I meant like making `fluid_profile_data_t` a member of fluid_synth_t, to being able to track it down to multiple instances. But I admit I currently dont see a usecase for this and it's not trivially done as you mentioned. So nevermind.

> Sorry, It seems that in 2aca948 i have restored fluid_synth.c from profile branch by mistake rather using that of master branch.

That's fine either, the overall diff looks good now: https://github.com/FluidSynth/fluidsynth/pull/345/files#diff-6f2023b9a54ace87ed75daa30f93ed9b

jjceresa:
This patch have now a `cancellation `capability that works on Windows and `should works on linux` but not yet fully tested.
The test is trivial doing this simple check:

Type `profile start` command for some measurement (i.e 20 measures):
`> prof_start  20`

Then during printing strike `<cr>` key. If the message "`Profiling canceled.`" is printed, it means that's fine for linux.
Many thanks for report.

derselbst:
After a first test cancellation works fine on Linux.

jjceresa:
>After a first test cancellation works fine on Linux.

Thanks.

derselbst:
Thanks for fixing my lousy incomplete modifications.

Something off topic: Why do we even have `fluid_synth_t::active_voice_count`? Shouldnt we better look up `fluid_rvoice_mixer_t::active_voices`, so that `fluid_synth_get_active_voice_count()` is also accurate when not being called from synthesis context?

src/utils/fluid_sys.c
```diff
@@ -386,11 +386,19 @@ unsigned int fluid_curtime(void)
 double
 fluid_utime (void)
 {
+#if defined(WIN32)      /* Window using intel performance counters  */
+	LARGE_INTEGER Freq;	/* Performance Frequency */
+	LARGE_INTEGER PerfCpt;
+	QueryPerformanceFrequency(&Freq);  /* Frequency value */
+	QueryPerformanceCounter(&PerfCpt); /* Counter value */
+	return PerfCpt.QuadPart * 1000000.0/Freq.QuadPart; /* time in micros */
+#else
```
derselbst: I think we should get rid of the special treatment for win32 by switching to g_get_monotonic_time(), which should provide the required resolution: https://github.com/GNOME/glib/blob/1e6803be3b3ec5005335f1fa7d53e57bbc0b8bba/glib/gmain.c#L2744

jjceresa: > switching to g_get_monotonic_time(), which should provide the required resolution:

I have tried this to verify if the result are the same as actual `fluid_utime() (L392 to L394)`.
Unfortunately the value returned by  `g_get_monotonic_time()` is not the same and the resolution is far less than `fluid_utime()` .  Some results follow:
```
now_fluid_utime():                        16055873753.508 micros
now_g_get_monotonic_time():          1520527285078125.000 micros 
```
```
// Duration measured using fluid_utime():
delta_fluid_utime = now_fluid_utime - prev_fluid_utime

// Duration measured using get_monotonic_time():
delta_g_get_monotonic_time = now_g_get_monotonic_time - g_get_monotonic_time

// Difference between these durations:
diff_glib_fluid_utime = delta_fluid_utime - delta_g_get_monotonic_time

delta_fluid_utime:                            1001003.479 micros
delta_g_get_monotonic_time:                   1000000.000 micros
diff_glib_fluid_utime:                           1003.479 micros  <-  error 1 ms 
```
`fluid_utime()` return the result in micros as a real keeping the resolution of the hardware counter. For example with a frequency of  3579545 Hz , the resolution is  0.279 micros.
Using `get_monotonic_time()` we get often an error about 1 ms between duration of 1 second. 
The value returned by `get_monotonic_time()` is not the value of the actual hardware counter. This function doesn't behave as we expect.

jjceresa:
>Something off topic: Why do we even have fluid_synth_t::active_voice_count? Shouldnt we better look up fluid_rvoice_mixer_t::active_voices, so that fluid_synth_get_active_voice_count() is also accurate when not being called from synthesis context?`

I guess that direct look up to `fluid_rvoice_mixer_t::active_voices` from `fluid_synth_get_active_voice_count()` requires access through `atomic operator `from both task side (the `application task `calling `fluid_synth_get_active_voice_count()` and the the `audio task `calling the audio rendering API `fluid_synth_write_xxxx()`. This makes the `execution duration `of `audio rendering API` dependent of the  `application task` particularly in the case of intensive call to fluid_synth_get_active_voice_count().

Actually, having this `fluid_synth_t::active_voice_count` updated  (when a voice has been finished  via fluid_synth_check_finished_voices())  avoid this dependency and avoid the use of atomic function.

Following general figure (life_cycle_of_voice.jpg) illustrates  how (execution)  independence has been implemented. (Sorry for the missing comments (those are still in french!)).

life_cycle_of_voice

src/utils/fluid_sys.c
derselbst: But if you measure for 1s and `delta_g_get_monotonic_time` gives you the perfect second, why do you think it's not exact? I believe that `fluid_utime()` is not exact then. Perhaps because retrieving `QueryPerformanceFrequency(&Freq);` every time is expensive? Also microsofts doc recommend to cache `Freq`, so glibs  `g_get_monotonic_time()` seems like the correct implementation.

derselbst:
> This makes the execution durationof audio rendering API dependent of the application task particularly in the case of intensive call to fluid_synth_get_active_voice_count().

I dont quite share this concern as I do not consider "polling" `fluid_synth_get_active_voice_count()` to be a use-case. In fact with the current implementation you would keep locking the synth API. And for normal usage of this function, I cant imagine an atomic increment on the `rvoice_mixer` side would hurt too much.

Looking up `fluid_rvoice_mixer_t::active_voices` would avoid data redundancy and allow us to get rid of this 

`voice->channel->synth->active_voice_count++;`

dance in `fluid_voice_start()`. I just dont know when a voice is *active*. Right after a call to `fluid_synth_start_voice()` or when it's actually being processed by the dsp loop?

src/utils/fluid_sys.c
jjceresa: I must admit that previous example is not clear and don't reveals what happens with 'g_get_monotonic_time()'.

I investigated further calling actual `fluid_utime()`, `g_get_monotonic_time()` and `g_get_current_time()` at the same time:

```
GTimeVal gtimeval;
double gctime;
double now_fluid_utime_cpt = fluid_utime();
double now_glib_cpt = (double) g_get_monotonic_time();
g_get_current_time(&gtimeval);

printf("fluid_utime():              %30.3f micros\n",now_fluid_utime_cpt);
printf("g_get_monotonic_time():     %30.3f micros\n",now_glib_cpt);
gctime = (double)gtimeval.tv_sec * 1000000.0 + (double)gtimeval.tv_usec;
printf("g_get_current_time():       %30.3f micros \n",gctime);
```
The result of print is:
```	
fluid_utime():                             23961936852.868 micros
g_get_monotonic_time():               1520640580187500.000 micros
g_get_current_time():                 1520640580187500.000 micros
```	
`g_get_monotonic_time()` returns the same result that `g_get_current_time()` and not the  same as `fluid_utime()` (that uses  `QueryPerformanceCounter()` ). So it seems that `g_get_monotonic_time()` make use of the same system call than `g_get_current_time()`.

`g_get_current_time()` make use of `GetSystemTimeAsFileTime()` that obtains the current system date and time that is less precise than `QueryPerformanceCounter()`.
***
>Perhaps because retrieving QueryPerformanceFrequency(&Freq); every time is expensive? Also microsofts doc recommend to cache Freq.

Yes, it a good practice to cache Freq even if `QueryPerformanceFrequency()` isn't too expensive (1.509 micros).

The  duration results show that caching is a good thing:
- fluid_utime() without caching: 2.982 micros.
- fluid_utime() with caching: 1.550 micros.
- QueryPerformanceFrequency(): 1.523 micros.

jjceresa:
>  Looking up fluid_rvoice_mixer_t::active_voices would avoid data redundancy and allow us to get rid of this voice->channel->synth->active_voice_count++;
dance in fluid_voice_start(). I just dont know when a voice is active. Right after a call to fluid_synth_start_voice() or when it's actually being processed by the dsp loop? I tend to vote for the latter option.

I understand completely your point of view but i think that actually  `synth::active_voice_count` is misspelling and should be renamed (i.e `started_voice_count`)  because it is the only information useful for the application. Once a voice has been started (on noteOn), Internally,  it will become quickly `active` as soon it will processed by the dsp loop . 

If an application is looking for `fluid_rvoice_mixer_t::active_voices` to know the number of started voices this could lead in an unexpected result. For example:
Let a an application  that want to know the number of voices corresponding to a note been played (using a fresh created synth).
The application plays the note using calling` noteon()` , then immediately the application call  `get_active_voice_count()` (looking `fluid_rvoice_mixer_t::active_voices`) . It is possible that this API return 0 (i.e in the case of the started voices have not yet received by dsp loop (in the audio task))

Looking for this  internal `asynchronous fluid_rvoice_mixer_t::active_voices` has no sense for an application (imho).

src/utils/fluid_sys.c
derselbst: > g_get_monotonic_time() returns the same result that g_get_current_time()

I kept looking at the implementation on github and just couldnt believe it. But you're right. And now I have the answer: You need at least glib 2.53.3 to get accurate results. Here's the bug report: https://bugzilla.gnome.org/show_bug.cgi?id=783340

So, what shall we do? I still believe reprogramming `g_get_monotonic_time()` is not a good way to go. Bumping the minimum required glib version to 2.53.3 would be an overkill. And after all, this topic is only relevant for us developers who want to profile from time to time. How about the following solution:

```c
double
fluid_utime (void)
{
#if GLIB_MAJOR_VERSION == 2 && GLIB_MINOR_VERSION >= 28
  /* use high precision monotonic clock if available and warn
   * if this clock is actually implemented as low prec. clock,
   * i.e. in case glib is too old,
   * see: https://bugzilla.gnome.org/show_bug.cgi?id=783340
   */
#if defined(WITH_PROFILING) &&\
    defined(WIN32) &&\
    GLIB_MINOR_VERSION <= 53 &&\
    GLIB_MICRO_VERSION < 3
#warning "Time measurements may be inaccurate. Get at least glib 2.53.3 for high precision monotonic time."
#endif
    
  return g_get_monotonic_time();
  
#else
  // fallback to less precise clock
  GTimeVal timeval;
  g_get_current_time (&timeval);

  return (timeval.tv_sec * 1000000.0 + timeval.tv_usec);
#endif
}
```

derselbst:
> Internally, it will become quickly active as soon it will processed by the dsp loop

"Quickly" in case an audio driver is used. But much more unpredictably if the application renders audio manually.

But I see your point, I will just clarify the docs of `fluid_synth_get_active_voice_count()` then.

jjceresa:
> But I see your point, I will just clarify the docs of fluid_synth_get_active_voice_count() then.

Clarifying the docs will be useful as you have mentioned `"...But much more unpredictably if the application renders audio manually"`. Thanks.

src/utils/fluid_sys.c
jjceresa: How about this solution for which `glib have precedence` unless the used version is unable to do the job.
```
double
fluid_utime (void)
{
#if GLIB_MAJOR_VERSION == 2 && GLIB_MINOR_VERSION >= 28
/* use high precision monotonic clock if available and warn
   * if this clock is actually implemented as low prec. clock,
   * i.e. in case glib is too old,
   * see: https://bugzilla.gnome.org/show_bug.cgi?id=783340
   */
#if defined(WITH_PROFILING) &&  defined(WIN32) &&\
	/* glib < 2.53.3 */\
    (GLIB_MINOR_VERSION < 53 || (GLIB_MINOR_VERSION == 53 &&  GLIB_MICRO_VERSION < 3))
#pragma message( "Warning: fluid_utime()" )
#pragma message( "This glib version is unable to use high precision monotonic time." )
#pragma message( "Get at least glib 2.53.3 for high precision monotonic time." )
#pragma message( "Using WIN32 API high-resolution performance counter instead." )
	static LARGE_INTEGER Freq_static = {0,0};	/* Performance Frequency */
	LARGE_INTEGER PerfCpt;
	if (! Freq_static.QuadPart)
	{
		QueryPerformanceFrequency(&Freq_static);  /* Frequency value */
	}
	QueryPerformanceCounter(&PerfCpt); /* Counter value */
	return PerfCpt.QuadPart * 1000000.0/Freq_static.QuadPart; /* time in micros */
#else
	return (double) g_get_monotonic_time();
#endif
#else
  /* fallback to less precise clock */
  GTimeVal timeval;

  g_get_current_time (&timeval);

  return (timeval.tv_sec * 1000000.0 + timeval.tv_usec);
#endif
}
```

derselbst: Ok, do it if you want. But in this case I consider the `#pragma messages` to be more confusing to the user than helpful. It just works in the end, no matter how.

derselbst:
I applied the glib change. I think this is ready to merge now. If you want you can play around with the interactive rebase tool to cleanup the commit history and forcefully push your changes. Otherwise I would squash this as well.

I think there are no other pending patches once this is merged. So this might be a good time to enforce our style guide  (#179) by executing clang-format and clang-tidy?

jjceresa:
> I applied the glib change. I think this is ready to merge now.

>I think there are no other pending patches once this is merged. So this might be a good time to enforce our style guide (#179) by executing clang-format and clang-tidy?

Ok, thanks. I would just update the pdf about the reason of fluid_utime() change. This will be the last patch i think.

> If you want you can play around with the interactive rebase tool to cleanup the commit history and forcefully push your changes. Otherwise I would squash this as well.

I still never used the interactive rebase tool, but i will try. However i haven' t yet  looked what could be cleaned in the history (probably only the beginning because of a late master merge ?).

Please, could we wait for 2 or 3 days  ?

derselbst:
The basic idea of rebasing would be to get rid of unnecessary try and
error history as well as incomplete changes (like mine
058afb2de3b1cda1bcf06b5508b2195461669273), so that whenever someone
uses git blame in the future, it is clear what parts of the code were
changed by what activity. You could do this by squashing most of my
and your commits into the very first "Add profiling command interface"
commit. Or you could create a dedicated "Cleanup profile patch commit"
that combines all the cleanups.

Sure, just let me know when you're ready.

mawe42:
@derselbst 
> I think there are no other pending patches once this is merged. So this might be a good time to enforce our style guide (#179) by executing clang-format and clang-tidy?

I've got quite a large refactor patch for the soundfont loading mechanism. Cleaning it up, splitting low-level file access from default soundfont loader. All with the aim of adding a "lazy loading" feature. I think it'll be another 1 to 2 weeks before it's ready for review. Could you wait until then with the source formatting?

derselbst:
Sure. Will announce it on the mailing list duly.

jjceresa:
I have made some cleanup here locally (squashing using rebase -i ).
Could this be pushed now ?.

derselbst:
Yes.

derselbst:
It seems like you "polluted" the profile branch with some commits that were already on master, perhaps by issuing a `git merge master` some when. Was that intended?

I'm not sure what would happen if I try to merge this and I don't dare to try it out, sorry. I will restore the original commit history and squash it to master.

jjceresa:
> It seems like you "polluted" the profile branch with some commits that were already on master, perhaps by issuing a git merge master some when. Was that intended?

No, it was not intended. Locally i  only pulled the origin/profile since the beginning . But may be this commit https://github.com/FluidSynth/fluidsynth/pull/345/commits/244abcbc5c2c233d7c4fd38bcdac5bfd0071a153 is guilty.

>I'm not sure what would happen if I try to merge this and I don't dare to try it out, sorry. I will restore the original commit history and squash it to master.

Ok, sorry for this disorder.

derselbst:
No problem. Hard to tell for me what went wrong. Still believe there was some unintended merge. Nevermind.

Squash merged in 182e593665ab30c068cb2342cc48e3f4d360889e.

PR #347, opened 2018-03-06
Don't assume the paths are relative.

krop:
When building fluidsynth, we can use relative or absolute paths for
LIB_INSTALL_DIR or INCLUDE_INSTALL_DIR (or let the build system decide).

Let's check whether one of these two vars was defined before writing
fluidsynth.pc.

Fixes https://bugzilla.opensuse.org/1084081


derselbst:
Ah I see. This actually never worked before and was now uncovered because @scarabeusiv switched to `%cmake` macro [in OBS req. 577003](https://build.opensuse.org/request/show/577003) which uses absolute paths.

Thanks. I'll submit the patch to OBS.

PR #350, opened 2018-03-15
Force enable/disable OSS support

Ne01eX:
+ Force enable/disable OSS support;
+ OSS, PulseAudio and Jack disabled by default;
+ Sort options by alphabet :-).


derselbst:
Why do you change the default behaviour of those audio drivers? This will cause trouble for everybody else. I was trying to say that oss should be enabled by default by adding `option(enable-oss "..." ON)`.

Also note that all CI builds failed because of a typo `enif`.

Ne01eX:
> Why do you change the default behaviour of those audio drivers? This will cause trouble for everybody else. I was trying to say that oss should be enabled by default by adding option(enable-oss "..." ON).

Who this users with problems? :-D On Linux platforms, ALSA is the native audio subsystem. OSS - DEPRECATED.
OSX uses CoreAudio / CoreMidi.
On Windows, the DS is used.
The rest - they themselves know - what to use, and what not. :-) My settings are optimal. With the current settings, you allow users you to compile fluidsynth with what they really do not need. With mine - this does not happen. :-)

>Also note that all CI builds failed because of a typo enif.

Yes, I hurried, sorry. Fixed.

derselbst:
The current settings allow to automatically query dependencies already
installed on the system and compile fluidsynth to support those. With your
settings cmake is not even looking for them. There are numerous build
scripts out there that do not explicitly enable anything. Instead, features
are disabled explicitly if need be. Most users want a most functional
build. So, while your settings might be optimal for you, it unnecessarily
breaks established conventions, which I cannot accept.

Ne01eX:
Ok. Default settings has been returned. :-\
If you want to know my opinion, this is a bloat-way. : - \
See https://en.wikipedia.org/wiki/Software_bloat for more details.

derselbst:
This "bloated" fluidsynth will only be built if cmake finds the respective
development packages / libraries installed on the system. The existence of
those suggests that the user has an interest in the functionality provided
by those libs, which in turn justifies to build fluidsynth with additional
features implicitly, i.e. without any explicit "enable-feature" call. So
far I haven't seen any project that applies your "disabled-by-default"
logic. If you need a lightweight build, you'll need to start thinking by
explicitly disabling unneeded features.

After all the user stays under full control of whether to build a "bloated"
fluidsynth or not. Thus I disagree with you.

Thanks for your patch, will apply it later.

PR #353, opened 2018-03-23
initial review

langitktv:


derselbst:
So, where to start... great job for creating your first PR. Neither did you provide a description / motivation for this, nor does the title tell me anything. I suppose you neither read our [CONTRIBUTING notes](https://github.com/FluidSynth/fluidsynth/blob/master/CONTRIBUTING.md), nor did you consult @loki666. I can only guess that you want to add android support.  The reason because @loki666 has not submitted this "solution" to upstream is because it

1. is unmaintainable as it duplicates various files, and
2. breaks compilation for all non-posix OSs by removing glib.

The only option for upstream to add android support is by using the existing build system **and** glib as OSAL. If you want fluidsynth without glib on android, you have to use a custom fork. There is nothing further to review here, that's why I close this.

loki666:
:smile:

PR #355, opened 2018-03-25
Build fixes.

carlo-bramini:


derselbst:
Ok. Is this meant to fix any specific build error, i.e. steps to reproduce available?

carlo-bramini:
Hello,
well, it's not really eaxctly a "bug". I would say that's an improvement that I noticed with my DJGPP port, because I used a custom Makefile, since CMAKE does not recognize this platform. For the fix in DSOUND, I just replicated the same thing that it had been made into all other drivers.

derselbst:
Alright, thanks.

PR #357, opened 2018-03-30
fix wrong index of 'for' statement in (fluid_rvoice.c)

fulinux:
469   for (i = buffers->count; i <= bufnum; i++) {
470     buffers->bufs[bufnum].amp = 0.0f;     
471     buffers->bufs[bufnum].mapping = i;     
472   }

change to 

469   for (i = buffers->count; i <= bufnum; i++) {
470     buffers->bufs[i].amp = 0.0f;     
471     buffers->bufs[i].mapping = i;     
472   }

Is it right?


derselbst:
I think technically you are correct. But for the current implementation it doesnt matter, because `amp` and `mapping` are overwritten later on in `fluid_rvoice_buffers_set_amp()` and `fluid_rvoice_buffers_set_mapping()` anyway.

In fact I believe this loop is only meant to zero initialize those "new" buffer. This however is redundant, because the buffer is declared as global array and therefore guaranteed to be zero initialized. Anyway, thanks for the hint, I noted this in #197. But for now I dont want to change it.

PR #361, opened 2018-04-06
Introduce unit tests

derselbst:
Start using cmake to build and exec some unit tests if requested (target `check`). Begin with a simple test for the sample cache.


mawe42:
Yes! Very nice! I kept thinking "we could really do with unit tests" quite often while doing refactors. Thanks, this will help a lot.

derselbst:
I think this should be good now. I disabled unit tests for macOS because the linker doesnt find `libintl` for some reason (the workaround from stackoverflow didnt work either...). And appveyor without vcpkg [doesnt find gthread-2.0.lib](https://ci.appveyor.com/project/derselbst/fluidsynth/build/834) when it comes to unit tests. Dont really care any further, vcpkg build works fine with unit tests.

It is also possible to use fluidsynths private headers in tests, simply put `#include "synth/fluid_synth.h"` or so...

test/README.md
```diff
@@ -0,0 +1,6 @@
+ 
+This directory contains small executables to verify fluidsynths correct behaviour, i.e. unit tests. Although some tests might serve as educational demonstration of how to use certain parts of fluidsynth, they are **not** intended to do so! It is most likely that those tests will consist of many hacky parts that are necessary to test fluidsynth (e.g. including fluidsynth's private headers to access internal data types and functions). For user applications this programming style is strongly discouraged! Do **not** blindly use the tests as template for your application! Keep referring to the documentation and code examples listed in the [API documentation](http://www.fluidsynth.org/api/).
```
mawe42: Could you line wrap this file? Keeps it in line with all the other doc files we have (I think...) and makes life easier in general :-)

derselbst:
> because it's missing ```/LIBPATH:C:/deps/lib" in the linker command.

Yes, but dunno why...

PR #364, opened 2018-04-08
Unify calling conventions for rvoice update functions 

derselbst:
* Introduce a proper function typedef for functions that need to be called by `rvoice_eventhandler`
* Get rid of that huge macro crap in fluid_rvoice_event.c and all its branch indirection it caused

During this cleanup I recognized that there are many duplicate executions, like setting the sample rate of a voice on synth creation which is done three (!) times. Partly fixed that, however this made other bugs and crashes come to light. Will fix them later, just wanting to provide this for reference and feedback now.

Addresses #197.

### DIFF
--- a/src/rvoice/fluid_rvoice_event.h
+++ b/src/rvoice/fluid_rvoice_event.h
@@ -88,11 +88,6 @@ int fluid_rvoice_eventhandler_push(fluid_rvoice_eventhandler_t* handler,
 int fluid_rvoice_eventhandler_push_ptr(fluid_rvoice_eventhandler_t* handler, 
                                 fluid_rvoice_function_t method, void* object, void* ptr); 
 
-int fluid_rvoice_eventhandler_push5(fluid_rvoice_eventhandler_t* handler, 
-                                fluid_rvoice_function_t method, void* object, int intparam, 
-                                fluid_real_t r1, fluid_real_t r2, 
-                                fluid_real_t r3, fluid_real_t r4, fluid_real_t r5);
-
 int fluid_rvoice_eventhandler_push_param(fluid_rvoice_eventhandler_t* handler,
                                          fluid_rvoice_function_t method, void* object,
                                          fluid_rvoice_param_t param[EVENT_PARAMS]);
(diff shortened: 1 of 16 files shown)


src/utils/fluidsynth_priv.h
```diff
@@ -205,6 +205,22 @@ typedef struct _fluid_server_socket_t fluid_server_socket_t;
...
 
+typedef union _fluid_rvoice_param_t
+{
+    void* ptr;
+    int i;
+    fluid_real_t real;
+} fluid_rvoice_param_t;
+typedef void (*fluid_rvoice_function_t)(void* obj, fluid_rvoice_param_t* param);
+
+enum { EVENT_PARAMS = 6 };
```
mawe42: Maybe call this MAX_EVENT_PARAMS? I think that would make the function signatures a bit clearer, indicating that not all params are used by all event handlers...

src/rvoice/fluid_rvoice_event.h
```diff
@@ -89,16 +82,15 @@ fluid_rvoice_eventhandler_get_finished_voice(fluid_rvoice_eventhandler_t* handle
 
 
 int fluid_rvoice_eventhandler_push(fluid_rvoice_eventhandler_t* handler,
```
mawe42: Maybe rename this to ```fluid_rvoice_eventhandler_push_int_real``` and ```fluid_rvoice_eventhandler_push_param``` to ```fluid_rvoice_eventhandler_push```. The latter is now the general case and could do with a generic name, I think.

derselbst:
> the function parameters are now sort of hidden

The macro makes it evident that it's an rvoice update function. It makes it simple to add new functions and retaining the same calling conventions by avoiding typos. I think that makes it worth keeping the macro rather than declaring functions as usual.

jjceresa:
> just  wanting to provide this for reference and feedback now.

Many thanks for this cleanup. This  makes the source more readable and brings the performance of `fluid_rvoice_event_dispatch` to its natural maximum. Nice job !

PR #370, opened 2018-04-22
Only import Soundfont instruments once

mawe42:
This is a follow up PR to the dynamic-sample-loading feature. It reduces the memory footprint of imported Soundfonts by ensuring that instruments are only imported once and then re-used in each preset zone that links to them.

Previously, each preset zone got its own copy of the complete instrument information from the Soundfont, which increased memory consumption dramatically for Soundfonts with lots of preset/instrument zones. For example, the memory footprint of the standard FluidR3_GM Soundfont is reduced by 30 MB after this change.

In order to allow this change, I had to split the mutable parts of the instrument zones (i.e. the zone ranges) from the imported instrument, creating a new fluid_voice_zone_t structure.

### DIFF
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -39,6 +39,8 @@ static int unload_preset_samples(fluid_defsfont_t *defsfont, fluid_preset_t *pre
 static void unload_sample(fluid_sample_t *sample);
 static int dynamic_samples_preset_notify(fluid_preset_t *preset, int reason, int chan);
 static int dynamic_samples_sample_notify(fluid_sample_t *sample, int reason);
+static int fluid_preset_zone_create_voice_zones(fluid_preset_zone_t* preset_zone);
+static fluid_inst_t *find_inst_by_idx(fluid_defsfont_t *defsfont, int idx);
 
 
 /***************************************************************
@@ -247,6 +249,11 @@ int delete_fluid_defsfont(fluid_defsfont_t* defsfont)
       fluid_defpreset_preset_delete(preset);
   }
   delete_fluid_list(defsfont->preset);
+
+  for (list = defsfont->inst; list; list = fluid_list_next(list)) {
+      delete_fluid_inst(fluid_list_get(list));
+  }
+  delete_fluid_list(defsfont->inst);
 
   FLUID_FREE(defsfont);
   return FLUID_OK;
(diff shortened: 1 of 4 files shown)


mawe42:
Thanks for the review!

derselbst:
Hm, there seems to be a strange bug introduced by this PR. Loading a Soundfont and playing a midi works fine the first time, unloading and destroying the synth, recreating the synth, reloading the same font and playing a midi may cause missing audio at the beginning of the synthesization. But not always. Happens very randomly and currently I cant give steps to reproduce. All I know is that it worked in 354bea9cfc34af38c728c40fe15832d6e7d9ed93.

Will further investigate in the next day(s) unless you have a brainstorm.

mawe42:
Weird... did you notice the effect on current master, or already immediately after merging this PR in 9ad07430ea710a5bf346880cde444e4a8e23b22a?

mawe42:
And does it happen with or without dynamic sample loading?

mawe42:
The fact that it works once and breaks after deleting and recreating the synth sounds a bit like an uninitialized pointer. One idea: initialize the ```zone_range``` pointer to NULL in ```new_fluid_voice```. Not really sure how that could cause problems, but it should still be initialized properly anyway.

derselbst:
> did you notice the effect on current master, or already immediately after merging this PR

Reproducible ever since this PR has been merged.

> And does it happen with or without dynamic sample loading?

Doesnt matter, always happens randomly. Even right in between of a song, when a voice of some instrument is played for the first time  as it seems.

> One idea: initialize the zone_range pointer to NULL in new_fluid_voice.

Nope, still happens :( will continue examining...

mawe42:
> Doesnt matter, always happens randomly. Even right in between of a song, when a voice of some instrument is played for the first time as it seems.

Only if you create, throw away and then recreate a synth? Or can you also reproduce it with the fluidsynth binary?

mawe42:
Another idea: could you comment out the creation of the unload timer in ```fluid_synth_sfont_unref``` and see if the problem persists?

derselbst:
Found it! When creating `FLUID_NEW(fluid_voice_zone_t)`, `voice_zone->range.ignore` is dangling, causing `fluid_zone_inside_range()` to return false in `fluid_defpreset_noteon()`. Committing fix.

> Are you doing fast file rendering, or real-time midi playback?

Fast rendering with ANMP.

mawe42:
> Found it! When creating FLUID_NEW(fluid_voice_zone_t), voice_zone->range.ignore is dangling, causing fluid_zone_inside_range() to return false in fluid_defpreset_noteon(). Committing fix.

Great, thanks!

PR #371, opened 2018-04-23
Preload presets in fluid_player before playback (if using dynamic sample loading)

mawe42:
**As discussed in the mail thread, the idea behind this pull request is too much of a hack and most likely not even needed. Will be closed in a few days***

This PR implements the preset sample preload function discussed in
http://lists.gnu.org/archive/html/fluid-dev/2018-04/msg00000.html

Preloading is always active on the MIDI player if dynamic sample loading is enabled.

This feature is best tested with a MIDI file containing lots of program changes and a fairly large SF3 Soundfont like the FluidR3_GM.sf3 from MuseScore, as uncompressing the Vorbis samples takes quite a wihle, even on fast computers.

Here is an example MIDI file that cycles through 10 presets in quick succession: [progchange-fast.zip](https://github.com/FluidSynth/fluidsynth/files/1939691/progchange-fast.zip) 

It should output a steady stream of notes. Without preloading, the playback is interrupted quite often.

### DIFF
--- a/include/fluidsynth/sfont.h
+++ b/include/fluidsynth/sfont.h
@@ -241,6 +241,27 @@ typedef int (*fluid_preset_get_num_t)(fluid_preset_t* preset);
 typedef int (*fluid_preset_noteon_t)(fluid_preset_t* preset, fluid_synth_t* synth, int chan, int key, int vel);
 
 /**
+ * Method to load a virtual SoundFont preset when using dynamic sample loading
+ * @param preset Virtual SoundFont preset
+ * @return #FLUID_OK on success (0) or #FLUID_FAILED (-1) otherwise
+ */
+typedef int (*fluid_preset_load_t)(fluid_preset_t* preset);
+
+/**
+ * Method to unload a virtual SoundFont preset when using dynamic sample loading
+ * @param preset Virtual SoundFont preset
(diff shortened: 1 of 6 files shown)


jjceresa:
> As discussed in the mail thread, the idea behind this pull request is too much of a hack and most likely not even needed.

Despite the fact that `"sample preload idea"` is too much specific to the need of each application it is an interesting idea that could be applied also in application for` live performance "preset preloading"` as the program changes needed for each song (among others things)  could be prepared and saved in MIDI files.

I just wanted to thanks for the works done and the useful examples given by this PR.

jjceresa:
> Leave out the hackish MIDI scanning. And then again... nobody has expressed an interest in that feature yet. Probably better to wait until someone really needs it.

I understand this. As an example i don't usually use internal fluidsynth sequencing and MIDI file because i use MidiShare library for theses jobs. I really think that `preloading `should be an application functionality. However `dynamic samples loading` is a fluidsynth nice feature. Good job.

mawe42:
>  I really think that preloading should be an application functionality.

Yes, but for an application to be able to control preloading, we would need an interface for it.

Anyway, I'm closing this now as announced. I'll keep the branch locally, just in case someone comes up with a need for it.

PR #372, opened 2018-04-25
rvoice_mixer refactorings + cleanups

derselbst:
The final cleanup of `rvoice_mixer` and related components. No functional change intended.

Addresses #197.


mawe42:
**Core i5 laptop (buffer size 128)**
```
                   ------------------------------------------------------------------------------
                   nVoices| total(%)|voices(%)| reverb(%)|chorus(%)| voice(%)|estimated maxVoices
                   -------|---------|---------|----------|---------|---------|-------------------
master, cores=1       1200|   83.003|   82.237|     0.444|    0.322|    0.068|             1450
refactor, cores=1     1200|   83.031|   82.263|     0.451|    0.317|    0.068|             1449
master, cores=4       1200|   43.065|   41.340|     1.163|    0.563|    0.034|             2865
refactor, cores=4     1200|   43.248|   41.503|     1.182|    0.562|    0.034|             2853
```

**ARM Dual-Core Cortex-A7 (always 1 core, buffer size 64) **
```
                   ------------------------------------------------------------------------------
                   nVoices| total(%)|voices(%)| reverb(%)|chorus(%)| voice(%)|estimated maxVoices
                   -------|---------|---------|----------|---------|---------|-------------------
master                 120|   90.227|   83.823|     4.691|    1.713|    0.690|              135 
refactor               120|   90.033|   83.711|     4.547|    1.774|    0.689|              135
```

derselbst:
> So all in all this refactor seems to have no effect on the performance, which is perfect.

Same here.

Ok, will leave it open for a few days.

jjceresa:
Sometimes tracking lost of performance isn't easy.

- tip 1: measurement of `very small performance difference`  must be done at least on  `MinSizeRel `release versions  (because RelWithDebInfo  on Windows XP or Debian masks theses differences).
- tip 2: For `voice cpu load` measurement, always `turn off chorus and reverb` (because voices cpu load are a bit dependent of presence of reverb or chorus fx unit).

For information, here the tracking of lost of performance (MinSizeRel) somewhere probably before  any `rvoice refactor` and  https://github.com/FluidSynth/fluidsynth/pull/372/commits/384f05e77c6583fe6b23c59dfa4dfeebb83bc370 (disable padding warnings).

```
branchs:       master before any refactor             rvoice-refactor-3	
--------------------------------------------------------------------------
Voice %	       0.160			               0.171

lost of performance: 7 %
```

jjceresa:
To avoid signed/unsigned warning,  `int ` could be changed to `unsigned int` type.

https://github.com/FluidSynth/fluidsynth/blob/384f05e77c6583fe6b23c59dfa4dfeebb83bc370/src/sfloader/fluid_defsfont.c#L269

mawe42:
> To avoid signed/unsigned warning, int could be changed to unsigned int type.

That code hasn't been touched as part of this PR, but comes from the dynamic sample loading.

mawe42:
> tip 1: measurement of very small performance difference must be done at least on MinSizeRelrelease versions (because RelWithDebInfo on Windows XP or Debian masks theses differences).

The measurements I made were done with "Release" builds.

> tip 2: For voice cpu load measurement, always turn off chorus and reverb (because voices cpu load are a bit dependent of presence of reverb or chorus fx unit).

Wouldn't it be just as good to always turn chrous and reverb on?

> For information, here the tracking of lost of performance (MinSizeRel) somewhere probably before any rvoice refactor and 384f05e (disable padding warnings).

Can you specify exactly which commit you used as reference?

jjceresa:
> Wouldn't it be just as good to always turn chrous and reverb on?

In a perfect word, normally `voices cpu load ` should be independent of effect unit presence(reverb,chorus).
However, the send parameter (for reverb or chorus) is computed only if  the corresponding buffers are
prepared (in `fluid_mixer_buffers_prepare()`) and used (in `fluid_rvoice_buffers_mix()`), so durations `synth_one_block:one_voice`  and `synth_one_block:all voices`  are  a bit
dependent of presence of reverb or chorus fx unit (i.e theses duration are a bit higher when reverb or chorus reverb is on due to send parameter computation).
It is easy to show this using the mode "full profiling printing" using the command: `prof_set_print 1`
Note this tip is only pertinent when measuring very small difference.
Saying that we could also always turn chorus and reverb on.

> Can you specify exactly which commit you used as reference?

Not  sure but it was a commit just after 182e593 ("add profiling command interface").

PR #374, opened 2018-05-05
Enable / Improve vectorizion in rvoice_mixer

derselbst:
This PR basically proposes to align the sample buffers of `rvoice_mixer` to the 64 byte boundary in order to make mixing voices more efficient.

Aligned data access will help the compiler to vectorize loops and therefore generate more efficient code. In particular using aligned arrays allows to get rid of the so called peel loop that would precede any vectoized loop in order to make sure to have aligned data access in the vectorized loop. Because fluidsynth usually only mixes `FLUID_BUFSIZE` samples each run, it is important to get rid of those peel loops IMO since they are unnecessary and might dominate mixing.

Almost every compiler (expect MSVC) has its own custom set of built-in macros that allow the programmer to instruct the compiler that certain arrays can be assumed to be aligned to some boundary, see section 3 for some more details: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0886r0.pdf

Instead of going this vendor specific way, I thought it would be better to use an open standard, i.e. openMP, to instruct the compiler of aligned arrays and tell it to vectorize the loop. This is achieved by using

`#pragma omp simd aligned(arr:64)`

to assert that array pointed to by `arr` is aligned, i.e. `(uintptr_t)arr % 64 == 0`. This feature requires openMP 4.0 to work. Compilers that dont know about that (like MSVC) should just ignore the pragma and continue to compile as before.

Using the intel compiler I get the following optimization report of `fluid_rvoice_buffers_mix()` before having done this fancy alignment:

```
LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice.c(455,7)
<Peeled loop for vectorization, Multiversioned v1>
   remark #25015: Estimate of max trip count of loop=7
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice.c(455,7)
<Multiversioned v1>
   remark #25228: Loop multiversioned for Data Dependence
   remark #15300: LOOP WAS VECTORIZED
   remark #15442: entire loop may be executed in remainder  <-- NOTE!
   remark #15448: unmasked aligned unit stride loads: 1 
   remark #15449: unmasked aligned unit stride stores: 1 
   remark #15450: unmasked unaligned unit stride loads: 1   <-- NOTE!
   remark #15475: --- begin vector cost summary ---
   remark #15476: scalar cost: 10 
   remark #15477: vector cost: 1.620 
   remark #15478: estimated potential speedup: 4.950 
   remark #15488: --- end vector cost summary ---
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice.c(455,7)
<Remainder loop for vectorization, Multiversioned v1>
   remark #15301: REMAINDER LOOP WAS VECTORIZED
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice.c(455,7)
<Remainder loop for vectorization, Multiversioned v1>
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice.c(455,7)
<Multiversioned v2>
   remark #15304: loop was not vectorized: non-vectorizable loop instance from multiversioning
   remark #25439: unrolled with remainder by 2  
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice.c(455,7)
<Remainder, Multiversioned v2>
LOOP END
```

As you can see there is a peel loop. Additionally we have multiple versions of the same `for` loop but for different alignment and data dependence scenarios. Using alignment and restricted pointers as this PR proposes this reduces to:

```
LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice_mixer.c(285,7) inlined into /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice_mixer.c(895,5)
    remark #15301: OpenMP SIMD LOOP WAS VECTORIZED
    remark #15448: unmasked aligned unit stride loads: 2 
    remark #15449: unmasked aligned unit stride stores: 1 
    remark #15475: --- begin vector cost summary ---
    remark #15476: scalar cost: 10 
    remark #15477: vector cost: 0.870 
    remark #15478: estimated potential speedup: 7.970   <-- NOTE!
    remark #15488: --- end vector cost summary ---
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice_mixer.c(285,7) inlined into /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice_mixer.c(895,5)
<Remainder loop for vectorization>
    remark #15301: REMAINDER LOOP WAS VECTORIZED
LOOP END

LOOP BEGIN at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice_mixer.c(285,7) inlined into /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/src/rvoice/fluid_rvoice_mixer.c(895,5)
<Remainder loop for vectorization>
LOOP END
```

Note that I've compiled with `enable-floats=1` and `-mtune=native`. Please, feel free to test this and give me feedback how it works for you. CMake should correctly report whether openMP 4.0 is available or not. Only in case it is available, this may give you a performance boost of a few seconds (depending how many voices are active).


mawe42:
Seems to work fine on Intel and ARM Linux. Speedup on both platforms is less than 1% in estimated voices.

PR #375, opened 2018-05-05
Cleanup pointless check in fluid_rvoice_check_sample_sanity

jjceresa:
The check is already done by the unique caller `fluid_rvoice_write()` as this.
https://github.com/FluidSynth/fluidsynth/blob/787f5e3302fa7fe1faa61a678508ad9d35e50cf2/src/rvoice/fluid_rvoice.c#L277-L278


derselbst:
Ok.

PR #380, opened 2018-05-07
new reverb

jjceresa:
This PR proposes a new reverb which is a `late `reverb (like the actual reverb) which is convenient for Fluidsynth sound engine.
The `parameters` are the same than the actual reverb. Also the `default response of these parameters` are rather the same.
The intended goal was:
- producing a better quality reverberation tail than actual reverb with far less "`ringing`" using rather no more `cpu load` if possible.
- reducing the memory size consumed by the delay lines ( results give: -110 kbytes(-54%)  to -144 kbytes(-72%)).

The macro `NBR_DELAYS` allows to choose 8 or 12 delays lines.
Although 8 lines give good result, using 12 delay lines brings the overall quality a bit higher. This quality augmentation is noticeable particularly when using long reverb time (roomsize = 1) on solo instrument with long release time. Of course the cpu load augmentation is +50% relatively  to 8 lines. 

As a general rule the reverberation tail quality is easier to perceive by ear when using: 
  - percussive instruments (i.e piano and others).
  - long reverb time (roomsize = 1).
  - no damping (damp = 0).


mawe42:
Cool, looking forward to testing this! Have you done any performance measurements yourself?

jjceresa:
> Have you done any performance measurements yourself?

Yes, using `NBR_DELAYS = 8` (default) you should expect rather the same performance than actual reverb.
The `default` MACROS should be sufficient to give an `immediate possibility of compare` with actual reverb.
The full description of compare with the actual reverb is given in a `table `in the top of fluid_rev.c.

mawe42:
Would it maybe be beneficial to be able to disable the denormal handling with a compile-time switch and rely on ```-ffast-math``` and maybe also (```-msse -mfpmath=sse```)? See http://carlh.net/plugins/denormals.php for more info.

jjceresa:
Thanks for the link.

> Would it maybe be beneficial to be able to disable the denormal handling ?

Actual denormal handling introduces  a loss of -6% of the `reverb cpu load` (i.e  20 % of only `one voice cpu load` ) (on cpu with math co-processor). 
This should be tested on others platform. I haven't an ARM without math co-processor . May be it worth to test this ?

derselbst:
Nice work. For me a difference is barely audible. Hope this is intended?

jjceresa:
> For me a difference is barely audible.

Please, could you develop a bit which difference ?

derselbst:
> Please, could you develop a bit which difference ?

No, because I dont really hear a difference (barely == almost not).

mawe42:
With a roomsize of 1, I can hear quite a difference. The existing reverb has a lot of high frequency ringing, probably an artifact of the comb filters. The new implementation has no unnatural ringing that I can hear. So much better in that regard!

But the new implementation has more "wobble" than the old one. I set roomsize to 1 and leave all other settings unchanged. Then I play a single short note of piano and listen to the reverb. The new reverb sounds a little bit like there is a phaser or chorus on it. The existing reverb sounds cleaner and more natural, in my opinion. But maybe it's just that the default parameters need tweaking?

derselbst:
> But the new implementation has more "wobble" than the old one.

Yes, the only difference I noticed so far. I think it's the `sinus_modulator` built into each delay line. A little less "wobble" might be good.

jjceresa:
>The existing reverb has a lot of high frequency ringing,probably an artifact of the comb filters.

Any reverb with only 8 comb filters will produces audible resonant frequencies (`ringing`). This is due to the fact that there not enough numbers of resonant frequencies and the frequency are hear individually. If the reverb will have at least 30 combs filters (with sufficient lengths) (as it should have !) the resonant frequencies will be so near each other that they could not be heard individually.

>The new implementation has no unnatural ringing that I can hear. 

Using only 8 comb filters a possible solution to cancel the resonant frequencies (without putting damp to 0.2) is to modulate a bit the comb filter (this simulates more peaks of frequency moving very near each to other). This allows to hear the full spectrum when intended (putting damp to 0) without hearing the building of unwanted resonant frequencies.

> The new reverb sounds a little bit like there is a chorus on it.

> I think it's the sinus_modulator built into each delay line

This is the inconvenient of a modulated delay line. It produces a unwanted frequency modulation (like chorus). However if the modulation depth (`MOD_DEPH`)  is very low(< 6 samples) and the frequency of modulation (`MOD_FREQ`) is very low( < 2 Hz) but sufficient ( > 0.5 Hz) this should be acceptable when we intend to listen the full spectrum (damp to 0). 
When using damp to 0.2 (default) we intend to diminish the reverb time of high frequency and in this case the `unwanted chorus` become more perceptible (i.e less acceptable). 

> A little less "wobble" might be good.

In summary:
When full spectrum is intended (damp at 0):

- Using NBR_DELAYS to 8,   MOD_DEPTH must be > 4 to cancel the unwanted "ringing". 
- Using NBR_DELAYS to 12, MOD_DEPTH to 3 is sufficient to cancel the unwanted "ringing". 

When less reverb time is intended for high frequency (damp > 0):

- high frequency are less often heard so maybe `MOD_DEPTH` could be diminished a bit. So may be `MOD_DEPTH` should be dependent of `damp` to get acceptable chorus. 

Will see what can done.

jjceresa:
Since now modulation depth (`MOD_DEPTH`) gives a compromise between both the unwanted resonant  frequencies ("**ringing**") and the unwanted **chorus**. However, the chosen value depend also of the instrument under reverberation. So it appears that a fixed value for MOD_DEPTH isn't appropriate.

Although  i am fan of less parameters as possible,  i think it worth that `modulation depth` should be chosen by the user via a new parameter that could called `moddepth` . Any thoughts ?

derselbst:
One important issue: I'm missing a copyright notice. Did you wrote this completely by yourself or is it based on some implementation? If you did it yourself, please add your name + copyright + hint for LGPL 2.1+.

>  i think it worth that modulation depth should be chosen by the user via a new parameter that could called moddepth

Wouldnt go that far for now. Let us continue rehearsal first.

jjceresa:
One important issue: I'm missing a copyright notice. Did you wrote this completely by yourself or is it based on implementation?.

All the concepts and ideas comes completely from all the research papers we found on w.w.w  on this subject since 40 years. In this source there no ideas or concepts that comes from me. I am responsible of the translation of theses ideas in C language. 
I have chosen a suited reverb model described here:

- https://ccrma.stanford.edu/~jos/Reverb/FDN_Late_Reverberation.html
 (indicated in the header of fluid_rev.c ). 
- The inner design math formulas comes from these ccrma  jos pages:
https://ccrma.stanford.edu/~jos/Reverb/First_Order_Delay_Filter_Design.html.
https://ccrma.stanford.edu/~jos/pasp/First_Order_Allpass_Interpolation.html.
- The stereo unit comes from the actual fluidsynth reverb.

I must admit that my knowledge about copyright statements is very limited. Also don't intend to claim any copyright of my own. You can probably help on this copyright subject.

derselbst:
Right, I'm aware that the concepts exists and are public domain as it seems. I was only referring to the source code.

> I am responsible of the translation of theses ideas in C language.

So does that mean that you wrote the source code completely yourself, without using or translating any existing source? In this case, just add the existing copyright notice found in fluidsynths other source files to the top of `fluid_rev.c`, replacing Peter Hanappe with your name. This is just to make evident that you (the creator and therefore copyright holder of this source code) license this code to fluidsynth under LGPL2.1+. 

...Simply to avoid this issue we had at last year with Juergen Muellers chorus implementation.

jjceresa:
> Let us continue rehearsal first.

Sure. Here the conditions and results of many experiments:
`Conditions`:
For many instruments under reverberation (roomzize=1, damp=0):

- Minimum Keyboard range (if possible) C2 to C6 octaves. A simple test is playing each white key consecutively (one pass at low velocity, then another pass at high velocity).
1) 2 pianos (one percussive then one less percussive). 
2) Then no percussive instrument ( oboe, flute,sax,..) 

`Results`.
For many instruments, the range for MOD_FREQ and MOD_DEPTH should be:
- MOD_FREQ: [0.5 ..2.0] (in Hz).
- MOD_DEPTH: [3..6]  (in samples).

Values below the lower limits are often not sufficient to cancel  unwanted "**ringing**"(resonant frequency).
Value above upper limits augment the unwanted "**chorus**".

jjceresa:
> So does that mean that you wrote the source code completely yourself, without using or translating any existing source?

Not completely because i have read a lot of research papers and read in this papers examples of implementations written in others languages. So in this regard the source i have written (3 years ago) is based on the works of the author of this papers.
So if i understand correctly the copyright rules, we should pay attention to mention the initial authors of this initial works (as it is the case in the actual fluidsynth fluid_rev.c that mention:  `"Written by Jezar...." translated to C by Peter Hannape`) ? . am  i correct ?.

derselbst:
> So if i understand correctly the copyright rules, we should pay attention to mention the initial authors of this initial works

Yes, the example implementations that you've used must be credited and they must compatible with LGPL (e.g. public domain).

I'm not sure if it's necessary to do this that detailed. If the examples you've used are part of the papers you've already cited it's probably ok to silently use them by translating them to C. Just look them through again so we dont miss anything.

In every case I'd suggest to apply fluidsynth's copyright note.

jjceresa:
> I'm not sure if it's necessary to do this that detailed. If the examples you've used are part of the papers you've already cited it's probably ok to silently use them by translating them to C. Just look them through again so we dont miss anything.

I have looked in the papers. The code examples i have read are just only part of the papers. There are no copyright mention, no licence mention. There are only the author name of the paper.

Should be a good idea to contact the author of the papers ?

>In every case I'd suggest to apply fluidsynth's copyright note.

You mean, of course  just add the existing copyright notice found in fluidsynths other source files to the top of fluid_rev.c.

derselbst:
> Should be a good idea to contact the author of the papers ?

No, I dont think that's necessary.

> You mean, of course just add the existing copyright notice found in fluidsynths other source files to the top of fluid_rev.c.

Yes please.

src/rvoice/fluid_rev.c
```diff
@@ -1,517 +1,1313 @@
...
+	*/
+
+	for(i = 0; i < NBR_DELAYS; i++)
+	{
+		/* sets local delay lines's parameters */
+		result = set_mod_delay_line(&late->mod_delay_lines[i], 
+		                             delay_length[i], MOD_DEPTH, MOD_RATE );
+		if (result == FLUID_FAILED)
+		{
+			delete_fluid_rev_late(late);
```
derselbst: Remove this delete_* call pls. Cleanup should be done by parent constructor function `new_fluid_revmodel()`.

jjceresa: All cleanup done.

>Please add NULL-guards to destructor functions: fluid_return_if_fail(rev != NULL);

Testing rev to  not NULL but not using fluid_return_if_fail(), as this is an internal  reverb interface API , not a public API. Hope it is correct ?

derselbst:
Finally coming back to this. I had a listen to several MIDI files and overall I think fdn-reverb does a pretty good job.

In case you're interest, [here's a demo tune](https://drive.google.com/open?id=1kC6tmsi6A-rlUIsGUfPSKuOrRtLAeHEN) rendered with `freeverb`, `fdn MOD_DEPTH 5` and `fdn MOD_DEPTH 6`. Interestingly there seems to be some strange resonant frequency for the fdn reverb when the short flute starts playing at 53 seconds. However this is the only "strange behaviour" I could reproduce with fdn.

All in all I cannot tell the difference between 8 or 12 delay lines. However I would vote for reducing the MOD_DEPTH to 5 in order to avoid any unnecessary "wobble".

jjceresa:
> Interestingly there seems to be some strange resonant frequency for the fdn reverb when the short flute starts playing at 53 seconds.

Sorry, i cannot capture a resonant frequency around 53s. What i hear at 53s is the `flute `that end on a C5 note then the `wood block`(or something like that) restart the theme. Anyway this "exploration in the underground" is funny to listen :).

> All in all I cannot tell the difference between 8 or 12 delay lines.

Yes, it is very difficult to tell the difference. I have tried also to see a difference between 16 lines  and 12 lines (in another implementation). It was also much more difficult to tell a difference by ear.

>However I would vote for reducing the MOD_DEPTH to 5 in order to avoid any unnecessary "wobble".

This seems  a good choice. I have noticed that MOD_DEPTH should be:

- Using NBR_DELAYS to 8, MOD_DEPTH must be >= 4 to cancel the unwanted "ringing".[4..6]
- Using NBR_DELAYS to 12, MOD_DEPTH to 3 is sufficient to cancel the unwanted "ringing".[3..6]

So using the `default NBR_DELAYS to 8` we could set the `default MOD_DEPTH to 5`.

derselbst:
> Sorry, i cannot capture a resonant frequency around 53s.

Maybe it's no resonant frequency, it's just that the reverb of this short flute sounds somewhat damped compared to freeverb... at least to me. Perhaps it's better to hear at 1:23.

derselbst:
It looks like your tweaks were correct, now it sounds better to me.

However your latest change has made the wobble become more dominant again. I suggest to reduce the MOD_RATE to 4, which still sounds fine to me. Do you agree?

jjceresa:
> It  looks like your tweaks were correct, now it sounds better to me.

Thanks for checking. 

> However your latest change has made the wobble become more dominant again
I suggest to reduce the MOD_RATE to 4, which still sounds fine to me. 

You mean probably about `MOD_DEPTH `change (not MOD_RATE) ?

>Do you agree?

Talking about MOD_DEPTH, of course i agree because it depends of the instrument, ears and taste. So it can be reduced to 4.

jjceresa:
>However your latest change has made the wobble become more dominant again.

Please, as i dont know how to reproduce the "wobble" you have heared,  i guess (and hope) the latest change you refer (while the wobble domination augmentation was observed) is `Change on internal gain scale`(https://github.com/FluidSynth/fluidsynth/pull/380/commits/dead0efcc8f9c5bead26e8bbb325ea601b3afbfd) ?.
Thanks for confirming this previous observation. (This will help to note the interaction or singularity  between presumed distinct inner parameters).

derselbst:
Yes I was referring to the internal gain scale. Specifically audible for high frequent samples. Here's the midi demo attached. Use it [with this soundfont](https://github.com/derselbst/ANMP-data/raw/master/soundfonts/N64/DK64.sf2).


[DK64-JJ-Underground.zip](https://github.com/FluidSynth/fluidsynth/files/2080291/DK64-JJ-Underground.zip)

derselbst:
All in all I'm fine with this. Ringing is gone, code is well documented, tiny wheeny drawback is the whobble, but that's only audible for a very few sounds, so nevermind. However I dont want to merge this for 2.0 as it already brings a lot new features. Instead I want to keep this for whatever comes after 2.0.

@jjceresa Could you perhaps sooner or later bring this feature to the mailing list?

jjceresa:
> tiny wheeny drawback is the whobble, but that's only audible for a very few sounds, so nevermind.

Yes, i understand. This compromise between ringing tone and wobble (know also as flutter) is really difficult and doesn't work in all situation.

>However I dont want to merge this for 2.0 as it already brings a lot new features. Instead I want to keep this for whatever comes after 2.0.

Yes of course. I understand this.
Also tweaking isn't not yet finished, i have noticed that: 
- low input gain (FIXED_GAIN < 1.0) should be reintroduced.
- suppressed stereo gain should be reintroduced.
Both these should enhance  flatness of frequency response.

>Could you perhaps sooner or later bring this feature to the mailing list?

Yes, i think it is necessary that more ears give a try. For example, i already know that jOrgan-user was interested to try this. Graham Goode (jOrgan-user) have said that it should be easier for them to try a fluidsynth executable (Windows). Did you kow if CI travis is able to build executable of new-reverb branch ? (i don't remember).

derselbst:
> Did you know if CI travis is able to build executable of new-reverb branch ? (i don't remember).

Sorry, just realized I haven't respond to this. Travis and AppVeyor continiously build all branches that are being pushed to. Currently Travis does not produce build artifacts. However as it seems you want windows binaries, you can just use the build artifacts from AppVeyor, e.g.: https://ci.appveyor.com/project/derselbst/fluidsynth/build/1192/job/8adfyve7lrsfgw4g/artifacts

jjceresa:
> However as it seems you want windows binaries, you can just use the build artifacts from AppVeyor.

I already know that jOrgan-user was interested to try this. Graham Goode (jOrgan-user) have said that it should be easier for them to try a fluidsynth executable (Windows).

Note that the binaries build by AppVeyor doen't run on Windows XP. These probably run on Windows 7 and above.

jjceresa:
> Could you please try this build:https://ci.appveyor.com/project/derselbst/fluidsynth-g2ouw/build/970/job/ut2rvpvic96bftfl/artifacts

This build starts on XP but unfortunately it introduces new unknow dlls (with a very complicated dlls dependencies). A lot of these are missing (listed in upper case). This dependency is:

**libfluidsynth-2.dll**  depends of:  

- glib-2.dll , and missing dll
`VCRUNTIME140.DLL`, `API-MS-WIN-CRT-HEAP-L1-1-0.DLL`, `API-MS-WIN-CRT-STRING-L1-1-0.DLL`, `API-MS-WIN-CRT-MATH-L1-1-0.DLL`, `API-MS-WIN-CRT-CONVERT-L1-1-0.DLL`, `API-MS-WIN-CRT-STDIO-L1-1-0.DLL`, `API-MS-WIN-CRT-UTILITY-L1-1-0.DLL`, `API-MS-WIN-CRT-RUNTIME-L1-1-0.DLL`.

 **glib-2.dll**  depends of:
- pcre.dll, libiconv.dll, libint.dll and missing dll:
`API-MS-WIN-CRT-TIME-L1-1-0.DLL`,  `API-MS-WIN-CRT-LOCALE-L1-1-0.DLL`, `API-MS-WIN-CRT-FILESYSTEM-L1-1-0.DLL`,  `API-MS-WIN-CRT-ENVIRONMENT-L1-1-0.DLL`,  , `API-MS-WIN-CRT-PROCESS-L1-1-0.DLL`, `API-MS-WIN-CRT-MATH-L1-1-0.DLL`.
- and some missing dll already listed above

**pcre.dll** depends of: some missing dll already listed above  

**libint.dll** depends of:  
- libcharset.dll and some missing dll already listed above.

****
On my own build for XP the classic minimum know dependency is:

**libfluidsynth-2.dll**  depends of:  

- glib dll (manually installed): LIBGTHREAD-2.0-0.DLL, LIBGLIB-2.0-0.DLL, INTL.DLL, 
- windows dll (already part of OS):DSOUND.DLL (audio API) ,WINMM.DLL (MIDI driver API), WS2_32.DLL (API), KERNEL32.DLL(API), USER32.DLL(API), MSVCR100.DLL (runtime library).

Note that i am not looking for AppVeyor binaries for my own. However, this could be useful for others users (not familiar with compilation) for a quick try. 
I don't know if others people (may be Marcus @mawe42 or Carlo @carlo-bramini ) have already ran AppVeyor  windows binaries on Win 7, 8, 10 ?

derselbst:
Strange. Please try this one as well: https://ci.appveyor.com/project/derselbst/fluidsynth/build/job/bdbhogdn4v1qwqda/artifacts

jjceresa:
> Strange. Please try this one as well: https://ci.appveyor.com/project/derselbst/fluidsynth/build/job/bdbhogdn4v1qwqda/artifacts

For this build, libfluidsynth-2.dll now depends of the same dll that for my build  but unfortunately still dependant of missing dll:
`VCRUNTIME140.DLL`, `API-MS-WIN-CRT-HEAP-L1-1-0.DLL`, `API-MS-WIN-CRT-STRING-L1-1-0.DLL`, `API-MS-WIN-CRT-MATH-L1-1-0.DLL`, `API-MS-WIN-CRT-CONVERT-L1-1-0.DLL`, `API-MS-WIN-CRT-STDIO-L1-1-0.DLL`, `API-MS-WIN-CRT-UTILITY-L1-1-0.DLL`, `API-MS-WIN-CRT-RUNTIME-L1-1-0.DLL`.

Please, see 
dll_dependency

jjceresa:
> Ok, seems like VS2015 pollutes the binaries with unnecessary dependencies even in xp compatibility mode

Yes, it seems. Just for infos, below the dependency of my build. We see the presence of MSVCR100.DLL insteed of the missing dlls described in previous comment.
jjc_build_dependency

PR #382, opened 2018-05-13
fix double free when deleting fluid_seqbind_t

derselbst:
This fixes a double free when manually deleting the internally created `fluid_seqbind_t` using an UNREGISTERING event. It actually gets down to a design question whether this unregister-functionality should really be exposed to the user.

### Intro
Suppose the user registers a sequencer client using `fluid_sequencer_register_fluidsynth()`. A `fluid_seqbind_t` is created internally, holding the synth instance, the sequencer and the sample timer. When deleting the sequencer, this client is automatically and properly unregistered by calling `fluid_sequencer_unregister_client()` for every client.

**If** the user however wanted to manually unregister the fluidsynth client, he must send an unregistering event to the sequencer (please find the  [unit test](https://github.com/FluidSynth/fluidsynth/blob/unregister/test/test_seqbind_unregister.c) to see how it works). This caused the internally allocated `fluid_seqbind_t` to be freed, but **not** the sequencer client to be unregistered (because of the prior manipulation of `seqbind->client_id = -1`. If the user then deleted the sequencer, the client would be still there, attempting to free the `fluid_seqbind_t` again. Crash.

### Proposal

With this change every user must cleanup the client registered with `fluid_sequencer_register_fluidsynth()` **manually** to avoid leaking memory. This is inconsistent to the usage of `fluid_sequencer_register_client()`, which does not need to be freed manually. I'm aware of that.

Leading to the design question: Do we really want to expose the `fluid_event_unregistering()` to the user? Or should we silently clean it up internally?

Pro for allowing manual cleanup: User has control to unregister from the sample timer at any time. There might be scenarios where that matters.

Pro for silent internal cleanup: Consistency with automatic unregistering behaviour of `fluid_sequencer_register_client()`. KISS.

@mawe42 @jjceresa  Opinions?

### DIFF
--- a/doc/fluidsynth-v11-devdoc.txt
+++ b/doc/fluidsynth-v11-devdoc.txt
@@ -103,6 +103,7 @@ Changes in FluidSynth 2.0.0 concerning developers:
 - all public \c fluid_settings_* functions that return an integer which is not meant to be interpreted as bool consistently return either FLUID_OK or FLUID_FAILED
 - all public delete_* functions return void and are safe when called with NULL
 - all public functions consistently receive signed integers for soundfont ids, bank and program numbers
+- explicit client unregistering is required for fluid_sequencer_register_client() and fluid_sequencer_register_fluidsynth()
 - the shell command handler was decoupled internally, as a consequence the param list of new_fluid_server() and new_fluid_cmd_handler() was adapted
 - reverb: roomsize is now limited to an upper threshold of 1.0 to avoid exponential volume increase
 - use unique device names for the "audio.portaudio.device" setting
--- a/src/midi/fluid_seq.c
+++ b/src/midi/fluid_seq.c
@@ -302,36 +302,23 @@ void
 fluid_sequencer_unregister_client (fluid_sequencer_t* seq, fluid_seq_id_t id)
 {
 	fluid_list_t *tmp;
-	fluid_event_t* evt;
 
 	if (seq->clients == NULL) return;
-
-	evt = new_fluid_event();
-	if (evt != NULL) {
-		fluid_event_unregistering(evt);
-		fluid_event_set_dest(evt, id);
-	}
 
 	tmp = seq->clients;
 	while (tmp) {
   		fluid_sequencer_client_t *client = (fluid_sequencer_client_t*)tmp->data;
 
   		if (client->id == id) {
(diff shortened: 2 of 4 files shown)


mawe42:
I've never used this part of the API myself, so I can't really comment on the actual use case.

But as far as the API design goes, I would always vote for the (Python) mantra "explicit over implicit". When users need to explicitly register something, then they should also explicitly unregister it again when they're done.

And I think hiding the ```fluid_event_unregistering``` from the user makes a lot of sense. It's an internal implementation detail. The user should not have to deal with that.

So I guess I'm saying:
- Register the sequencer with the existing ```fluid_sequencer_register_fluidsynth```, unregister with a new ```fluid_sequencer_unregister_fluidsynth``` which hides the ```fluid_event_unregistering``` dance from the user
- Register the client with the existing ```fluid_sequencer_register_client``` and require explicit unregister with ```fluid_sequencer_unregister_client```

derselbst:
Thanks for your input so far. Marcus is basically right: if there is a `fluid_sequencer_register_fluidsynth()` there should also be a `fluid_sequencer_unregister_fluidsynth()`. And JJC is right that an app must be able to schedule an unregister at any asynchonous date. For the latter I can hardly think of a use-case though (a singleton sequencer and multiple synths??).

Anyway, I think we should keep `fluid_event_unregistering()` then. It is however surprising that we never got a bug report that it doesnt work.

Will think about how to fix this bug then:

* Either via my proposal to explicitly require the user to use this unregistering event (which I currently prefer)
* or via JJC's proposal (which is more user friendly).

derselbst:
The sequencer itself is  actually pretty useful if you need more flexibility than the player offers. E.g. I need callbacks to custom processing functions on certain MIDI CCs that fill the sequencer with new events.

But thinking about a specific use-case for asynchronous client unregistering doesnt come to my mind. But I neither want to remove or hide this functionality to the user. The fact that apparently no one uses `fluid_event_unregistering()` with the synth as client, doesnt mean that there are custom clients with custom processing functions listening for unregistering events to do something with.

So I think the best way to go is
* to require explicit unregisters for fluid_sequencer_register_fluidsynth() (due to internal memory alloc)
* and to recommend explicit unregisters for `fluid_sequencer_register_client()`.

Unless there are objections, I will update the API docs and merge this in the next days.

PR #385, opened 2018-05-15
Add reverb and chorus settings

### LINKED ISSUE #49: Add FluidSynth settings for reverb and chorus
Adding [FluidSynth](/apps/trac/fluidsynth/wiki/FluidSynth) settings parameters for reverb and chorus would allow for the valid ranges to be queried and also clean up the command interface (chorus and reverb commands could be deprecated).  

Example setting names:  

  * synth.reverb.room_size 
  * synth.reverb.damp 
  * synth.reverb.width 
  * synth.reverb.level 
  * synth.chorus.nr 
  * synth.chorus.level 
  * synth.chorus.speed 
  * synth.chorus.depth 

There is already synth.reverb.active and synth.chorus.active.  

Reported by: elementgreen

Original Ticket: [fluidsynth/tickets/49](https://sourceforge.net/p/fluidsynth/tickets/49)

derselbst:
This introduces reverb and chorus realtime settings as suggested by #49. It also deprecates reverb and chorus shell commands and removes the `FLUID_*_DEFAULT_*`  macros from the public API, as they can now be retrieved via the settings object.

Closes #49.


test/test_synth_chorus_reverb.c
```diff
@@ -0,0 +1,39 @@
...
+
+// this test should make sure that sample rate changed are handled correctly
+int main(void)
+{
+    fluid_synth_t *synth;
+    fluid_settings_t *settings = new_fluid_settings();
+    TEST_ASSERT(settings != NULL);
+    
+    TEST_SUCCESS(fluid_settings_setnum(settings, "synth.reverb.roomsize", 1.0));
+    TEST_SUCCESS(fluid_settings_setnum(settings, "synth.reverb.damp", 1.0));
```
mawe42: If you would use different values for the parameters, the tests could actually check that the correct value is returned/set for a named parameter.

test/test_synth_chorus_reverb.c
```diff
@@ -0,0 +1,39 @@
...
+    TEST_ASSERT(synth != NULL);    
+    
+    TEST_ASSERT(fluid_synth_get_reverb_roomsize(synth) == 1.0);
+    TEST_ASSERT(fluid_synth_get_reverb_damp(synth) == 1.0);
+    TEST_ASSERT(fluid_synth_get_reverb_width(synth) == 1.0);
+    TEST_ASSERT(fluid_synth_get_reverb_level(synth) == 1.0);
+    
+    TEST_ASSERT(fluid_synth_get_chorus_nr(synth) == 99);
+    TEST_ASSERT(fluid_synth_get_chorus_level(synth) == 1.0);
+    TEST_ASSERT(fluid_synth_get_chorus_speed_Hz(synth) == 1.0);
```
mawe42: I just noticed that there is this capital "H" in ```fluid_synth_get_chorus_speed_Hz```. Maybe we should clean that up as well before the 2.0 release?

src/synth/fluid_synth.c
```diff
@@ -3367,6 +3428,53 @@ fluid_synth_render_blocks(fluid_synth_t* synth, int blockcount)
   return blockcount;
 }
 
+/*
+ * Handler for synth.reverb.* settings.
```
mawe42: Maybe also mention synth.chorus.* settings and explain the _num (floating point) and _int (int) distinction?

test/test_synth_chorus_reverb.c
derselbst: > the tests could actually check that the correct value is returned/set for a named parameter.

This is done below by asking the synth to return these values (which he must have read from the settings object to be correct). For the correct saving of the settings, relying on the return value of the settings API is sufficient here.

test/test_synth_chorus_reverb.c
derselbst: Not sure. This function is already in use for 1.1.x for long. I dont like it as well, if you'd asked me I would simply call it `*_speed()`. However one might consider this as an unnecessary API change, motivated by personal taste.

src/synth/fluid_synth.c
derselbst: ok

test/test_synth_chorus_reverb.c
mawe42: Sorry, I mean by settings damp to 1.0, roomsize to 0.9, etc we could test that ```fluid_synth_get_reverb_damp``` does actually return the value set by synth.reverb.damp, and not the one set via synth.reverb.roomsize.

test/test_synth_chorus_reverb.c
derselbst: Alright, consistency is a good argument. Shall we call it `fluid_synth_get_chorus_speed()` then?

mawe42:
> synth.reverb.roomsize or synth.reverb.room-size?

room-size would be more consistent with the other settings, I think. We already have things like media-role, period-size, sample-format, cpu-cores, min-note-length.

derselbst:
Ok, should be good now.

derselbst:
Dont think that's necessary, it's mentioned in the overall 2.0 changes.

PR #393, opened 2018-06-02
Avoid bypassing visibility control

### LINKED ISSUE #392: Inconsistent symbol export
Currently, it depends on dbus being available at build time if the `fluid_rtkit_make_realtime()` function is exported as part of the public API or not. This means that applications could never simply call this function without risking an "invalid symbol" kind of crash.

The fix is merely a matter of moving some defines around in `fluid_rtkit.{c,h}`. If dbus is unavailable, the function could be defined to the same dummy that it currently is if dbus is available but the build system is neither Linux nor Apple. 

Please consider making `fluid_rtkit_make_realtime()` part of the stable fluidsynth 2.0 API.

derselbst:
Visibility control can be easily bypassed by using a cmake build type different to `Release | RelWithDebInfo | MinSizeRel`. With this change `CMAKE_C_FLAGS` is set by us to use our mandatory build flags. Users are free to append additional flags.

`CMAKE_C_FLAGS_RELEASE` and `CMAKE_C_FLAGS_RELWITHDEBINFO` are not used anymore by us.

Fixes #392.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,15 @@ mark_as_advanced ( LIB_SUFFIX )
 # the default C standard to use for all targets
 set(CMAKE_C_STANDARD 90)
 
+# the default global visibility level for all target
+# no visibility support on OS2
+if ( NOT OS2 )
+    set ( CMAKE_C_VISIBILITY_PRESET hidden )
+endif ( NOT OS2 )
+# enforce visibility control for all types of cmake targets
+cmake_policy ( SET CMP0063 NEW )
+
+
 # Default install directory names
 include ( DefaultDirs )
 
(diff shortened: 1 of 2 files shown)


mawe42:
I've tested this and compared the used flags to master and 1.1.x... looks good! So we can use CFLAGS to add additional flags in the beginning, CMAKE_C_FLAGS_{BUILD TYPE NAME} to replace the build specific ones at the end of the argument list. Nice.

One thing that was a surprise to me was that current master uses ```-O3``` for Release builds, but 1.1.x uses ```-02 -finline-functions```. Just mentioning this here to make sure it was intentional...

derselbst:
Ok, thanks for testing.

> One thing that was a surprise to me was that current master uses -O3 for Release builds,

You mean with this change the default is -O3? Yes, an intended coincidence if you want, because -O3 implies -finline-functions [(and a few more)](https://gcc.gnu.org/onlinedocs/gcc-4.5.2/gcc/Optimize-Options.html), so this should yield (nearly) the same result as before.

mawe42:
> You mean with this change the default is -O3?

Yes, but also on master.

mawe42:
To clarify: master sets ```-02 -finline-functions .... -03```. 1.1.x doesn't set the ```-O3``` at the end. But never mind, as it's intended and probably a good idea anyway.

derselbst:
Oh yes ofc, that's because you asked me to allow the user to override `CMAKE_C_FLAGS_RELEASE` some time ago. So while master appends to any existing (default) `CMAKE_C_FLAGS_RELEASE`, 1.1.x overwrites it.

PR #395, opened 2018-06-03
Fluid sample count branch

carlo-bramini:
Basically, it just improves the readability and the maintenance of the code.
Although those multiplications are resolved at compile time, so there are not performance improvements, in my opionion the code looks a bit easier to understand after the change.
Sincerely.


derselbst:
I wanted to avoid hiding the buffer stride behind a single macro. Previously it was obvious that the buffer contains MAX_BUFFERS, of which each contains FLUID_BUFSIZE samples.

If you want it that way, I'd prefer:

```c
#define FLUID_BUFFER_STRIDE (FLUID_BUFSIZE * FLUID_MIXER_MAX_BUFFERS_DEFAULT)
```

PR #396, opened 2018-06-09
implement true zero gain for maximum attenuation

### LINKED ISSUE #319: True zero gain for maximum attenuation
During the discussion on #317 about which range to use for the new balance control, we also discussed the fact that FluidSynth currently doesn't make channels completely silent when attenuation is at maximum. Looking through the Sound Font 2.04 specs, we found the following paragraph:

>  The volume envelope operates in dB, with the attack peak providing a full scale output, appropriately scaled by the initial volume. The zero value, however, is actually zero gain. The implementation in the EMU8000 provides for 96 dB of amplitude control. When 96 dB of attenuation is reached in the final gain amplifier, an abrupt jump to zero gain (infinite dB of attenuation) occurs. In a 16-bit system, this jump is inaudible.

So, to be true to the EMU8k hardware, we should probably do something like the following in our "final gain amplifier" (which I think is effectively fluid_rvoice_buffers_mix):
```diff
 fluid_rvoice_buffers_mix(fluid_rvoice_buffers_t* buffers, 
                          fluid_real_t* dsp_buf, int samplecount, 
                          fluid_real_t** dest_bufs, int dest_bufcount)
 {
+  const fluid_real_t MIN_GAIN = pow(10.0, (double) 960.0 / -200.0);
   int bufcount = buffers->count;
   int i, dsp_i;
   if (!samplecount || !bufcount || !dest_bufcount) 
     return;
 
   for (i=0; i < bufcount; i++) {
     fluid_real_t* buf = get_dest_buf(buffers, i, dest_bufs, dest_bufcount);
     fluid_real_t* next_buf;
     fluid_real_t amp = buffers->bufs[i].amp;
-    if (buf == NULL || amp == 0.0f)
+    if (buf == NULL || amp <= MIN_GAIN)
       continue;
 
     /* Optimization for centered stereo samples - we can save one 
        multiplication per sample */
     next_buf = (i+1 >= bufcount ? NULL : get_dest_buf(buffers, i+1, dest_bufs, dest_bufcount));
     if (next_buf && buffers->bufs[i+1].amp == amp) {
       for (dsp_i = 0; dsp_i < samplecount; dsp_i++) {
         fluid_real_t samp = amp * dsp_buf[dsp_i]; 
         buf[dsp_i] += samp;
         next_buf[dsp_i] += samp;
       }
       i++;
     }
     else {
       for (dsp_i = 0; dsp_i < samplecount; dsp_i++)
         buf[dsp_i] += amp * dsp_buf[dsp_i];
     }
   }
 }
```

But like the paragraph explains, that is only an inaudible jump on 16-bit systems. We have support for 24-bit samples and also provide >= 24-bit outputs. So I'm a little unsure on how to resolve this. I would really like to be able to switch off channels completely. It will benefit performance, but also make channel separation more exact. But the hardcoded 96 dB threshold seems to be wrong for > 16 bit.

Should fluid_rvoice_buffers_mix maybe use a different threshold for 16 bits (96 dB) and >= 24 bits (144 dB) output? It would be really great if somebody with access to emu8k and emu10k hardware could experiment with this. Is the 10k1 or 10k2 even capable of rendering 24-bit samples? And where does it make the jump to infinite attenuation? 

@mrbumpy409 Sorry to bother you again, but you seem to have a great wealth of knowledge about the EMU hardware. Could you maybe give us your opinion on this subject?

derselbst:
This implements true zero gain according to SF2.04 9.1.7.

Fixes #319.


src/rvoice/fluid_rvoice_mixer.c
```diff
@@ -310,6 +310,21 @@ fluid_rvoice_buffers_mix(fluid_rvoice_buffers_t* buffers,
...
+   *  an abrupt jump to zero gain (infinite dB of attenuation) occurs."
+   * 
+   * fluidsynth is a 24 bit synth, it could (should??) use 144 dB of attenuation.
+   * However the spec makes no distinction between 16 or 24 bit synths, so use
+   * 96 dB here.
+   * 
+   * This magic number was obtained by
+   * powl(10.0L, -96.0L / 20.0L);
+   */
+  static const fluid_real_t MIN_GAIN = 0.000015848931924611135L;
```
jjceresa: **Just a remark related to this PR about the the hard coded 96 db of attenuation.**

Actually **fluid_rvoice_calc_amp() , fluid_rvoice_noteoff_LOCAL** and **fluid_rvoice_multi_retrigger_attack()** make use **also** of hard coded **96 db** value:

https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/rvoice/fluid_rvoice.c#L54

https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/rvoice/fluid_rvoice.c#L510

https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/rvoice/fluid_rvoice.c#L572


These hard coded `960.0` should be replaced by a constant (e.g PEAK_ATTENUATION_RANGE = 960.0). 
`static const fluid_real_t PEAK_ATTENUATION_RANGE = 960.0;`

Also we notice that `MIN_GAIN` defined is this PR should be related to `PEAK_ATTENUATION_RANGE` as this:
`MIN_GAIN = powl(10.0L, -PEAK_ATTENUATION_RANGE / 200.0L)`

Probably, the above remark is applicable to the amount of default modulator here in fluid_synth.c  ?:
https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/synth/fluid_synth.c#L308
https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/synth/fluid_synth.c#L320
https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/synth/fluid_synth.c#L384
https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/synth/fluid_synth.c#L412
https://github.com/FluidSynth/fluidsynth/blob/b844dd1f04870f288b7a7d8499ff4a1e08d9539f/src/synth/fluid_synth.c#L469

derselbst: Yes I aggree we could introduce this `PEAK_ATTENUATION_RANGE` constant. However this should be part of another PR. Feel free to do so.

I also wanted to write `MIN_GAIN = pow(10.0, -960.0 / 200.0)`, however as `MIN_GAIN` being `static const` it requires a constant expression to init at compile time. Thus I just hardcoded the value because I didnt want to init some static var in fluid_conv.c and pull them over at runtime.

mawe42:
And it looks to me like this is (at least conceptually) related:
https://github.com/FluidSynth/fluidsynth/blob/05c2d57e65f3fdc264c53fcaa4947cda2eb508a8/src/rvoice/fluid_rvoice.h#L42

Maybe we should try to combine those two defines?

derselbst:
Ok, thanks. For now I think the best place for MIN_GAIN is in rvoice_mixer. There we can be sure that it's the "final" gain. Combining it with `FLUID_NOISE_FLOOR` could be done, however it's currently not on my schedule. Hope I can test this soon and will  merge it then.

PR #397, opened 2018-06-09
introducing ATTENUATION_RANGE_CB macro.

jjceresa:
This macro allows to maintain consistent dependency between different parts of code using this value.

### DIFF
--- a/src/utils/fluid_conv.c
+++ b/src/utils/fluid_conv.c
@@ -67,7 +67,7 @@ fluid_conversion_config(void)
      implemented according to the pictures on SF2.01 page 73. */
 
   for (i = 1; i < 127; i++) {
-    x = -20.0 / 96.0 * log((i * i) / (127.0 * 127.0)) / M_LN10;
+    x = (-200.0 / ATTENUATION_RANGE_CB) * log((i * i) / (127.0 * 127.0)) / M_LN10;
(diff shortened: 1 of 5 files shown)


derselbst:
There was one more appropriate use-case in fluid_conv.c, so I moved the macros' declaration to fluid_conv.h.

I dont quite like the naming yet. I would prefer `FLUID_PEAK_ATTENUATION`. Your opinion?

src/utils/fluid_conv.c
```diff
@@ -67,7 +67,7 @@ fluid_conversion_config(void)
      implemented according to the pictures on SF2.01 page 73. */
 
   for (i = 1; i < 127; i++) {
-    x = -20.0 / 96.0 * log((i * i) / (127.0 * 127.0)) / M_LN10;
+    x = (-200.0 / ATTENUATION_RANGE_CB) * log((i * i) / (127.0 * 127.0)) / M_LN10;
```
jjceresa: It seems i was introduced an enormous bug here. Thanks for reviewing this!

jjceresa:
>  dont quite like the naming yet. I would prefer FLUID_PEAK_ATTENUATION. Your opinion?

Yes, FLUID_PEAK_ATTENUATION is a better name as it refers to the dynamic of volume adsr envelope.

derselbst:
Ok thanks. Will leave them open a few days to give Marcus a chance.

PR #398, opened 2018-06-16
Revert "implement true zero gain for maximum attenuation"

derselbst:
Reading the spec again, the final gain amp should actually applied for the whole voice, not just single buffers (cf. sect. 9.1.8 Figure 2). Thus `rvoice_mixer` is not the correct place to implement this. It should instead be done in `fluid_rvoice_calc_amp()`, where we already have the noise floor calculation. So this MIN_GAIN approach might turn out to be redundant.

### DIFF
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -310,21 +310,6 @@ fluid_rvoice_buffers_mix(fluid_rvoice_buffers_t* buffers,
                          int start_block, int sample_count, 
                          fluid_real_t** dest_bufs, int dest_bufcount)
 {
-  /*
-   * SF2.04 9.1.7 Envelope Generators:
-   * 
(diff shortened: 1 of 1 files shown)


mawe42:
I don't think it's redundant, but it definitely seems to need a more thorough approach. The problem is that controls like balance and panning directly affect the mixer buffers amp value, they are not taken into consideration in ```fluid_rvoice_calc_amp```.

derselbst:
> controls like balance and panning directly affect the mixer buffers amp value, they are not taken into consideration in fluid_rvoice_calc_amp.

Looking at Figure 2, I believe that this behaviour is intended by the spec. The "Amplifier" in this picture is what I understand as final gain amp, which is handled in `fluid_rvoice_calc_amp()`. The buffers amplitudes are handled separately afterwards and from my understanding they are not taken into account as final gain amp.

PR #402, opened 2018-06-21
Reformatting

### LINKED ISSUE #179: Mixed whitespace

derselbst:
This proposes the reformatting based on our [styleguide](https://github.com/FluidSynth/fluidsynth/blob/master/CONTRIBUTING.md#style-guide).

Turns out that blindly trusting `clang-tidy` to apply readability braces may break build... so for this I used `astyle` to get the bracing right and `clang-format` for the withe spaces.

One drawback remains: `clang-format` does not consequently respect the column limit of 120 chars. Some lines esp. function delarations are longer. But I can live with that.

@jjceresa @mawe42 Comments?

Closes #179.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -7,6 +7,7 @@ AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: false
+BinPackArguments: false
 BinPackParameters: false
 BreakBeforeBinaryOperators: false
 BreakBeforeTernaryOperators: false
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,6 @@
 AccessModifierOffset: 0
 AlignEscapedNewlinesLeft: true
-AlignTrailingComments: false
+AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortFunctionsOnASingleLine: false
 AllowShortIfStatementsOnASingleLine: false
@@ -25,12 +25,12 @@ BraceWrapping:
   BeforeElse: true
   IndentBraces: false
 
-ColumnLimit: 100
+ColumnLimit: 120
 CommentPragmas: ''
 ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ConstructorInitializerIndentWidth: 0
 ContinuationIndentWidth: 0
-Cpp11BracedListStyle: true
+Cpp11BracedListStyle: false
 DerivePointerAlignment: false
 IndentCaseLabels: true
 IndentWidth: 4
@@ -58,3 +58,4 @@ SortIncludes: false
 Standard: Cpp11
 TabWidth: 4
 UseTab: Never
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,9 +334,9 @@ find_program( CLANG_TIDY
     message ( STATUS "Found clang-tidy" )
         # whenever clang-tidy is available, use it to automatically add braces after ever "make"
         if ( WITH_PROFILING )
-#             set ( CMAKE_C_CLANG_TIDY "clang-tidy" )
+            set ( CMAKE_C_CLANG_TIDY "clang-tidy" )
         else ( WITH_PROFILING )
-#             set ( CMAKE_C_CLANG_TIDY "clang-tidy;-checks=-*,readability-braces-around-statements;-format-style=file" )
+            set ( CMAKE_C_CLANG_TIDY "clang-tidy;-checks=-*,readability-braces-*;-format-style=file" )
         endif ( WITH_PROFILING )
     endif ( CLANG_TIDY )
 endif ( CMAKE_VERSION VERSION_GREATER "3.6.0" )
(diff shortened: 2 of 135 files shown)


mawe42:
Great, thanks for doing this. I'm really looking forward to a more readable codebase!

I have one major concern, though:
Including the format commands as make targets looks to me like you would want to call them on a regular basis. If that is the case, then I would like to make a case against that. I strongly think that automatic reformatting of the whole codebase should be a one-off operation. Something we do once now and (hopefully) never again. After the global reformat, we watch new contributions for style and request changes or make them ourselves. So those long lines left over by the reformat are not a problem and can be fixed manually.

Doing reformatting on a regular basis leads to lots of problems. It means we can't format the code for best readability, but need format it according to what clang-format supports. And it heavily depends on the clang-format version. I've checked out the format5 branch and ran ```make format```. The result were lots of white-space changes (using ```clang-format version 4.0.1-6```). And future versions of clang-format might format the code in yet another way.

If I got the wrong impression and you also think this should only be a one off, then please ignore this comment :-)

But in that case I would argue to remove the ```format``` make target.

derselbst:
I was not intending to ever do (i.e. commit) a global reformat again. I intended those make targets to execute on code made by contributors. E.g. before merging (eventually squashing) the new reverb #380, I thought to `make format` the new file to have a clean and consistent start. For all minor code changes (either from contributors or by ourselves) one could take care of formatting manually.

And yes, clang-format seems to be a bit volatile currently. I've used 5.0.1 here.

jjceresa:
Thanks for doing this. This should encourage and educate anyone to follow our simple styleguide.
Good idea to apply this on new file only. I think that it is a good think that anyone be able to appropriate our styleguide manually also. We will see.

mawe42:
> I was not intending to ever do (i.e. commit) a global reformat again.

Ok, very glad that I got the wrong impression! :-)

I had a look through the reformatted code and the only thing that really sticks out is the line length. Is there really no way that clang-format can properly format the function declarations?

And I must admit that I'm not very fond of the 120 char line length. On my devel setup I use two code windows side by side on a regular basis, and 100 chars seems to be a better upper limit for that. But this is all personal taste, I know...

jjceresa:
> And I must admit that I'm not very fond of the 120 char line length. On my devel setup I use two code windows side by side on a regular basis, and 100 chars seems to be a better upper limit for that. But this is all personal taste, I know...

As an other example, i usually use only 80 char maximum length which is convenient to fit A4 paper width using font size between 11 and 12 points. Also, reading font size below 11 is very difficult for me due to eyes limitation.
****
Please, just a question: If one wrote a function declaration on more than one line (saying 4 lines ,each line being 80 char maximum length), how this declaration look once reformatted by clang-format ?. Is it transformed in a less number of lines (each line being 120 char length maximum) ?

derselbst:
I dont know why clang-format doesnt respect ColumnLimit in this case. I
will further investigate tomorrow at latest.

If you ask me, I dont need that ColumnLimit at all. I'm using a 32" screen,
I can easily fit 200 char in one line. And for the rest I'm using dynamic
line breaks. I've chosen 120 because that's the maximum possible when
viewing a file in the github source browser. I'm open to limit it to 100,
but 80 really is a bit too narrow IMO.

@jjc I'd expect clang-format to split a very long declaration (that doesnt
fit into a single 120 char line) that way that all arguments are listed in
one line (not necessarily aligned):

void someveryveryveryveryveryverylongname(int arg1,
    int arg2,
    int arg3,
...)

jjceresa:
>I'm open to limit it to 100, but 80 really is a bit too narrow IMO.

No problem with line 120 or 100 car length. I'm fine with this too .

derselbst:
Just updated to clang 6.0 and it looks a little different (better?), see 922f68eed5033a42809cd4296cf23bab395b9a22.

derselbst:
I would squash d1983aefc9945dc12666b1e7adf767741d19f4d6 922f68eed5033a42809cd4296cf23bab395b9a22 ece1d21c13b28299891cabad31dfa3f0a152b0cb and then merge this PR, may I?

mawe42:
> Just updated to clang 6.0 and it looks a little different (better?), see 922f68e.

Not sure... it leads to even more very long lines. And it joins some parameters on a single line that were previously split correctly into multiple lines.

mawe42:
> I would squash d1983ae 922f68e ece1d21 and then merge this PR, may I?

Well, if you both don't care about 120 or 100 chars line length, then I would state again that I would much prefer the 100.

The automatic reformatting made most of the code easier to read, but some parts definitely got harder to read. How should we deal with this? We could either fiddle around with clang-format more to find the optimum settings or fix the most glaring problems by hand. If we fix a lot by hand, I think those changes should be squashed into this PR and merged as a single commit. That reduces the noise when doing bisects or blames on the codebase.

It seems to me that lines which end in a (long) comment are especially problematic...

PR #403, opened 2018-06-24
clang-format ColLimit 100

derselbst:
Reformatting using clang-format 6 with column limit of 100.

### DIFF
--- a/src/synth/fluid_gen.c
+++ b/src/synth/fluid_gen.c
@@ -25,70 +25,70 @@
 
 /* See SFSpec21 $8.1.3 */
 static const fluid_gen_info_t fluid_gen_info[] = {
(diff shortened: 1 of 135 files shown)


src/synth/fluid_gen.c
```diff
@@ -25,70 +25,70 @@
 
 /* See SFSpec21 $8.1.3 */
 static const fluid_gen_info_t fluid_gen_info[] = {
-        /* number/name             init  scale         min        max         def */
```
mawe42: Is there a way to protect these this nicely formatted code from clang-format?

derselbst: Only via explicit `/* clang-format off */`, or if we only used astyle, see #404.

PR #404, opened 2018-06-24
reformat using astyle

### LINKED ISSUE #179: Mixed whitespace

derselbst:
Proposes to reformat the code solely using `astyle` (f373e87) and additionally forcing a column limit of 100 (5a5d794).

Resolves #179.
Closes #402.
Closes #403.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,12 +1,13 @@
 AccessModifierOffset: 0
 AlignEscapedNewlinesLeft: true
-AlignTrailingComments: false
+AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortFunctionsOnASingleLine: false
 AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
 AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: false
+BinPackArguments: false
 BinPackParameters: false
 BreakBeforeBinaryOperators: false
 BreakBeforeTernaryOperators: false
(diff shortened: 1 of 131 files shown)


mawe42:
That looks *much* better in my opinion! Not as complete as the clang-format, but not as intrusive either. Some inconsistencies like ```if (args)``` vs. ```if(args)``` are not cleaned up, but those are not really bad for readability.

So, thumbs-up for this version!

derselbst:
> That looks much better in my opinion!

With or without 100 char limit?

derselbst:
> inconsistencies like if (args) vs. if(args) are not cleaned up,

Can be fixed with the `-H` option.

mawe42:
The forced 100 char column break does seem to mess up some things again, though. Maybe leave that step out and do line length formatting manually if code needs to be touched anyway?

derselbst:
Ok, forced pushed, now without column limit and consistent `if(arg)`. First used astyle to format, then used clang-tidy to add missing braces that astyle didnt find, fixed the compilation bugs introduced by clang-tidy and ran astyle again... duh. So I kept the .clang-format config for now, as it may be used by clang-tidy to correctly apply missing braces.

jjceresa:
I have not fully understood. Is this processing pipeline breaks line above a given length limit ?. Which is this limit ?.

derselbst:
Right now there is no column limit, i.e. no forced line breaks here.

mawe42:
This looks very good, in my opinion!

I would prefer a space after ```if (arg)```, ```while (arg)``` etc to separate them from "normal" function invokations, but that is again very personal.

PR #405, opened 2018-06-24
Complete fluid_synth_process()

### LINKED ISSUE #384: complete fluid_synth_process()
Implement `fluid_synth_process()` according to https://lists.nongnu.org/archive/html/fluid-dev/2018-04/msg00075.html before final 2.0 release.

derselbst:
This proposes the implementation of `fluid_synth_process()` according to related discussion on the mailing list. I'm not quite happy with the handling of the effects channels, when it comes to multiple reverb or chorus channels: The goal is to allow using `fluid_synth_process()` to render effects of each midi channel to its own stereo channel, just as it's possible for dry audio. ~I'm thinking to just combine reverb and chorus of one midi channel and let them render to the same stereo channel... but that's another story.~

This also deprecates `fluid_synth_nwrite_float()`, but I'll keep it for now.

Fixes #384.

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3473,40 +3473,64 @@ fluid_synth_nwrite_float(fluid_synth_t *synth, int len,
 }
 
 /**
- * @brief Synthesize floating point audio to planar audio buffers.
+ * @brief Synthesize floating point audio to stereo audio channels (implements the default interface #fluid_audio_func_t).
  *
- * Synthesize and <strong>mix</strong> audio to a given number of stereo audio channels.
- * Therefore pass <code>nout = i*2</code> float buffers to \p out in order to render
- * the synthesized audio to \p i stereo channels. Each float buffer must be
+ * Synthesize and <strong>mix</strong> audio to a given number of planar audio buffers.
+ * Therefore pass <code>nout = N*2</code> float buffers to \p out in order to render
+ * the synthesized audio to \p N stereo channels. Each float buffer must be
  * able to hold \p len elements.
  *
  * \p out contains an array of planar buffers for normal, dry, stereo
  * audio (alternating left and right). Like:
 @code{.cpp}
-out[0]  = left_buffer_channel_1
-out[1]  = right_buffer_channel_1
-out[2]  = left_buffer_channel_2
-out[3]  = right_buffer_channel_2
+out[0]  = left_buffer_audio_channel_0
+out[1]  = right_buffer_audio_channel_0
+out[2]  = left_buffer_audio_channel_1
+out[3]  = right_buffer_audio_channel_1
 ...
-out[ (i-1) * 2 + 0 ]  = left_buffer_channel_i
-out[ (i-1) * 2 + 1 ]  = right_buffer_channel_i
+out[ (i * 2 + 0) % nout ]  = left_buffer_audio_channel_i
+out[ (i * 2 + 1) % nout ]  = right_buffer_audio_channel_i
 @endcode
  *
- * for one-based channel index \p i.
- * Same buffer layout is used for \p fx for storing effects
- * like reverb and chorus audio.
- * This function implements the default interface #fluid_audio_func_t.
+ * for zero-based channel index \p i.
+ * The buffer layout of \p fx used for storing effects
+ * like reverb and chorus looks similar:
+@code{.cpp}
+fx[0]  = left_buffer_channel_of_reverb_unit_0
+fx[1]  = right_buffer_channel_of_reverb_unit_0
+fx[2]  = left_buffer_channel_of_chorus_unit_0
+fx[3]  = right_buffer_channel_of_chorus_unit_0
+fx[4]  = left_buffer_channel_of_reverb_unit_1
+fx[5]  = right_buffer_channel_of_reverb_unit_1
+fx[6]  = left_buffer_channel_of_chorus_unit_1
+fx[7]  = right_buffer_channel_of_chorus_unit_1
+fx[8]  = left_buffer_channel_of_reverb_unit_2
+...
+fx[ ((k * fluid_synth_count_effects_channels() + j) * 2 + 0) % nfx ]  = left_buffer_for_effect_channel_j_of_unit_k
+fx[ ((k * fluid_synth_count_effects_channels() + j) * 2 + 1) % nfx ]  = right_buffer_for_effect_channel_j_of_unit_k
+@endcode
+ * where <code>0 <= k < fluid_synth_count_effects_groups()</code> is a zero-based index denoting the effects unit and
+ * <code>0 <= j < fluid_synth_count_effects_channels()</code> is a zero-based index denoting the effect channel within
+ * unit \p k.
+ *
+ * Any voice playing is assigned to audio channels based on the MIDI channel its playing on. Let \p chan be the
+ * zero-based MIDI channel index an arbitrary voice is playing on. To determine the audio channel and effects unit it is
+ * going to be rendered to use:
+ *
+ * <code>i = chan % fluid_synth_count_audio_groups()</code>
+ *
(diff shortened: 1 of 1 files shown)


derselbst:
88fda6f introduces a new setting `synth.effects-groups` that allows using multiple effects units (typically one for each midi channel) so that all the effects can be rendered to separate audio channels as well. I've done that now as prove of concept for `fluid_synth_process()` becoming fluidsynths generic rendering function. By default there only is one effects units (i.e. one reverb and chorus unit), so this stays backward compatible.

`synth.effects-channels` for now stays hardcoded to 2 as being the number of effects available per unit (i.e. reverb and chorus).

Once having merged this, I'd consider 2.0 to be feature complete and release a last beta. Documentation to follow.

@mawe42 @jjceresa General thoughts? Is the name of the new setting appropriate?

jjceresa:
>88fda6f introduces a new setting synth.effects-groups that allows using multiple effects units (typically one for each midi channel) so that all the effects can be rendered to separate audio channels as well.

Assuming:

- 6 MIDI channels used (0 to 5).
- **audio_groups** set to 0.
- **audio_channels** set to 3. 
- **synth.effects-channels** set to 2. 
-  fluid_synth_process() called for 3 audio outputs and 3 effects outputs
The number of audio buffers will be: buf_count = max(audio_group,audio_channels)` egal to 3. And 
the know mapping behavior was:

- MIDI channel 0,3 mapped to audio buffer 0, then to output 0 (by fluid_synth_process())
- MIDI channel 1,4 mapped to audio buffer 1, then to output 1 (by fluid_synth_process())
- MIDI channel 2,5 mapped to audio buffer 2, then to output 2 (by fluid_synth_process())

- Reverb and chorus (for all MIDI voices) mapped (wrapped) to  output 0,1,2
 

Based on the above settings, please could you give via a short examples what happens with different values of `synth.effects-groups` ?. Does **synth.effects-channels** is related to **synth.effects-groups** ?

derselbst:
There are a few bugs in your explanations.

> audio_groups set to 0.

It cannot be set to zero and would stay one instead.

> fluid_synth_process() called for 3 audio outputs and 3 effects outputs

Speaking of its previous behaviour: It would return FLUID_FAILED, because you've provided more effects outputs than `fluid_synth_count_effects_channels()` available (which was and still is 2).
`synth.effects-channels` only tells you how much effects there are available per effects unit (i.e. reverb and chorus only. and in the old implementation we only had one such effects unit).

Apart from that, the channel mappings are correct as you've explained for both: the previous implementation and the one suggested by this PR.

> Reverb and chorus (for all MIDI voices) mapped (wrapped) to output 0,1,2

Assuming that you had called `fluid_synth_process()` correctly with only 2 effects outputs, the reverb would have been dispatched to effects output 0 and chorus to output 1. This was the old implementation (and still default behaviour). You can think of this old implementation as `synth.effects-groups=1` as there only was one reverb and one chorus unit for everything.

The new setting now allows you to use more effects units. E.g.  `synth.effects-groups=3` allows you to pass  `synth.effects-groups * synth.effects-channels == 6` outputs into `fluid_synth_process()` so that you can render reverb and chrous split across 3 effects units to separate outputs. That is:


* Reverb of MIDI channel 0,3 mapped to rvoice audio buffer 3, then to effects output 0 (by fluid_synth_process())

* Chorus of MIDI channel 0,3 mapped to rvoice audio buffer 4, then to effects output 1 (by fluid_synth_process())

* Reverb of MIDI channel 1,4 mapped to rvoice audio buffer 5, then to eff. output 2

* Chrous of MIDI channel 1,4 mapped to rvoice audio buffer 6, then to eff. output 3

* Reverb of MIDI channel 2,5 mapped to rvoice audio buffer 7, then to eff. output 4

* Chrous of MIDI channel 2,5 mapped to rvoice audio buffer 8, then to eff. output 5

If you had called `fluid_synth_process()` with 3 outputs, you just take the above `effect outputs % 3` (not sure how this could be useful though...)

jjceresa:
>audio_groups cannot be set to zero and would stay one instead.

Thanks for correcting my crappy hypothesis. 

Many thanks for this nice and useful explanations. This shows that `synth.effects-groups`  gives the user more control on which `user effect output` a MIDI voice `effect` (reverb,chorus,others..) will be send (by fluid_synth_process()). Also the name **synth.effects-groups** seems to be fine.

So in summary from a user perspective to understand mapping, this user should only know the following arithmetic formula  (F1,F2):

```
Let:
- midi_chan, the MIDI channel # to map (0 to 15).
- midi_effect, the midi effect # to map (0 to fluid_synth_count_effects_channels() -1).
- audio_count_dry_outputs, the number of dry outputs given by the user (1 to max(synth.audio-channels, synth.audio-groups).
- audio_count_effects_outputs, the number of effect outputs given by the user (1 to (synth.effects-groups * synth.effects-channels)).
- audio_dry_output, the user dry output # used to map midi_chan.
- audio_effect_output, the user effect output # used to map midi_effects # for midi_chan #.

The mapping in user output will be:
(F1) audio_dry_output = midi_chan % audio_count_dry_outputs.
(F2) audio_effect_output = (((midi_chan * synth.effects-channels) + midi_effect) % (synth.effects-groups * synth.effects-channels)) % audio_count_effects_outputs.
```
Hope those formulas are correct in regard of this PR'  behaviour ?

src/rvoice/fluid_rvoice_mixer.c
```diff
@@ -668,18 +702,25 @@ new_fluid_rvoice_mixer(int buf_count, int fx_buf_count, fluid_real_t sample_rate
...
     mixer->buffers.buf_count = buf_count;
-    mixer->buffers.fx_buf_count = fx_buf_count;
+    mixer->buffers.fx_buf_count = fx_buf_count * fx_units;
 
     /* allocate the reverb module */
-    mixer->fx.reverb = new_fluid_revmodel(sample_rate);
-    mixer->fx.chorus = new_fluid_chorus(sample_rate);
-
-    if(mixer->fx.reverb == NULL || mixer->fx.chorus == NULL)
+    mixer->fx = FLUID_ARRAY(fluid_mixer_fx_t, fx_units);
```
jjceresa: May be the returned pointer this should be checked ?

src/rvoice/fluid_rvoice_mixer.c
```diff
@@ -126,65 +127,78 @@ static int fluid_rvoice_mixer_set_threads(fluid_rvoice_mixer_t *mixer, int threa
...
+        out_ch_l = out_rev_l = fluid_align_ptr(mixer->buffers.left_buf, FLUID_DEFAULT_ALIGNMENT);
+        out_ch_r = out_rev_r = fluid_align_ptr(mixer->buffers.right_buf, FLUID_DEFAULT_ALIGNMENT);
 
         reverb_process_func = fluid_revmodel_processmix;
         chorus_process_func = fluid_chorus_processmix;
 
     }
     else
     {
+        // mix effects to respective stereo effects channel
```
jjceresa: The comment should be: // replace effects to respective stereo effects channel

jjceresa:
Just a question a bit out of subject. What could be the function of `synth.audio-groups` ?. May it is related to more audio produced by LADSPA ?

src/rvoice/fluid_rvoice_mixer.c
derselbst: Thanks!

derselbst:
> What could be the function of synth.audio-groups ?

Not quite sure either. It has to do with LADSPA, though I never really investigated. All I know is that it's the modulo operand responsible for voice to channel mapping in `fluid_voice_init()`.

derselbst:
Documentation done, ready to merge.

> Hope those formulas are correct in regard of this PR' behaviour ?

I think it's basically right. I attempted to simplify it a little more.

derselbst:
Yes, however 1 MB per synth.effects-groups isnt really worth to mention.

PR #406, opened 2018-06-25
Convex shape for modEnv

### LINKED ISSUE #153: Modulation envelope attack is linear instead of convex
According to the [2.01/2.04 spec](http://freepats.zenvoid.org/sf2/sfspec24.pdf), both the volume and modulation envelopes should have a convex attack phase. FluidSynth correctly uses a convex attack for the volume envelope, but uses a linear attack for the modulation envelope.

You can hear the difference by comparing the two "modulation envelope test" OGG recordings (inside attached archive) between the Audigy 2 ZS and FluidSynth. In the recording, you will hear a single note playing through all stages of a modulation envelope that is mapped to the note's pitch. Hear the difference in the speed/curve of the initial rise in the note's pitch. The similarly-named .mid and .sf2 files can be used to test this yourself.

Reported by: mrbumpy409

Original Ticket: [fluidsynth/tickets/155](https://sourceforge.net/p/fluidsynth/tickets/155)

derselbst:
This implements the convex shape of modEnvAttack as per soundfont spec.

Tested with the data set by @mrbumpy409 in issue #153. First picture shows the spectrogram of the  audigy's version, second is the new convex version of fluidsynth.

@jjceresa Could you please look into correct handling during legato playing, such as in `fluid_rvoice_multi_retrigger_attack()`?

audigy-convex
fluidsynth-convex

Fixes #153.

### DIFF
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -592,7 +592,7 @@ fluid_rvoice_noteoff_LOCAL(fluid_rvoice_t *voice, unsigned int min_ticks)
         /* A voice is turned off during the attack section of the volume
          * envelope.  The attack section ramps up linearly with
          * amplitude. The other sections use logarithmic scaling. Calculate new
-         * volenv_val to achieve equievalent amplitude during the release phase
+         * volenv_val to achieve equivalent amplitude during the release phase
          * for seamless volume transition.
          */
         if(fluid_adsr_env_get_val(&voice->envlfo.volenv) > 0)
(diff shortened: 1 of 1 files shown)


jjceresa:
>Could you please look into correct handling during legato playing, such as in fluid_rvoice_multi_retrigger_attack()?

Yes, i will look and add the adequate handling, during the next days,weeks!

derselbst:
Ok, no hurry. I've scheduled this for 2.1, so we have plenty of time.

PR #407, opened 2018-06-26
Remove ramsfont

derselbst:
This proposes to remove the ramsfont, because it is

* unmaintained,
* a huge code duplication of defsfont,
* highly limited in regards to modulators and instrument zones, and
* considered to be not really useful at all.

I think the API cleanup is a good opportunity to just remove it and see if anybody downstream uses it. If yes, they should contact the mailing list and eventually my motivation to maintain and think about optimizing this piece of code may rise.


mawe42:
I was wondering about the state of ramsfont when doing the loading refactor. I think now is a good time to remove it.

Actually, maybe we could go one step further and also remove the Max/DSP and JNI bindings? At least they look like they are unmaintained...

derselbst:
> maybe we could go one step further and also remove the Max/DSP and JNI bindings?

Sure, will do. They use a couple of fluidsynth's internal types and since the fvisibility=hidden of 1.1.7 they shouldnt compile anyway. Also I dont think those plugins fit in the context of this repository. If someone really still needs them we could setup separate repositories for those plugins.

PR #410, opened 2018-07-01
Fix single thread build

carlo-bramini:
Compilation fails if ENABLE_MIXER_THREADS is set to zero, because a mismatch with parenthesis.


derselbst:
Ah yes, it's a bug in clang-tidy. Thanks!

PR #417, opened 2018-07-18
Compile time constant lookup tables

carlo-bramini:
Prologue: I'm doing a fix for calculating the matrices inside fluid_rvoice_dsp.c at configure time by cmake and I wrote a small utility program for this purpose. This program generates a fluid_rvoice_dsp.h with these const tables, which is included if the related macro is emitted in config.h. Since the RAM of an embedded system is a limited resource, moving these arrays in the read-only section is very helpful.

The fix: I was wondering if it would be possible to add a macro (I called it as FLUID_INTERP_SINC_ORDER) for defining the order of the interpolation filter. A similar macro called SINC_INTERP_ORDER is already defined in the source; this fix moves this definition outside and now sinc_table7[] makes also use of it.

Epilogue: in my generation program, I could also hardcode the value '7', but in this way things looks a bit better, at least in my opinion.

Sincerely.


derselbst:
The  `SINC_INTERP_ORDER` macro connotes a flexibility that doesnt exist, because it isnt being used by the interpolation routine. And moving it to `fluid_phase.h` doesnt feel like a good solution either.

However having those tables compile time constant sounds interesting. And I could imagine that a couple of users might profit from that. Perhaps you could elaborate a little more on this approach and I'll see if it can be incorporated to upstream.

jjceresa:
> Since  the RAM of an embedded system is a limited resource

@carlo-bramini. For your need, also the table `sinc_table[]` (5120 bytes) of chorus unit (fluid_chorus.c) could be in rom only. I guess you have already see that.

jjceresa:
> Since the RAM of an embedded system is a limited resource

Also, just for information and in the same goal of reducing RAM memory consumption , the 
`struct _fluid_chorus_t:lookup_tab ` (588000 bytes at speed 0.3 Hz and 1764000 bytes at 0.1 Hz !) could be completely removed and replaced by an other LFO waveform technique. But this is another story.

carlo-bramini:
I uploaded all my preliminar changes that I made for handling the generation of the tables at compile time. By calling cmake with "-Denable-ctables=on" you will activate the feature, otherwise it works like in the past. However, this change has the limitation that it cannot be used when you are cross compiling. Actually the generation program MUST be compiled for running in the build system. Unfortunately, I have no idea at the moment to bypass this limitation. I'm wondering if you have some suggestions for this.

@jjceresa sinc_table[] could be also generated at compile time, unfortunately I did not it yet.
Actually, the biggest waste of RAM is the buffer used for sound dithering, IMO.

I hope that you will find some bits useful in this change.

Sincerely.

mawe42:
Just out of curiosity: what platform are you targeting, where things like 5k of RAM for the sinc_table makes a difference?

derselbst:
Looks pretty good, thanks for your input Carlo! However I think it would be better if the auto generated tables would contain the actual values, rather than a series of preprocessor macros. This would also avoid the gcc limitation.

The only solution I came up with is to use C++17.  The idea would be to let cmake check whether a C++ compiler is available and check whether `#if __cplusplus == 201703L` is true. The `fluid_rvoice_dsp.c` could then be compiled in C++17 mode, while exporting all its interpolation functions as `extern "C"`.
The code for the linear interpolation table would look like:

```c++
using interp_coeff_linear_t = ::std::array<fluid_real_t, FLUID_INTERP_MAX * 2>;

constexpr interp_coeff_linear_t init_interp_coeff_linear()
{
    interp_coeff_linear_t result = {0};

    for(int i = 0; i < FLUID_INTERP_MAX; i++)
    {
        double x = (double) i / (double) FLUID_INTERP_MAX;

        result[i*2 + 0] = (fluid_real_t)(1.0 - x);
        result[i*2 + 1] = (fluid_real_t)x;
    }
    
    return result;
}

/* Linear interpolation table (2 coefficients centered on 1st) */
static constexpr interp_coeff_linear_t interp_coeff_linear = init_interp_coeff_linear();
```

Ofc, this would be optional, i.e. falling back to runtime init if no C++17. Advantage would be that it works for cross compiling. Disadvantage: `sin()` and `cos()` are not constexpr, so we would need to implement them our self.

Any thoughts which way to go? Carlos auto generated tables (with real values) or C++17 constexpr arrays?

mawe42:
> Any thoughts which way to go? Carlos auto generated tables (with real values) or C++17 constexpr arrays

Cross-compiling is very important for my use of fluidsynth, so I would personally prefer a solution that is compatible with that. What exactly is it that makes Carlos solution not work with cross-compiling?

And how much memory does this change save? Does the amount of memory we save by one of these two approaches actually justify another compile-time switch and (seemingly) more code complexity? Don't get me wrong... the idea of generating static data at compile-time seems to make a lot of sense. And reducing memory consumption is always a good thing. I'm just trying to understand the motivation for this change. Hence my question about the target system...

carlo-bramini:
@mawe42 I'm using an LPC1788, ARM Cortex M3, 120 MHz, 512 KB Flash ROM, 96KB of high speed SRAM. On my prototype board I have also an 8MB of SDRAM (I use it for dynamic memory allocation) and a 64 MB NOR Flash not used ATM with fluidsynth because soundfonts loaded from an SD card.

carlo-bramini:
@derselbst ATM I have used the builtin intrinsic functions provided by GCC because in that way I was sure that there was no precision loss with a less work. That was the only reason, but it would be also possible to write big tables filled with numbers with enough number of decimal digits, hoping that they are enough.

@mawe42 this solution does not work when cross compiling because you must generate also the tool that creates the tables and run it at compile time. For example, if you are running on an x86 PC and the cross compiler generates executables for ARM, the tool that I have called "make_tables" must be compiled for running on x86. But when cross compiling, the tool "make_tables" is also created with the cross compiler, which is not good, because that cannot run on your x86 PC. ATM I bypassed this by compiling a local build of fluidsynth for x86, then I copied the generated include file in the binary directory of the embedded ARM and finallly I could compile the ARM version too without troubles. But I think that there should be a way to do it, I know to be not expert enough with cmake probably I'm just missing the way to do it.

derselbst:
> I was sure that there was no precision loss with a less work

I wouldnt worry too much about that. One can print out doubles pretty accurately either by

`printf("%.17g", d);`

or by using C99 hex floats

`printf("%a", d);`

After all, the result of `sin()` and `cos()` would esp. depend on the currently active rounding mode, which the `__built_in` versions probably just ignore.

> Does the amount of memory we save by one of these two approaches

There are many more tables in question, e.g. all the lookup tables of `fluid_conv.c`.

> ...actually justify another compile-time switch

If possible I would also prefer to avoid another cmake flag here, let's see.

Can you guys actually use C++ at all when (cross-)compiling for  embedded system, as it would require to link against libstdc++ as well?

mawe42:
@carlo-bramini thanks for your target platform details, very interesting! It's great to have someone use fluidsynth on such limited hardware. And now I fully understand the need to save every precious byte.

Regarding the cross-compiling... Couldn't we generate those files and make them part of the source distribution? Still provide the program to regenerate them, but have a default set checked in so that there's no c++ dependency and to ease cross-compiling?

derselbst:
> Couldn't we generate those files and make them part of the source distribution?

We could. But checking in auto generated files is so lame :(

> so that there's no c++ dependency

What exactly strikes you? C++ in general or the dependency to the standard library?

carlo-bramini:
> > Couldn't we generate those files and make them part of the source distribution?
> 
> We could. But checking in auto generated files is so lame :(
> 

You made me an idea, perhaps there could be a solution.

When I cross compile WINE with autotools, I have to do two steps: in the first one I configure and then I launch a make buildtools. Next, I configure again with option `--with-wine-tools=<subdir>`.
`subdir` will point to the directory with the tools compiled for the build platform in the previous step. In that way, the tools are not compiled, but they are taken from there.

Perhaps, we can do the same thing with fluidsynth: if an option is specified, it will compile the utility program for table generation, otherwise it will compile fluidsynth. The possible scenario can be:
- no options, it is compiled as it is now.
- enable constant table option activated: if the extra option with the path to the pre-compiled generator is given, it uses that one. Otherwise, check if it is a cross compile: if yes, it signals a warning an it disables the feature, otherwise, since it is a native compilation, it builds also the tool.

If I'm not mistaken, The trick itself seems quite easy, with something like this inside CMakeLists.txt:

```
if (option)
add_subdirectory ( tools)
else()
add_subdirectory ( src )
add_subdirectory ( doc )
endif()
```

What do you think?

mawe42:
> But checking in auto generated files is so lame :(

Agreed :-) Well, calling one extra build step with host gcc and the rest with target gcc is not really a problem, I would think. Especially if using a build framework like buildroot or similar.

C++ itself is not a problem, I was just thinking that adding another lib sounds counter productive when the goal is to save as many bytes as possible. But as the main problem is working memory, I guess it doesn't really matter.

derselbst:
> Well, calling one extra build step with host gcc and the rest with target gcc is not really a problem,

But how can you access the host compiler in cmake? When cross compiling you specify this toolchain file that does something like

`FIND_PROGRAM(CMAKE_C_COMPILER NAMES ${TOOLCHAIN_PREFIX}-gcc)`

which is the target compiler, where's the host?

> C++ itself is not a problem, I was just thinking that adding another lib sounds counter productive when the goal is to save as many bytes as possible.

Right. In this case I want to point out that we could avoid any dependency to libstdc++ by writing our own lightweight constexpr std::array class. It would still require a C++14 compliant compiler though.

derselbst:
I kept thinking about the C++ approach: When taking also the  lookup tables of `fluid_conv.c` into account, we would not only need `sin()` but also `log()` and `pow()`.

* `sin()` is simple to approximate through a Taylor series
* `log()` can be approximated through a Taylor series as well, or via [Newtons method with cubic convergence](https://en.wikipedia.org/wiki/Natural_logarithm#High_precision) (this however would require a general power function)
* `pow()`... trivially to implement for integral exponents. But for calculating a general power you have to use a power, argh. The only viable solution I see would be to split up the power into a multiplication of a power with integral exponent and a remaining float exponent, i.e.:

```
powi(double base, int exp);
powf(double base, double exp);

pow(double base, double exp)
{
  int iexp = (int)exp;
  float rem_exp = exp - iexp;

  return powi(base, iexp) * powf(base, rem_exp);
}
```

where powi() is trivial and powf() would be

\sum_{n%3D0}^\infty%20\frac{(b%20\ln%20a)^n}{n!}^n}{n!})

where the first 10 iteration would be sufficiently accurate and `-1 < b < 1`. If we were not splitting this up and perform everything by this infinite sum we would need at least 50 iterations, where the factorial get way too big.

Dont get me wrong: I'm not insisting on the C++ approach. I'm pointing out that it's challenging and no quickly straight forward solution. And dont forget about the custom `constexpr array class` and possibly unit tests to assert a certain accuracy. A lot to do, but it would work without additional cmake flags and also for cross compilation (a C++14 compiler provided). This makes it so interesting for me and I would be willing to do this... just need to know your preference.

derselbst:
Here's a minimal example of how the C++ approach would look like.

[constexpr.txt](https://github.com/FluidSynth/fluidsynth/files/2238687/constexpr.txt)

Compile with: `g++ constexpr.cpp -o const -std=c++17 -Wall`

The above command implicitly links against libstdc++. If you however comment out all the `std::cout`s you should be able to get rid of this dependency, i.e. there is no need of a custom `std::array` implementation. Verify by:

```bash
$ gcc constexpr.cpp -o const -std=c++17 -Wall
$ ldd const
        linux-vdso.so.1 (0x00007ffe4e6c1000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f537f666000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f537fa20000)
```

Thus we could just integrate this into the existing codebase with a bunch of `#if __cplusplus >= 201703L`.

Opinions, thoughts?

carlo-bramini:
I made some improvements to this topic:
* the generators are made by compiling and linking the code in FluidSynth.
* it uses hex floating point constants, without precision loss. It needs C99 if activated.
* I also added the tables inside fluid_conv.c.

Usage 1: it builds the generator and FluidSynth in one pass, both are made to be used on the build platform.
`cmake -Denable-ctables=on`

Usage 2: natively compile the generator and then cross compile fluidsynth. It uses the cmake script that I had posted here some time ago, with the settings for compiling with mingw compiler under cygwin shell.

```
cd fluid_native
cmake ../fluidsynth -Denable-ctables=on -Denable-generator=on
cd ../fluid_cross
cmake ../fluidsynth -Denable-ctables=on -DWITH_GENERATOR=$HOME/fluid_cross/src/tools/maketables.exe -DCMAKE_TOOLCHAIN_FILE=$HOME/mingw32.cmake
```

Without options, everything will work like in the past.

derselbst:
Frankly, I'm not too happy with this solution. You introduced three extra cmake flags, two of them are only needed to somehow circumvent the problem we are facing when cross compiling. I dont think this is justified for this issue. It makes things unnecessarily complex and I doubt that anyone except you will actually make use of this.

Is there anything about the C++ approach that bothers you?

derselbst:
This surprises me, as you were the first one making a PR for gcc7. Could you elaborate a little more why this is the case? Is C++ in general not applicable for you? Would it help to relax the requirements to C++14?

jjceresa:
> Perhaps, a tool with its standalone build script could be made. It is compiled only if the user wants and it creates a source file with all the constant tables.

As the goal is to get all constants values in ROM (for embedded needs), it seems suitable that the user should be compelled for this choice. Imho,  the right place for this choice should be done 
only at Cmake time (if possible ofc) and **only** via a build option (e.g ReleaseTablesInRom or other name) to remain the most understandable as possible.

Hope this comment will not interfere with interesting Tom's questions.

carlo-bramini:
About C++, MSVC and Mingw are not a problem... the problem is my cross compiler, because I'm forced to use GCC 4 for using precompiled USB libraries provided by NXP:

https://community.nxp.com/thread/450385

carlo-bramini:
I did a new solution, without the need of an external tool program.
The tables are generated at compile time by the preprocessor.
Works on native, cross compile and also on C90 compilers.
Only one option inside CMakeLists.txt.

derselbst:
Ok, lets go with the preprocessor. However compilation with clang currently fails because it doesnt evalute the __builtin calls at compiletime. I suggest to completely remove this special treatment for GCC and simply use the approximations.

carlo-bramini:
While trying to think about how to simplify some expressions for making some future improvments at compile time, I noticed that this line:

https://github.com/FluidSynth/fluidsynth/blob/55a4bb559736f4b1369d27774307fb9fa6456db8/src/utils/fluid_conv.c#L77

calculates a logarithm of a square value.
In my opinion, this code:

`log((i * i) / (fluid_real_t)((FLUID_VEL_CB_SIZE - 1) * (FLUID_VEL_CB_SIZE - 1))`

could be rewritten as:

`2 * log((fluid_real_t)i / (fluid_real_t)(FLUID_VEL_CB_SIZE - 1))`

Now, the division inside the argument could be separated:

`2 * (log(i) - log(FLUID_VEL_CB_SIZE - 1))`

into two logarithms of natural interger numbers.
At the moment, for starting I would like to suggest to extract the square out of the argument of the logarithm, which seems a redundant multiply.

jjceresa:
> could  be rewritten as: 2 * log((fluid_real_t)i / (fluid_real_t)(FLUID_VEL_CB_SIZE - 1))

Correct. Please note also when computing the inverse of fluid_convex (in fluid_rvoice_multi_retrigger_attack) , this simplification has been already taken into account in (https://github.com/FluidSynth/fluidsynth/pull/406).

> I would however vote for making the enable-ctables option defaulting to TRUE or remove it entirely (as we've found a C90 solution that will work everywhere and for everyone).

In the case of **enable-ctables** option still will exist (defaulting to TRUE), may i suggest to logout the effect of this option at Cmake time ?.

derselbst:
> may i suggest to logout the effect of this option at Cmake time ?.

Sure, currently we are informing the user anyway.

carlo-bramini:
Although this solution allows to generate a good amount of tables at compile time:

1) the code is less readable than before and harder to maintain.

2) we are forced to use some hardcode values. For example, `interp_coeff[]` has `FLUID_INTERP_MAX` items, and into the declaration I used `AUTO_GEN_ARRAY_256()` because its size is actually 256. But if somebody will change the value of `FLUID_INTERP_MAX`, let's say to 512, the user has to modify also the declaration to `AUTO_GEN_ARRAY_512()`, because it won't otherwise.

derselbst:
Regarding 1: That's the necessary evil of this solution, I'm willing to buy it.

Regarding 2:

#define FLUID_INTERP_MAX 256

#define CONCAT(x,y) x ## y
#define EVAL(x,y)  CONCAT(x,y)

#define DECLARE_LOOKUP_TABLE(NAME, TYPE, SIZE, COEFF) \
static const TYPE NAME[SIZE] = { EVAL(AUTO_GEN_ARRAY_, SIZE)(COEFF) };

DECLARE_LOOKUP_TABLE(interp_coeff_linear, fluid_real_t,
FLUID_INTERP_MAX, INTERP_COEFF_LINEAR)

carlo-bramini:
About (2) I was thinking the same, but that requires C99.

derselbst:
Oh, right. Dont like it either anyway, I dont mind having this slightly redundant writing as we do now.

derselbst:
Superseded by #423.

PR #423, opened 2018-08-21
Compile time constant lookup tables (updated)

derselbst:
Calculate lookup tables during compile time, as intended by #417.

* cleaned up commit history
* use a more accurate constexpr sin()
* fix calculation of sinc table
* add a unit test to verify accuracy to legacy runtime initialized lookup tables

#### TODO:
- [ ] memory consumption for all compilers very high (>3GiB)
- [x] ~~MSVC entirely fails (out of heap space)~~ (wont fix)
- [x] clang-tidy may be killed fails


derselbst:
Sry, I dont know to which change you are referring. However, all the interpolation tables of fluid_rvoice_dsp.c are correctly placed in the .rodata section, provided that you set the cmake flag.

carlo-bramini:
Ouch, you are right, I have seen where the const keyword is. ^^;

Perhaps, since it could be used by more than one file, it could be emitted by cmake into config.h something like this:
`#define CONST_CTABLE const`
or:
`#define CONST_CTABLE /* empty */`
depending on the flag at configure time, to simplify a bit more.

derselbst:
Ok, will do so.

carlo-bramini:
I had also seen the crash with MSVC... but perhaps it can be fixed without much work if we want: instead of using a cumulative base 10 inside auto_gen_array.h, the same thing but made with base 2 can be used instead. I will try to do some experiments.

derselbst:
Any success with MSVC?

carlo-bramini:
I did some experiments and I did a solution that in my opinion should use less memory, but MSVC6 still does not work. I have not tried newer versions of the compiler at the moment.

PR #425, opened 2018-09-04
Implement fluid_get_userconf() on windows

derselbst:
Suggests to implement `fluid_get_userconf()` on win32 starting with fluidsynth 2.0 to point to `%userprofile%\\fluidsynth.ini`. Not sure why it was avoided to implement. At least some user complained on the mailing list a couple of months ago.

@jjceresa Are you fine with this?

### DIFF
--- a/doc/fluidsynth-v20-devdoc.txt
+++ b/doc/fluidsynth-v20-devdoc.txt
@@ -81,6 +81,7 @@ FluidSynths major version was bumped. The API was reworked, deprecated functions
 - rename chorus getters to match naming conventions: fluid_synth_get_chorus_speed() and fluid_synth_get_chorus_depth()
 - fluid_synth_remove_sfont() returns FLUID_OK or FLUID_FAILED
 - introduce a separate data type for sequencer client IDs: #fluid_seq_id_t
+- fluid_get_userconf() has been implemented for Windows
 
 <strong>New Features and API additions:</strong>
 
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -623,22 +623,27 @@ fluid_source(fluid_cmd_handler_t *handler, const char *filename)
 char *
 fluid_get_userconf(char *buf, int len)
 {
-#if defined(WIN32) || defined(MACOS9)
-    return NULL;
-#else
-    char *home = getenv("HOME");
-
+    const char *home = NULL;
+    const char *config_file;
+#if defined(WIN32)
+    home = getenv("USERPROFILE");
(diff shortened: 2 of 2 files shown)


mawe42:
Good idea to implement this. I have tested the VS17-Win64 build from AppVeyor on Windows 10 and it seems to work as intended.

I would recommend to use a different filename, though. The .ini extension suggest that it's in [INI-format](https://en.wikipedia.org/wiki/INI_file), which is it not. Maybe go with `fluidsynth.cfg` instead?

And it seems like there are many ways to specify the user home directory on Windows. So maybe instead of writing this ourselves, we could use g_get_home_dir, or just look at their implementation: https://github.com/GNOME/glib/blob/master/glib/gutils.c#L820

mawe42:
> Ok thx. I dont like that g_get_home_dir() returns the sysroot as "ultima ratio". That doesnt quite fit for our usecase.

Good point.

> I think relying on %USERPROFILE% should be fine.

Yes, probably good enough. And it can always be extended, if people with weird profile setups have troubles.

jjceresa:
> Suggests to implement fluid_get_userconf() on win32 starting with fluidsynth 2.0 to point to %userprofile%\\fluidsynth.ini

Good idea.
The .ini extention should be avoided for the reason described by Marcus. Use of .cfg (or other) seems fine.

> I think relying on %USERPROFILE% should be fine.

I think it is a good choice as this folder is really related to the current user (not the other users). 

For example actually on my Windows machine my user name is "Administrateur".
And  %USERPROFILE%  should be the home directory of "Administrateur" (which is usually `C:\Documents and Settings\Administrateur` ).
1)When a command line console is open, we see the following prompt:
```
Microsoft Windows XP [version 5.1.2600]
(C) Copyright 1985-2001 Microsoft Corp.

C:\Documents and Settings\Administrateur>
```
2)The following command verify that %USERPROFILE% is the home directory (named also by the prompt):
```
C:\Documents and Settings\Administrateur>dir "%userprofile%\"
```
The display is:
```
 Le volume dans le lecteur C s'appelle Systeme
 Le numéro de série du volume est ACC7-B57D

 Répertoire de C:\Documents and Settings\Administrateur

28/07/2018  10:09    <REP>          .
28/07/2018  10:09    <REP>          ..
12/08/2018  21:21             9 862 .bash_history
27/12/2017  12:05                63 .gitconfig
....
```
3)The following command display the content of working directory:
`C:\Documents and Settings\Administrateur>dir`
And the display is the same as for 2).

derselbst:
Alright, thanks to both of you.

PR #426, opened 2018-09-06
Change the method of loading the source files.

fulinux:
Hi, 
   Change the method of loading the source files in src/CMakeLists.txt
tks.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -107,78 +107,14 @@ endif ( MIDISHARE_SUPPORT )
 
 set ( config_SOURCES ${CMAKE_BINARY_DIR}/config.h )
 
-set ( libfluidsynth_SOURCES
-    utils/fluid_conv.c
-    utils/fluid_conv.h
-    utils/fluid_hash.c
-    utils/fluid_hash.h
-    utils/fluid_list.c
-    utils/fluid_list.h
-    utils/fluid_ringbuffer.c
-    utils/fluid_ringbuffer.h
-    utils/fluid_settings.c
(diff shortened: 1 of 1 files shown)


derselbst:
Please explain why, i.e. what are you intending?

fulinux:
we can avoid adding source file name in cmakelist one by one, my english was poor 0.0

fulinux:
@derselbst we can avoid adding source file name in cmakelist one by one, my english was poor 0.0

derselbst:
The problem with your solution is that `aux_source_directory()` ignores header files. While this works fine for compilation only, cmake is also able to generate project files for e.g. VisualStudio. If we used `aux_source_directory()` any generated project file would lack all header files. 

Also note that the cmake docs discourage your proposed solution:

> It is tempting to use aux_source_directory() to avoid writing the list of source files for a library or executable target. While this seems to work, there is no way for CMake to generate a build system that knows when a new source file has been added. Normally the generated build system knows when it needs to rerun CMake because the CMakeLists.txt file is modified to add a new source. When the source is just added to the directory without modifying this file, one would have to manually rerun CMake to generate a build system incorporating the new file.

A similar warning exists for `file(GLOB)`:

> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

This in turn means that I would have to re-generate the build system everytime when using `git bisect` in the future. I'm not willing to do this. Thus I prefer to stay with manually listed files.

fulinux:
@derselbst Thank you for your explanation

Issue #414, opened 2018-07-07
extend implementation of "midi.autoconnect"


derselbst:
Implement "midi.autoconnect" for alsa_raw, jack and potentially other midi drivers.

- [x] CoreMidi
- [x] jack
- [x] winmidi
- [ ] alsa raw

jjceresa:
I don't quite understand this "autoconnect" need. Please does anyone could explain anymore what is expected ?

derselbst:
Pls. review PR #266 and let me know if this clarifies it.

jjceresa:
>let me know if this clarifies it.

Thanks, now it is clear. As far i know i dont think that "autoconnect" can be implemented for Windows (i mean without a dependency taking account of multiple MID input devices).

carlo-bramini:
Excuse me, I was wondering if you could clarify few things that are not clear to me.

1) The user starts FluidSynth. After that, the user adds or removes a plug and play MIDI device, typically connected with USB. When this happens, FluidSynth updates its list of MIDI devices. Should this update happen only when autoconnect feature is activated?

2) In the current code, only one MIDI device is open each time, at least with winmidi. Now, when autoconnect feature is activated, does this mean that FluidSynth must open ALL avaliable devices when it starts?

3) In my opinion, the list of available audio output devices should be updated,  in the same manner of MIDI devices, but not automatically opened like in the autoconnected feature. For example, the user can plug or unplug USB speakers and the list of available devices is updated according that action. What do you think?

derselbst:
> When this happens, FluidSynth updates its list of MIDI devices. Should this update happen only when autoconnect feature is activated?

The autoconnect feature only says that suitable MIDI devices should be automatically connected to. Whether this makes it necessary to maintain a list of devices or get a callback are details of the implementation. However, I currently don't see a need to keep that list up to date when autoconnect is disabled.

> In the current code, only one MIDI device is open each time, at least with winmidi. Now, when autoconnect feature is activated, does this mean that FluidSynth must open ALL avaliable devices when it starts?

I'm not familiar with winmidi. The autoconnect feature originated from alsa, where fluidsynth connects to ports. Whenever a new port appears, fluidsynth gets a callback and connects to it. I hoped that smth. similar to this is available for winmidi as well. If not, the autoconnect feature could either be dropped for winmidi or, as compromise like you said, connect to every readable port that is available when fluidsynth starts up.

[...]

PR #427, opened 2018-09-10
coremidi input port and autoconnect

ColinKinloch:
Adds a port to the coremidi driver by default and optionally automatically connects this to all midi sources (related to #414).

### DIFF
--- a/src/drivers/fluid_coremidi.c
+++ b/src/drivers/fluid_coremidi.c
@@ -57,7 +57,9 @@ typedef struct
     fluid_midi_driver_t driver;
     MIDIClientRef client;
     MIDIEndpointRef endpoint;
+    MIDIPortRef input_port;
     fluid_midi_parser_t *parser;
+    int autoconn_inputs;
 } fluid_coremidi_driver_t;
 
 void fluid_coremidi_callback(const MIDIPacketList *list, void *p, void *src);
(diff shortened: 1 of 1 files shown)


derselbst:
Oh, very nice. Thank you.

Could you please explain why you are creating this midi input port? Or what's the difference to `MIDIDestinationCreate()`? Sry, I don't have any experience with CoreMidi and a quick reading didn't cleared it up, so just asking.

ColinKinloch:
To be honest, I don't know. 

I've never been able to get fluidsynth to interact with a MIDI keyboard on macOS before now.

I spent a while searching the web and looking through the CoreMIDI docs without much luck.

CoreMIDI via `MIDIGetNumberOfSources` and `MIDIGetSource` allows you to enumerate input devices (virtual / hardware) which, like the output of `MIDIDestinationCreate`, are "endpoints".

`MIDIPortConnectSource` allows a client to connect a source endpoint to a "port". This doesn't require a destination for the client to exist and there doesn't seem to be an analogue in CoreMIDI for connecting destinations and sources.

mawe42:
I dont' know CoreMidi either, but from reading some explanations I think it works like this: 

MIDIDestinationCreate` creates a virtual MIDI device that other applications can connect to. So that gets used if you want to use fluidsynth as a virtual instrument in Garageband etc. Garageband will enumerate all destinations (virtual or hardware) known to the system. As soon as you choose one, it creates an output port to send MIDI messages to that destination. So in essence MIDIDestinationCreate is used to open up to connections initiated by other apps.

An input port created with `MIDIInputPortCreate` is simply the receiving end of a connection to a source, be it virtual or hardware. So that is used when we ourselves initiate the connection to a MIDI source and want to receive events from it.

Anyway, as I said... just a wild guess. CoreMIDI has some strange concepts when compared to the Linux world of things.

src/drivers/fluid_coremidi.c
```diff
@@ -67,6 +69,31 @@ void fluid_coremidi_driver_settings(fluid_settings_t *settings)
     fluid_settings_register_str(settings, "midi.coremidi.id", "pid", 0);
 }
 
+static void fluid_coremidi_autoconnect(fluid_coremidi_driver_t *dev, MIDIPortRef input_port)
+{
+    unsigned int i;
+    int source_count = MIDIGetNumberOfSources();
+    for(i = 0; i < source_count; ++i)
```
derselbst: Now, the  comparision `i < source_count` differs in signedness. Prefer using int.

PR #430, opened 2018-09-15
Removing useless invalid preset generators checking in fluid_defpreset_noteon()

jjceresa:
At fluid_defpreset_noteon() some invalid preset zones generators should be ignored. But if the soundfont loader has already ignored these, it seems useless that fluid_defpreset_noteon() redo this check ?.

### DIFF
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -832,20 +832,20 @@ fluid_defpreset_noteon(fluid_defpreset_t *defpreset, fluid_synth_t *synth, int c
                     for(i = 0; i < GEN_LAST; i++)
                     {
 
- 						/* SF 2.01 section 8.5 page 58: If some generators are
+                        /* SF 2.01 section 8.5 page 58: If some generators are
                          encountered at preset level, they should be ignored.
-						 However this check is not necessary when the soundfont
-						 loader has ignored invalid preset generators.
-						 Actually load_pgen()has ignored these invalid preset
-						 generators:
+                         However this check is not necessary when the soundfont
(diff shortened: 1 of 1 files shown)


src/sfloader/fluid_defsfont.c
```diff
@@ -832,43 +832,40 @@ fluid_defpreset_noteon(fluid_defpreset_t *defpreset, fluid_synth_t *synth, int c
...
-                                && (i != GEN_STARTADDRCOARSEOFS)
-                                && (i != GEN_ENDADDRCOARSEOFS)
-                                && (i != GEN_STARTLOOPADDRCOARSEOFS)
-                                && (i != GEN_KEYNUM)
-                                && (i != GEN_VELOCITY)
-                                && (i != GEN_ENDLOOPADDRCOARSEOFS)
-                                && (i != GEN_SAMPLEMODE)
-                                && (i != GEN_EXCLUSIVECLASS)
-                                && (i != GEN_OVERRIDEROOTKEY))
+ 						/* SF 2.01 section 8.5 page 58: If some generators are
```
mawe42: Can you format this comment properly and make sure the indentation is correct?

jjceresa: Yes.

derselbst:
> Those checks might have made sense when we still had different loader implementations (i.e. the ramsfont loader)

Nope, each sfloader has its own `noteon()` function, ramsfont as well. This is a much older leftover, even older than "initial revision" (039e880a9437fde543bb4d1da7f91c8bf2f3c175).

Anyway, this should be fine. But give me a few more days to test a little.

PR #435, opened 2018-09-28
Update fluid_dsound.c

carlo-bramini:
I did some quick improvements to DirectSound driver:

- fluid_dsound_audio_run() and fluid_win32_error() are now static functions.

- WAVEFORMATEX is not required by DirectSound during the playback. We do not need it anymore after calling ::SetFormat(), so allocating this structure dynamically inside driver structure is useless.

- Implemented support for float sample type.

- Uses an event object for handling the end of thread, it allows to combine the quit event with the later wait in milliseconds in a single block.

- Calculates the amount of milliseconds to sleep rather than sleeping always for one millisecond.

- Fix an error into a FLUID_LOG() call.

- Fix handle leak of the thread, now it is correctly closed with CloseHandle().

- ExitThread() is a nonsense in that position, since the thread is already exiting.

### DIFF
--- a/src/drivers/fluid_dsound.c
+++ b/src/drivers/fluid_dsound.c
@@ -29,6 +29,8 @@
 
 #include <mmsystem.h>
 #include <dsound.h>
+
+#define NOBITMAP
 #include <mmreg.h>
 
 static DWORD WINAPI fluid_dsound_audio_run(LPVOID lpParameter);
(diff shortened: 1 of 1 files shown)


carlo-bramini:
I'm using it right now, it seems to be fine.

derselbst:
Ok, thanks!

PR #436, opened 2018-09-30
Avoid use of C++ reserved keyword in variables

carlo-bramini:
If it is possible, I would like to suggest to not use the 'new' reserved keyword in this way: hopefully this structure is used only internally, so it is not too difficut to fix it.

### DIFF
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -28,7 +28,7 @@
 typedef struct _fluid_audriver_definition_t
 {
     const char *name;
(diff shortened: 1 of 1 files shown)


carlo-bramini:
I did not tried to compile that in C++ mode, it's just a "cosmetic" change, because the editor underlined that as possible syntax error :expressionless: however, it can be closed without problems, it's not very important afterall...

derselbst:
Problem is that I'm not feeling comfortable to start applying "cosmetic" C++ related changes. So, sorry.

PR #437, opened 2018-10-02
NULL derefs and mem leaks

derselbst:
Fixes various NULL dereferences and two memory leaks in sfont loader.


mawe42:
Looks good!

PR #441, opened 2018-10-13
Check if at least a driver is supported/enabled.

carlo-bramini:
I had noticed that it is possible that FluidSynth does not compile if none of the audio out drivers is supported. I would like to suggest introduce a check, in the same manner it has been done for the MIDI input drivers. Actually, the absence of an audio driver does not mean that FluidSynth is not usable, it will always act as a rendering engine while the user will generate the samples with fluid_synth_write_float()/fluid_synth_write_s16() and play them with the audio API provided by his operating system.

### DIFF
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -21,6 +21,18 @@
 #include "fluid_adriver.h"
 #include "fluid_settings.h"
 
+#undef FLUID_AOUT_SUPPORT
+
+#if JACK_SUPPORT || ALSA_SUPPORT || OSS_SUPPORT || PULSE_SUPPORT || \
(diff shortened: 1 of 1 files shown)


derselbst:
This breaks `fluid_audio_driver_register()` as it would always return `FLUID_OK`. But it should only do that if the user passed NULL or an empty array. If the user requested a different driver that fluidsynth doesn't provide it should return `FLUID_FAILED`.

carlo-bramini:
Excuse me... I did a mistake and I deleted my branch and it seems that I cannot update this anymore, so I'm updating this fix in a different PR. Sorry.

carlo-bramini:
Superseded by PR #444 .

PR #442, opened 2018-10-13
Rounded samples do not need to be float.

carlo-bramini:
Actually, the function roundi() already returns an "int" type value, so in my opinion there is no need to use a floating point value for saturating the values in the range -32768/+32767. The generated assembly code looks more efficient after that.


derselbst:
Yes. You could even go one step further and do the clipping right in `roundi()` itself, using `fluid_clip()`.

derselbst:
Ok.

PR #443, opened 2018-10-13
Fix fluid log

carlo-bramini:
For people who wants to strip a bit the size of the binary, it is possible to reduce the size of the executable by commenting the content of the FLUID_LOG macro. However, I discovered that there are some points in the code that are calling directly the function fluid_log() instead of using FLUID_LOG macro. So, this patch fixes the sources that still need to replace fluid_log() with FLUID_LOG macro.

### DIFF
--- a/src/midi/fluid_seq.c
+++ b/src/midi/fluid_seq.c
@@ -117,7 +117,7 @@ new_fluid_sequencer2(int use_system_timer)
 
     if(seq == NULL)
     {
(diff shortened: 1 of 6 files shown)


derselbst:
You forgot the explanatory comment at the definition of the `FLUID_LOG` macro.

carlo-bramini:
Sorry, I had understood that the explanatory comment had to be only in the GIT log. I will do it.

PR #444, opened 2018-10-14
Check if at least a driver is supported/enabled (2).

carlo-bramini:
I had noticed that it is possible that FluidSynth does not compile if none of the audio out drivers is supported. I would like to suggest introduce a check, in the same manner it has been done for the MIDI input drivers. Actually, the absence of an audio driver does not mean that FluidSynth is not usable, it will always act as a rendering engine while the user will generate the samples with fluid_synth_write_float()/fluid_synth_write_s16() and play them with the audio API provided by his operating system.

EDIT: this PR improves #441 by returning FLUID_FAILED into function fluid_audio_driver_register(), if the library cannot provide a supported audio out driver.

### DIFF
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -21,6 +21,18 @@
 #include "fluid_adriver.h"
 #include "fluid_settings.h"
 
+#undef FLUID_AOUT_SUPPORT
+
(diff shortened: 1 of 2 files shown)


derselbst:
Always returning `FLUID_FAILED` isn't correct either. Pls, read my previous comment again or ask if something is unclear.

carlo-bramini:
Excuse me... I was wondering if the last change was correct or I did mistakes. Thank you.

PR #447, opened 2018-10-20
Fix build when no audio drivers are supported

derselbst:
Same intention as #444 but with less code.

Closes #444.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -105,6 +105,10 @@ if ( MIDISHARE_SUPPORT )
   include_directories ( ${MidiShare_INCLUDE_DIRS} )
 endif ( MIDISHARE_SUPPORT )
 
+if ( AUFILE_SUPPORT )
+  set ( fluid_aufile_SOURCES drivers/fluid_aufile.c )
+endif ( AUFILE_SUPPORT )
+
 set ( config_SOURCES ${CMAKE_BINARY_DIR}/config.h )
 
 set ( libfluidsynth_SOURCES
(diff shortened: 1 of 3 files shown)


carlo-bramini:
In my opinion, while this solution too resolves the troubles when no drivers are supported, I have some doubts that it requires less code... it keeps an some kilobytes of useless code (I have not measured how much) into the library.

derselbst:
By "less code" I was referring to source code. All the loops will never be traversed and will be optimized away. I don't think it's worth to remove all the code with macros just to save a few more bytes for the compiled binary.

carlo-bramini:
Actually, both solutions solve the trouble of compiling FluidSynth when none of the included drivers is supported, so at the end it does not matter too much which one you will decide to apply... If I can say my opinion, 3 or 4 `#ifdef..#endif` are not the ones that change the complexity of the code, so it may be better to keep FLUID_MIDI_SUPPORT and FLUID_AOUT_SUPPORT and let cmake to define them directly into `config.h`. Perhaps, since the library is used by several applications just as a rendering engine because they already provide their output drivers, in the future the user may have also the chance to decide at compile time to exclude the drivers with an option (it's just an idea).

PS: about the "less code", I did a change in my branch, for removing all the conditional definitions into `fluid_midi_driver_settings()` and `fluid_audio_driver_settings()`, by moving them into the `for(;;)` cycle. Into `fluid_mdriver.c`, I also changed a bit `fluid_midi_drivers[]`, to keep the same priority order used into the code. Do you think it may be useful?

PR #448, opened 2018-10-23
Less code drivers

carlo-bramini:


src/drivers/fluid_adriver.h
```diff
@@ -29,7 +29,7 @@
 
 struct _fluid_audio_driver_t
 {
-    const char *name;
+    const void *define;
```
derselbst: I would prefer `const fluid_audriver_definition_t* define;` here. Which you could do by only moving the typedef to this header while keeping the struct´s implementation opaque.

src/drivers/fluid_mdriver.h
```diff
@@ -29,7 +29,7 @@
 
 struct _fluid_midi_driver_t
 {
-    const char *name;
+    const void *define;
```
derselbst: Same issue as in fluid_adriver.h

src/drivers/fluid_mdriver.c
```diff
@@ -25,18 +25,18 @@
 /*
  * fluid_mdriver_definition
  */
-struct fluid_mdriver_definition_t
+typedef struct _fluid_mdriver_definition_t
```
derselbst: Duplicated typedef. As well as for `fluid_audriver_definition_t`.

PR #450, opened 2018-10-27
jack midi autoconnect

ColinKinloch:
"midi.autoconnect" option for the jack driver.

Related to #414.

### DIFF
--- a/src/drivers/fluid_jack.c
+++ b/src/drivers/fluid_jack.c
@@ -82,6 +82,8 @@ struct _fluid_jack_midi_driver_t
     int midi_port_count;
     jack_port_t **midi_port; // array of midi port handles
     fluid_midi_parser_t *parser;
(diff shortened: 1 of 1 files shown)


ColinKinloch:
I changed it to autoconnect during the jack process callback which may not be ideal as anything in there should be "suitable for real-time execution". It can't be run in the port registration callback. 
Is there a more appropriate time to have it run?

derselbst:
> It can't be run in the port registration callback. Is there a more appropriate time to have it run?

The process callback is indeed unsuitable, as there is memory allocation going on in `jack_get_ports()`. Why isn't it possible in the registration callback itself?

ColinKinloch:
Jack gives the error "Cannot callback the server in notification thread!" on jack_connect an the connection isn't established.

derselbst:
Tested and works as expected. Thanks!

PR #454, opened 2018-11-01
Allow soundfonts to be searched from multiple directories

### LINKED ISSUE #453: Look for soundfonts by default in /usr/share/sounds/sf2
### FluidSynth version 
1.1.11

### Current behavior
Nowadays it looks like we need to load files like `/usr/share/sounds/sf2/GeneralUser.sf2` which is a bit annoying and it makes the need to write the full path every time.

### Expected behavior
It would be way beter to have it look for sound fonts in `/usr/share/sounds/sf2` by default so you only need to load the soundfont with `GeneralUser.sf2` instead. This would make lide a lot easier :smile: 

From `sudo fluidsynth -a alsa -s -g 1.5 /usr/share/sounds/sf2/GeneralUser.sf2`
to `sudo fluidsynth -a alsa -s -g 1.5 GeneralUser.sf2`

remyabel:
Fixes #453

### DIFF
--- a/cmake_admin/DefaultDirs.cmake
+++ b/cmake_admin/DefaultDirs.cmake
@@ -1,6 +1,16 @@
 # Several directory names used by FluidSynth to install files
 # the variable names are similar to the KDE4 build system
 
+# SOUNDFONT_DIRS - list of directories to search for soundfonts
+if ( WIN32 )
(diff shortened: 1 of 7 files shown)


Jacalz:
Absolutely perfect, a great improvement on my initial idea 💡

derselbst:
@remyabel I appreciate your commitment, but, again, this needs to be
discussed at the mailing list first.

PR #455, opened 2018-11-01
Small math simplify

carlo-bramini:


derselbst:
`fluid_hz2ct` looks a little over-engineered, given the fact that it isn’t used anywhere (?). Don’t know how to deal with that. The rest looks fine.

mawe42:
`fluid_hz2ct` was in the initial code drop by Peter and seems to have never been used anywhere. Maybe simply remove it? If we decide to keep it, I would vote to leave it as is, not optimise it.

carlo-bramini:
I'm sorry, I did not notice that fluid_hz2ct was unused. You can drop all the changes on that function if you want.

derselbst:
> Maybe simply remove it?

I would like to keep it for educational purpose, possibly commenting it out. IIRC the spec is a little complicated when it comes to cents <--> hertz.

mawe42:
Sound good.

PR #457, opened 2018-11-03
Fix an incorrectly aligned result when converting between pointer types 

derselbst:
Clang-3.x raises the following warning:

https://travis-ci.org/FluidSynth/fluidsynth/jobs/450156015#L991

This is because the `char idlist[]` is accessed like a 32bit integer, but as being a character array, it may not be aligned as such. This results in undefined behaviour. To fix it I put it in a `union` along with a uint32 telling the compiler to suitably align it for integer access. Now, the warning is gone:

https://travis-ci.org/FluidSynth/fluidsynth/jobs/448425040#L841

@mawe42 Do you agree with this fix?


mawe42:
That looks good! Nifty trick to use a union with a 4-byte integer to force proper alignment. But maybe it would be good to add a short comment above the union to explain it's purpose?

jjceresa:
Cannot we use "const char c[ ]" instead of "const char c[116]" ?

derselbst:
Sure, comment added.

> Cannot we use "const char c[ ]" instead of "const char c[116]" ?

No. In C89, arrays wraped in unions must have a fixed size. Otherwise the size of the union would depend on the initialization of the first member, which results in different sizes for different instances of the same union type.

src/sfloader/fluid_sffile.c
```diff
@@ -77,9 +77,25 @@ enum
     SM24_ID
 };
 
-static const char idlist[] = {"RIFFLISTsfbkINFOsdtapdtaifilisngINAMiromiverICRDIENGIPRD"
-                              "ICOPICMTISFTsnamsmplphdrpbagpmodpgeninstibagimodigenshdrsm24"
-                             };
+/*
+ * This declares a char array containing the SF2 chunk identifiers. This
+ * array is being access like an uint32 below to simplify id comparision.
```
mawe42: accessed

mawe42: comparison

jjceresa:
>Otherwise the size of the union would depend on the initialization of its first member, which results in different sizes for different instances of the same union type.

Oh, thanks , I had forgotten for this ! 

Next is out of subject: Also, sorry for the absence of  "permanent link" , "quote" and  "formatting" message. Since today I cannot use these possibilities anymore (these are now disabled). Guys at GitHubSoldOffByMicrosoft now have complicated things. Here GitHub web page  was notified requesting for upgrading another recent web browser or OS !.

PR #461, opened 2018-11-09
Avoid useless dsp parameters update upon modulation.

jjceresa:
Actually on a single CC change, both 'fluid_voice_modulate()' and 'fluid_voice_modulate_all()' have the risk to call 'fluid_voice_update_param()'  several times for the same generator if several modulators have that generator as destination. 

Example:
-------------
  CCx --> mod 1 --> gen y
  CCx --> mod 2 -->Gen y
Both modulators (mod 1 and 2) have the same destination generator (gen y). On a simple CCx change, gen y will be updated twice (instead of only one).

This is not a bug, but only a risk of lost of performance particularly on intensive modulation situation. This PR avoid this risk.

### DIFF
--- a/src/synth/fluid_voice.c
+++ b/src/synth/fluid_voice.c
@@ -1161,18 +1161,23 @@ fluid_voice_update_param(fluid_voice_t *voice, int gen)
  * iteration of the audio cycle (which would probably be feasible if
  * the synth was made in silicon).
  *
- * The update is done in three steps:
+ * The update is done in four steps:
  *
- * - first, we look for all the modulators that have the changed
+ * - step 1: first, we look for all the modulators that have the changed
  * controller as a source. This will yield a list of generators that
  * will be changed because of the controller event.
  *
- * - For every changed generator, calculate its new value. This is the
- * sum of its original value plus the values of al the attached
+ * - step 2: For every changed generator, calculate its new value. This is the
+ * sum of its original value plus the values of all the attached
  * modulators.
  *
- * - For every changed generator, convert its value to the correct
- * unit of the corresponding DSP parameter
+ * - step 3: We need to avoid the risk to call 'fluid_voice_update_param' several
+ * times for the same generator if several modulators have that generator as
+ * destination. So every changed generators are registered in a list.
+
+ * - step 4: When finished with all the modulators, for every registered generator
+ * in the list, convert its value to the correct unit of the corresponding DSP
+ * parameter (fluid_voice_update_param()).
  */
 int fluid_voice_modulate(fluid_voice_t *voice, int cc, int ctrl)
 {
@@ -1180,6 +1185,8 @@ int fluid_voice_modulate(fluid_voice_t *voice, int cc, int ctrl)
     fluid_mod_t *mod;
     int gen;
     fluid_real_t modval;
+    int gen_changed[GEN_LAST]; /* registered list of changed generators */
+    int gen_count = 0;         /* count of generators in gen_changed */
 
     /*    printf("Chan=%d, CC=%d, Src=%d, Val=%d\n", voice->channel->channum, cc, ctrl, val); */
 
@@ -1209,12 +1216,27 @@ int fluid_voice_modulate(fluid_voice_t *voice, int cc, int ctrl)
             fluid_gen_set_mod(&voice->gen[gen], modval);
 
             /* step 3: now that we have the new value of the generator,
-             * recalculate the parameter values that are derived from the
-             * generator */
-            fluid_voice_update_param(voice, gen);
+               we register this generator only once in the list  */
+            for (k = 0; k < gen_count; k++)
+            {
+                if (gen_changed[k] == gen)
+                {
+                    break; /* gen already in the list */
+                }
+            }
+            if (k == gen_count)
+            {   /* registers gen as it isn't yet in the list */
+                gen_changed[gen_count++] = gen;
+            }
         }
     }
 
+    /* spep 4: for every registered generator in the list, convert its value to
+       the correct unit of the corresponding DSP parameter */
+    for (k = 0; k < gen_count; k++)
+    {
+        fluid_voice_update_param(voice, gen_changed[k]);
+    }
     return FLUID_OK;
 }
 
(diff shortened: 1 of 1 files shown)


mawe42:
Thinking more about this... maybe we could also do this with a different approach: compare the new generator amount to the previously set one and only call `fluid_voice_modulate` if the generator value has actually changed. That would not only help in the case where multiple modulators target the same generator in reaction to a single MIDI event, but also when consecutive MIDI events would result in the same generator value.

But again this would probably also be at the expense of the more common case that the generator value has actually changed.

jjceresa:
>I'm just wondering if we're optimising what I assume is the corner case (multiple modulators targeting the same generator) at the expense of the common case (only one modulator for a generator).

The modulators connection topology (ie one modulator to one generator or multiples modulators added to the summing input node of the same generator) is fully dependent of the soundfont designer .
I dont think that a SF2 synthesizer should paid attention to the coherency of this modulator topology (but checking modulators identity). Also the synthesizer should not cope with a modulator topology in favor of an another one. Also this will be the case if later fluidsynth will implement SF2 linked modulator.

Simply the aim of this PR is only to avoid multiple call to fluid_voice_update_param() apon a single CC change because useless fluid_voice_update_param() overload the dsp engine job. This is obvious to see after a close look to the inner code. So, i have not yet done performance analysis (before/after) the changes.

jjceresa:
> maybe we could also do this with a different approach: compare the new generator amount to the previously set one and only call fluid_voice_update_param() if the generator value has actually changed.

Your approach (i.e avoiding redundancy call to call fluid_voice_update_param())  is fine too. It works for actual version that use only unlinked modulator. Unfortunately it will not sufficient if later we implement linked modulators (sorry for the lack of explanations ). I am currently working on implementing linked modulators. If implementation of linked modulators is intended to be accepted, i suggest to keep the actual approach of this PR otherwise we could adopt yours.
Any opinions about the perspective of linked modulator's addition ?

mawe42:
> The modulators connection topology [...] is fully dependent of the soundfont designer .
[...] Also the synthesizer should not cope with a modulator topology in favor of an another one.

Yes, of course. But you are changing existing code and in the description you said that the motivation behind this change is performance. I'm not suggesting that it will have horrible effects. I actually like that the generator is not calculated multiple times anymore. I'm just saying performance optimisation without actual verification feels wrong. For the case where two modulators target the same generators, the performance gain seems obvious. What is not obvious is how the other case is impacted.

mawe42:
> Any opinions about the perspective of linked modulator's addition ?

My two cents: they are in the spec, so it would be great to support them!

jjceresa:
>My two cents: they are in the spec, so it would be great to support them!

I think too, because i know at least one soundfont editor (polyphone) that support linked modulators.
I am waiting for other feedback approbation and would propose a PR later; probably during the bad tempered days of the winter :).

src/synth/fluid_voice.c
```diff
@@ -1161,25 +1175,32 @@ fluid_voice_update_param(fluid_voice_t *voice, int gen)
...
+ * in the list, convert its value to the correct unit of the corresponding DSP
+ * parameter (fluid_voice_update_param()).
  */
 int fluid_voice_modulate(fluid_voice_t *voice, int cc, int ctrl)
 {
     int i, k;
     fluid_mod_t *mod;
     int gen;
     fluid_real_t modval;
+    int gen_changed[GEN_LAST]; /* registered list of changed generators */
```
derselbst:
```suggestion
    enum fluid_gen_type gen_changed[GEN_LAST]; /* registered list of changed generators */
```

Consider to specify the enum type explicitly. The `enum` could be one byte wide, while the `int` is 4 or 8 bytes, in which case we would waste many bytes on the stack.

jjceresa:
>I like the change in fluid_voice_modulate(), one minor note below.

Good point (thanks for your good eyes). I have accidentally resolved the conversation and cannot show it anymore. Due to recent stupid changes in github behaviour, interleaved conversation now cannot display correctly. Only top level conversation (like this message) still works !. I don't remember the enum name. Please recall me.

derselbst:
Conversation unresolved.

jjceresa:
>My concern in fluid_voice_modulate_all().....

Very good concern about the appropriate cpu's cache usage. At the time of the change,I had forgotten this. I was focusing only about the fact that the original implementation is making an inner loop inside an outer loop which leads to a number of iteration equal to pow(voice->mod_count,2).

Probably the best would be to keep the original implementation (improved by memorizing which generators were already updated). Also as the logic is pretty the same as for fluid_voice modulate() may be we could introduce a common function ?.

derselbst:
> Probably the best would be to keep the original implementation (improved by memorizing which generators were already updated). Also as the logic is pretty the same as for fluid_voice modulate() may be we could introduce a common function ?.

Sounds good.

jjceresa:
>To avoid hassle, you could just use uint32_t for the bitmask.

thanks, i don'like hassle.

>I don't understand why you wrap the bitmask in a struct. I don't see any alignment issues here.

Right.

derselbst:
### Profiling of this PR:

 Cpu loads(%) (sr: 44100 Hz, sp: 22.68 microsecond) and maximum voices

 nVoices| total(%)|voices(%)| reverb(%)|chorus(%)| voice(%)|estimated maxVoices
 -------|---------|---------|----------|---------|---------|-------------------
1833|   98.423|   98.423|     0.000|    0.000|    0.054|             1864


### Profiling of recent `master`:

Cpu loads(%) (sr: 44100 Hz, sp: 22.68 microsecond) and maximum voices

 nVoices| total(%)|voices(%)| reverb(%)|chorus(%)| voice(%)|estimated maxVoices
 -------|---------|---------|----------|---------|---------|-------------------
1833|   99.466|   99.466|     0.000|    0.000|    0.054|             1844

Slight performance increase, looks good!

jjceresa:
Thanks a lot , nice report. I guess that "profiling commands" was used during this test. However these commands produce noteon events but not CC events triggering modulation. Note that "Cpu load(%)" is only a time measurement of audio rendering. Is that  specifics "CC ->modulators" stimulus was used during this test ?

jjceresa:
Thanks a lot , nice report. I guess that "profiling commands" was used during this test. However these commands produce noteon events but not CC events triggering modulation. Note that "Cpu load(%)" is only a time measurement of audio rendering. Is that  specifics "CC ->modulators" stimulus was used during this test ?

PR #462, opened 2018-11-10
Float fixes

carlo-bramini:
I tried to check the generated ASM code when WITH_FLOAT macro is defined and when it is undefined.
In both cases I noticed that there are some points where the math number are converted from float to double and viceversa. Actually this happens whether the FPU is present or not, perhaps for granting the right precision in digits.
I also noticed few points that could be simplified a bit by using integer math, so I also included them.

### DIFF
--- a/src/rvoice/fluid_iir_filter.c
+++ b/src/rvoice/fluid_iir_filter.c
@@ -275,8 +275,8 @@ fluid_iir_filter_calculate_coefficients(fluid_iir_filter_t *iir_filter,
          * into account for both significant frequency relocation and for
          * bandwidth readjustment'. */
 
-        fluid_real_t omega = (fluid_real_t)(2.0 * M_PI *
-                                            (iir_filter->last_fres / ((float) output_rate)));
+        fluid_real_t omega = (fluid_real_t)(2.0 * M_PI) *
+                                            iir_filter->last_fres / output_rate;
         fluid_real_t sin_coeff = (fluid_real_t) sin(omega);
         fluid_real_t cos_coeff = (fluid_real_t) cos(omega);
         fluid_real_t alpha_coeff = sin_coeff / (2.0f * iir_filter->q_lin);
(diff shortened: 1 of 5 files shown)


src/rvoice/fluid_iir_filter.c
```diff
@@ -275,8 +275,8 @@ fluid_iir_filter_calculate_coefficients(fluid_iir_filter_t *iir_filter,
          * into account for both significant frequency relocation and for
          * bandwidth readjustment'. */
 
-        fluid_real_t omega = (fluid_real_t)(2.0 * M_PI *
-                                            (iir_filter->last_fres / ((float) output_rate)));
+        fluid_real_t omega = (fluid_real_t)(2.0 * M_PI) *
+                                            iir_filter->last_fres / output_rate;
```
derselbst:
```suggestion
                                            (iir_filter->last_fres / output_rate);
```

src/rvoice/fluid_rev.c
```diff
@@ -34,7 +34,7 @@
  * output.  There is a very small turn-on transient response, which should not
  * cause problems.
  */
-#define DC_OFFSET 1e-8
+#define DC_OFFSET (fluid_real_t)1e-8
```
derselbst:
```suggestion
#define DC_OFFSET ((fluid_real_t)1e-8)
```

src/rvoice/fluid_rvoice.h
```diff
@@ -39,7 +39,7 @@ typedef struct _fluid_rvoice_t fluid_rvoice_t;
  * 24 bits => 144-4 = 140 dB dynamic range => 1.e-7
  * 1.e-7 * 2 == 2.e-7 :)
  */
-#define FLUID_NOISE_FLOOR 2.e-7
+#define FLUID_NOISE_FLOOR (fluid_real_t)2.e-7
```
derselbst:
```suggestion
#define FLUID_NOISE_FLOOR ((fluid_real_t)2.e-7)
```

PR #463, opened 2018-11-11
Enable mixer threads

carlo-bramini:


carlo-bramini:
If I can say my opinion, I disagree. The threading support should be transparent to the user and an application should work in the same way, indipendently by the fact that it is executed with a library compiled in one way or another.

jjceresa:
>The threading support should be transparent to the user and an application should work in the same way, indipendently by the fact that it is executed with a library compiled in one way or another.

Yes of course. This is why perhaps it should be preferable that:
1) registering "synth.cpu-cores" setting should be done with an upper limit of 1 instead of 256 when enable-threads option is set to false. 
2)An other way could be not registering this setting when enable-threads option is set to false. 
Note that case (2) will mask to the application the multi-cores capability of the library.

jjceresa:
>The threading support should be transparent to the user and an application should work in the same way, indipendently by the fact that it is executed with a library compiled in one way or another.

yes of course, this is why perhaps "synth.cpu-cores" setting's upper limit should be 1 instead of 256 when enable-threads option is set to false.

derselbst:
>  perhaps "synth.cpu-cores" setting's upper limit should be 1 instead of
256 when enable-threads option is set to false

Agreed.

jjceresa:
Option name 'enable-threads' seems a bit missing meaning. I would vote for a more explicit name like 'enable-extra-threads'  or enable-cores-threads' or others name related to multi cpu-cores use. Any opinions  ?

derselbst:
I suggested a general naming like `enable-threads` or
`enable-threading` to allow this option not being limited to parallel
voice synthesis. It could be extended to other optional threading that
we may introduced in the future (though I can hardly think of any at
the moment).

If this cmake option really shall only control parallel voice
synthesis, one might question whether it's worth introducing it at
all.

jjceresa:
>I suggested a general naming like `enable-threads` or `enable-threading` to allow this option not being limited to parallel voice synthesis.

Ok, i see. So actual name is fine enough. In fact as the number of options is growing, it is not obvious to remember the meaning of each. May be cmake invocation offers a switch to display meaning text about all options ?.

derselbst:
> May be cmake invocation offers a switch to display meaning text about all options ?.

Try `cmake -LAH` or `ccmake`.

jjceresa:
>Try cmake -LAH or ccmake.

cmake -LAH works, thanks.

PR #464, opened 2018-11-15
Add Android audio drivers based on OpenSLES and Oboe

atsushieno:
( A related but very outdated PR on this topic https://sourceforge.net/p/fluidsynth/code-git/merge-requests/3/ )

This set of changes brings audio drivers for Android, either OpenSLES or [Oboe](https://github.com/google/Oboe). The changes in the original sources are kept minimal so that it should be easily maintained, as long as the build system works (which is actually somewhat complicated).

There are still couple of changes in the original tree, regarding custom sfloader for Android. Hopefully they are harmless to others who don't care about Android.

(I had a couple of other fixes when Android NDK had switched to clang, but they were already fixed in master, which thus reduced significant amount of my changes!)

I know there is another Oboe-based implementation by @VolcanoMobile, but their sources are quite different from master branch here and I see no open code, so I didn't care about any conflict (e.g. `android` directory collision).

See `android/README.Android.md` for more details.

(squashed merge recommended, there's a handful of experimental changes and reverts.)


derselbst:
Very nice, so far so good. Before going into a detailed review, I have a few issues:

1. Oboe uses opensles under hood, doesn't it? So why do we need an explicit opensles driver?

1. I understand that oboe only provides a C++ API. And I comprehend that you wrap it in a C wrapper library. However, this is unmaintainable for fluidsynth, as we would need to adjust both the wrapper AND the audio driver implementation whenever the API changes. Why not writing the audio driver in C++ and exporting the relevant audio driver functions as `extern "C"` to fluidsynth?

1. I have a similar concern with the Makefiles. fluidsynth is not responsible for maintaining Makefiles that build cerbero and oboe. oboe should be built by the user in a proper cross-compilation environment, so that fluidsynth can handle it using a pkg-config file, just like every other dependency library we need. With your proposed method a user is compelled to use your Makefiles in order to get oboe support by magically linking it in; rather than using cmake and querying for it at compile time.

1. I don't think `new_fluid_android_asset_sfloader()` should be added to fluidsynth's API. fluidsynth shouldn't be responsible for trying to create a sfloader that hopefully turns out to be useful on android.
Instead, this "asset sfloader" should be placed in the [API documentation](http://www.fluidsynth.org/api/) as code example describing a possible way of creating a suitable sfloader for android (along with the related README section). After all, the user's application is responsible for creating a suitable sfloader.

atsushieno:
Thanks for the quick reviews!

I'm all for making it maintainable. Take whatever useful into tree, and keep extra bits out. I can provide things from my repo under my maintenance too, if it turns out to be better.

> 1. Oboe uses opensles under hood, doesn't it? So why do we need an explicit opensles driver?

It is actually connected to the concern 2. We don't need to. However, as you saw, Oboe driver build is quite wacky and I think not all people whoever are interested in these Android drivers want to try to build Oboe. For such people, OpenSLES build is way simpler.

> 2. (...) Why not writing the audio driver in C++ and exporting the relevant audio driver functions as extern "C" to fluidsynth? 

There are multiple reasons why I ended up to write C wrapper, but in a short answer, yes it's doable. I was not sure if making build structure changes to also build C++ code is acceptable. Bringing it back to C++ is easy (it was actually written in C++ before https://github.com/atsushieno/fluidsynth-fork/commit/898eb9f68a0a6c9c94c38d1f89beb7730554daed#diff-7f2d970d7da1d49de720528d5c30fdc1 )

However, with related to that, I had to build two shared libraries: libfluidsynth.so and liboboe-c.so. It is actually my wishlist item to elimiante latter https://github.com/atsushieno/fluidsynth-midi-service-j/issues/10 , and I have somewhat longer explanation at https://dev.to/atsushieno/fluidsynth-hack-continued-oboe-driver-10fg ("gnustl and libc++") section, but in short, it is due to NDK migration from gnustl to libc++ and cerbero not moving away from gnustl.

Building liboboe.so is possible, but surprisingly?, Oboe does not offer shared library build, so we will still have to either write additional build script to build shared libs (like my changes do) or modify Oboe sources (I didn't want to). 

> 3. I have a similar concern with the Makefiles. fluidsynth is not responsible for maintaining Makefiles that build cerbero and oboe. (...)

I totally understand the concern and I also believe this should be improved, yet I had no idea on what is the best to achieve.

There is no such thing "pkg-config" in Android development world and thus Oboe itself does not offer oboe-1.0.pkgconfig(.in). I can "hack" that too (actually I once tried that too) but since Android NDK itself does not come up with `sysroot/usr/lib/pkgconfig`, PKG_CONFIG_PATH needs to be hackily customized anyways. There are couple of other fluidsynth driver dependencies that are not based on pkgconfig, I thought the proposed approach is the way to go.

On being a Makefile - Android build needs to run as many times as the number of the supported ABIs, and (to my entry-level understanding) Android CMake support is not designed to iterate them all. But maybe it can be changed.

> 4. I don't think new_fluid_android_asset_sfloader() should be added to fluidsynth's API. fluidsynth shouldn't be responsible for trying to create a sfloader that hopefully turns out to be useful on android.

Basically I'm for that, but I have some complicated feeling...

I still included this in the PR mostly because, building custom native code for Android is quite annoying task and no one wants to build their own custom code, and asset sfloader is probably almost mandatory to use.

But I also hate adding some API to the public surface, especially Android-specific ones. And JNI-related requirement code is ugly (I wrote about that in this blog post https://dev.to/atsushieno/fluidsynth-20x-for-android-4j6b , at "Custom SoundFont loader for Android Assets").

I can create another tiny repo that builds this custom sfloader so that not all people have to mess with the loader. But IF it seems almost mandatory, maybe it's better included in fluidsynth itself.

(Wow, it is going to be lengthy. Excuse me for that!)

atsushieno:
> I wouldn't mind. But we should convince cmake to only query for a C++ compiler if really needed. 

Alright, that's a good idea. It will be added when this PR goes back to C++ driver implementation.

> It somewhat surprises me that fluidsynth links gnustl. It's also doing this on my system, but I currently don't see a C++ dependency here.

I have no idea on which system you are on, but maybe some drivers depend on it?

For Android, it is an explicitly added dependency only to Android in cerbero: https://cgit.freedesktop.org/gstreamer/cerbero/tree/packages/base-system-1.0.package?id=49f47bf

> Why are you using cerbero for all builds? To me that seems like making your life unnecessarily complicated. Have you tried the procedure described in our wiki?

Actually I have been building glib etc. for many years (yes, I used to have scripts like that wiki page before too), and the build scripts always get broken every time NDK updates or build systems in dependencies change. Cerbero, the latest build system I found that respects Android, does not ignore ABIs that matter, respects pkg-config (compared to any other build scripts that tried to workaround it), and most importantly, developers maintain it. What I learned from that experience was that it is still better to let GNOME-related project maintainers to let the job and I make use of it.

But since it's not looking very maintainable to others, maybe this PR would also have better to rather hack a build script, as long as some people can fix that occasionally breakable bits...

It seems there is also CI build scripts for Android within glib and it may be a better option. I will have a look and find simpler solution.

> The NDK doesn't need to. But the libraries should. 

In Android world, pkg-config does not really exist, just like autotools has no presence. No library developers for Android respect pkg-config. Oboe is one of the best examples. It is just an external library that is not NDK itself, and never provides `liboboe-1.0.pc`. As earlier I wrote, I can hack it too, but that means `*.pc` file is written by non library maintainers, which is not great.

> We always need a way to make sure the library is present(...)

I'm all for that. I couldn't find a way to achieve that yet (various attempts to get `find_library()` etc. failed), but maybe because I really don't know cmake yet.

> I don't understand... you would only need to remove STATIC from this line to make it build a shared lib.

This is what I wrote, "modify Oboe sources". If I do so, it will bring in a lot of git merge failure annoyance on upgrades.

atsushieno:
Now the issues are getting clearer, let me list up action items (let me know if there are more):

- [ ] remove android asset related API from the public surface.
- [ ] switch to C++ implementation.
- [ ] hopefully remove gnustl dependency and build one single libfluidsynth.
- [ ] find simpler build scripts and hopefully remove Makefile.

I'm on travel and may need some longer time to fill them all, but will be back to code hopefully sooner.

derselbst:
The list is fine. Let's first concentrate on the build and oboe related issues and take care of the asset loader later. While you are on it: Pls. try to get rid of `CMAKE_POSITION_INDEPENDENT_CODE_OVERRIDE` and `CMAKE_C_VISIBILITY_OVERRIDE`. Not only do I consider them as hacks, they also encourage people to build and link a static version of fluidsynth into their likely proprietary apps, which [infringes the LGPL](https://github.com/FluidSynth/fluidsynth/wiki/LicensingFAQ). The goal should be to build fluidsynth as shared library.

> I have no idea on which system you are on, but maybe some drivers depend on it?

Jack does, indeed.

> the build scripts always get broken every time NDK updates

For the few builds I've done, gcc was to blame for the breakages. My hope is that with clang things become more stable. To me, cerbero doesn't seem suitable for our usecase (e.g. gnustl). Note that it says: "Cerbero build system used to build the official upstream GStreamer 1.0 SDK binaries" and not "Cerbero build system (ab-)used to cross-compile every other third party library for every platform available".

> It seems there is also CI build scripts for Android within glib and it may be a better option. I will have a look and find simpler solution.

We may setup a TravisCI job some day that cross compiles fluidsynth for Android. At least this would avoid having multiple build scripts and makefiles lying around that nobody really cares about.

> As earlier I wrote, I can hack it too, but that means *.pc file is written by non library maintainers, which is not great.

Perhaps it would be worth to propose a pkg-config file upstream? As well as to allow building a shared lib? I'm not intending to "pollute" the "android world" with pkg-config, it's just to make the cross compilation process easier. Else we need to find a solution that uses `find_library()`. Just ask if you need help.

> I'm on travel and may need some longer time to fill them all,

No need to hurry. I've scheduled this for 2.1 release, which I expect in Q2 2019.

carlo-bramini:
What is the minimum API level of this implementation?

carlo-bramini:
After a quick look, I would like to suggest to:
1) remove inclusion of `<sys/time.h>`, this is already handled with conditional define inside fluidsynth_priv.h
2) several structs like `loc_buffer_queue`,` format_pcm`, etc, are declared as C99. Probably, it is good anyways, even if not handled in the "old way", however I signal it.
3) Perhaps if would be worth to use FLUID functions/macros instead of `clock_gettime()` and `usleep()`.

derselbst:
@atsushieno Any news?

atsushieno:
I have been away from this since early December as it wouldn't be late if I come back like in Spring. I'm doing busy this month and won't be able to resume it soon.

IIRC the latest status was like:

- separate build for oboe C code is gone.
- Android specific code is gone from libfluidsynth.
- stuck at building android asset loader which needs to reference libfluidsynth (I am by no means CMake expert).

There is another development branch after this PR at https://github.com/atsushieno/fluidsynth-fork/tree/oboe-2

atsushieno:
I'm back to hacking world wit enough time. Before I resume bringing this back, I have a question to resolve to go ahead: how do your Android build scripts relevant to the actual request for changes? Those build scripts are based on Autotools, but you asked to do everything in CMake. Shouldn't those build scripts also be based on CMake, which will be completely different build script?

Are those build scripts CI-friendly? Should I really follow them, or discard everything and rewrite from scratch? If latter, the actual request for the changes are more like "set up entire Android build system", not just enabling audio drivers for Android, which is much beyond my intent for the changes (it may be still something I might be able to help, but I'm not necessarily super familiar with neither of CMake nor fluidsynth builds).

Depending on the answer(s), the actual changes I'll make will change e.g. (1) prebuilt glib and related bits as well as liboboe.a can be premises, (2) I don't have to work on writing an entire build system to maintain (that is actually the strong reason why I preferred cerbero build system), or (3) I contribute only implementation and drop the entire build script stuff so that anyone can write any build systems. (Or (4) something else.)

atsushieno:
Actually I found another level of the problem: glib does not contain `autogen.sh` or even `configure` anymore, so the existing scripts won't work at all. It had migrated to [meson](https://github.com/mesonbuild/meson).

derselbst:
> how do your Android build scripts relevant to the actual request for changes? Those build scripts are based on Autotools, but you asked to do everything in CMake. Shouldn't those build scripts also be based on CMake, which will be completely different build script?

Not quite sure if I understand you correctly. The android build scripts in our wiki are a leightweight example of how to build fluidsynth and its dependencies for one ABI.  The scripts use the offically supported build system of the respective projects. Any user is free to automate, change or integrate them into his/her build environment or use a completely custom approach (as you seem to prefer with cerbero).

> Are those build scripts CI-friendly?

I don't know what you consider as "CI-friendly". However I am confident to use those scripts e.g. on TravisCI to cross compile for Android.

> Should I really follow them, or discard everything and rewrite from scratch? If latter, the actual request for the changes are more like "set up entire Android build system", not just enabling audio drivers for Android, which is much beyond my intent for the changes (it may be still something I might be able to help, but I'm not necessarily super familiar with neither of CMake nor fluidsynth builds).

This PR shall not "set up an entire Android build system". It only should make sure that fluidsynth's cmake build system is able to find opensles and oboe when building for Android (and that it actually works and plays audio, ofc). This is a little tricky due to the lack of pkgconfig as we found earlier. If you need advice, feel free to ask. Our scripts shall give an idea of how to use fluidsynth's cmake build system to compile for android. You may follow them, but I'm also open for different (lightweight) approaches.

> Actually I found another level of the problem: glib does not contain autogen.sh or even configure anymore, so the existing scripts won't work at all. It had migrated to meson.

I don't see a problem here. The user chooses which version of glib to use and build.

atsushieno:
Alright, I took it as (a) it would just suffice if the changes to `CMakeLists.txt` assumes that there are packages that can be probed by `PKG_CONFIG_PATH`. Those who build libfluidsynth for Android can choose whatever build scripts they prefer e.g. the one on the fluidsynth wiki, or mine (in this changeset).

If users prefer my build script, there is `oboe-1.0.pc` which is based on local oboe builds for each ABI. It assumes liboboe.so which does not build in the official repo, so I changed the pointed URL to `atsushieno/oboe` from `google/oboe` as well as `atsushieno/cerbero` from the official freedesktop repo.

Also (b) there will not be official CI server for Android, so I don't worry about that. If anyone wants to set up a CI server, **they** are responsible to provide package path to glib, oboe, and so on. My cerbero-based `Makefile.android` is still in the changeset, but it is not necessary for users to run it. It should be still super helpful to Android app devs (I keep receiving inquiries on how to build fluidsynth for Android).

I believe I reflected all the requested changes i.e.

- unified C++ sources / no oboe-c dependency anymore. No liboboe-c.so anymore too.
- no Android-specific public API anymore.
  - Android specific code is under `android` directory, which is not part of libfluidsynth build anymore. It exists just as a "hint" for Android app devs.

Please let me know if we need more changes.

If everything looks good I would create another simplified PR - there are too many unnecessary commits in the PR.

derselbst:
Ok, looks good so far, thanks. Two issues:

1. That clang specific flag `-fsanitize-trap=undefined` [breaks the build](https://travis-ci.org/FluidSynth/fluidsynth/jobs/503110007#L993). Is it dispensable?
1. It should be ensured that cmake does not query for a C++ compiler by default.

I will try to build and test this PR in the days, by using both, your cerbero based build system and the scripts in our wiki (may take some time, maybe I'll need assistance). Once successful, I would take care of 2. and any missing documentation that I'll come across (wiki, elsewhere).

Feel free to provide a rebased version of this PR in a new PR. Else, I'd would inform you before squashing it.

atsushieno:
This is what happens without `-fsanitize-trap=undefined` (same as [this SO thread](https://stackoverflow.com/questions/45869368/android-nkd-r15c-and-ubsan-clang):

```
make[5]: Entering directory '/sources/fluidsynth-midi-service-j/external/fluidsynth/android/build/x86'
[ 96%] Linking CXX shared library libfluidsynth.so
/sources/fluidsynth-midi-service-j/external/fluidsynth/src/drivers/fluid_aufile.c:77: error: undefined reference to '__ubsan_handle_type_mismatch_v1'
/sources/fluidsynth-midi-service-j/external/fluidsynth/src/drivers/fluid_aufile.c:78: error: undefined reference to '__ubsan_handle_type_mismatch_v1'
/sources/fluidsynth-midi-service-j/external/fluidsynth/src/drivers/fluid_aufile.c:80: error: undefined reference to '__ubsan_handle_type_mismatch_v1'

<snip>

clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'm going to add another commit for this fix. On `project ( FluidSynth C CXX )` it seems that changing that part is going to mess the entire sequence of `CMakeLists.txt` - I would leave it to you. Thanks.

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+  
+  if (!dev->cont)
+    return DataCallbackResult::Stop;
+  
+  /* FIXME: it seems that fluidsynth stores results in full float
+   * value range, while Oboe expects the results between -1 and 1.
+   * To make things worse, fluidsynth callback only accepts float*
+   * (while non-callback accepts short*) which makes it unsure if
+   * the stream is processed appropriately.
+   * So far, disable custom callback for short.
```
derselbst: That callback mode used with `new_fluid_audio_driver2()` only supports floats as buffers. [See the documentation of  fluid_audio_func_t](http://www.fluidsynth.org/api/audio_8h.html#a4c108916c658973df5242ae77fefa3d0). A custom callback mode for short isn't supported.

src/drivers/fluid_opensles.c
```diff
@@ -0,0 +1,414 @@
...
+{
+  fluid_opensles_audio_driver_t* dev = (fluid_opensles_audio_driver_t*) p;
+
+  if (dev == NULL)
+    return;
+
+  dev->cont = 0;
+
+  if (dev->thread)
+    fluid_thread_join (dev->thread);
```
derselbst: You are lacking a call to `delete_fluid_thread()`.

src/drivers/fluid_opensles.c
```diff
@@ -0,0 +1,414 @@
...
+      dev->player_buffer_queue_interface, out_short, period_frames * sizeof (short) * NUM_CHANNELS);
+    }
+    if (result != 0)
+    {
+      err = result;
+      /* Do not simply break at just one single insufficient buffer. Go on. */
+    }
+  }	/* while (dev->cont) */
+
+  return NULL;
```
derselbst: return FLUID_THREAD_RETURN_VALUE;

src/drivers/fluid_opensles.c
```diff
@@ -0,0 +1,414 @@
...
+    }
+  }
+  else
+  {
+    if (dev->callback)
+    {
+      short_callback_buffers [0] = out_short;
+      short_callback_buffers [1] = out_short;
+
+      (*dev->callback)(dev->synth, period_frames, 0, NULL, 2, (float**) short_callback_buffers);
```
derselbst: Same as with oboe: shorts aren't supported for custom callbacks. You shouldn't attempt to support it.

src/drivers/fluid_opensles.c
```diff
@@ -0,0 +1,414 @@
...
+void delete_fluid_opensles_audio_driver(fluid_audio_driver_t* p);
+void fluid_opensles_audio_driver_settings(fluid_settings_t* settings);
+static fluid_thread_return_t fluid_opensles_audio_run(void* d);
+static void opensles_callback(SLAndroidSimpleBufferQueueItf caller, void *pContext);
+void adjust_latency(fluid_opensles_audio_driver_t* dev);
+void process_fluid_buffer(fluid_opensles_audio_driver_t* dev);
+
+void fluid_opensles_audio_driver_settings(fluid_settings_t* settings)
+{
+  fluid_settings_register_int(settings, "audio.opensles.use-callback-mode", 1, 0, 1,
```
derselbst: What is your motivation to provide a "non-callback" mode, given the workaround you are attempting in the `adjust_latency` function?

derselbst:
Apart from that I've tested your changes and they compile fine with the build scripts from our wiki. I had no luck with cerbero though: It complained that my distribution is unsupported. When trying to convince it that it's just a usual opensuse it still complained that it doesn't know how to bootstrap. Yet, it asked me for root privileges, which I am not willing to grant.

atsushieno:
That is the reason why I preferred to have CI servers :-)

Thanks for the other bunch of code review comments. I will fix them soon (but not likely in a day or two).

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+  if (dev == NULL) {
+    FLUID_LOG(FLUID_ERR, "Out of memory");
+    return NULL;
+  }
+
+  FLUID_MEMSET(dev, 0, sizeof(fluid_oboe_audio_driver_t));
+  
+  dev->synth = synth;
+  dev->callback = func;
+  dev->oboe_callback = new OboeAudioStreamCallback (dev);
```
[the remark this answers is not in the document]
atsushieno: fixed, I was fairly newbie to find this new thing...!

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+  return new_fluid_oboe_audio_driver2 (settings,
+                                        (fluid_audio_func_t) fluid_synth_process,
+                                        (void*) synth);
+}
+
+/*
+ * new_fluid_oboe_audio_driver2
+ */
+fluid_audio_driver_t*
+new_fluid_oboe_audio_driver2(fluid_settings_t* settings, fluid_audio_func_t func, void* data)
```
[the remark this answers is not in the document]
atsushieno: This too, thanks for the hint!

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+  if (dev == NULL) {
+    return;
+  }
+
+  dev->cont = 0;
+  
+  dev->stream->stop ();
+  
+  dev->stream->close ();
```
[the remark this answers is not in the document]
atsushieno: ditto

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+  fluid_oboe_audio_driver_t *dev;
+  OboeAudioStreamCallback *oboe_callback;
+  
+  oboe_callback = (OboeAudioStreamCallback*) callback;
+  dev = (fluid_oboe_audio_driver_t*) oboe_callback->user_data;
+  
+  if (!dev->cont)
+    return DataCallbackResult::Stop;
+  
+  /* FIXME: it seems that fluidsynth stores results in full float
```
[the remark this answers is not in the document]
atsushieno: I seem to have left some old comment. It fills the buffer as expected in the latest code.

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+  fluid_audio_driver_t driver;
+  fluid_synth_t *synth;
+  int32_t cont;
+  fluid_audio_func_t callback;
+  OboeAudioStreamCallback *oboe_callback;
+  AudioStream *stream;
+} fluid_oboe_audio_driver_t;
+
+static fluid_thread_return_t fluid_oboe_audio_run(void* d);
+static fluid_thread_return_t fluid_oboe_audio_run2(void* d);
```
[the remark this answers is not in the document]
atsushieno: Yes, removed.

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+  OboeAudioStreamCallback (void *userData)
+    : user_data (userData)
+  {
+  }
+
+  void *user_data;
+
+  DataCallbackResult onAudioReady (AudioStream *oboeStream, void *audioData, int32_t numFrames)
+  {
+    return on_audio_ready (this, oboeStream, audioData, numFrames);
```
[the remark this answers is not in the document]
atsushieno: It was leftover from "liboboe-c.so" build in my first "consolidate C API" approach. Moved into the class itself now.

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,255 @@
...
+#include "fluid_settings.h"
+
+} // extern "C"
+
+#if OBOE_SUPPORT
+
+#include <sys/time.h>
+#include <oboe/Oboe.h>
+
+extern "C" {
```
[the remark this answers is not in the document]
atsushieno: this too is a leftover from liboboe-c buid. Removing.

src/drivers/fluid_oboe.cpp
atsushieno: Thanks, good to know.

src/drivers/fluid_opensles.c
atsushieno: Actually I'm removing the entire non-callback code (mentioned in another comment).

src/drivers/fluid_opensles.c
atsushieno: Alright. The entire function is being removed anyways.

src/drivers/fluid_opensles.c
atsushieno: I reviewed the entire things and figured out that I don't really need this one, so removed the entire "non-callback" related code. That resulted in lots of removal of code. (more comment at 5034a57)

CMakeLists.txt
```diff
@@ -598,16 +605,34 @@ else ( enable-midishare )
   unset ( MidiShare_LIBS CACHE )
 endif ( enable-midishare )
 
+unset ( OPENSLES_SUPPORT CACHE )
+unset ( OpenSLES_LIBS CACHE )
+if ( enable-opensles )
+  check_include_file ( SLES/OpenSLES.h OPENSLES_SUPPORT )
+  if ( OPENSLES_SUPPORT )
+    find_library ( OpenSLES_LIBS OpenSLES )
```
derselbst: Sry about 89b96a616e08364efeb495c3feb37982d53ec4b4. I was concerned that opensles support only depends on whether it's headers were found. But if the library isn't found it would still be activated. Suggesting to change it as follows:

```suggestion
    find_library ( OpenSLES_LIBS OpenSLES )
    if ( NOT OpenSLES_LIBS )
      unset ( OPENSLES_SUPPORT )
    endif ( OpenSLES_LIBS )
```

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+#include "fluid_synth.h"
+#include "fluid_adriver.h"
+#include "fluid_settings.h"
+
+} // extern "C"
+
+#if OBOE_SUPPORT
+
+#include <sys/time.h>
```
derselbst: This include seems to be redundant now.

src/drivers/fluid_opensles.c
```diff
@@ -0,0 +1,285 @@
...
+ *
+ */
+
+#include "fluid_synth.h"
+#include "fluid_adriver.h"
+#include "fluid_settings.h"
+
+#if OPENSLES_SUPPORT
+
+#include <sys/time.h>
```
derselbst: Same.

src/drivers/fluid_oboe.cpp
```diff
@@ -0,0 +1,218 @@
...
+  dev->oboe_callback = new (std::nothrow) OboeAudioStreamCallback(dev);
+  if (!dev->oboe_callback) {
+    FLUID_LOG(FLUID_ERR, "Out of memory");
+    goto error_recovery;
+  }
+
+  fluid_settings_getint(settings, "audio.period-size", &period_frames);
+  fluid_settings_getnum(settings, "synth.sample-rate", &sample_rate);
+  is_sample_format_float = fluid_settings_str_equal (settings, "audio.sample-format", "float");
+  fluid_settings_getint(settings, "audio.oboe.device-id", &device_id);
```
derselbst: When going through the other settings I realized that `audio.oboe.device-id` might be a little inconsistent. Usually we have settings like `audio.NAME.device` as string. Since oboe receives an integer I'd suggest to rename it to `audio.oboe.id`. I know we have `audio.jack.id` being already inconsistent as it receives a string, but `audio.oboe.id` seems to be the better choice in my opinion.

atsushieno: Alright, if you say so :-)

src/drivers/fluid_oboe.cpp
atsushieno: fixed

src/drivers/fluid_opensles.c
atsushieno: ditto

CMakeLists.txt
atsushieno: indeed, thanks for the suggestion

derselbst:
I made cmake to only query for C++ if `-Denable-oboe=1`. Also, all those linker errors when building with ubsan occurred due to the missing `-fsanitize` flag when linking. This should be fixed as well. However, ubsan runtime lib requires NDK r17 or newer. This should be fine though.

The fluid_opensles.c source isn't strictly C90 complaint. But since usually clang or gcc will be used to compile this and as long as nobody complains, this should be fine as well.

I have a slight preference to move the `android/` folder into the `doc/` folder. Let me know if would cause too much trouble for you.

Apart from that I consider this to be ready to merge. In any case, please give it a last test and let me know whether you want to rebase the commit history. If you don't want to, I would squash it. In every case: before I press the merge button, I will format the code to match our style guide.

atsushieno:
Thanks for a bunch of code cleanup.

> I have a slight preference to move the android/ folder into the doc/ folder. Let me know if would cause too much trouble for you.

No, it is totally fine. If those android-specific code is awkward or in the way, I can publish them somewhere outside. I also have some work-in-progress code for new Android Q AMIDI that currently stays near it (also relocatable).

atsushieno:
And forgot to mention, I'm okay with either style for rebasing or squashing. Thanks.

derselbst:
> If those android-specific code is awkward or in the way, I can publish them somewhere outside.

Not at all. The doc/ directory is absolutely fine.

> And forgot to mention, I'm okay with either style for rebasing or squashing.

Ok then. I will squash this in 2 to 3 days, having some time for a final test run.

---

This really is a great feature for fluidsynth, thank you very much! Looking forward to Q AMIDI!

atsushieno:
I think e7496a9 has caused some regression:

```
2019-03-26 19:57:26.514 2683-2683/name.atsushieno.fluidsynthmidideviceservicej E/AndroidRuntime: FATAL EXCEPTION: main
    Process: name.atsushieno.fluidsynthmidideviceservicej, PID: 2683
    java.lang.UnsatisfiedLinkError: Unable to load library 'fluidsynth':
    dlopen failed: library "libclang_rt.ubsan_standalone-i686-android.so" not found
    dlopen failed: library "libclang_rt.ubsan_standalone-i686-android.so" not found
    dlopen failed: library "libclang_rt.ubsan_standalone-i686-android.so" not found
```

`libclang_rt.ubsan_standalone-i686-android.so` exists but it is like 2.5MB of shared library that we app developers wouldn't like to add as extra dependency. Could we bring back old ubsan related build options?

atsushieno:
This actually fixed it (but probably you want different solution).

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 240ab1d..1cf7dc8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,13 @@ if ( CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang" OR CMAKE_C_
   
 endif ( CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang" OR CMAKE_C_COMPILER_ID STREQUAL "Intel" )
 
+# Options for CLANG only
+if ( CMAKE_C_COMPILER_ID STREQUAL "Clang" )
+  set ( CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fsanitize-trap=undefined" )
+  set ( CMAKE_SHARED_LINKER_FLAGS_DEBUG "${CMAKE_SHARED_LINKER_FLAGS_DEBUG} -fsanitize-trap=undefined" )
+endif ( CMAKE_C_COMPILER_ID STREQUAL "Clang" )
+
+
 # Windows
 unset ( WINDOWS_SUPPORT CACHE )
 unset ( WINDOWS_LIBS CACHE )
```

derselbst:
> libclang_rt.ubsan_standalone-i686-android.so exists but it is like 2.5MB of shared library that we app developers wouldn't like to add as extra dependency.

This would only affect debug builds, which app developers probably wouldn't deploy.

But anyway, since ubsan is currently only useful for me, let's get rid of this hack and add an option to enable it explicitly on demand. Doing so will also compel you to provide it at runtime: edb8707668dbe8749cb167896abcf47d266491ca

atsushieno:
Okay, I confirmed that the latest changes worked for me.

derselbst:
Great, thanks!

PR #465, opened 2018-11-16
Avoiding illegal CC modulation

jjceresa:
Some CC aren't intended to modulate (mainly  CC lsb).
Even these CC aren't found as modulator's source by fluid_voice_modulate(), it worth to avoid useless call to fluid_voice_modulate().

Also, for now this prevent unwanted modulation effect when illegal modulators are in the soundfont.
Please note that illegal modulators in soundfont will be fixed soon in another PR.

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -1312,7 +1312,7 @@ fluid_synth_damp_voices_by_sostenuto_LOCAL(fluid_synth_t *synth, int chan)
  * @param mod Modulator info (values copied, passed in object can be freed immediately afterwards)
  * @param mode Determines how to handle an existing identical modulator (#fluid_synth_add_mod)
  * @return #FLUID_OK on success, #FLUID_FAILED otherwise
- * 
+ *
  * @note Not realtime safe (due to internal memory allocation) and therefore should not be called
  * from synthesis context at the risk of stalling audio output.
  */
@@ -1382,7 +1382,7 @@ fluid_synth_add_default_mod(fluid_synth_t *synth, const fluid_mod_t *mod, int mo
  * @param synth synth instance
  * @param mod The modulator to remove
  * @return #FLUID_OK if a matching modulator was found and successfully removed, #FLUID_FAILED otherwise
- * 
+ *
  * @note Not realtime safe (due to internal memory allocation) and therefore should not be called
  * from synthesis context at the risk of stalling audio output.
  */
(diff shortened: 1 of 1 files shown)


jjceresa:
>To me this sounds like those contributions should be done automatically, e.g. if CC1 changes it should take CC33 into account during modulation. 

Right, this is what MIDI specs says .CC LSB (32 to CC 63) are LSB contributions of CC MSB (0 to 31). It is up to the synthesizer to decide to take account of this lsb value or not. It is true that actually Fluidsynth (FS) doesn't use CC lsb value  inside fluid_fluid_update_param() once fluid_voice_modulate has been triggered. This is because actually FS needs only 7 bits resolution (and not 14 bits) from these CCs. So FS is using only 7 bit MSB (but portamento time). In regard to MIDI specs Fluidsynth behaves correctly. 

To deal correctly with theses CC (regardless any synth will use CC MSB alone(7 bit) or both CC MSB,LSB(14 bits)  during synthesis), the SF specifs recommends  to use only CC MSB(not CC LSB) in modulators's sources to trigger modulation (i.e modulators with CC LSB connected to sources inputs should be ignored).
This  specifs is good particularly for synth that use 14bits. In this case , the MIDI transmitter sends CC LSB first followed by CC MSB. The MIDI synth receives both CC LSB and CC MSB but only CC MSB will trigger the modulation. This will produce correct synthesis parameters from a correct 14 bit CC. If in SF specifs, modulators'sources with CC LSB had been accepted, both CC LSB and CC MSB will triggers 2 modulations. This leads to ramdon synthesis parameters followed by a correct synthesis parameters.

As FS will uses only CC 7 bits resolution, it is safe to ignore this PR.

>Also I have a personal reason: I have a couple of MIDIs that use CC33 for FilterQ and CC34 for FilterFC ....

I understand fully your needs and don't want to break anythings here. 

Be aware that CC33 and CC34 are not recommended in SF file (assuming that Soundfont editor will accept CC LSB for modulators'input).

Please why did you choose these CC LSB   instead of any general purpose CC MSB (16-19 or 80-83) ?

derselbst:
Ok, thanks for your explanations.

> Please why did you choose these CC LSB instead of any general purpose CC MSB (16-19 or 80-83) ?

I didn't. Those are MIDIs extracted from video games, their developers are to blame :)

jjceresa:
In https://github.com/FluidSynth/fluidsynth/pull/465/commits/a3571c1811cce86081e4f9681f39199703e3bae7 CC LSB are now allowed to modulate (but not DATA_ENTRY_LSB). So now this behaves as before (apart DATA_ENTRY_LSB). Comments have been added.

derselbst:
That's fine, thanks!

PR #467, opened 2018-11-20
Add modulators checking at soundfont loading time

jjceresa:
Actually some basic modulators check are done at noteon time (in fluid_voice_add_mod()). That means that we know if a modulator (modx) is invalid only when a MIDI noteon is received and only for a preset modx belongs to. This is not appropriate.

This PR moves the modulator checking at soundfont loading time. Enhancements are:
1) A better verbose modulator integrity check, for any soudnfont loaded at appropriate time.
1.1) All modulators are checked (preset zone (local/global), instrument zone (local/global).
1.2.1) Modulators check are enforced to source src1 and src2  (for non-CC and CC sources) (following SF specs (except for CC LSB) ( see comment in fluid_synth_cc_LOCAL()).
Modulators CC sources checking is coherent with the actual behaviour in fluid_synth_cc_LOCAL() in regard of modulation triggering.
1.2.2)Also, identic modulator in the same zone are detected. 
1.2.3)Any invalid modulator(sources invalid, or modulator identic) is removed at loading time with a warning message displaying the cause and name of the modulators.

2) This fix a bug in noteon, in the case of identic modulators in global preset zone.
Assuming 2 identics modulator (m1 and m2) in a preset global zone, the actual noteon doesn't check this case. (the actual code detect identic modulator in all others zones (instrument (local or global), preset(local)) but not preset global).

3) NoteOn is faster.
2.1)There is no more modulators checks at noteon making this more efficient.
2.2 As there are no identic modulator in the same zone, there is no more identity modulator check (i.e local zone against local zone), (i.e global zone against global zone). This result in a faster code and the bug described in (2) is gone.

4) Modulators sources checking as been added in API functions fluid_synth_add_default_mod() and fluid_voice_add_mod(). Please note that actually fluid_voice_add_mod() has no return code, but this could be perhaps changed ?


src/synth/fluid_mod.c
```diff
@@ -486,6 +486,131 @@ size_t fluid_mod_sizeof()
...
+                && ((src < NRPN_LSB) || (RPN_MSB < src))
+                /* is src not ALL_SOUND_OFF, ALL_CTRL_OFF, LOCAL_CONTROL, ALL_NOTES_OFF ? */
+                /* is src not OMNI_OFF, OMNI_ON, POLY_OFF, POLY_ON ? */
+                && (src < ALL_SOUND_OFF)
+                /* CC lsb shouldn't allowed to modulate (spec SF 2.01 - 8.2.1) 
+                   However, as long fluidsynth will use only CC 7 bits resolution,
+                   it is safe to ignore these SF recommendations on CC receive.
+                   See explanations in fluid_synth_cc_LOCAL() */
+                   /* uncomment next line to forbid CC lsb  */
+	           	/* && ((src < 32) || (63 < src)) /* src isn't CC lsb */
```
derselbst: Compiler complains "comment within a comment". Pls remove the second `/*`.

src/sfloader/fluid_defsfont.c
```diff
@@ -1186,6 +1198,77 @@ static int fluid_preset_zone_create_voice_zones(fluid_preset_zone_t *preset_zone
...
+        {  /* the modulator is useless so we remove it */
+            if (prev_mod)
+            {
+                prev_mod->next =next;
+            }
+            else
+            {
+                *list_mod = next;
+            }
+            FLUID_FREE(mod); /* freeing */
```
derselbst:
```suggestion
            delete_fluid_mod(mod); /* freeing */
```

src/sfloader/fluid_defsfont.c
```diff
@@ -682,6 +682,118 @@ fluid_defpreset_next(fluid_defpreset_t *defpreset)
...
+                mod_list[i] = NULL;
+            }
+        }
+
+        /* Finally add the new modulator to to the list. */
+        mod_list[mod_list_count++] = local_mod;
+        local_mod = local_mod->next;
+    }
+
+    /* Step 2: global + local modulators are added to the voice unsing mode. */
```
derselbst: `unsing`?

jjceresa:
>Unfortunately this function doesn't return an int. But this shouldn't matter too much, because if the user adds an invalid modulator, the current implementation would ignore it anyway on a noteon, wouldn't it?

Yes, actually current implementation fluid_voice_add_mod() has no return code. For now, this PR reproduces same behaviour. However, assuming this PR accepted, maybe future implementation of this function API could return an int (FLUID_OK, FLUID_FAILED) ?. Of course doing this will introduce an API change.

src/sfloader/fluid_defsfont.c
```diff
@@ -682,6 +682,118 @@ fluid_defpreset_next(fluid_defpreset_t *defpreset)
...
+ * Preset zone list (local/global) must be added using FLUID_VOICE_ADD.
+ * 
+ * @param voice voice instance.
+ * @param global_mod global list of modulators.
+ * @param local_mod local list of modulators.
+ * @param mode Determines how to handle an existing identical modulator.
+ *   #FLUID_VOICE_ADD to add (offset) the modulator amounts,
+ *   #FLUID_VOICE_OVERWRITE to replace the modulator,
+*/
+void static
```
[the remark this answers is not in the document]
derselbst: There are more occurrences, pls see the provided build log.

src/synth/fluid_voice.c
```diff
@@ -1486,32 +1486,42 @@ fluid_voice_stop(fluid_voice_t *voice)
...
-     * Some soundfonts come with a huge number of non-standard
-     * controllers, because they have been designed for one particular
-     * sound card.  Discard them, maybe print a warning.
-     */
+/**
+ * Adds a modulator to the voice.
+ * local version of fluid_voice_add_mod function. Called at noteon time.
+ * @param voice, mod, mode, same as for fluid_voice_add_mod() (see above).
+ * @param check_limit_count is the modulator number limit to handle with existing
+ *   identical modulator(for mode FLUID_VOICE_OVERWRITE or FLUID_VOICE_OVERWRITE only).
```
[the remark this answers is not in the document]
derselbst: I meant why are you listing `FLUID_VOICE_OVERWRITE` twice?

derselbst:
> However, assuming this PR accepted, maybe future implementation of this function API could return an int

Yes. But since the current implementation ignores it anyway, I currently consider this nice-to-have.

jjceresa:
Sorry for commit pollution. Commits 5659cc5 and ab9c8a2 are useless. These are due to my local faulty command "git commit --amend". :(

jjceresa:
>Just let me know when you think it's ready.

I am fine with this PR. I consider this ready.

>I would schedule this for 2.1 if you don't mind?

Ok. For information, i am currently working on adding `linked modulators` which will be based on the result of this PR. In short, this will involve:
- linked modulators checking at soundfont loading time.
- taking these new linked modulators in the voice at noteon time (using same or similar methods that for actual unlinked modulators).
- taking account of linked modulators inside voice_modulate() and voice_get_lower_boundary_for_attenuation().
I would prefer waiting for this PR merged before starting a new branch for linked modulators.

derselbst:
Ok. I'll try to test and merge this within the next 2 or 3 days.

jjceresa:
>I'll try to test and merge this within the next 2 or 3 days.

Please, no need to hurry.

derselbst:
Works well for me. Thanks!

PR #468, opened 2018-11-22
Publish MIDI event constants and add functions to get text/lyrics

RangelReale:
* Move MIDI event definitions to public include
* Functions to get text/lyrics text

The MIDI events are useful when using the library with a player callback, otherwise every user would need to copy all the events to its our source every time.


derselbst:
Please explain what you are trying to do in the player callback or provide
usage example(s).

derselbst:
In this case it should be sufficient to only export `enum midi_meta_event`. Why do suggest to make all the other enums public?

RangelReale:
My thought was that all these constants have the possibility of being used on a player callback, like for example displaying the notes. As all this data can be extracted from the "fluid_midi_event_t", like for example "fluid_midi_event_get_control", I think the constants can be public, as using this method would't be useful if you don't know the values for the event control.

Maybe 'fluid_midi_event_type', 'fluid_midi_control_change' and 'midi_meta_event' should be enough.

derselbst:
> Maybe 'fluid_midi_event_type', 'fluid_midi_control_change' and 'midi_meta_event' should be enough.

Ok. I can't decide this by myself though. Discussion pending:

https://lists.nongnu.org/archive/html/fluid-dev/2018-11/msg00027.html

derselbst:
I will add the two getters you suggested. But currently I can't publish the enums, sry.

derselbst:
Perhaps as a compromise I've made the getters typesafe, so that you wouldn't need to check the type explicitly for text and lyric events.

PR #470, opened 2018-11-23
Implement the SNDIO driver (upstreamed FreeBSD port patches)

yurivict:
I only moved the patches into the pull request from the port.
Somebody implemented it before, but it was never upstreamed.


yurivict:
Items 1-3 are caused by inadvertent mixing in of BSD patches, I'll update the patch.
For item 4, I contacted Jacob Meuser and asked him to re-license.

yurivict:
I updated the patch that is supposed to fix items 1-3.

t6:
> I updated the patch that is supposed to fix items 1-3.

The CI builds still fail. Let's either
1. disable the sndio driver by default
2. properly detect libsndio and sndio.h in CMakeLists.txt (make the `enable-sndio` option description not a lie)

> Most notably: The implementation of the sndio driver uses a custom license rather than the LGPL.

It's not a custom license.  It's the [ISC License](http://www.gnu.org/licenses/license-list.html#ISC) (OpenBSD's preferred license).  Why would this block upstreaming `fluid_sndio.c`? This was not a problem for other LGPL licensed projects such as [jack1](https://github.com/jackaudio/jack1/blob/master/drivers/sndio/sndio_driver.c) or [SDL 1.2](http://hg.libsdl.org/SDL/file/67aeabf947b2/src/audio/sndio/SDL_sndioaudio.c).

derselbst:
I am no lawyer. But as far as I know, this license is controversial due to its wording:

> Permission to use, copy, modify, and distribute

whereas the current version of this license states:

> Permission to use, copy, modify, and/or distribute

Jacob Meuser has chosen the first wording. Again, I'm no laywer, but this seems to make some juridical difference. Also note that the link you've provided says:

> [...] However, to help make sure this language cannot cause any trouble in the future, we encourage developers to choose a different license for their own works.

As fluidsynth is LGPL licensed, contributions should be made under this license. I can't speak for other projects, but I don't want fluidsynth to become a license jungle, and ultimately a subject to potential legal offenses. We already had a similar issue with the `fluid_chorus.c` implementation. If there is an urgent and comprehensible reason stated _by the author_ not to use LGPL, I might accept a compatible license instead... however, not the ISC.

> Let's [...] properly detect libsndio and sndio.h in CMakeLists.txt (make the enable-sndio option description not a lie)

I second this.

yurivict:
I Jacob Meuser doesn't answer I would have to take this patch back to FreeBSD port.

derselbst:
Closed due to incompatible license.

jobbautista9:
FYI I am using this PR as a base for patching sndio support into Debian's packaging of fluidsynth. For those interested, please see [bug 1011425](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1011425). I will try to maintain this as long as I can.

fabiangreffrath:
To be honest, I don't quite understand why adding a source file under a license that is way more liberal than the LGPL itself should cause any issues. 🤷

PR #482, opened 2018-12-04
Fourcc support

carlo-bramini:
I was wondering what you think about having direct comparison to the fourcc code rather than searching all times with function chunkid() if the code is right or not. I do not know if there was a particular need for doing that. I also imported two functions from utils/fluid_sys.c, fluid_is_midifile() and fluid_is_soundfont(), so that they could also use it.

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -82,6 +82,37 @@ static int fluid_midi_file_get_division(fluid_midi_file *midifile);
  */
 
 /**
+ * Check if a file is a MIDI file.
+ * @param filename Path to the file to check
+ * @return TRUE if it could be a MIDI file, FALSE otherwise
+ *
+ * The current implementation only checks for the "MThd" header in the file.
+ * It is useful only to distinguish between SoundFont and MIDI files.
+ */
+int fluid_is_midifile(const char *filename)
+{
+    FILE    *fp = FLUID_FOPEN(filename, "rb");
+    uint32_t id;
+    int      retcode = 0;
+
+    do
+    {
+        if(fp == NULL)
+            break;
+
+        if(FLUID_FREAD(&id, sizeof(id), 1, fp) != 1)
+            break;
+
+        retcode = (id == FLUID_FOURCC('M','T','h','d'));
+    }
+    while (0);
+
+    FLUID_FCLOSE(fp);
+
+    return retcode;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


derselbst:
I'm struggling with this change. Yes, the original version searches
through the chunkids all the time. Although I consider it to be
inexpensive I understand that searching them them seems redundant.
Unfortunately your proposal doesn't eliminate the search completely.
Right now, if you ask me, I tend to consider it as premature optimization.

carlo-bramini:
No problem, we can keep this change in standby and eventually evaluate it again in the future. Actually, I also wanted to tell that the function chunkid() is tipically expanded inline when the optimizer is enabled, so the generated code is a bit bigger in this way.

mawe42:
I like this change, not because of the possible performance improvements, but because it makes the code easier to read and modify. For example, in the original code the enum definition and the strings in the idlist need to be in exactly the same order, otherwise stuff breaks. And this is not immediately clear when reading the code... at least it wasn't for me. The new code defines the IDs and then adds them to a list. Much easier to understand and easier to maintain. And it also removes the need for the "union alignment trick".

For more performance benefits, we could even consider reordering the FCC according to the frequency that the ID can appear in a Soundfont. The gains will probably be very minimal, though.

derselbst:
While it removes the alignment trick, the union trick is still there, in `process_info()`

> it makes the code easier to read and modify

Ok. Who could test this on a big endian machine? (I.e. execute `make check`)

> For more performance benefits, we could even consider reordering the FCC according to the frequency that the ID can appear in a Soundfont.

Too premature.

carlo-bramini:
> While it removes the alignment trick, the union trick is still there, in `process_info()`

I could do it also with an "L-value" assignment, but in this way it looked cleaner, at least to me. Afterall, the `union` statement exists for this purpose.

derselbst:
I don't mind using unions.

mawe42:
> Ok. Who could test this on a big endian machine? (I.e. execute make check)

I've tested this branch on a big-endian PowerPC emulation (via qemu) with Debian wheezy. ```make check``` returns success and a manual fast render with the Fluid3R_GM.sf2 soundfont worked fine.

derselbst:
> I've tested this branch on a big-endian PowerPC emulation (via qemu) with Debian wheezy.

Good idea, thanks for your effort!

PR #485, opened 2018-12-15
Only retrieve modification time once and remove warning if failed

### LINKED ISSUE #483: Bogus warning when loading soundfont from memory: Unable to read modificaton time of soundfont file
FluidSynth 2.0.2

When using a fluid_synth_t with a soundfont loaded from memory (see http://www.fluidsynth.org/api/fluidsynth_sfload_mem_8c-example.html), this is printed to stderr:

`fluidsynth: warning: Unable to read modificaton time of soundfont file.`

Fluidsynth does not have a way to distinguish between loading a soundfont from a file and loading one from memory. Perhaps it would have been better to have a different function for loading soundfonts from memory rather than using `fluid_synth_sfload()` for both? This would either allow fluidsynth to internally mark the soundfont as being in memory, or the new function could define a new callback function for this purpose (`fluid_sfloader_callback_stat_t()`). Or a callback that returns a `FILE*` handle, which is `NULL` if there's no file handle associated with the soundfont.

For now, I tried to kludge away the warning by making the filename string appear to be empty:

```
char bogus_filename[64];
bogus_filename[0] = 0;
sprintf(bogus_filename + 1, "%p", ptr);
fluid_synth_sfload(synth, bogus_filename, 1);
```

but fluidsynth still tries to do file access on it and prints the warning. Perhaps it would be better if fluidsynth would not attempt any file access and don't print a warning about it if the string appears to be empty.

mawe42:
This change removes the warning message if retrieving the soundfont file modification time fails in the sample cache loader. The warning made sense while we didn't have mtime support on all platforms, but after switching to GLibs g_stat, it's no longer needed.

To reduce the number of calls to fluid_get_file_modification_time and also get rid of a possible race condition, getting the mtime has been moved to fluid_samplecache_load.

This change also fixes #483 because it removes the bogus warning messages if a soundfont was loaded from memory by abusing the filename to store a pointer.

In #483 it was also discussed if we should bypass the sample caching completely for soundfonts that
were loaded from memory. After going through the code, I think this is not really sensible. The current 
soundfont loader implementation always creates a copy of the data provided by the file callbacks, it also
takes care of uncompressing SF3 samples before storing them in the cache. So caching is quite benefitial, as it reduces the number of times the data gets copied and/or decompressed.

### DIFF
--- a/src/sfloader/fluid_samplecache.c
+++ b/src/sfloader/fluid_samplecache.c
@@ -60,8 +60,10 @@ struct _fluid_samplecache_entry_t
 static fluid_list_t *samplecache_list = NULL;
 static fluid_mutex_t samplecache_mutex = FLUID_MUTEX_INIT;
 
-static fluid_samplecache_entry_t *new_samplecache_entry(SFData *sf, unsigned int sample_start, unsigned int sample_end, int sample_type);
-static fluid_samplecache_entry_t *get_samplecache_entry(SFData *sf, unsigned int sample_start, unsigned int sample_end, int sample_type);
+static fluid_samplecache_entry_t *new_samplecache_entry(SFData *sf, unsigned int sample_start,
+        unsigned int sample_end, int sample_type, time_t mtime);
+static fluid_samplecache_entry_t *get_samplecache_entry(SFData *sf, unsigned int sample_start,
+        unsigned int sample_end, int sample_type, time_t mtime);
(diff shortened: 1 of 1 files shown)


mawe42:
@realnc would you mind giving this a quick test using your setup?

realnc:
Yep, this removes the warning on my end.

PR #487, opened 2018-12-16
Fix minor bug in lower_boundary_for_attenuation()

jjceresa:
- v variable is renamed min_val for clarity.
- This fix a bug when calculating the possible minimun  value (min_val) of a modulator. min_val should only
  be dependent of:
  - signs of mapped sources (ie bipolar [-1..+1] or unipolar [0..+1]).
  - sign of amount.
  - absolute value of amount.

### DIFF
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -1354,7 +1354,20 @@ fluid_zone_mod_import_sfont(fluid_mod_t **mod, SFZone *sfzone)
              * Deactivate the modulator by setting the amount to 0. */
             mod_dest->amount = 0;
         }  
-        /* *** Dest *** */
+
+        /* Note: When primary source input (src1) is set to General Controller 'No Controller',
+           output will be forced to 0.0 at synthesis time (see fluid_mod_get_value()).
+           That means that the minimum value of the modulator will be always 0.0.
+           We need to force amount value to 0 to ensure a correct evaluation of the minimum
+           value later (see fluid_voice_get_lower_boundary_for_attenuation()).
+        */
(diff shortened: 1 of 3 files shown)


mawe42:
I found it difficult to understand which bug you are referring to. If I understand correctly, then the problem is that the extra check for ```mod->src1 == FLUID_MOD_PITCHWHEEL``` is wrong because the modulator could have been overwritten by a soundfont/user to have a unipolar mapping?

jjceresa:
>If I understand correctly, then the problem is that the extra check for mod->src1 == FLUID_MOD_PITCHWHEEL is wrong because the modulator could have been overwritten by a soundfont/user to have a unipolar mapping?

Right. And when evaluating possible `min_val` of any modulator, this should be independent of the controller palette (i.e CC palette or any other like FLUID_MOD_CHANNELPRESSURE, FLUID_MOD_PITCHWHEEL...).
In my opinion, only the `amplitude/sign` of the resulting `mapping` (i.e [0..1] or [-1..+1]) and the sign/amplitude of `amount` value are relevant. This is simply due to the SF formula of the modulator output: mod out = map1(src1) * map2(src2) * amount.

derselbst:
> If I understand correctly, then the problem is that the extra check for `mod->src1 == FLUID_MOD_PITCHWHEEL` is wrong because the modulator could have been overwritten by a soundfont/user to have a unipolar mapping?

Indeed, that's my bug. It seems that I have mistakenly assumed `FLUID_MOD_PITCHWHEEL` to be always bipolar, probably because it's usually used as such. But ofc. it's just a midi controller like the others (with higher precision though). Thanks for taking care of this!

PR #489, opened 2018-12-16
Add two new tuning shell commands to tune all octaves and all notes

mawe42:
While FluidSynth supports the Midi Tuning Standard and reacts to single note and octave tuning SysEx messages, the shell currently only supports single note tunings. This pull requests adds a new tuning command to tune the 12 notes of all octaves simultaneously with a given cent offset from the standard tuning.

In contrast to the existing tuning command, there is also the option to apply the tuning in real-time, i.e. to also affect sounding notes immediately via the "--realtime" argument.

Edit: previously there were two new commands. But the "tuneall" command duplicates existing functionality as there is already a RPN for channel wide fine tuning.

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -262,11 +262,6 @@ static const fluid_cmd_t fluid_commands[] =
         "                           standard tuning for all 12 notes, starting from C\n"
     },
     {
-        "tuneall", "tuning", fluid_handle_tuneall,
-        "tuneall bank prog offset   Tune all notes with the given cent offset from\n"
-        "                           standard tuning\n"
-    },
-    {
         "settuning", "tuning", fluid_handle_settuning,
         "settuning chan bank prog   Set the tuning for a MIDI channel"
     },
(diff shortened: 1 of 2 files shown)


mawe42:
> The --realtime option is currently not documented. Should we add this option to the other tuning commands as well? Or should we change all tuning commands to be realtime by definition?

Making all realtime automatically would change existing behaviour. So I maybe just add the option to the "tune" command as well. I think that would make the interface consistent again.

> Do you agree to keep this as feature for 2.1?

Yes, sure. It's a new feature, after all.

mawe42:
And I think I would like to change the command interface before we consider this for merging. Having to specify all 12 note offsets even if you only want to tune one specific note is not so nice. And I don't really like such long parameter lists anyway.

So I would like to change it to:
```
tuneall <bank> <prog> <note (0 - 11)> <cent-offset> [--realtime]
```

That way it works very similar to the ```tune``` command,  just on all octaves of a note and with offsets instead of exact pitches. What do you think?

derselbst:
Sure, having a command that tunes a specific note for all octaves sounds useful.

mawe42:
I currently don't have enough time to polish this up. And as I don't really need this feature and nobody requested it, I'm going to retract this pull request. I keep the branch locally, just in case somebody comes forward who needs this.

PR #490, opened 2018-12-17
Need fluid_synth_nwrite_float

schnitzeltony:


derselbst:
> It is the only render function allowing clients to use own buffer layout.

It's not. `fluid_synth_process()` much more flexible. It already allows null buffer.

> Instead of overwriting buffers it mixes rendered samples ontop  of the values in buffers.

Which is intended to allow the application to mix the buffers on-the-fly.

> That requires buffer cleanup before call to avoid garbage and is a perfomance sink.

I can't believe that a `memset()` is such a significant bottleneck (if you meant that). Which platform is it? Have you done any measurements?

... and after all: Why are you comming up with this now? We had a beta phase for such issues.

schnitzeltony:
How comes that I expected an answer like this - go on brake may use case and waste my time

derselbst:
I'm not intending to brake anybody's use case. It does however surprise me that you bring up this issue now, although you were already actively participating since May.

If you are serious in what you say, you should start a discussion about that on the mailing list. Perhaps giving a little more insights of your use case would also be helpful.

derselbst:
Thanks for your explanations.

> Instead of just doing memcpy (as fluid_synth_nwrite_float does for floats) fluid_synth_process mixes all samples which is more expensive and useless in my case.

Yes. But it is not clear to me how this affects the performance in your case? Is 2.0.2 completely unusable for you when using `fluid_synth_process()`?

> After getting my images to build with 2.0.1 I discovered that fluidsynth is no more usable and asked again [8]. Response was not exactly helpful

I didn't have enough information to actually help, which is why I asked for them.

> Conclusion: fluid_synth_nwrite_float just performs best which is important on weak machines and I see no solution which will not cause performance regression.

`fluid_synth_process()` provides the chance to mix the rendered audio directly onto the user's buffers. I.e. similar to what currently is done in `fluid_rvoice_buffers_mix()`. While this should be simple for single threaded rendering, it may become tricky for multi threaded rendering. Still I see the potential to eliminate the final copy currently done by `fluid_synth_nwrite_float()` and ultimately being even faster than `fluid_synth_nwrite_float()`.

schnitzeltony:
A new year and I would like to come back to this. Meanwhile I updated to 2.0.3 and it is usable although it still does not perform as good as 1.1.11. A preliminary test result:

With 1.1.11 I can run the MIDI file mentioned above with 0 underruns. For 2.0.3 jack reports 10(127) underruns. Both tests were performed within qtractor and fluidsynth-dssi  using fluid_synth_nwrite_float. This would be totally acceptable: dynamic load is the hot feature I am looking for. It seems that MIDI file is a 'close to the edge challenge' for RaspberyPi 3B+ and therefore a good test case.

Next step for me would be to come up with performance numbers for different render functions.

Before I go into this - it is something which is not done within few hours - I would like to ask: In case of better performance for fluid_synth_nwrite_float vs. fluid_synth_process: **Is there a chance for these patches to be applied?** If not (looks to me that the decision was already made) the tests would be just waste of time.

derselbst:
> With 1.1.11 I can run the MIDI file mentioned above with 0 underruns. For 2.0.3 jack reports 10(127) underruns. Both tests were performed within qtractor and fluidsynth-dssi using fluid_synth_nwrite_float.

The fact that both tests were made with `fluid_synth_nwrite_float()` means that a performance regression might have been introduced. The core problem most likely lies in `rvoice_mixer` then. I can look into this. However, this is not solved by undeprecating and, excuse me, "hacking" a public render function to perform better, as you suggested.

Which soundfont have you used btw?

It would still be interesting to know how `fluid_synth_process()` performs. If you use jack anyway and don't want to port fluidsynth-dssi, you could just use fluidsynth's jack audio driver to test this. No need to spend hours on this.

> This would be totally acceptable: dynamic load is the hot feature I am looking for.

Are you referring to `synth.dynamic-sample-loading`? Have you set it to TRUE when testing 2.0.3?

mawe42:
> The fact that both tests were made with fluid_synth_nwrite_float() means that a performance regression might have been introduced.

Just a thought: I'm quite sure we also made lots of changes to the cmake scripts between 1.1.11 and 2.0.3. Maybe there are weird side effects that cause your build process to use different compiler flags? Could you check that both versions are compiled with the same compiler and flags?

> Are you referring to synth.dynamic-sample-loading? Have you set it to TRUE when testing 2.0.3?

Just to explain why this would be relevant: the dynamic sample loading in combination with program change events in a midi file will likely cause dropouts on systems with fairly low IO throughput, especially if the system is under heavy load anyway. We thought about adding a "prefetch" option to parse a midi file before playback to load all used samples in advance, but decided against it because nobody had a proper use for this.

Interesting would be when the dropouts occur. Only at the start of playback? Or during or shortly after program change events? Randomly?

schnitzeltony:
There were more questions but I don't have the time currently to check for all.
@derselbst 
> The core problem most likely lies in rvoice_mixer then. I can look into this. However, this is not solved by undeprecating and, excuse me, "hacking" a public render function to perform better, as you suggested.

I am aware that this does not fix the root cause of performance regression. Friendly spoken: It is a bit strange to me that you consider checking for zero pointer a hack. Again: It gives client the chance to avoid fluidsynth performing useless operations.

Please remember what this pull request is about

* not removing a function which is there for many years and avoiding useless buffer clear/mix and handles multi-threading rendering perfectly well.
* give client the chance to enhance it further performance wise. 

So please forget performance regression for a moment - it is a different issue. BTW: I haven't understood why you want to get rid of fluid_synth_nwrite_float - commit bb413011f7bc21b1c2db047fe4f4fcca0dc89a4a is not helpful exactly. Could you enlighten me please?

@mawe42 
> Just to explain why this would be relevant: the dynamic sample loading in combination with program change events in a midi file will likely cause dropouts on systems with fairly low IO throughput, especially if the system is under heavy load anyway. We thought about adding a "prefetch" option to parse a midi file before playback to load all used samples in advance, but decided against it because nobody had a proper use for this.

The drop-out values are similar for synth.dynamic-sample-loading set to 1 or 0 - there is no significant difference.

> Interesting would be when the dropouts occur. Only at the start of playback? Or during or shortly after program change events? Randomly?

There are certain areas in the song when many instruments/note are active.

Ahh before I forget - it might be of interest: When I create one track in qtractor with fluidsynth, the idle CPU load (not playing anything) of fluidsynth 2.x is higher significantly than it was with 1.1.11.

mawe42:
> The drop-out values are similar for synth.dynamic-sample-loading set to 1 or 0 - there is no significant difference.

Thanks, that is very good to know.

> Ahh before I forget - it might be of interest: When I create one track in qtractor with fluidsynth, the idle CPU load (not playing anything) of fluidsynth 2.x is higher significantly than it was with 1.1.11.

Can you give a little more detail about the scale of the idle load increase with 2.x? Just a rough idea of the absolute CPU loads of 1.1.11 and 2.x.

I understand that you think that this PR is not about this particular performance regression. But it touches the render function you would like to de-deprecate and the reason you would like to continue using the deprecated function is based on concerns about loss of performance. So I think it's highly relevant.

mawe42:
And I would still be very interested to know if 1.1.11 and 2.x used different compilers and/or compiler flags on your platform.

schnitzeltony:
> And I would still be very interested to know if 1.1.11 and 2.x used different compilers and/or compiler flags on your platform.

Same - but I had to replace hard-disks in my build-machine and am still fighting to get full images build again. I really hope to come back with more answers/numbers soon...

derselbst:
> Friendly spoken: It is a bit strange to me that you consider checking for zero pointer a hack.

In the context of your problem I consider this a hack: It introduces a functionality that mitigates the symptoms of a deeper problem, which wasn't necessary before and only works for those users, whose use-case actually allows to nullify buffers.

> Please remember what this pull request is about
> not removing a function which is there for many years

The fact that fluid_synth_nwrite_float() has been deprecated does not mean in any way that there are plans to remove it. (Imagine: If I were to cause trouble I would have done this last major version bump.) For now (and likely the next year(s)), the only purpose of the deprecation is to indicate to developers that they should use fluid_synth_process(). Why? Because the [old API docs](http://www.fluidsynth.org/api-1.x/index.html#UsingSynth) said so (not my words):

> > The function fluid_synth_process() is still experimental and its use is therefore not recommended but it will probably become the generic interface in future versions.

What now exists as fluid_synth_process() is the result of a discussion at the mailing list:

http://lists.nongnu.org/archive/html/fluid-dev/2018-04/msg00075.html
http://lists.nongnu.org/archive/html/fluid-dev/2018-05/msg00000.html 

Deprecating fluid_synth_nwrite_float() was a natural step for me to go as fluid_synth_process() is intended to replace its use-case. Yet, there's currently nothing that keeps an existing implementation from continuing to use fluid_synth_nwrite_float().

> and avoiding useless buffer clear/mix

The motivation of this uselessness was to enable rendering the audio directly onto the user's buffer(s), while allowing the user to keep e.g. recorded audio in those buffers. 

> give client the chance to enhance it further performance wise.

Nullifying buffers has been taken into account for fluid_synth_process(). I am struggling to change the behaviour of a depreacted function. Still I'm currently not intending to de-deprecate fluid_synth_nwrite_float() because I see the chances fluid_synth_process() brings. Sure, it is possible that I am mistaken and that fluid_synth_process() is doomed to fail, in which case we can still fallback to fluid_synth_nwrite_float().

> and handles multi-threading rendering perfectly well.

Multi-threading is implemented in rvoice_mixer. I don't see any connection to fluid_synth_nwrite_float() or fluid_synth_process() here.


> Please remember what this pull request is about

I must admit that this was very unclear to me initially. You demanded to de-deprecate a public function and change it's behaviour, which irritated me. You seem to have provided a solution or workaround, but the problem was missing. So far, I've understood the problem as that the performance is not satisfying compared to 1.1.11. Great. I'll look into it when I'm free. 

> And I would still be very interested to know if 1.1.11 and 2.x used different compilers and/or compiler flags on your platform.

Second.

mawe42:
I just wrote a small test program that uses ```fluid_synth_nwrite_float``` and does a similar setup as fluidsynth-dssi does. I basically copied the synth creation code, including the settings with the default values from fluidsynth-dssi. Then I did a bisect between current master and v1.1.11. I actually noticed a dramatic performance regression, roughly factor 5 - 7 between good and bad commits. And the first bad commit is 6ab0cc3024a18b9691b57880b101915c6d3f764d

In that commit, we cleaned up the settings handling and removed the possibility to set integer settings via ```fluid_settings_setstr```. And fluidsynth-dssi does the following:
```
fluid_settings_setstr(settings, "synth.reverb.active", "no");
fluid_settings_setstr(settings, "synth.chorus.active", "no");
```

In Fluidsynth 2.x, those two lines have no effect. They will actually return an error, but as fluidsynth-dssi does not check the return value, reverb and chorus simply remain active. Which accounts for the added CPU load. When I change it to the proper
```
fluid_settings_setint(settings, "synth.reverb.active", 0);
fluid_settings_setint(settings, "synth.chorus.active", 0);
```
performance is back to normal, even after this commit.

mawe42:
And unless I overlooked something, we failed to communicate this change of behaviour of the ```fluid_settings_setstr``` function and of the other backwards incompatible change from #293... :-/

mawe42:
You're welcome. Gave me a reason to try out automatic git bisect, which I've never done before. What a fantastic tool...

@derselbst I think we need to discuss if, where and how to add a note about those two backward incompatible changes from #293. But probably in a different issue...

schnitzeltony:
> @derselbst I think we need to discuss if, where and how to add a note about those two backward incompatible changes from #293. But probably in a different issue...

Have not idea of implementation details/efforts but how about creating a warning in case a setting is accessed with incompatible type - similar to those warnings created when soundfont have issues?

mawe42:
> how about creating a warning in case a setting is accessed with incompatible type

Good idea. I've opened a pull request #500 with some documentation about it and added warnings for the ```fluid_settings_set*``` functions.

schnitzeltony:
Lionel Ritchie's 'Hello' MIDI runs buffer underrun free again. Thanks again.

What concerns me: I should have heard a difference with chorus/reverb active - never mind...

So close this?

mawe42:
> Lionel Ritchie's 'Hello' MIDI runs buffer underrun free again.

\o/ Great :-)

> What concerns me: I should have heard a difference with chorus/reverb active - never mind...

I don't think you should... you are passing NULL pointers for the effects buffers, and ```fluid_synth_nwrite_float``` doesn't mix effects with the dry audio, as far as I can tell.

PR #492, opened 2018-12-22
Fix error with MSVC

carlo-bramini:
FLUID_INLINE must be used instead of inline, otherwise MSVC fails to build fluidsynth.


derselbst:
Thanks.

Issue #320, opened 2018-01-06
Proposal: DLS file loading


frabert:
I am writing a DirectMusic compatibility layer of sort, and I found myself in need of a softsynth for DLS files, which are in principle similar to SoundFonts. Currently, my approach consists of feeding the DLS files to a converter that outputs SF2 files, which are then loaded normally. However, this procedure is expensive both in terms of memory and in time. If I were to develop a patch to enable FluidSynth to directly load DLSs, would it be accepted (if it met the quality requirements, ofc)? Please note that I have not yet looked into the FluidSynth architecture, so it's all speculative for now, but my main concern was for the "Based on SoundFont 2" tagline.

derselbst:
That would be nice feature indeed, esp. for a major 2.0 release. I'm not familiar with DLS, but I think starting a development from scratch wouldnt be appropriate. You should rather have a look at [libinstpatch](https://sourceforge.net/p/swami/code/ci/master/tree/libinstpatch/). The ["Future" wiki page](https://github.com/FluidSynth/fluidsynth/wiki/Future#proposed-features) suggests to make this library an (optional) dependency and by that bring DLS and GigaSampler support.

@elementgreen What is libinstpatch's current status of supporting DLS? (How) Could it help fluidsynth to implement this feature? Could you give some pointers for further reference / starting points?

frabert:
Is the Glib requirement a hard one or is it optional? Seems like a big dependency, especially on Windows...

derselbst:
For fluidsynth it's a hard one and has been extensively discussed before, see mailing list and [#145](https://github.com/FluidSynth/fluidsynth/issues/145#issuecomment-337562513). But that seems to be off-topic,  where's the reference to DLS loading?

frabert:
I was referring to libinstpatch, I didn't actually notice that FluidSynty already needs it, sorry.

elementgreen:
libInstPatch is completely dependent on glib/GObject and does support DLS loading/editing/saving.  The DLS synthesis model is very similar to SoundFont (I suspect they borrowed it from SoundFont), but has increased parameter resolution (32 bit parameters instead of 16 bit).  Seemed like a nice format to me when I wrote support for it in libInstPatch.  However, it seems like the format is not that popular?  Maybe that is my own limited impression though.  Because of this though I never really did add full editing capabilities to Swami.  It was also a shame that the specification cost money to download previously (not sure about now, seems like it may just require registering an account?).

[...]

PR #493, opened 2018-12-30
Support Loading DLS Files

derselbst:
This is a RFC PR that attempts to support DLS as requested in #320 (perhaps also GigaSampler and SLI). I intend to bring this feature to the mailing list when it's ready, but for now, see this as alpha test.

DLS support is brought by [libinstpatch](https://sourceforge.net/p/swami/code/ci/master/tree/), as suggested by the ["Future" wiki page](https://github.com/FluidSynth/fluidsynth/wiki/Future#proposed-features). CMake queries for it during compilation. If found, it adds a libinstpatch sfloader to the synth (`fluid_instpatch.c`).

`fluid_instpatch.c` basically works, but could be improved:

* ~currently ignores `synth.dynamic-sample-loading`~ (IMO, this is the responsibility of libinstpatch)
* ~allocates presets on-the-fly as needed, i.e. no preset pool~
* when loading a DLS file the defsfont loader complains that it's no SF2 file, which is a little annoying... one might need to rethink that

I've so far only tested this with a few DLS files. I'm not familiar with GigaSampler or SLI. It would be great if someone could give it a test. Ofc. general feedback is appreciated as well.

---

### Why using libinstpatch?

Pro:
* It's LGPL licensed (although the COPYING claims GPL)
* It provides a C API
* Most of the necessary implementation for this sfloader already exists in the [fluidsynth-plugin for swami
](https://sourceforge.net/p/swami/code/ci/master/tree/swami/src/plugins/fluidsynth.c)
* It provides a converter that converts all its supported files to SF2 voice model, which makes it easy to use for fluidsynth (generators, modulators,... everything is there after a single _convert_)
* Uses refcounting for every object, which makes it uncomplicated on cleanup

Con:
* There are many issues in the code: Neither the stable 1.0.0 release nor recent git version is working
  * mostly due to incorrect GValue assignments and incorrect mutex (un-)locking
  * as a workaround I patched [my own working version](https://build.opensuse.org/package/show/home:derselbst:branches:multimedia:libs/libinstpatch) (you know that it works when you don't get any `Glib - CRITICAL ** assertions` on DLS loading)
* Libinstpatch claims DLS and eps. GigaSampler support to be incomplete, but I don't know how severe this incompleteness is

### Any alternative to libinstpatch? Yes: [libgig](https://www.linuxsampler.org/libgig/)

Pro:
* Seems to be more complete and mature for DLS and GigaSampler

Con:
* GPL licensed
* Only provides a C++ API (haven't looked to close at it)

Both Cons are not too severe, but might restrain the number of users making use of it.


jjceresa:
>This is a RFC PR that attempts to support DLS as requested in #320 (perhaps also GigaSampler and SLI). I intend to bring this feature to the mailing list when it's ready, but for now, see this as alpha test.

This opens the door to other soundfont file formats by selecting at Cmake time the sfloader(s) need by the application. This is nice.

>when loading a DLS file the defsfont loader complains that it's no SF2 file, which is a little annoying... one might need to rethink that

When fluid_synth_sfload() is called it uses all the `fluid_sfloader_t loader` (in the list) until loading is successful. Assuming the following loaders list: `sfload1, dls loader, defsfont`. When loading a file fails for `sfload1`, and `dls loader` this failure should be done silently. But when `defsfont` fails, as it is a sf2 loader and the last in the list it seems normal for a message to be displayed. Isn't it ?

mawe42:
To clarify my thoughts: my opinion is that adding loaders for advanced formats like GigaSampler would only make sense if we would also change and extend the synthesis model at the same time. In other words: if we were to move more into the direction of LinuxSampler and implement different synthesis engines for the different formats. Otherwise we would simply add code that would make Fluidsynth a bad GigaSampler synth. And there is already a very good GigaSampler synth out there: LinuxSampler.

derselbst:
> And I'm afraid that if we open up to other formats, we will loose some of that focus.

Sure, I understand this. I think that's a matter of documentation. We could claim smth. like limited or partial DLS support. Similar to like it's said in the [sfont loader API](http://www.fluidsynth.org/api/sfont_8h.html#details):

> This API allows for virtual SoundFont files to be loaded and synthesized, which may not actually be SoundFont files, as long as they can be represented by the SoundFont synthesis model.

We can keep this limited to DLS support for now, as this was the intention of #320. [Citing ElementGreen](https://github.com/FluidSynth/fluidsynth/issues/320#issuecomment-355844440), DLS and SF2 are very similar, except that DLS is "more" expressive because it supports 32 bit parameters, rather than 16 bit as SF2 does.

The idea of GigaSampler only came up because libinstpatch supports them as well by loading DLS files and then recognizing those as GigaSampler files. But I can't tell anything about the relation between GigaSampler and DLS.

> In other words: if we were to move more into the direction of LinuxSampler and implement different synthesis engines for the different formats.

You're right. And I'm not intending to shift our focus. Fluidsynth should stay a SF2 synth. Supporting DLS just seems to be a nice idea as much/most/all (?) of DLS can be represented by the SoundFont synthesis model. And if there are any incorrect envelopes or attenuation calculations for DLS, keep in mind that there is limited / partial support.

As next steps I suggest:

1. Test more DLS files to figure out the current state of support (I have not been very successful in finding them on the web, if you have any it would be nice to submit them)
1. Test a very few GigaSampler files just to see how it works (well/barely/not at all ?) and how to treat them
1. Decide whether to go with libinstpatch or not (I'm confident though)
1. Fix libinstpatch upstream + new release

> PS: At least the linuxsampler code is a strange form of GPL, because the authors forbid the use for commercial projects (even if those projects comply with the GPL). Not sure if that also applies to the libgig code.

[No, it doesn't.](http://www.linuxsampler.org/downloads.html)

mawe42:
> Fluidsynth should stay a SF2 synth. Supporting DLS just seems to be a nice idea as much/most/all (?) of DLS can be represented by the SoundFont synthesis model.

I agree, if DLS really is like "high-resolution Soundfont", then it seems like a very good fit and adding support is a good idea, especially as there is actual user demand for this feature.

> I have not been very successful in finding (DLS files) on the web, if you have any it would be nice to submit them

To be honest, I had never even heard of DLS before reading the ticket. Maybe @frabert (the original reporter of the ticket) could supply some?

Your proposed way forward sounds good. And going with libinstpatch instead of libgig feels better to me (regardless of code quality / maturity) as libgig looks to be a GigaSampler library first and foremost. And using such a lib when we can't completely support that format would send the wrong message, I think. Granted, a message that not many people will hear or understand... :-)

frabert:
@mawe42 Unfortunately I do not have any freely redistributable DLS file, but every Windows copy ships with c:\windows\system32\gm.dls, which is used for General MIDI patches by the system's softsynth.

derselbst:
Meanwhile, I've tested a bunch of DLS files and they all play fine. Gig files are problematic though, because they currently cause lots of crashes in libinstpatch due to a nasty int to pointer cast.

derselbst:
As this seems to be a very popular feature, open TODOs right now:

- [ ] implement IO callbacks
- [ ] make sure it doesn't crash when dls font is unloaded while its samples are still used for rendering
- [ ] check for memory leaks (samples unloaded correctly)

derselbst:
Considering this ready for squashing.

@mawe42, @carlo-bramini Any of you available for a review?

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+
+    /* open a file, we get a reference */
+    if((file = ipatch_dls_file_new()) == NULL)
+    {
+        FLUID_FREE(patchfont);
+
+        return NULL;
+    }
+
+    /* TODO
```
mawe42: Maybe add a short comment explaining why that is still a TODO and currently not really needed.

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+
+        g_object_unref(file);
+        FLUID_FREE(patchfont);
+
+        return NULL;
+    }
+
+    /* get rid of the reference we own, we dont need it any longer */
+    g_object_unref(file);
+    file = NULL;
```
mawe42: g_object_clear?

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+        FLUID_FREE(patchfont);
+
+        return NULL;
+    }
+
+    patchfont->dls = ipatch_dls_reader_load(reader, &err);
+
+    /* unref the reader directly afterwards, not needed any longer */
+    g_object_unref(reader);
+    reader = NULL;
```
mawe42: g_object_clear?

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+        do
+        {
+            fluid_preset_t *preset;
+            IpatchSF2VoiceCache* cache;
+            int bank, prog;
+            ipatch_dls2_inst_get_midi_locale(inst, &bank, &prog);
+            
+            if((cache = convert_dls_to_sf2_instrument(patchfont, inst)) == NULL)
+            {
+                FLUID_LOG(FLUID_WARN, "Failed to cache DLS instrument bank %d , prog %d", bank, prog);
```
mawe42: "Failed to **convert** DLS instrument ..."?

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+            }
+            else
+            {
+                int isPercussion = (ipatch_item_get_flags(inst) & IPATCH_DLS2_INST_PERCUSSION) != 0;
```
mawe42: Naming style: is_percussion

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+
+static int delete_fluid_instpatch(fluid_instpatch_font_t *pfont)
+{
+    guint16 index_array[MAX_INST_VOICES];		/* voice index array */
+    int sel_values[IPATCH_SF2_VOICE_CACHE_MAX_SEL_VALUES];
+    fluid_list_t* list;
+    
+    fluid_return_val_if_fail(pfont != NULL, FLUID_OK);
+
+    /* look through all fluid samples and return error if failed */
```
mawe42: **loop** through all fluid samples to do what? And return error if what failed?

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+                                            fluid_instpatch_preset_get_name,
+                                            fluid_instpatch_preset_get_banknum,
+                                            fluid_instpatch_preset_get_num,
+                                            fluid_instpatch_preset_noteon,
+                                            fluid_instpatch_preset_free);
+
+                if(preset == NULL)
+                {
+                    delete_fluid_instpatch_preset(preset_data);
+                    FLUID_LOG(FLUID_ERR, "Out of memory.");
```
mawe42: No fullstop

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+    /* loop over voices and load sample data into RAM */
+    count = cache->voices->len;
+
+    for(i = 0; i < count; i++)
+    {
+        IpatchSF2Voice *voice = &g_array_index(cache->voices, IpatchSF2Voice, i);
+
+        if(!ipatch_sf2_voice_cache_sample_data(voice, NULL))
+        {
+            FLUID_LOG(FLUID_ERR, "ipatch_sf2_voice_cache_sample_data() failed");
```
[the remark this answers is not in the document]
derselbst: It looks quite complicated, see [ipatch_sample_data_get_cache_sample()](https://github.com/swami/libinstpatch/blob/2ba87b9307264342ea625d7c53906923af542540/libinstpatch/IpatchSampleData.c#L937-L1100) which is called subsequently.

src/sfloader/fluid_instpatch.c
derselbst: Sry. That's a leftover, which I've dropped, see my first comment of this PR.

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+        FLUID_LOG(FLUID_ERR, "Unable to find a voice cache converter for this type");
+        return NULL;
+    }
+
+    cache = ipatch_sf2_voice_cache_new(NULL, 0);
+
+    if(cache == NULL)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
+        g_object_unref(conv);
```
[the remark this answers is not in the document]
derselbst: Here not, because `conv` isn't set to NULL.

src/sfloader/fluid_instpatch.c
```diff
@@ -0,0 +1,667 @@
...
+        }
+
+        /* set only those generator parameters that are set */
+        gen_array = &voice->gen_array;
+
+        for(i = 0; i < IPATCH_SF2_GEN_COUNT; i++)
+        {
+            if(IPATCH_SF2_GEN_ARRAY_TEST_FLAG(gen_array, i))
+            {
+                fluid_voice_gen_set(flvoice, i, (float)(gen_array->values[i].sword));
```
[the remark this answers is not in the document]
derselbst: Good point. I stole that code from swami. Even `ipatch_sf2_gen_amount_to_value()` only uses `sword`. After a quick look into libinstpatch I can't find relevant accesses to `uword` either.

src/sfloader/fluid_instpatch.c
derselbst: Yeah, but it's also assigned to a cache, idk.

derselbst:
Thanks for the quick review Marcus! Hope I've addresses all issues. (I should have executed `make format` earlier)

Will keep this open for another 3 or 4 days. Anybody, feel free to test or give feedback.

src/sfloader/fluid_instpatch.c
mawe42: "convert and cache" ? :-)

src/sfloader/fluid_instpatch.c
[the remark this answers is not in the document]
mawe42: The Soundfont spec doesn't specify which generator uses which format. But when we look at our own loading code, the only two generators that use uword are ```Gen_Instrument``` and ```Gen_SampleId```. And as FluidSynth (and I assume also libinstpatch) translates those generators during loading of the font, they are not relevant here. So I guess we could remove the FIXME from fluid_defsfont.c.

src/sfloader/fluid_instpatch.c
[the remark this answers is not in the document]
mawe42: Ok. What tripped me was that the error message sounds even more technical than our other error messages, mentioning a specific function name in a library. But it's probably not really important.

src/sfloader/fluid_instpatch.c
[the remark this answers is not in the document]
mawe42: Sure

src/sfloader/fluid_instpatch.c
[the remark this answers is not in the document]
derselbst: You're right. I cleaned this up: 91a5e9526e6ed76c9bdce5dcba9b00383fcb4d9e

MaKiPL:
Hello everyone,
I'm having some issues or either I am doing something wrong. I cloned the most up-to-date commit and built with libinstpatch support. Therefore now in API the function ```fluid_synth_sfload(synth, char*, int)``` has a loader support from list and tries to read the file until success or until running out of available loaders. I've called ```fluid_synth_sfload``` function pointing to my DLS file and ended up with int= -1, therefore indicating error. Log reports "Not a SoundFont file" and "Failed to load SoundFont ...". Do I have to "register" of some kind the DLS loader or it should just work out-of-box with ```fluid_synth_sfload``` function? 
Running fluidsynth executable and passing the casual commands to play MID it reports the same error with DLS

PR #500, opened 2019-01-11
Document settings changes

mawe42:
This pull request attempts to document the backward incompatible changes introduced in #293 for version 2.0. It adds a note about the two changes to the 2.0 devdocs and fixes an invalid code example.

It also adds a warning message if trying to set a parameter that doesn't exist or which has an invalid type.

### DIFF
--- a/doc/fluidsynth-v20-devdoc.txt
+++ b/doc/fluidsynth-v20-devdoc.txt
@@ -96,6 +96,8 @@ FluidSynths major version was bumped. The API was reworked, deprecated functions
 - fluid_synth_remove_sfont() returns FLUID_OK or FLUID_FAILED
 - introduce a separate data type for sequencer client IDs: #fluid_seq_id_t
 - fluid_get_userconf() has been implemented for Windows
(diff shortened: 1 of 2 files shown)


derselbst:
Overall looks good.

I personally don't like the idea of having a log message in the setters of the fluid_settings. (Why only the setters, why not the getters and anywhere else as well...). People should properly check the return values rather than relying on any log message that we may or may not print. ...Admittedly, I am a bad role model here [since I'm not even doing this myself](https://github.com/derselbst/ANMP/blob/c9c9bc0cbb75f9e4171c5b094e0dd809062657bf/src/InputLibraryWrapper/FluidsynthWrapper.cpp#L225-L247).

Long story short: I don't mind having a log message there, it won't hurt. However, I would vote for making it `FLUID_ERR` rather than just a warning. And calling it `setting '%s'` rather than `parameter`.

mawe42:
> Why only the setters, why not the getters and anywhere else as well...

I'm also in two minds about the warnings, but I can think of no other way to make sure that what happened to the fluidsynth-dssi problem doesn't go unnoticed for other people as well. As the change in behaviour has been active since the first 2.x version, I think it's important that we take an active role in helping people spot a possible problem, and not just rely on them reading the updated devdocs.

> I would vote for making it FLUID_ERR rather than just a warning. And calling it setting '%s' rather than parameter.

Yes, sounds good.

derselbst:
Ok. I'll mess around with your commits and merge it in.

mawe42:
Great, thanks!

derselbst:
Applied to 2.0.x, rebased and merged back to master.

PR #505, opened 2019-02-01
Add support for linked modulators

### LINKED ISSUE #497: Proposal: Adding Soundfont linked modulators
1)Overview.
Linked modulators is a SF specification (rev 2.04 - 9.5.4) that extends the modulators capabilities of classic non-linked modulators. A non-linked modulator have 2 MIDI event inputs sources and a generator destination.
Linked modulators allows a soundfont designer to think and design a "complex" modulator tailored to the need of a particular sound articulation during playing performance. This `complex` modulator can have any MIDI events inputs and a generator destination. Its internal structure is free. This way the soundfont designer can easily substitute a complex modulator by another. Also the musician at performance time can select  a sound articulation or another on the fly without express need to select another preset (at least in most situation). In others words "complex" modulator helps to think, design, and play sound articulated instruments in real time. 

2)Implementation in actual branch `linked-modulators`
This brings full SF linked modulators to fluidsynth

2.1) Checking linked modulators at `soundfont loading` time. 
- Identical complex modulators in the same zone are naturally removed (as it is the case for non-linked modulators).
- Any invalid complex modulators are removed.

2.2)At `noteon` time.
The concept of modulators precedence (instrument zone(local/global) , preset zone(local,global)) and the `additive` effect of presets modulators works also for complex modulators the same way it works for non-linked modulators. So the behaviour is straightforward in regard to actual know SF specifications.

2.3)Real time modulation `on CC change` is implemented.

Note that taking account of linked modulators at noteon time (2.2) or during CC change (2.3) is done without lost of performance in regard to non-linked modulators (i.e the overload is very low).

3)Soundfont editor abilities
This is a major issue. Few free soundfont editor have linked modulators capabilities. Polyphone editor actually has partial capabilities (probably due to misunderstanding about SF specifications).
I hope that editor developer would add this capabilities !.

Opinions, comments are welcome.

derselbst:
Implements and closes #497.

TODO:

- [x] set up unit tests
- [ ] code duplication `fluid_defpreset_noteon_add_mod_to_voice`  `fluid_defpreset_noteon_add_linked_mod_to_voice`
- [ ] `make format`


src/sfloader/fluid_defsfont.c
```diff
@@ -1269,6 +1450,383 @@ fluid_zone_is_mod_identic(fluid_mod_t *mod, char *name)
...
+ * @mod modulator.
+ * @return TRUE if mod is in path table, FALSE otherwise.
+ */
+static int
+fluid_is_mod_in_path(fluid_mod_t *path[], int count, fluid_mod_t *mod)
+{
+    int i;
+    for (i = 0; i < count; i++)
+    {
+        if (path[i] == mod)
```
derselbst: `path[i]` will cause a buffer overflow when executing `test_modulator_links`. This probably has to do with the fact that the `path` pointer will be advanced when called from `fluid_defsfont.c:1556`, but `count` is no adjusted to reflect the remaining length of the buffer. Haven't been able to fully comprehend it yet, so I'm posting it here, so we don't forget about it. Here is the output of AddressSanitizer:

```
==11474==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603000000360 at pc 0x00000044c22d bp 0x7fffffffd6f0 sp 0x7fffffffd6e8
READ of size 8 at 0x603000000360 thread T0
    #0 0x44c22c in fluid_is_mod_in_path fluid_defsfont.c:1466
    #1 0x44c712 in fluid_check_linked_mod_path fluid_defsfont.c:1556
    #2 0x44c52b in fluid_check_linked_mod_path fluid_defsfont.c:1524
    #3 0x44cd7f in fluid_zone_check_linked_mod fluid_defsfont.c:1641
    #4 0x40f103 in main test/test_modulator_links.c:145
    #5 0x7ffff3a2af89 in __libc_start_main (/lib64/libc.so.6+0x20f89)
    #6 0x40d229 in _start (build/test/test_modulator_links+0x40d229)

0x603000000360 is located 0 bytes to the right of 32-byte region [0x603000000340,0x603000000360)
allocated by thread T0 here:
    #0 0x7ffff6f00510 in malloc (/usr/lib64/libasan.so.4+0xdc510)
    #1 0x44ccef in fluid_zone_check_linked_mod fluid_defsfont.c:1634
    #2 0x40f103 in main test/test_modulator_links.c:145
    #3 0x7ffff3a2af89 in __libc_start_main (/lib64/libc.so.6+0x20f89)

SUMMARY: AddressSanitizer: heap-buffer-overflow fluid_defsfont.c:1466 in fluid_is_mod_in_path
Shadow bytes around the buggy address:
  0x0c067fff8010: fd fd fa fa fd fd fd fd fa fa fd fd fd fa fa fa
  0x0c067fff8020: fd fd fd fd fa fa 00 00 00 00 fa fa 00 00 00 00
  0x0c067fff8030: fa fa 00 00 00 00 fa fa 00 00 00 00 fa fa 00 00
  0x0c067fff8040: 00 00 fa fa 00 00 00 00 fa fa 00 00 00 00 fa fa
  0x0c067fff8050: fd fd fd fa fa fa fd fd fd fa fa fa fd fd fd fa
=>0x0c067fff8060: fa fa fd fd fd fd fa fa 00 00 00 00[fa]fa fa fa
  0x0c067fff8070: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff8080: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff8090: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff80a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff80b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==11474==ABORTING
```

jjceresa:
Just added some `printf` in `test_modulator_links.c`.

The output look like this:

'''
D:\WinProjet\Git\GitHub\fluidsynth-VS10\test\Release>test_modulator_links.exe
test 1: set up a list of simple modulators
test 2: set up a valid list of complex modulators
test 3: same list, but changed order
test 4: same list, but with additional mod4 that points to mod2 without mod2 having FLUID_MOD_LINK_SRC
fluidsynth: warning: invalid destination test zone with linked modulators/mod1
fluidsynth: warning: path without destination test zone with linked modulators/mod3
test 5: same list, with additional mod4 and valid mod2 this time
test 6: circular complex modulators m3->m2->m3
fluidsynth: warning: invalid circular path test zone with circular linked modulators/mod1
fluidsynth: warning: path without destination test zone with circular linked modulators/mod3
fluidsynth: warning: invalid isolated path test zone with circular linked modulators/mod1
test 7: invalid list of complex modulators: the first modulator should not have a linked destination
fluidsynth: warning: invalid destination test zone with invalid linked modulators/mod0
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod0
fluidsynth: warning: invalid destination test zone with invalid linked modulators/mod0
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod2
fluidsynth: warning: invalid isolated path test zone with invalid linked modulators/mod1
test 8: invalid list of complex modulators: valid first modulator but invalid destinations
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod1
'''

jjceresa:
just added some `printf` in ttest_modulator_links.c

The output look like this:

`D:\WinProjet\Git\GitHub\fluidsynth-VS10\test\Release>test_modulator_links.exe
test 1: set up a list of simple modulators
test 2: set up a valid list of complex modulators
test 3: same list, but changed order
test 4: same list, but with additional mod4 that points to mod2 without mod2 having FLUID_MOD_LINK_SRC
fluidsynth: warning: invalid destination test zone with linked modulators/mod1
fluidsynth: warning: path without destination test zone with linked modulators/mod3
test 5: same list, with additional mod4 and valid mod2 this time
test 6: circular complex modulators m3->m2->m3
fluidsynth: warning: invalid circular path test zone with circular linked modulators/mod1
fluidsynth: warning: path without destination test zone with circular linked modulators/mod3
fluidsynth: warning: invalid isolated path test zone with circular linked modulators/mod1
test 7: invalid list of complex modulators: the first modulator should not have a linked destination
fluidsynth: warning: invalid destination test zone with invalid linked modulators/mod0
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod0
fluidsynth: warning: invalid destination test zone with invalid linked modulators/mod0
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod2
fluidsynth: warning: invalid isolated path test zone with invalid linked modulators/mod1
test 8: invalid list of complex modulators: valid first modulator but invalid destinations
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod1`

>path[i] will cause a buffer overflow when executing test_modulator_links. ....

Sry ,I got no `buffer overflow` here !!.

jjceresa:
Just added some `printf` in ttest_modulator_links.c

The output look like this:

D:\WinProjet\Git\GitHub\fluidsynth-VS10\test\Release>test_modulator_links.exe

`test 1: set up a list of simple modulators`
`test 2: set up a valid list of complex modulators`
`test 3: same list, but changed order`
`test 4: same list, but with additional mod4 that points to mod2 without mod2 having FLUID_MOD_LINK_SRC`
fluidsynth: warning: invalid destination test zone with linked modulators/mod1
fluidsynth: warning: path without destination test zone with linked modulators/mod3
`test 5: same list, with additional mod4 and valid mod2 this time`
`test 6: circular complex modulators m3->m2->m3`
fluidsynth: warning: invalid circular path test zone with circular linked modulators/mod1
fluidsynth: warning: path without destination test zone with circular linked modulators/mod3
fluidsynth: warning: invalid isolated path test zone with circular linked modulators/mod1
`test 7: invalid list of complex modulators: the first modulator should not have a linked destination` fluidsynth: warning: invalid destination test zone with invalid linked modulators/mod0
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod0
fluidsynth: warning: invalid destination test zone with invalid linked modulators/mod0
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod2
fluidsynth: warning: invalid isolated path test zone with invalid linked modulators/mod1
`test 8: invalid list of complex modulators: valid first modulator but invalid destinations`
fluidsynth: warning: path without destination test zone with invalid linked modulators/mod1

>path[i] will cause a buffer overflow when executing test_modulator_links. ....

Sry ,I got no buffer overflow here !?.

jjceresa:
Please note, in previous comment, internal fluidsynth warning message displays modulator index with 0 based convention.
Example:

fluidsynth: warning: invalid destination test zone with linked modulators/mod1

`mod1` is modulator with index 1 (zero based). This correspond to the second modulator in the list.

src/synth/fluid_voice.c
```diff
@@ -1182,28 +1195,34 @@ fluid_voice_update_param(fluid_voice_t *voice, int gen)
...
 
-    for(i = 0; i < voice->mod_count; i++)
+    for (i = 0; i < voice->mod_count; i++)
     {
+
         mod = &voice->mod[i];
+        /* memorize gen destination */
+        if(!(mod->dest & FLUID_MOD_LINK_DEST))
+        {
+            gen = mod->dest;
```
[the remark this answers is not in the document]
derselbst: You are assuming that the first modulator is always a non-complex one. This is dangerous. While the sfloader makes sure that this is the case, the user is free to (mis-)use fluidsynths API:

```c
fluid_mod_t* mod = new_fluid_mod();
fluid_mod_set_dest(mod, FLUID_MOD_LINK_DEST | 0);
fluid_voice_t* voice = new_fluid_voice();
fluid_voice_add_mod(voice, mod, FLUID_VOICE_OVERWRITE);
// synth calls fluid_voice_modulate() --> undefined behaviour
```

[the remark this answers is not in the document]
derselbst: Just pushed a fix for it.

src/synth/fluid_voice.c
```diff
@@ -1530,14 +1603,40 @@ fluid_voice_add_mod_local(fluid_voice_t *voice, fluid_mod_t *mod, int mode, int
...
+                    voice_mod->dest += offset;
+                    /* chain previous to this modulator.Useful for future identity 
+                       test of complex modulator */
+                    prev_mod->next = voice_mod; 
+                }
+                prev_mod = voice_mod; /* previous voice mod */
+                mod = mod->next;      /* next modulator member to clone */
+            }
+            /* last modulator member will be recognized by next field set to NULL*/
+            voice_mod->next = NULL;
```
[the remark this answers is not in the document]
derselbst: Safe for now, but dangerous if we ever changed `fluid_get_num_mod()` to return zero. Fixed.

jjceresa:
>You are assuming that the first modulator is always a non-complex one. This is dangerous. While the sfloader makes sure that this is the case, the user is free to (mis-)use fluidsynths API:

>Safe for now, but dangerous if we ever change fluid_get_num_mod() to return zero. 

Yes, as the user is free to (mis-)use fluidsynths API , may be it would be preferable to do basic sanity check inside `API fluid_voice_add_mod()` only, but not in non-api function (e.g fluid_voice_add_mod_local(),fluid_voice_modulate() ). 
For now, a simple check for API `fluid_voice_add_mod()` (before calling internal fluid_voice_add_mod_local()) could be to accept only a `simple modulator` and refuse any complex modulator because  internal complex modulator's structure should be opaque to the application ?.

derselbst:
> For now, a simple check for API fluid_voice_add_mod() could be to accept only a simple modulator and refuse any complex modulator ?

No, because it does not solve the problem. It is _not_ an invalid usage to pass a complex modulator via the API. It is only invalid if it is the first modulator of a voice. Technically, I know, `FLUID_MOD_LINK_DEST` is private currently. If we however in the future make it public and allow complex modulators via the API, I would like to avoid searching through the source code again, disabling all those "no-complex-modulator-pass-the-API" checks and rethink about how to fix those bugs that we are already looking into.

Also, I usually prefer to have those checks at that place where it can go wrong. If the implementation of `fluid_voice_modulate()` requires the first modulator to be a simple one, it should complain if this is not the case. This way, it keeps not only API users from breaking stuff. It also keeps a drunken maintainer from creating a broken unit test using some hackish internal-member-access-magic without noticing that this breaks things.

That said, yes, it might be necessary to enforce some more safety check via the API that currently only the sfont loader covers. I'm only talking about _necessary checks_ that avoid any undefined behaviour in the code, like passing a complex modulator with insanely high destination and causing a buffer overrun in fluid_voice_add_mod(). No checks that enforce the rules of the soundfont spec via the API, e.g. no need to care if the user adds the same modulator to a voice over and over again.

derselbst:
Regarding your change in b74efca2832f21b8c2d69503f3b45db981e85e62: I consider it incorrect. The soundfont spec may not have complex modulators as default, but the user could have set one via `fluid_synth_add_default_mod()` which would then be added to every voice. And if for some reason the user then wants to override that complex modulator of that one specific voice, it would not work.

derselbst:
Rearding the bufferoverrun: a segfault does not occur, neither for me locally, nor anywhere online at TravisCI or AzurePipelines. And still, the `path` buffer is overrun by one element. Here is the full call stack including function parameters:

```c
(gdb) bt
#0  0x000000000044c286 in fluid_is_mod_in_path (path=0x603000000350, count=3, mod=0x6030000001f0) at fluid_defsfont.c:1466
#1  0x000000000044c7c3 in fluid_check_linked_mod_path (zone_name=0x5108a0 "test zone with linked modulators", list_mod=0x6030000001c0, dest_idx=129, path=0x603000000340, cur_path_idx=2, path_idx=0x7fffffffd810)
    at fluid_defsfont.c:1556
#2  0x000000000044c5dc in fluid_check_linked_mod_path (zone_name=0x5108a0 "test zone with linked modulators", list_mod=0x6030000001c0, dest_idx=-1, path=0x603000000340, cur_path_idx=2, path_idx=0x7fffffffd810)
    at fluid_defsfont.c:1524
#3  0x000000000044ce30 in fluid_zone_check_linked_mod (zone_name=0x5108a0 "test zone with linked modulators", list_mod=0x6030000001c0) at fluid_defsfont.c:1641
#4  0x000000000040f190 in main () at /home/tom/Programme/fluidsynth/fluidsynth/fluidsynth/test/test_modulator_links.c:150
```

Within frame 3 `fluid_zone_check_linked_mod()` the `path` buffer is allocated with 4 elements.

In frame 1, `fluid_is_mod_in_path()` is called:

```c
#1  0x000000000044c7c3 in fluid_check_linked_mod_path (zone_name=0x5108a0 "test zone with linked modulators", list_mod=0x6030000001c0, dest_idx=129, path=0x603000000340, cur_path_idx=2, path_idx=0x7fffffffd810)
    at fluid_defsfont.c:1556
1556                if (fluid_is_mod_in_path(&path[cur_path_idx], *path_idx, mod ))
```

At this call, the `path` pointer is advanced by 2 elements:

```c
(gdb) print cur_path_idx
$2 = 2
```

and its remaining length is told to be 3:

```c
(gdb) print *path_idx
$3 = 3
```

Which causes an overrun of `path` in `fluid_is_mod_in_path()` by one element.

jjceresa:
>No, because it does not solve the problem. It is not an invalid usage to pass a complex modulator via the API.
Also, I usually prefer to have those checks at that place where it can go wrong.....

Yes. I understand completely your points.

>That said, yes, it might be necessary to enforce some more safety check via the API that currently only the sfont loader covers. I'm only talking about necessary checks that avoid any undefined behaviour in the code....No checks that enforce the rules of the soundfont spec via the API, e.g. no need to care if the user adds the same modulator to a voice over and over again.

Yes, I know that actual PR code is incomplete about safety check for API fluid_voice_add_mod().
I am currently thinking about calling  checks parts actually used by the sfont loader, particularly for complex modulators.

jjceresa:
>Regarding your change in b74efca: I consider it incorrect. The soundfont spec may not have complex modulators as default, but the user could have set one via fluid_synth_add_default_mod() which would then be added to every voice.

Ok. 

Actual code in this PR  is unable to work for adding a complex modulator by calling fluid_voice_add_mod_local() in mode FLUID_VOICE_OVERWRITE. I need to think how to add this functionality.

Also, calling API `fluid_synth_add_default_mod()` to set a complex modulator means that sanity checks must be added for complex modulator (this sound similar to  `fluid_voice_add_mod()` checks presented  in previous comment). Need time to see how this can be done.

jjceresa:
>Rearding the bufferoverrun: a segfault does not occur, neither for me locally, nor anywhere online at TravisCI or AzurePipelines. And still, the path buffer is overrun by one element.

Thanks for details report !. Will look further to fix this issue.

derselbst:
> Also, calling API fluid_synth_add_default_mod() to set a complex modulator means that sanity checks must be added for complex modulator (this sound similar to fluid_voice_add_mod() checks presented in previous comment). Need time to see how this can be done.

Ok, cool. Just keep in mind they are low level interfaces. And when you come to the question how to deal with circular references of complex modulators, you could make your life very easy and say that sound articulation is undefined in this case.

jjceresa:
Since commit https://github.com/FluidSynth/fluidsynth/pull/505/commits/4be60e6fc05d3a138b9d50356b1b0e9a2ec065f7 overrun of path should be gone now.

jjceresa:
Some notes about `test_modulators_links.c` (1),(2) and fluidsynth internal use of complex modulator (3).

1)
-Test 6 is a false circular path test.
-Test 7 may be, sound like a misunderstood: The ordering of all members of a complex modulator could be completely arbitrary (SF specifications). That means that any destination of any member could be a generator or a modulator regardless of the position of this member (i.e its index) in the list.

Next commits will attempt to clarify this by formatting the output of each test. These commits will come sooner.

2)
Also please, note that  `fluid_zone_check_linked_mod()` make only validity checks and never modify the ordering of modulators in the input list. That means that after calling this function, TEST_ASSERT check on any ordering changes in the list are useless.

3)
Once  SF modulator's lists are processed by `fluid_zone_check_mod()` this produces output `linked_mod` list for internal use by fluidsynth. In these internal lists, all modulators member of the same complex modulator are ordered in a specific way (by `fluid_zone_copy_linked_mod()`) for the following reasons:
- (1) fast performance on  noteon or CC modulation events.
- (2) foreseeable ordering allowing fast identity checking between 2 complex modulators (via `fluid_linked_mod_test_identity()` ).
Obviously, this internal specific ordering is vital and we should pay attention to never broke this.

derselbst:
> Since commit 4be60e6 overrun of path should be gone now.

Confirmed, thanks!

>  -Test 6 is a false circular path test.

I was planning to add a few more circular path test cases.

> Also please, note that fluid_zone_check_linked_mod() make only validity checks and never modify the ordering of modulators in the input list. That means that after calling this function, TEST_ASSERT check on any ordering changes in the list are useless.

I don't think it's useless. As you said, this function is not supposed to modify the order, so in my opinion it is legal to test that order is not modified.

> Once SF modulator's lists are processed by fluid_zone_check_mod() this produces output linked_mod list for internal use by fluidsynth.

Initially I was thinking whether to write tests against `fluid_zone_check_mod()`, but then decided to go with the lower level function, because `fluid_zone_check_mod()` only tells me either "out-of-memory" or "everthing is fine, I did the job". But when it comes to correct ordering of modulators for fluidsynth's internal usage, it is probably worth to add a few test cases for `fluid_zone_check_mod()` as well.

jjceresa:
Probably test 7 should be removed as modulator ordering is free 'see comment point (1) above' ?

derselbst:
> Probably test 7 should be removed as modulator ordering is free 'see comment point (1) above' ?

I'm not sure, I leave the choice to you.

derselbst:
The latest change broke the build because there are multiple definitions of `fluid_dump_modulator()`. Pls. use the one in `fluid_mod.c`. Note that `printf()` should be called with string literals, otherwise this results in a bunch of compiler warnings: https://travis-ci.org/FluidSynth/fluidsynth/jobs/564745684#L1348-L1376

Also, I don't like the idea of selecting which test cases to run. If you need this, fine, but by default, i.e. no additional arguments given, the unit test should execute all test cases. As of now, `test_fluid_zone_check_mod.c` doesn't do anything when executed during the CI build and just exits with 0, which is treated by ctest as "success".

Also pls. use the `TEST_ASSERT` and `TEST_SUCCESS` macros rather than checking return values yourself. The macros call `abort()` in case of failure. When attached with a debugger this causes a break at where the condition failed.

derselbst:
fb148e6e7db69e6f75b5c0f632377b6e3e9a18b7 : This is redundant, `CMAKE_C_FLAGS_DEBUG` already have DEBUG defined. Just remove the `#ifdef DEBUG` around `fluid_dump_modulator()`.

jjceresa:
>This is redundant, CMAKE_C_FLAGS_DEBUG already have DEBUG defined.

Sry, but without the change in fb148e6, using MSVC generator choosing `debug` build (or `enable-debug` option), then opening MSVC, DEBUG macro is still missing in the list of pre-processor macro for target  libfluidsynth-OBJ. I do not understand why.

derselbst:
Ok, I'll look into this and fix it on the master branch. This PR however is
the wrong place to address this. It is not needed as well: To fix the
build, just remove the #ifdef around fluid_dump_modulator() .

derselbst:
Regarding your latest change in 5f4f1180603866e468fe9e261c103575050c1a82: I understand that you cannot modify the modulators since the argument of `fluid_synth_add_default_mod()` is const. I actually didn't noticed that before, sry. With that change however it's getting even harder to test whether the internal implementation logic correctly detects invalid modulators. I would prefer to keep modulator invalidation within the `fluid_mod_t` structure as done before, rather than trying to make complex modulators work via the API at any price. We can introduce something like `fluid_synth_add_complex_default_mod()` with a non-const mod argument at any time if need be.  Perhaps this function should better receive an array of modulators rather than a list. And `fluid_voice_add_mod()` actually has no way to tell whether the operation succeeded or failed, which means we would need to introduce some `fluid_voice_add_mod2()` which does return an `int`.

...I must admit that the more I think about this the more I believe it was a bad idea to intermix API design with this feature PR. I think your initial idea to simply reject complex modulators via the API really is the better way to go for the moment. I'm sorry to say this now, after the work you've put in :(

jjceresa:
>With that change however it's getting even harder to test whether the internal implementation logic correctly detects invalid modulators.I would prefer to keep modulator invalidation within the fluid_mod_t structure as done before.

You mean that any invalid modulator should be marked with amount forced to 0. This is effectively the only simple way for the caller to get the necessary information. This can be reverted as before without no problem :).

>I think your initial idea to simply reject complex modulators via the API really is the better way to go for the moment. I'm sorry to say this now, after the work you've put in

No problem, however, adding a complex modulator in a voice via  any `fluid_voice_add_mod_x() API`  is the unique feature allowing application edition of complex modulator with an immediate flexible feedback when having fluidsynth as internal synthesizer.

`fluid_voice_add_mod()` could continue to behave as actual for simple modulator. Simply we could return invalid modulator by setting amount to 0.0. This should compensate for the lack of return `int` without breaking API.

>I must admit that the more I think about this the more I believe it was a bad idea to intermix API design with this feature PR.

My first idea was motivated by clarity for the point of view of application developer, I think also that a separate API for adding a complex modulator in a voice would be better. We could think of  something like:
`int fluid_voice_add_complex_mod(voice, mod, mode)` with a non-const mod argument. returning 
- FLUID_FAILED when the complex modulators is fully invalid (no linked path valid).
- FLUID_OK when the complex modulator is partially or fully valid (at least one linked path valid). 
  Any invalid modulators will be returned with amount forced to 0.0
- mod list could be an array of modulators or a list (to be more flexible). Being an array it can be a 
  vla  in stack (for speed).
- link ordering of the modulators should be free (to be more flexible for easy importation from
  application's  soundfont loader).
- mode would be the same as actual fluid_voice_add_mod(). 

As you said, we could postpone `fluid_synth_add_complex_default_mod()` when needed.

jjceresa:
>This can be reverted as before without no problem :).

I mean, I will do that !.

derselbst:
> > This can be reverted as before without no problem :).
> 
> I mean, I will do that !.

Thanks.

> As you said, we could postpone `fluid_synth_add_complex_default_mod()` when needed.

Alright then, let's postpone it.

jjceresa:
I need doing a pause a few days/week, then will continue this.

derselbst:
Great to see that you are back! I'll review this asap, but I'm afraid I won't be able to join development before October.

test/test_fluid_zone_check_mod.c
```diff
@@ -0,0 +1,1365 @@
...
+	print_list_mod("list_mod",list_mod);
+	
+	print_list_linked_mod(NULL,"linked_mod",linked_mod);
+}
+
+/*
+ Print modulator member of a complex modulator.
+ @param offset, offset to add to each index member.
+*/
+static void fluid_dump_linked_mod(fluid_mod_t *mod, int offset)
```
derselbst: Reminder for myself: This function is duplicated from `fluid_defsfont.c`.

src/sfloader/fluid_defsfont.c
```diff
@@ -1268,23 +1450,548 @@ fluid_zone_is_mod_identic(fluid_mod_t *mod, char *name)
...
+    }
+    /* initialize path: 
+	   - reset bits FLUID_PATH_VALID, FLUID_PATH_CURRENT
+	   - set bits FLUID_MOD_VALID
+	*/
+    FLUID_MEMSET(path, FLUID_MOD_VALID, count);
+
+    /* checks valid modulator sources (specs SF 2.01  7.4, 7.8, 8.2.1).*/
+    /* checks identic modulators in the list (specs SF 2.01  7.4, 7.8). */
+    if(list_name != NULL)
```
derselbst: Should the number of checks this function makes really depend on whether or not the first argument to this function is `NULL`? Wouldn't it be better to let this function always execute all checks?

src/sfloader/fluid_defsfont.c
```diff
@@ -1268,23 +1450,548 @@ fluid_zone_is_mod_identic(fluid_mod_t *mod, char *name)
...
+ *  - valid complete paths (FLUID_PATH_CURRENT, FLUID_PATH_VALID set to 1) or
+ *  - invalid incomplete paths (FLUID_PATH_CURRENT set to 1, FLUID_PATH_VALID
+ *    set to 0).
+ *
+ * @return
+ *  - the number of linked modulators if any valid linked path exists.
+ *  - 0 if no linked path exists.
+*/
+static int
+fluid_check_linked_mod_path(char *list_name, fluid_mod_t *list_mod,
```
derselbst: Reminder for myself: Think of a better naming. `fluid_check_linked_mod_path` and `fluid_list_check_linked_mod` are too confusing.

jjceresa:
>To avoid any misunderstanding: This PR should not add and implement this new, potentially-implemented-in-the-future-API.

I am with you. Ofc, the new future API must be implemented in another PR. Actual PR should keep its functionality  (i.e adding complex modulator (from soundfont loader).

>To me fluid_list_check_linked_mod() actually looks quite suitable already to be useful for such a new API however. Ofc, if you have some more improvements in mind, go ahead.

Yes, some simple improvements are necessary. Particularly for performance reason when it  will be called from future API `fluid_voice_add_mod2()` (because this new API is expected be called in realtime situation at noteon time):
- Actually `fluid_list_check_linked_mod()` suffers of lack of performance due to memory allocation to return a linked list in `linked_mod`.
- The improvement (in next commit) is to add a `linked_count` parameter that indicates (if linked_count > 0) that `linked_mod` contains a table provided by the caller . This way the function return the linked list in this table without doing any memory allocation. Ofc, the table provided by the caller should be `vla` (or static) to keep maximum of performance.
If linked_count is 0, the function make internal allocation and return the list in linked_mod. This later case is appropriate when called from soundfont loader.

jjceresa:
More comments:

> static void `fluid_test_linked_mod_test_identity . It is not quite clear to me whether this function is supposed to just print some information to console ...`

Yes, it is only supposed to do  just print some information to console. The comment in the code isn't clear this will be fixed.
***
> line 1865    if(list_name != NULL)
Should the number of checks this function makes really depend on whether or not the first argument to this function is NULL? Wouldn't it be better to let this function always execute all checks?

Letting this function always execute all checks is better when calling from soundfont loader.

When the function will be called from future API fluid_voice_add_mod2() and for performance reasons (see comment above), it is better that the calling application be able to expect the best performance possible. Of course this is true in situation where the application know by advance that the complex modulator it intends to add to a voice is valid. In this case only minimum vital checks should be done.
The source comments need be completed. This will be fixed.
***
>static void fluid_dump_linked_mod(fluid_mod_t *mod, int offset)
Reminder for myself: This function is duplicated from fluid_defsfont.c.

To avoid duplication, perhaps this function in fluid_defsfont should be public, then  it will be callable from any tests. This function will be probably useful for future API tests.

***
>Reminder for myself: Think of a better naming. fluid_check_linked_mod_path and fluid_list_check_linked_mod are too confusing.

- `fluid_check_linked_mod_path()` is a low level function. It checks linked path only and return informations  in path table. It doesn't return a linked list.

- `fluid_list_check_linked_mod()` is a high level function. It checks and return a linked list.

For example, we could keep style naming function already existing: `fluid_name_verb()`
`fluid_check_linked_mod_path()` renamed `fluid_linked_path_check()`
`fluid_list_check_linked_mod()` renamed  `fluid_list_check_get_linked_mod()` 

similarly,
`fluid_limit_mod_list` renamed fluid_list_limit_mod`

derselbst:
> > static void fluid_dump_linked_mod(fluid_mod_t *mod, int offset)
> > Reminder for myself: This function is duplicated from fluid_defsfont.c.
> 
> To avoid duplication, perhaps this function in fluid_defsfont should be public, then it will be callable from any tests. This function will be probably useful for future API tests.

Sounds good to me.



> For example, we could keep style naming function already existing: `fluid_name_verb()`
> `fluid_check_linked_mod_path()` renamed `fluid_linked_path_check()`
> `fluid_list_check_linked_mod()` renamed `fluid_list_check_get_linked_mod()`
> 
> similarly,
> `fluid_limit_mod_list` renamed fluid_list_limit_mod`

I'm struggling with this one. Actually, the prefix `fluid_list_` is reserved for `fluid_list.c`. I will have to think about this, you can ignore it for now.

derselbst:
> * Actually `fluid_list_check_linked_mod()` suffers of lack of performance due to memory allocation to return a linked list in `linked_mod`.
> * The improvement (in next commit) is to add a `linked_count` parameter that indicates (if linked_count > 0) that `linked_mod` contains a table provided by the caller . This way the function return the linked list in this table without doing any memory allocation. Ofc, the table provided by the caller should be `vla` (or static) to keep maximum of performance.
>   If linked_count is 0, the function make internal allocation and return the list in linked_mod. This later case is appropriate when called from soundfont loader.

Good point regarding memory allocation. However, I would prefer to always demand a user-supplied buffer in this case. The caller can easily estimate the required memory by calling `fluid_get_count_mod()` and declare a VLA. IMO, this is better than doing any magic memory allocation inside `fluid_list_check_linked_mod()`, because allocated memory needs to be freed, which makes the calling convention for this function unnecessarily complicated.

jjceresa:
>However, I would prefer to always demand a user-supplied buffer in this case. The caller can easily estimate the required memory by calling fluid_get_count_mod() and declare a VLA....

1)When called from `soundFont loader`, the list of linked modulators returned need to be persistent during the life of the soundfont. Also the list can contain  any number of complex modulators. So this list should necessarily be memory allocated (it cannot be a VLA). Then the list will be freed at the when the preset will be deleted (as it it actually the case). 

2) When called from API `fluid_voice_add_mod2()` (adding a complex modulator to a voice). The list of linked modulator returned is just one complex modulator and need not to be persistent, it can be VLA.

3)When called from API `fluid_synth_add_default_mod2()` (add a default complex modulator to a synth). The list of linked modulator returned should be persistent during the life of the synth until the internal `default_mod` list be will be deleted. The case is similar to case (1).

4)When called from API `fluid_synth_remove_default_mod2()`, The list of linked modulator returned is just one complex modulator and need not to be persistent, it can be VLA. Similar to case 2.

This lead to 2 distinct situations:
- cases (1),(3) need persistent memory allocation.
- cases (2),(4) doesn't need memory allocation and can be VLA.

Please, note that memory allocation is done in `fluid_list_copy_linked_mod()` (not in fluid_list_check_linked_mod())

Despite this make the calling convention  a bit complicated, it is only an internal low level function not an API.

jjceresa:
> However, I would prefer to always demand a user-supplied buffer in this case. The caller can easily estimate the required memory by calling fluid_get_count_mod() and declare a VLA....

Your point would be preferable, but it seems that it isn't not possible:
1)When called from the `soundfont loader`, the list of linked modulators returned need to exist during the whole life of the soundfont's preset. Also the list can contain any number of complex modulators. So this list should necessarily be memory allocated (it cannot be a VLA). Then the list will be freed when the preset will be deleted (as it it actually the case).

2)When called from API `fluid_voice_add_mod2()` (adding a complex modulator to a voice). The list of linked modulators returned is short  and long life isn't not required , it can be a VLA.

3)When called from API `fluid_synth_add_default_mod2()` (adding a default complex modulator to the internal synth's `default_mod` list). The list of linked modulators returned should exist during the life of the synth until the `default_mod` list be will be deleted. The case is similar to case (1).

4)When called from API `fluid_synth_remove_default_mod2()`, The list of linked modulators returned is short and long life isn't not required , it can be a VLA. Similar to case 2.

This lead to 2 distinct situations:
    - cases (1),(3) need  memory allocation.
    - cases (2),(4) doesn't need memory allocation and can be VLAs.

Please, note that memory allocation is concentrated in `fluid_list_copy_linked_mod()` (not in fluid_list_check_linked_mod()). Freeing is done when any preset is deleted or when the synth is deleted.

Despite this make the calling convention a bit complicated, it is an internal low level function not an API.

derselbst:
Ok, go ahead. (Probably things will get clearer to me when actually seeing your changes.)

jjceresa:
Now, since commit `03d518e`, `fluid_list_check_linked_mod()` should be considerate appropriate
to be called from the soundfont loader or from the new api without difficulties (the implementation of these API  will be proposed in another PR).
However, some functions probably need better name or should be renamed for consistency.

jjceresa:
Now, since commit `03d518e`, `fluid_list_check_linked_mod()` should be considered appropriate
to be called from the soundfont loader or from the new api without difficulties (the implementation of these API will be proposed in another PR).
However, some functions probably need better name or should be renamed for consistency.

jjceresa:
Thanks for reviewing this.

According to last functions renaming, may I suggest following possible renaming also ?:
- fluid_limit_mod_list() --> fluid_mod_limit_list_count()
- fluid_list_check_linked_mod() --> fluid_mod_check_linked_mod()
- fluid_check_linked_mod_path() --> fluid_mod_check_linked_path()
- fluid_list_copy_linked_mod() --> fluid_mod_copy_linked_mod()

derselbst:
Your suggested renaming sounds good to me, I will take care of it.

I've just recognized that you are now using `alloca` in `fluid_list_check_linked_mod`. This is dangerous. A specially crafted soundfont (or an attacker via the future complex mod API) could abuse this and try to allocate `INT_MAX` modulators on the stack, which would result in a stack overflow. I'll therefore revert 3795c7dee9aad651bdffe9ccc57e3df94684f9ee.

jjceresa:
>I've just recognized that you are now using alloca in fluid_list_check_linked_mod.

Yes, this was chosen for speed reason for the future new complex mod API (for which it is intended to ensure that member count must be  < FLUID_NUM_MOD).

>A specially crafted soundfont (or an attacker via the future complex mod API) could abuse this and try to allocate INT_MAX modulators on the stack, which would result in a stack overflow.

Right, to avoid this issue, `fluid_list_check_linked_mod()` could  call  FLUID_MALLOC() when `linked_count` param is 0 , or `alloca()` otherwise. Before calling `alloca()` , `mod_count` need to be checked < FLUID_NUM_MOD. This leads to `fluid_list_check_linked_mod() being fast when called by future complex mod API and  memory secure when called by soundfont loader (or future complex mod API).

I will do that in next commit.

derselbst:
From my understanding, `fluid_list_check_linked_mod()` must cope with any invalid modulators. How can you limit it to `FLUID_NUM_MOD` (which is the job of `fluid_limit_mod_list()`) before even knowing whether there are invalid or even circular modulators?

jjceresa:
>How can you limit it to FLUID_NUM_MOD (which is the job of fluid_limit_mod_list()) before even knowing whether there are invalid or even circular modulators?

My comment isn't clear. You are right , it isn't the job of  `fluid_list_check_linked_mod()` to check that the count inside `list_mod` exceed internal limit FLUID_NUM_MOD.
- When called from future complex mod API, it is the job of these API to do this check before calling `fluid_list_check_linked_mod()` .
- When called from the soundfont loader, it is the job of `fluid_limit_mod_list()` to limit the count after calling `fluid_list_check_linked_mod()`.

derselbst:
> it isn't the job of fluid_list_check_linked_mod() to check that the count inside list_mod exceed internal limit FLUID_NUM_MOD. When called from future complex mod API, it is the job of these API to do this check before calling fluid_list_check_linked_mod() .

I disagree. `FLUID_NUM_MOD` is fluidsynth's limit of how many _valid_ modulators can be assigned to a voice. If the API is being called with `FLUID_NUM_MOD` invalid and `1` valid modulator, that single valid modulator should be assinged to the voice. I.e. we cannot limit the modulator count before `fluid_list_check_linked_mod()`, we must always do it after flagging the invalid ones. Just like it's done when called from the soundfont loader.

jjceresa:
The API in question is `"add one complex modulator to voice"`.
I just wanted to say that the API must ensure that the count inside `list_mod` must be reasonable before calling `fluid_list_check_linked_mod()`. This is because the API is (typically) called in real-time (in the context of "note on"). In this context,  the application should be compelled  to supply (sooner or later) only valid members for this complex modulator, and  `fluid_list_check_linked_mod()` must behave without memory allocation (for real-time performance). To cope with this case , for example it seems acceptable that the API return FAILED when the member count of this complex modulator is  above FLUID_NUM_MOD (or a reasonable multiple of FLUID_NUM_MOD). I  think that designing a complex modulator with less than 64 modulators inside should be sufficient.

>Just like it's done when called from the soundfont loader.
This isn't (typically) a real time context ("note on"), and as you said the soundfont could be crafted. In this case it seems preferable to reject invalid modulators in the soundfont, and accept that synthesizer remains still able to play only remaining valid modulators.

jjceresa:
The API in question is `"add one complex modulator to voice"`.
I just wanted to say that the API must ensure that the count inside `list_mod` must be reasonable before calling `fluid_list_check_linked_mod()`. This is because the API is (typically) called in real-time (in the context of "note on"). In this context, the application should be compelled to supply (sooner or later) only valid members for this complex modulator, and `fluid_list_check_linked_mod() must behave without memory allocation (for real-time performance). To cope with this case , for example it seems acceptable that the API return FAILED when the member count of this complex modulator is above FLUID_NUM_MOD (or a reasonable multiple of FLUID_NUM_MOD). I hope that designing a complex modulator with less than 64 modulators inside should be usually sufficient :).

>.. Just like it's done when called from the soundfont loader.

This isn't (typically) a real time context ("note on"), and as you said the soundfont could be crafted, a. In this case it seems preferable to reject invalid modulators in the soundfont, and accept that synthesizer remains still able to play only remaining valid modulators.

jjceresa:
>The API in question is "add one complex modulator to voice"

An implementation of this API and friends will be proposed in a new branch (`complex-mod-api`) sooner.

derselbst:
I understand your point. The problem I see is that `fluid_list_check_linked_mod()` requires memory allocation given the way it is currently implemented. And using `alloca` is just a hack to conceal this fact.

If you want to completely avoid memory allocation in `fluid_list_check_linked_mod()`, why not adding an `enum fluid_path_flags path` member to the `fluid_mod_t` struct? That way you could store those `FLUID_MOD_VALID` bit flags directly in the modulators itself.

jjceresa:
>If you want to completely avoid memory allocation in fluid_list_check_linked_mod(), why not adding an enum fluid_path_flags path member to the fluid_mod_t struct? That way you could store those FLUID_MOD_VALID bit flags directly in the modulators itself.

This would be by far a more elegant implementation. Also, that way the caller of `fluid_list_check_linked_mod()` could be informed that supplied modulators have been found invalid. This will allow keeping `amount` field intact (rather than forced to 0).

I will add this new `enum fluid_path_flags path` field to the `fluid_mod_t` struct.

derselbst:
There is a code duplication of `fluid_defpreset_noteon_add_mod_to_voice` and `fluid_defpreset_noteon_add_linked_mod_to_voice`. The only real difference I can see is the call to  `fluid_mod_test_linked_identity()`. The other differences are removings that are just costmetic issues, see below.

```diff
static void
fluid_defpreset_noteon_add_mod_to_voice(fluid_voice_t *voice,
                                        fluid_mod_t *global_mod, fluid_mod_t *local_mod,
                                        int mode)
{
    fluid_mod_t *mod;
    /* list for 'sorting' global/local modulators */
    fluid_mod_t *mod_list[FLUID_NUM_MOD];
    int mod_list_count, i;

    /* identity_limit_count is the modulator upper limit number to handle with
     * existing identical modulators.
     * When identity_limit_count is below the actual number of modulators, this
     * will restrict identity check to this upper limit,
     * This is useful when we know by advance that there is no duplicate with
     * modulators at index above this limit. This avoid wasting cpu cycles at
     * noteon.
     */
    int identity_limit_count;

    /* Step 1: Local modulators replace identic global modulators. */

    /* local (instrument zone/preset zone), modulators: Put them all into a list. */
    mod_list_count = 0;

    while(local_mod)
    {
        /* As modulators number in local_mod list was limited to FLUID_NUM_MOD at
           soundfont loading time (fluid_limit_mod_list()), here we don't need
           to check if mod_list is full.
         */
        mod_list[mod_list_count++] = local_mod;
        local_mod = fluid_mod_get_next(local_mod);
    }

    /* global (instrument zone/preset zone), modulators.
     * Replace modulators with the same definition in the global list:
     * (Instrument zone: SF 2.01 page 69, 'bullet' 8)
     * (Preset zone:     SF 2.01 page 69, second-last bullet).
     *
     * mod_list contains local modulators. Now we know that there
     * is no global modulator identic to another global modulator (this has
     * been checked at soundfont loading time). So global modulators
     * are only checked against local modulators number.
     */

    /* Restrict identity check to the number of local modulators */
    identity_limit_count = mod_list_count;

    while(global_mod)
    {
        /* 'Identical' global modulators are ignored.
         *  SF2.01 section 9.5.1
         *  page 69, 'bullet' 3 defines 'identical'.  */

        for(i = 0; i < identity_limit_count; i++)
        {
-            if(fluid_mod_test_identity(global_mod, mod_list[i]))
+            if(fluid_mod_test_linked_identity(global_mod, mod_list[i], FLUID_LINKED_MOD_TEST_ONLY))
            {
                break;
            }
        }

        /* Finally add the new modulator to the list. */
        if(i >= identity_limit_count)
        {
            /* Although local_mod and global_mod lists was limited to
               FLUID_NUM_MOD at soundfont loading time, it is possible that
               local + global modulators exceeds FLUID_NUM_MOD.
               So, checks if mod_list_count reachs the limit.
            */
-            if(mod_list_count >= FLUID_NUM_MOD)
-            {
                /* mod_list is full, we silently forget this modulator and
                   next global modulators. When mod_list will be added to the
                   voice, a warning will be displayed if the voice list is full.
                   (see fluid_voice_add_mod_local()).
                */
-                break;
-            }

            mod_list[mod_list_count++] = global_mod;
        }

        global_mod = fluid_mod_get_next(global_mod);
    }

    /* Step 2: global + local modulators are added to the voice using mode. */

    /*
     * mod_list contains local and global modulators, we know that:
     * - there is no global modulator identic to another global modulator,
     * - there is no local modulator identic to another local modulator,
     * So these local/global modulators are only checked against
     * actual number of voice modulators.
     */

    /* Restrict identity check to the actual number of voice modulators */
    /* Acual number of voice modulators : defaults + [instruments] */
    identity_limit_count = voice->mod_count;

    for(i = 0; i < mod_list_count; i++)
    {

        mod = mod_list[i];
        /* in mode FLUID_VOICE_OVERWRITE disabled instruments modulators CANNOT be skipped. */
        /* in mode FLUID_VOICE_ADD disabled preset modulators can be skipped. */

-        if((mode == FLUID_VOICE_OVERWRITE) || (mod->amount != 0))
        {
            /* Instrument modulators -supersede- existing (default) modulators.
               SF 2.01 page 69, 'bullet' 6 */

            /* Preset modulators -add- to existing instrument modulators.
               SF2.01 page 70 first bullet on page */
            fluid_voice_add_mod_local(voice, mod, mode, identity_limit_count);
        }
    }
}
```

jjceresa:
>There is a code duplication of fluid_defpreset_noteon_add_mod_to_voice and fluid_defpreset_noteon_add_linked_mod_to_voice.

Yes, I am aware of this duplication and will review this.

mawe42:
Thanks a lot to both of you for working on this, I think it will add some great new expressive capabilities to Fluidsynth!

I'm now trying to get my head around these linked modulators and to understand what we could do with them musically. I'll try to explain it to myself here... maybe you can tell me if I got it right or wrong :-)

The Soundfont spec only has a very short section on linked modulators, but there is also a picture that explains the concept (shown here along with the depiction of the modulator structure on the right):
sf24spec-mod-chaining

So for the example in the picture on the left, for both mod 1 and mod 2 the output value is calculated by taking the both inputs (src1 and src2) in _native units_, using the normalisation to convert them into the [-1;1] range, multiply them with the "amount" in _destination units_, passing them through the output transform (which we don't currently support) and then on to the summing node. The resulting value is then used as input to the normalisation stage of mod 3.

As mod 1 and mod 2 do not target a generator, they don't have direct _destination units_. But I guess it makes sense that all modulators in the graph share the same _destination units_: the units of the generator that the final modulator targets.

Now to the thing that tripped me up: I think the current implementation skips the input value normalisation stage for values coming from linked modulators. This is problematic because:
- we loose expressiveness because the final modulator (mod 3) can not make use of input mapping curves (unipolar positive, bipolar negative etc)
- the values coming from the linked modulators are already multiplied with the amount in _destination units_, meaning they might be outside of the normalised [-1;1] range.

If that really is the case and I haven't read the code wrong, then I would propose that we change the behaviour so that the input normalisation for values coming from linked modulators is not skipped. That means we have to have some way of knowing the min/max values to map them to the [-1;1] space. For that I see two possible solutions:
1. Limit the "amount" field of modulators that target another modulator to the [-1;1] range
2. Use the min/max range of the "destination units"

Solution 2. seems to be more intuitive, especially when I put my Soundfont designer hat on. But I'm not sure. The linked modulator concept as described in the Soundfont spec seems like it hasn't been fully thought through... or maybe I simply didn't understand it properly.

Sorry, very long post. But it's a really interesting but quite complicated beast :-)

mawe42:
One more comment: unless I missed something in the spec, then linked modulators are a little "under specified". Meaning that there is no single correct way to implement linked modulators. So what we're adding now is the "fluidsynth way", but there might be other and different implementations as well. Not necessarily a bad thing, I don't think there are other SF synth with support for linked modulators out there so we could be the "leader of the pack" here :-) But something we should be aware of when announcing this feature, I think.

derselbst:
> I think the current implementation skips the input value normalisation stage for values coming from linked modulators.

You are correct:

https://github.com/FluidSynth/fluidsynth/blob/664e97305a33ec5b6ba3404d74bae88d20e67277/src/synth/fluid_mod.c#L500-L504

The reason is, like you already said, that we don't have min/max ranges to perform normalization. 

Just to make it clear: You are now suggesting to pull that absolute non-normalized `src1 input` through all linked modulators until we come to the very last modulator which is linked to a destination generator, and then performing normalization according to the min/max of section `8.1.3 Generator Summary` ? In other words: What if modulator3 in the example had a descendant linked modulator4? How to treat the `src1 input` received by modulator3?

mawe42:
> Just to make it clear: You are now suggesting to pull that absolute non-normalized src1 input through all linked modulators until we come to the very last modulator which is linked to a destination generator, and then performing normalization according to the min/max of section 8.1.3 Generator Summary ?

No quite... my proposal 2. meant that we keep the implementation as suggested in the picture. In other words, that each modulator runs through all stages, including output multiplication with "amount" in _destinaton units and input normalisation back to [-1,1].  And that the min/max ranges for all input normalisation is taken from the generator that is the target of the whole graph.

But like I said, I'm not sure that is even possible for all generators. For example, what max value would we use for the SampleOffset generators? Maybe it would be possible to analyse the linked modulators to figure out their min/max values programatically from the configuration of the modulator?

The other option would be to skip the input normalisation stage (maybe really just normalisation but keep the mapping curves) for all modulators in the graph that take input from other modulators. But that would also require that we ignore the "amount" field for all but the last modulator, or force the amount to be in the [-1;1] range. In other words, the _destination units_ for linked modulators would be [-1,1].

derselbst:
>  And that the min/max ranges for all input normalisation is taken from the generator that is the target of the whole graph.

Min/max are usually defined by the source controllers. Doing it now the other way around just because we are dealing with linked modulators feels very unnatural and artificial to me.

> For example, what max value would we use for the SampleOffset generators?

Defined by the sample itself. And I'm afraid introducing a unique `fluid_mod_t` to `fluid_voice_t` to `fluid_sample_t` dependency is really challenging.

> Maybe it would be possible to analyse the linked modulators to figure out their min/max values programatically from the configuration of the modulator?

We could do anything, but after all, it must still be comprehensible to the soundfont designer how that complex modulator behaves. And walking thourgh all members to find out min/max (which I believe the designer doesn't even know himself) would be contraproductive, I believe.

---

I think what we need to make design decisions here are one or two real-world examples of how complex modulators can be used. Something useful that cannot be achieved with simple modulators.

mawe42:
> Min/max are usually defined by the source controllers. Doing it now the other way around just because we are dealing with linked modulators feels very unnatural and artificial to me.

Yes, it is definitely weird. But the whole linked modulators thing is weird from the start and doesn't really fit the rest of the modulator system design.

Maybe the most straight forward and easiest to understand approach would be to state that "destination units" of modulators that target another modulator are in the [-1,1] range. And have linked modulators apply the input mapping curves but leave the range untouched (just limit the values to [-1,1] to be on the safe side.

> I think what we need to make design decisions here are one or two real-world examples of how complex modulators can be used. Something useful that cannot be achieved with simple modulators.

Yes, that would be good. I'm still struggling to come up with a good use-case.

mawe42:
> Yes, that would be good. I'm still struggling to come up with a good use-case.

Thinking more about this, I think the way linked modulators are (under-)specified means there really is nothing they add to the expressiveness of a Soundfont synth. The only thing they add is what JJC has already mentioned: you could create multiple modulators that target the same generator and switch them all on/off with a secondary source on the final modulator. But as he also said: you can also do that without linked modulators. It's just a few more clicks.

So to be honest... I'm not convinced that that is an important enough use case to justify the added code complexity and possible performance impact that this feature might add.

But maybe @jjceresa can think of something that I've missed that might justify it anyway?

jjceresa:
>we loose expressiveness because the final modulator (mod 3) can not make use of input mapping curves (unipolar positive, bipolar negative etc)...

Not really. Expressiveness comes from CCs sources connected to mod1 and mod2 with their respective mapping curves. This expressiveness is still present at mod3's link input and resulting value (at link input) shouldn't  be mapped again because this new mapping could break the one done before. It would be quickly impossible for a designer to predict what could be the result of : mapx(...map3(map2((map1(CCs sources)))). (As already said all must still comprehensive for the soundfont designer). As any data at the output of a modulator (i.e mod1, mod2) is expressed in destination units (i.e the one of final generator), resulting value at link input should be still considered in destination unit. In fact mod3's link input node behaves similarly that the summing input node of this generator.
That means that mod1 and  mod2 destination could be directly connected to the final generator instead passing through another modulator.

That said "when could we need to  pass the output of a well designed modulator `mx` through another modulator `my` before reaching the ending generator ?". The first answer could be "each time we need to control the effect of this modulator `mx` to the generator without impacting the modulator `mx` itself".

For example, this case occurs at performance time when the musician want to select a "sound articulation" among others, on the fly while he is playing (without express need to select another preset). (These "sound articulations" will be called "effects" here).
Let a set of "effect"  {e1,e2}  for the current preset selected. 
Each "effect" is represented by one ore more modulators but for simplicity we assume here there is only one modulator per effect (i.e `m1` for e1 and `m2` or e2) .
Each effect is controlled on input by the musician by a set of  foot CC. For simplicity we assume that each effect is controlled by only one foot CC (on mod's src1) and a rotary knob (on mod's src2). The goal of the rotary knob being to set a variation of the effect before starting playing and this setting is independent of the foot CC.

The musician have a palette of 2 effects (e1,e2) and one foot CC and want to be able to play one effect at a time with the same foot CC. While the musician plays with this hands he can also use the foot CC that  plays the current selected effect (e.g e1). While continuing playing with one hand, he use the other hand to select  another effect (e2), effect e1 will be substituted by effect e2 (i.e m1 is disabled and m2 is enabled).

Effect selection logic can use a linked modulator `me1` inserted between `m1` output  and the relevant generator. Then, another linked modulator `me2` is inserted between `m2 output  and the relevant generator. 
Then `me1`s src2` and `me2 src2` inputs are feed in opposite direction by the appropriate selection CC.

This example shows that `m1` and `m2` modulator need both controlled without modification of any respective sources (src1,src2). Of course only destination field need to be changed at design time.

>So to be honest... I'm not convinced that that is an important enough use case to justify the added code complexity and possible performance impact that this feature might add.

No problem. Please, if you try this branch on your ARM machine let me know about any possible performance impact.

>Given all the new questions that arose while working on this, I agree. I think would make sense to unassign it from version 2.1 . If need be we can continue work at any time and eventually ship it for a later release.

I am fine with this choice.

jjceresa:
>While compiling this branch I noticed that gcc complains about a possible uninitialized variable (only with release type RelWithDebInfo):
fluidsynth/src/synth/fluid_mod.c: In function ‘fluid_mod_copy_linked_mod’:
fluidsynth/src/synth/fluid_mod.c:1573:36: warning: ‘mod_cpy’ may be used uninitialized

This is just a warning, the variable is really initialized.

mawe42:
Sorry, very late reply... 

> As already said all must still comprehensive for the soundfont designer

Yes, and that is probably my main concern with linked modulators: they are not well defined in the spec. So Soundfont designers will effectively create a "Fluidsynth Version" of their Soundfont that targets the way we implement linked modulators and probably only sounds correct in Fluidsynth. But maybe that doesn't matter as we are the only synth with actual support for linked modulators. In any case, if we decide to actually release this feature, we need to document our way of supporting linked modulators (because reading the SF2 spec is not helpful).

> For example, this case occurs at performance time when the musician want to select a "sound articulation" among others, on the fly while he is playing (without express need to select another preset).

That is a good example of a possible use-case, thanks! And for that special case, linked modulators are one possible solution. But my concern is exactly that: that it's just one of the possible solutions. Other solutions would be to use a hardware MIDI controller with "setup banks", like many controllers have. So you can configure, store and switch quickly which pedal or input sends which MIDI messages. Or you could use a software based MIDI router to filter and modify MIDI messages from your MIDI controller. Or you could implement different presets with different modulator sources in the Soundfont.

It all depends on the actual *real-world* use-case, I guess. And I might be wrong here, but your given use-case sounds theoretical, not driven by a real-world need that you or others have. So effectively we are implementing a solution that is looking for a problem it can solve. And a solution that is not well specified in the Soundfont spec.

Please don't get me wrong: I am really grateful for all the work you are putting into this feature! And I am strongly for extending and improving Fluidsynths expressiveness. And generally I'm in favour of linked modulators... I just don't know how to overcome the problem that the specification is unclear and the fact that we seem to have no real-world application for this feature.

jjceresa:
>my main concern with linked modulators: they are not well defined in the spec.

They are simply defined as a way to `route` modulator m1's output to the summing link input of another modulator m2, instead of routing m1's output to the summing input of a generator.

>In any case, if we decide to actually release this feature, we need to document our way of supporting linked modulators (because reading the SF2 spec is not helpful).

yes, the way linked modulators are supported need to be documented. 

>Other solutions would be to use a hardware MIDI controller with "setup banks", like many controllers have. So you can configure, store and switch quickly which pedal or input sends which MIDI messages. Or you could use a software based MIDI router to filter and modify MIDI messages from your MIDI controller.

I  see all this things well suited only for hardware configurations making MIDI controllers (Keyboard, CC pedal...) independent of the `synthesizers` and `soundfont preset` being intended to be played later in real time by the musician.

Which is interesting is to be able to define which  CC number will be used and what is the destination route of this CC (modulator or generator)  for the current preset. Having this `routing information` being a `property` of the `current soundfont preset` allows the designer having two "hats" jobs. With the first "hat"  the designer define the "preset instrument data" (i.e the sound of the instrument). With the second hat, the designer and the musician (intended to play the preset) define the "routing information" (linked modulators) intended to be selectable in real-time by the musician during the song. All these information  ("preset instrument data", "routing data") are part of the preset stored in the soundfont file format.  
The same application having knowledge of soundfont format can be used to aggregate these two "hats" jobs. This will eliminate others routing applications (incorporated between MIDI keyboard and synthesizers) with the difficulty to correlate different `routing configurations files` with `soundfont files`.

>And I am strongly for extending and improving Fluidsynth expressiveness. And generally I'm in favour of linked modulators... I just don't know how to overcome the problem that the specification is unclear and the fact that we seem to have no real-world application for this feature.

Expressiveness supplied by the musician is captured by the first modulator which has a CC or (GC) on source 1 . For me, linked modulators are there for `easy control effect`  purpose during real-time playing of the current note in  the current preset. I have always considered this feature important and not special.

>I just don't know how to overcome the problem that the specification is unclear..

Addressing modulators, soundfont specifications aren't not obvious and and this is why Emu write another document (sfapp21.pdf SoundFont 2.1 Application Note).
In this document Emu. Reading chapter "Compared to the Pros" should help to understand that Specification 2.1 should not be considered complete.

For example Soundfont spec suffers of lack of important generators like "crossfading velocity range" and "crossfading key range". Theses generators very easy to implement allows instrument zones to overlapp partially and crossfade smoothly at  the beginning and ending of zones. This simple things brings a serious expressiveness enhancement by making instruments more realistic particularly for monophonic instrument playing legato.

derselbst:
Postponed indefinitely.

jjceresa:
Yes, I postponed this. Mainly because there is only Polyphone that  supports linked modulators.
I noticed that Polyphone is unable to install to previous version of Windows. The authors said that this involves to much work to make installation program for Windows XP and older.
To overcome this issue, here I am adding support to swami for linked modulators.
Once this Swami support finished, I will continue this branch.

PR #516, opened 2019-02-26
Improve systemd integration

fleger:
When fluidsynth is run as a service using systemd, make sure
the service is considered started only when it is ready to process events.

In order to do so:
 - Add an optional runtime dependency to libsystemd to the fluidsynth executable
 - Change the systemd service type to "notify"
 - Have fluidsynth notify systemd that the service is started after the server is started
 - Have fluidsynth notify systemd that the service is stopping after joining the server thread


src/fluidsynth.c
```diff
@@ -954,10 +963,13 @@ int main(int argc, char **argv)
             fluid_server_join(server);
         }
 
+#ifdef SYSTEMD_SUPPORT
+        sd_notify(0, "STOPPING=1");
```
derselbst: Shouldn't this `STOPPING` notification happen before `fluid_server_join()`?

fleger: I've tried that and it doesn't work because the service would be in stopping state soon after it is started and systemd would ultimately kill it.

STOPPING=1 is used to tell the service manager that the service is beginning its shutdown.
fluid_server_join is a blocking call that waits for the server thread to quit.
systemd stops the fluidsynth process by sending a SIGTERM signal that causes the server thread to exit, which in turn makes the call to fluid_server_join return.
So from systemd's perspective the shutdown sequence starts _after_ the call to fluid_server_join.

derselbst:
(cherry-picked to 2.0.x)

fleger:
You're welcome.

PR #527, opened 2019-04-14
Fix buffering bug in fluid_synth_process().

swesterfeld:
The old buffering code assumes that synth->cur is between 0 and FLUID_BUFSIZE.
However since fluid_synth_process() can render more than one buffer at a time
this isn't always true, and the new code handles other values properly.

I found this bug while porting the deprecated fluid_synth_nwrite_float() calls in BEAST to 
the non-deprecated fluid_synth_process(). However, if I do this, our soundfont unit test fails, and indeed the wave starts sounding somewhat crappy. BEAST will call fluid_synth_process() with arbitary buffer sizes, so for instance it can do

```
...
process() 200
note on
process() 100
note off
process() 200
...
```

So these are note a multiple of FLUID_BUFSIZE (64), with fluid_nwrite_float() it works but with fluid_synth_process() it fails.

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3644,7 +3644,7 @@ fluid_synth_process(fluid_synth_t *synth, int len, int nfx, float *fx[],
     int nfxchan, nfxunits, naudchan;
 
     double time = fluid_utime();
-    int i, f, num, count;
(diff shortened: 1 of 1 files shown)


derselbst:
Indeed, thanks a bunch! (cherry-picked this to 2.0.x branch and smashed together a regression test)

PR #533, opened 2019-04-21
[Android] disable sdl2 and enable sndfile at cmake.

atsushieno:
The latest master looks for sdl2 via pkg-config, which needs to be explicitly disabled for Android build.
Also there was missing cmake argument to enable sndfile (overlooked when backporting changes from my amidi branch).

### DIFF
--- a/doc/android/Makefile.android
+++ b/doc/android/Makefile.android
@@ -95,7 +95,7 @@ build-fluidsynth:
 build-fluidsynth-one:
 	mkdir -p build/$(A_ABI) && cd build/$(A_ABI) && \
 	LD_RUN_PATH=$(DIST_PATH)/android-$(BUILD_ABI)/lib:$(OBOE_BUILD_PATH)/$(A_ABI) LD_LIBRARY_PATH=$(DIST_PATH)/android_$(BUILD_ABI)/lib PKG_CONFIG_PATH=$(DIST_PATH)/android_$(BUILD_ABI)/lib/pkgconfig/:$(OBOE_BUILD_PATH)/$(A_ABI) \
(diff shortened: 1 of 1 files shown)


derselbst:
> The latest master looks for sdl2 via pkg-config, which needs to be explicitly disabled for Android build.

I guess that's a cerbero specific reason why it needs to be disabled?

atsushieno:
Not really. it is because pkg-config is NOT a first class citizen in Android NDK world, and any packages that is in the *host* path will be taken into consideration. Normal Android NDK projects don't use pkg-config and usually this kind of problem does not happen, but Fluidsynth depends on it. So we have to live with it.

derselbst:
> any packages that is in the host path will be taken into consideration.

Yes, but that's why `PKG_CONFIG_PATH` and `PKG_CONFIG_LIBDIR` should be overridden when cross-compiling.

Anyway, thx for explaining.

atsushieno:
Hmm, yeah that's true. I thought those toolchains implicitly specify their sysroots somewhere, but now that the clang/llvm stuff is likely the identical ones as Linux host tools with little tricks, it might be possible to have PKG_CONFIG_PATH not adding the host sysroot. Although it is [explicitly specified without adding local paths](https://github.com/FluidSynth/fluidsynth/blob/656fe6e2d9c61688dae104466891857fa2af90c8/doc/android/Makefile.android#L97) and I have no further idea to limit more...

If I could figure out more I will create another PR.

derselbst:
> Although it is explicitly specified without adding local paths and I have no further idea to limit more...

This only specifies `PKG_CONFIG_PATH`. You need to specify `PKG_CONFIG_LIBDIR` as well, see comment 2:

https://bugs.freedesktop.org/show_bug.cgi?id=88992#c2

atsushieno:
Thanks, now it's simplified in the new PR.

PR #536, opened 2019-06-08
Make const strings

carlo-bramini:
By looking the map file, I noticed that some strings were allocated in RAM although they were supposed to be in the const area. This PR resolves it.


derselbst:
Ok, thanks.

Issue #537, opened 2019-06-10
Introduce usage of float math functions.


carlo-bramini:
Hello,
I wanted to ask what do you think about adding support for float math functions when `WITH_FLOAT` is defined. In this case, Fluidsynth may use something like this:

```
#ifdef WITH_FLOAT
#define FSIN(_x)    sinf(_x)
#define FCOS(_x)    cosf(_a)
#define FLOG(_x)    logf(_x)
...etc
#else
#define FSIN(_x)    sin(_x)
#define FCOS(_x)    cos(_a)
#define FLOG(_x)    log(_x)
...etc
#endif
```
These macros to functions could be used when the required type is `fluid_real_t` and they may be able to improve a bit the performance: in the current code there are several points where the traditional functions will not provide any additional precision because the parameters are promoted to `double` but the result is typically converted to `float`.
Of course, cmake is supposed to test if these functions exist before using them.
Opinions?

Sincerely.

jjceresa:
Hello,
Assuming this will be supported, the corresponding CMake option should have a "performance optimization" meaning. I just want to say that the user shouldn't be confused between "precision" meaning given by actual `enable-floats` option and "optimization" option given by this new feature.

derselbst:
> Opinions?

Makes sense to me, esp. for embedded system which may have their own optimized implementation of those C99 single precision functions. IIR filter, chorus and reverb come to my mind, anything else?

Seems like a good proposal to me. For consistency however we should call it `FLUID_SIN `rather than `FSIN`.

IMO it would be best to use `CMAKE_C_COMPILE_FEATURES` and search for `c_std_99`. Unfortunately this would require to bump `cmake_minimum_required` to 3.8. Too big step. Perhaps better use [CheckSymbolExists](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html#module:CheckSymbolExists) whether sinf, cosf and friends are available.

> the corresponding CMake option should have a "performance optimization" meaning

I don't see a need introduce more cmake build options here. If the user actively enables floats, he is aware of a potential loss in precision which on the other hand may cause some (severe) optimizations, e.g. avoiding the need to calculate trigonometric functions up to double precision while truncating the result to single precision.

carlo-bramini:
> Seems like a good proposal to me. For consistency however we should call it FLUID_SIN rather than FSIN.

Ok.

> I don't see a need introduce more cmake build options here. If the user actively enables floats, he is aware of a potential loss in precision.

[...]

PR #538, opened 2019-06-11
Fix math 1

carlo-bramini:
These fix are preliminar before fixing #537.

1) Since `log(x)=log10(x)*log(10)`, it seems absolutely redundant and harmful that code to me.
2) math.h should also provide M_PI_2 macro, which is M_PI/2. I added the test for its presence and used it in few points.

### DIFF
--- a/src/rvoice/fluid_rev.c
+++ b/src/rvoice/fluid_rev.c
@@ -773,7 +773,7 @@ static void update_rev_time_damping(fluid_late *late,
             /* gi = f(roomsize, gi_max, gi_min) */
             gi_tmp = gi_min + roomsize * (gi_max - gi_min);
             /* Computes T60DC from gi using inverse of relation E2.*/
-            dc_rev_time = -3 * delay_length[NBR_DELAYS - 1] * sample_period / log10(gi_tmp);
+            dc_rev_time = -3 * M_LN10 * delay_length[NBR_DELAYS - 1] * sample_period / log(gi_tmp);
         }
 #endif /* ROOMSIZE_RESPONSE_LINEAR */
         /*--------------------------------------------
@@ -782,7 +782,7 @@ static void update_rev_time_damping(fluid_late *late,
         /* Computes alpha from damp,ai_tmp,gi_tmp using relation R */
         /* - damp (0 to 1) controls concave reverb time for fs/2 frequency (T60DC to 0) */
         ai_tmp = 1.0 * damp;
-        alpha = sqrt(1 / (1 - ai_tmp / (20 * log10(gi_tmp) * log(10) / 80))); /* R */
+        alpha = sqrt(1 / (1 - ai_tmp / (20 * log(gi_tmp) / 80))); /* R */
     }
 
     /* updates tone corrector coefficients b1,b2 from alpha */
@@ -806,7 +806,7 @@ static void update_rev_time_damping(fluid_late *late,
                                             sample_period / dc_rev_time);
 
         /* iir low pass filter feedback gain */
-        fluid_real_t ai = (fluid_real_t)(20 * log10(gi) * log(10) / 80 *
+        fluid_real_t ai = (fluid_real_t)(20 * log(gi) / 80 *
                                          (1 -  1 / pow(alpha, 2)));
         /* b0 = gi * (1 - ai),  a1 = - ai */
         set_fdn_delay_lpf(&late->mod_delay_lines[i].dl.damping,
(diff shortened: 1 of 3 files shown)


derselbst:
Looks good. But do we really need `M_PI_2`? An optimizing compiler should do constant folding and optimize `M_PI / 2` away.

carlo-bramini:
I introduced `M_PI_2` for reducing the numbers because they also need to be converted to float when `WITH_FLOAT` is defined.
My idea was to do something like this:
```
#ifdef WITH_FLOAT
#define FLUID_M_PI    M_PI##f
...etc
#else
#define FLUID_M_PI    M_PI
...etc
#endif
```
and make some macros in this way but I can do it in a different way, if you prefer.

src/synth/fluid_mod.c
```diff
@@ -334,21 +334,21 @@ fluid_mod_transform_source_value(fluid_real_t val, unsigned char mod_flags, cons
      * is close enough.
      */
     case FLUID_MOD_SIN | FLUID_MOD_UNIPOLAR | FLUID_MOD_POSITIVE: /* custom sin(x) */
-        val = sin(M_PI / 2 * val_norm * 0.87);
+        val = sin(M_PI_2 * val_norm * 0.87);
```
derselbst: Aha, you want to avoid promotion of the literals to double, causing the whole expression to be evaluated as double, although `WITH_FLOAT` is defined. But then you could have done:

```suggestion
        val = sin((M_PI / 2.0f) * val_norm * 0.87f);
```

derselbst:
Ah, scratch that thought about static const. Including a bunch of static variables would pollute every compilation unit, ofc. Alright, stay with macros.

src/synth/fluid_mod.c
carlo-bramini: This change did not removed the double calculation from my code because M_PI. Unfortunately, I had to change it to something like this:

`val = sin((fluid_real_t)(M_PI / 2.0  * 0.87) * val_norm);`

When compiling for an ARM Cortex with VFPv4-SP FPU, which supports only single precision numbers, this effect was made very visible.

carlo-bramini:
You are right, it is much simpler to do something like this:

`#define FLUID_M_PI    ((fluid_real_t)M_PI)`

rather than making an `#ifdef...#else...#endif`

src/synth/fluid_mod.c
derselbst: Sure, if M_PI is double, it would be promoted to double again. Perhaps this explicit cast as you said
 
```
val = sin((fluid_real_t)(FLUID_M_PI / 2.0 * 0.87) * val_norm);
```

would be somewhat strange / redundant / hard to read if `FLUID_M_PI` is defined as cast itself:

```
val = sin((fluid_real_t)(((fluid_real_t)M_PI) / 2.0 * 0.87) * val_norm);
```

So, I have a slight preference for:

```
val = sin((FLUID_M_PI / 2.0f * 0.87f) * val_norm);
```

Btw, well spotted to place the 0.87 next to PI to allow proper constant folding and saving one multiplication at runtime.

src/rvoice/fluid_rev.c
```diff
@@ -782,7 +782,7 @@ static void update_rev_time_damping(fluid_late *late,
         /* Computes alpha from damp,ai_tmp,gi_tmp using relation R */
         /* - damp (0 to 1) controls concave reverb time for fs/2 frequency (T60DC to 0) */
         ai_tmp = 1.0 * damp;
-        alpha = sqrt(1 / (1 - ai_tmp / (20 * log10(gi_tmp) * log(10) / 80))); /* R */
+        alpha = sqrt(1 / (1 - ai_tmp / (20 * log(gi_tmp) / 80))); /* R */
```
derselbst:
```suggestion
        alpha = sqrt(1 / (1 - ai_tmp / ((20 / 80) * log(gi_tmp)))); /* R */
```

src/rvoice/fluid_rev.c
```diff
@@ -806,7 +806,7 @@ static void update_rev_time_damping(fluid_late *late,
                                             sample_period / dc_rev_time);
 
         /* iir low pass filter feedback gain */
-        fluid_real_t ai = (fluid_real_t)(20 * log10(gi) * log(10) / 80 *
+        fluid_real_t ai = (fluid_real_t)(20 * log(gi) / 80 *
```
derselbst:
```suggestion
        fluid_real_t ai = (fluid_real_t)((20 / 80) * log(gi) *
```

carlo-bramini:
Argh, I accidentally deleted my fork with these changes!
I will do a new one with the fixes that you suggested me.
I'm sorry, please excuse me.

derselbst:
> I accidentally deleted my fork with these changes

You do know that you have write permissions to this repo, do you? You just have to create a new branch for your commit(s), since you can't commit to master directly.

carlo-bramini:
Sorry, just to confirm... do you prefer to use `(M_PI / 2.0f)` rather than using `M_PI_2` macro, right?

derselbst:
Yes, I prefer  `(FLUID_M_PI / 2.0f)` .

PR #542, opened 2019-06-22
Restructure cmake build summary

derselbst:
Now looks like:

```
**************************************************************
Build Summary:
Build type:            RelWithDebInfo
Install Prefix:        /usr

Audio / MIDI driver support:
  ALSA:                  yes
  CoreAudio:             no
  CoreMIDI:              no
  DSound:                no
  JACK:                  yes
  MidiShare:             no
  Oboe:                  no
  OpenSLES:              no
  OS/2 DART:             no
  OSS:                   yes
  PortAudio:             no
  PulseAudio:            yes
  SDL2:                  yes
  WaveOut:               no
  WinMidi:               no

Support for SF3 files:   no (libsndfile has no ogg vorbis support)
Support for DLS files:   no (libinstpatch not found)

Audio to file rendering: yes
  libsndfile:            yes

Miscellaneous support:
  D-Bus:                 yes
  LADSPA support:        yes
  LASH support:          yes (NOTE: GPL library)
  NETWORK Support:       yes
    IPV6 Support:        yes
  Readline:              yes (NOTE: GPL library)
  systemd:               yes

Developer nerds info:
  Samples type:          double
  Multithread rendering: yes
  OpenMP 4.0:            yes
  Profiling:             no
  Debug Build:           no
  Trap on FPE (debug):   no
  Check FPE (debug):     no
  UBSan (debug):         no

**************************************************************
```

Also drops the useless `Windows: yes/no` hint and makes sure to write the build summary to stdout rather than stderr.


jjceresa:
>Now looks like:

This new thematic classification of CMake options is important. Nice things, thanks you.
I guess that `Sample type` option should replace actual `enable float` option currently in discussion:  https://github.com/FluidSynth/fluidsynth/issues/537.

derselbst:
> I guess that Sample type option should replace actual enable float option currently in discussion: 

`Sample type` is just the representation of the `enable-float` option. However, I'm not intending to change any cmake options at this time. One might think about a better wording than "sample type".

jjceresa:
>One might think about a better wording than "sample type"

May be `internal sample format` or `inner sample format` or `synthesis sample format`... Any naming that refers to the  sample format of the synthesizer engine should be fine.

derselbst:
Thanks, I'll keep thinking.

PR #543, opened 2019-06-26
Fix math 2

carlo-bramini:
This PR superseedes #538 and it includes some preliminar corrections before fixing #537.

Next PR should just include:
- Detection with CMAKE of C99 math float if `WITH_FLOAT` is set.
- Replace macros for math constants with FLUID macros (for example M_PI -> FLUID_M_PI)
- Replace math functions with FLUID macros (sin -> FLUID_SIN) if `WITH_FLOAT=1` and C99 math float is supported.

### DIFF
--- a/src/synth/fluid_voice.c
+++ b/src/synth/fluid_voice.c
@@ -1745,7 +1745,7 @@ fluid_voice_get_lower_boundary_for_attenuation(fluid_voice_t *voice)
                     || (mod->flags2 & FLUID_MOD_BIPOLAR)
                     || (mod->amount < 0))
             {
-                min_val = 0.f - min_val; /* min_val = - |amount|*/
+                min_val = -min_val; /* min_val = - |amount|*/
             }
             else
             {
--- a/src/synth/fluid_voice.c
+++ b/src/synth/fluid_voice.c
@@ -344,9 +344,9 @@ fluid_voice_init(fluid_voice_t *voice, fluid_sample_t *sample,
     voice->synth_gain = gain;
 
     /* avoid division by zero later*/
-    if(voice->synth_gain < 0.0000001)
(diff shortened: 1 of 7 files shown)


carlo-bramini:
Probably, several things can be made easier than this PR.
Actually, I changed several lines similar to this one:

```
if(out >= 1.0) /* reset in case of instability near PI/2 */
```

to this:

```
if(out >= 1.0f) /* reset in case of instability near PI/2 */
```

by adding the 'f' character at the end,
But it seems that there is a better way when handling integer constants with floating point.
All you need to do is to write:

```
if(out >= 1) /* reset in case of instability near PI/2 */
```

without the dot character and the decimal part. In this way, it seems that you do not need to worry anymore about the size of the calculation, the compiler will handle the size for you, depending on the variable you are handling.
This is valid only for integer constants and small enough to fit the mantissa of your type.

So probably some parts of this PR and many other constants inside the sources may be simplified by removing the decimal dot and the zero fractional part. However, I see that in the current code you are able to distinguish between an FPU and a CPU calculation by simply looking the way it is written.
But I do not know if it is a thing that you wanted or not.

Sincerely.

src/synth/fluid_voice.c
```diff
@@ -1745,7 +1745,7 @@ fluid_voice_get_lower_boundary_for_attenuation(fluid_voice_t *voice)
                     || (mod->flags2 & FLUID_MOD_BIPOLAR)
                     || (mod->amount < 0))
             {
-                min_val *= -1.0; /* min_val = - |amount|*/
+                min_val = 0.f - min_val; /* min_val = - |amount|*/
```
derselbst: For readability I would have preferred to use unary minus in this case. Have you considered this?

derselbst:
Thx.

PR #545, opened 2019-07-06
Fix math 3

carlo-bramini:


src/utils/fluidsynth_priv.h
```diff
@@ -126,19 +126,80 @@ typedef void (*fluid_rvoice_function_t)(void *obj, const fluid_rvoice_param_t pa
...
+#endif
+
+#if defined WITH_FLOAT && defined HAVE_SQRTF
+#define FLUID_SQRT  sqrtf
+#else
+#define FLUID_SQRT  (fluid_real_t)sqrt
+#endif
+
+#if defined WITH_FLOAT && defined HAVE_LOGF
+#define FLUID_LOGF  logf
```
derselbst: I think it should be `FLUID_LOG`, none of the others has the `F` suffix either.

carlo-bramini: As I have written in the comment, FLUID_LOG already exists in the source of FluidSynth, it is used for printing log messages. That's why I set FLUID_LOGF for the math function.
If you prefer something different, I can do the changes.

derselbst: Oops, completely missed that. Sry, nevermind. It's fine I guess.

Issue #553, opened 2019-08-16
Allow audio render to STDOUT


FalcoGer:
### Is your feature request related to a problem?
I wish to pipe the output of fluidynth audio data to another program. Writing to /dev/stdout however gives me this error:

```
fluidsynth -a pulseaudio -d -F /dev/stdout /usr/share/sounds/sf2/FluidR3_GM.sf2 sample.mid
FluidSynth version 1.1.9
Copyright (C) 2000-2018 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of E-mu Systems, Inc.

Rendering audio to file '/dev/stdout'..
RIFWAVEfmt D��dataRIFWAVEfmt D��data����fluidsynth: error: Failed to open audio file '/dev/stdout' for writing
```

### Describe the solution you'd like
Allow `-F -` or `-F STDOUT` or similar to put the stream as a byte stream to standard output.

mawe42:
Please see this libsndfile FAQ entry: http://www.mega-nerd.com/libsndfile/FAQ.html#Q017

As explained there, rendering WAV to a pipe is difficult, as the header needs to be filled in with information that is not known until the complete audio data has been rendered. But you can use other file formats to render, for example `raw`, `au`, or `flac`. Just specify the output format on your command line, for example with `-o audio.file.format=au`.

One thing to note is that currently the fluidsynth command line program **always** prints a welcome message to stdout. So your resulting file will have non-audio text garbage at the beginning. But it's of a predicable length, so you can either filter it out before using the output somewhere else... or you could send your output to a different pipe. A quick'n dirty test with stderr worked well for me:
`fluidsynth -a none -F /dev/stderr  /path/to/song.mid /path/to/font.sf2 -o audio.file.type=au 2> test.au`

Or... you can write your own program that uses the the Fluidsynth API to fast render to stdout.

mawe42:
@FalcoGer As #554 has just been merged, you could also try out the new `-q / --quiet` option. :-)

Oh, and bear in mind that rendering the audio to stdout and having it spew binary data to your terminal will probably break. But redirecting it to a file or other program works fine.

So this breaks:
```
$ fluidsynth -a none -F- test.mid test.sf2 -o audio.file.type=au -q
.snd�����Dfluidsynth: error: Failed to open audio file '-' for writing
```

But this works fine:
```
[...]

PR #554, opened 2019-08-17
Add --quiet option to fluidsynth binary

mawe42:
People have asked about piping the output of the file renderer to stdout a few times now. See #553 for the latest example. In principle this works quite well for certain formats (au, flac, raw), but the fluidsynth binary currently always prints a welcome message and a "Rendering to ..." line to stdout as well.

This PR adds a new command line option to silence those messages. To do that, the welcome_message printing had to be moved after the optarg parsing. With the side effect that errors in any arguments or options will now be printed without a leading welcome message.

I quite like this, as it makes the error messages more visible and less likely to be overlooked by users.

The second commit is just for consistency and cleanup.

### DIFF
--- a/src/fluidsynth.c
+++ b/src/fluidsynth.c
@@ -646,6 +646,15 @@ int main(int argc, char **argv)
 
         case 'q':
             quiet = 1;
+
+#if defined(WIN32)
+            /* Windows logs to stdout by default, so make sure anything
+             * lower than PANIC is not printed either */
+            fluid_set_log_function(FLUID_ERR, NULL, NULL);
+            fluid_set_log_function(FLUID_WARN, NULL, NULL);
+            fluid_set_log_function(FLUID_INFO, NULL, NULL);
+            fluid_set_log_function(FLUID_DBG, NULL, NULL);
+#endif
             break;
 
         case 'R':
@@ -1170,7 +1179,8 @@ print_help(fluid_settings_t *settings)
     printf(" -p, --portname=[label]\n"
            "    Set MIDI port name (alsa_seq, coremidi drivers)\n");
     printf(" -q, --quiet\n"
-           "    Do not print welcome message or other informational output\n");
+           "    Do not print welcome message or other informational output\n"
+           "    (Windows only: also suppress all log messages lower than PANIC\n");
     printf(" -r, --sample-rate\n"
            "    Set the sample rate\n");
     printf(" -R, --reverb\n"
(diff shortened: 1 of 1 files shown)


src/fluidsynth.c
```diff
@@ -1181,8 +1195,4 @@ print_help(fluid_settings_t *settings)
     {
         FLUID_FREE(midi_options);
     }
-
-    delete_fluid_settings(settings);
```
derselbst: Removing that call leaks the settings object.

mawe42: Well, there are plenty of places where we simply `exit(0)` without any cleanup. And freeing `settings` in a function called `print_settings()` just feels plain wrong to me.

mawe42:
> You need to disable the FLUID_LOG functions as well, because on windows they will print to stdout.

Ah, good idea! But I would only disable >= WARN... I think errors and panics should still print. What do you think?

derselbst:
> But I would only disable >= WARN... I think errors and panics should still print.

Ok, let the errors and panic print.

src/fluidsynth.c
derselbst: > Well, there are plenty of places where we simply exit(0) without any cleanup.

Probably, but you should not produce more leaks ;)

> And freeing settings in a function called print_settings() just feels plain wrong to me.

I understand, then just place it in the switch-case :)

mawe42: But does fluidsynth run on any platform where an `exit()` does not free all allocated memory anyway?

mawe42:
I've rebased the branch to remove the cleanup commit (no need to mix this cleanup with the quiet option) and to add the suggested fluid_set_log_function calls for < PANIC. I chose to suopress ERR as well, as we use that level to indicate problems in soundfonts that are no necessarily really bad.

The cleanup handling could be done another time. Looking through the code I've found **a lot** of places where we simply `exit()` without cleanup. And some of the exit codes look wrong as well. So something for later :-)

src/fluidsynth.c
derselbst: > But does fluidsynth run on any platform where an exit() (or rather the OS after the process exists) does not free all allocated memory anyway?

I don't like this kind of arguing. I don't believe in that kind of "Alles-wird-wieder-gut" process that collects our resource garbage. This is our job. But since you've asked for a specific OS, I'd suggest Windows95 (Carlo has tested it recently).

derselbst:
> So something for later

Ok.

src/fluidsynth.c
mawe42: Ah, didn't know that Win95 does not use process-based virtual memory. Ok, I get your point (and personally I like cleanup up better as well).

mawe42:
Great, thanks for the review and merge!

PR #555, opened 2019-08-17
Fix various memory leaks in the fluidsynth binary

derselbst:
Remove premature exit() calls and properly free memory before quitting.


derselbst:
Thanks!

> why did you first commit this to the 2.0.x branch and not to master?

`2.0.x` only receives bug fixes, like this one. The `--quite` option is a new feature, this goes to master.

> Should I do this as well from now on?

Just commit to master, I'll take care of this :)

PR #559, opened 2019-09-12
Add FluidSostenuto-004.pdf documentation

jjceresa:


derselbst:
Nice, thank you.

One question: Is there a reason why the pdf explains how to implement this in fluidsynth? Because as far as I know Sostenuto is already (fully?) implemented in fluidsynth, so section 1.4 might be confusing to the reader.

jjceresa:
> One question: Is there a reason why the pdf explains how to implement this in fluidsynth?

No, there is no reason. The implementation was done at the time of v 1.1.6.

>Because as far as I know Sostenuto is already (fully?) implemented in fluidsynth, 

Right, and Sostenuto is actually fully implemented as described by the specifications of this .pdf

>so section 1.4 might be confusing to the reader.

Yes, section 1.4 and 1.7 should be removed. I am doing this.

derselbst:
Great, thanks!

PR #562, opened 2019-09-23
Add a public wrapper function for free()

derselbst:
to allow proper deallocation for programming languages other than C/C++.


jjceresa:
May be `fluid_free()` could be renamed `fluid_free_mem()`, to avoid confusion with other `delete_fluid_**()` function ?

derselbst:
> to avoid confusion with other delete_fluid_**() function ?

That's why I've put an explicit comment for that to `fluid_free()`. I would state that `fluid_free_mem()` only makes naming unnecessarily complicated.

jjceresa:
Oh yes, the comment (i had not read before!) is clear, so `fluid_free()` name is fine.

PR #564, opened 2019-09-27
Calling fluid_free() in fluidsynth.c

jjceresa:
- On Windows, any memory allocated by a shared library (.dll)
  must be freed by this library. (Otherwise this leads to
  memory leak or memory violation detected in debug build).

### DIFF
--- a/src/fluidsynth.c
+++ b/src/fluidsynth.c
@@ -909,7 +909,16 @@ int main(int argc, char **argv)
                 fluid_synth_sfload(synth, s, 1);
             }
 
+#ifdef WIN32
+            /* On Windows any memory allocated by a shared library (.dll) must
+               be freed by this library */
+// To do: replace call to FLUID_FREE() by call to fluid_free() once fluid_free()
+// will be implemented in fluidsynth library.
+//          fluid_free(s);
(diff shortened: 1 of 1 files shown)


jjceresa:
This PR is related to PR #562.
Sorry, I just realize that this should be merged to branch `2.0.X` rather to `master` branch.

derselbst:
I don't see why this is necessary. `FLUID_FREE` is now defined as `fluid_free`. Currently only on 2.0.x branch, will be merged back to master later.

jjceresa:
>FLUID_FREE is now defined as fluid_free. in 2.0.X

Ah, i didn't see that. So  this PR isn't necessary. However , waiting for 2.0.x to be merged to master, we just keep aware that actually building `master` in `Debug` for Windows will produce fluidsynth application that raise possible exception and abort when invoked in the following case:
1) fluidsynth -help (exception on FLUID_FREE(audio_options)) and FLUID_FREE(midi_options)).
2) fluidsynth -F .......  (exception on FLUID_FREE(filename)).
3) exception on FLUID_FREE(s), when no soundfont is specified.

When building in `Release`, these exception doesn't occurs, the application doesn't abort.

PR #568, opened 2019-10-16
Fix incorrect usage of fluid_ct2hz()

### LINKED ISSUE #456: Incorrect usage of fluid_ct2hz()
### FluidSynth version 
<= 2.0.x

### Current behavior
[It seems that `fluid_ct2hz()` is mistakenly used in GEN_OVERRIDEROOTKEY.](https://github.com/FluidSynth/fluidsynth/blob/da6a2e7a91820bf97f89d3bcdb15b76e94e90bc2/src/synth/fluid_voice.c#L839)

### Expected behavior
Because `fluid_ct2hz()` secretly clips to 13500 cents (upper fc limit), `fluid_ct2hz_real()` should be used instead. (ok, `fluid_ct2hz_real()` clips it to 14100 cents...)

derselbst:
Fixes #456 .

### DIFF
--- a/src/synth/fluid_voice.c
+++ b/src/synth/fluid_voice.c
@@ -825,7 +825,7 @@ fluid_voice_update_param(fluid_voice_t *voice, int gen)
                 voice->root_pitch = voice->sample->origpitch * 100.0f - voice->sample->pitchadj;
             }
 
-            x = (fluid_ct2hz(voice->root_pitch) * ((fluid_real_t) voice->output_rate / voice->sample->samplerate));
+            x = (fluid_ct2hz_real(voice->root_pitch) * ((fluid_real_t) voice->output_rate / voice->sample->samplerate));
         }
         else
         {
@@ -838,7 +838,7 @@ fluid_voice_update_param(fluid_voice_t *voice, int gen)
                 voice->root_pitch = 0;
             }
 
-            x = fluid_ct2hz(voice->root_pitch);
+            x = fluid_ct2hz_real(voice->root_pitch);
         }
 
         /* voice->pitch depends on voice->root_pitch, so calculate voice->pitch now */


jjceresa:
`fluid_ct2hz(cent)` behaves monotonically in the range [1500..13500] cents.
`fluid_ct2hz_real(cent)` isn't monotonic (because for cent > 14100, the frequency jump back to 1Hz). 

`fluid_ct2hz()` is still used here in `fluid_iir_filter_calc()` (fluid_iir_filter.c- line 377).
-1) It seems that `here` is the correct usage of fluid_ct2hz(cent) because we need a monotonic behaviour ?.

- 2)I wonder why `fluid_ct2hz_real()` isinsight the right usage for `root_pitch_hz` initialization ?.

-3)I wonder why `fluid_ct2hz_real()` is the right usage during `phase calculation` (fluid_rvoice.c-line 375), instead of `fluid_ct2hz()` ?.

Any feedback will be appreciated.

jjceresa:
`fluid_ct2hz(cent)` behaves monotonically in the range [1500..13500] cents.
`fluid_ct2hz_real(cent)` isn't monotonic (because for cent > 14100, the frequency jump back to 1Hz).

`fluid_ct2hz()` is still used here in `fluid_iir_filter_calc()` (fluid_iir_filter.c- line 377).
-1) It seems that here is the correct usage of `fluid_ct2hz()` because we need a monotonic behaviour ?.

-2) I wonder why `fluid_ct2hz_real()` is the right usage for `root_pitch_hz` initialization ?.

-3) I wonder why `fluid_ct2hz_real()` is the right usage during `phase calculation` (fluid_rvoice.c-line 375), instead of `fluid_ct2hz()` ?.

Any feedback will be appreciated.

derselbst:
For me what matters is that `fluid_ct2hz_real` is our general cents to hertz conversion function, that we basically should use everywhere. `fluid_ct2hz` limits the convertible range to satisfy `initialFilterFc` generator requirements, so it's only appropriate for IIR filter stuff. Why fluid_ct2hz_real jumps to 1Hz I can't tell. Doesn't really make sense to me, I'm open to change this behaviour.

jjceresa:
>For me what matters is that fluid_ct2hz_real is our general cents to hertz conversion function, that we basically should use everywhere. fluid_ct2hz limits the convertible range to satisfy initialFilterFc generator requirements, so it's only appropriate for IIR filter stuff.

Ok, we must think of `fluid_ct2hz()` being a sub range of `fluid_ct2hz_real()`.

>Why fluid_ct2hz_real jumps to 1Hz I can't tell. Doesn't really make sense to me, I'm open to change this behaviour.

This doesn't make sense for me also:
We could replace fluid_conv.c- line 85
 `return (fluid_real_t) 1.0; /* some loony trying to make you deaf */`
by 
 `return (fluid_real_t) 28160.0; /* some loony trying to make you deaf */`

derselbst:
...or fall-back to "on-the-fly computation" rather than some hardcoded value.

jjceresa:
...`fluid_ct2hz_real()` being used at synthesis time during `phase calculation` (fluid_rvoice.c-line 375),  we should pay attention that "computing-on-the-fly" is cpu-fpu expensive.

derselbst:
I was referring to the very rare case when cent > 14100 .

jjceresa:
In current implementation each `hardcoded frequency value` is obtained by multiply the previous one by 2:......7040.0, 14080.0, 28160.0 (the last one).
For the very rare case (last one) we could also do "on-the-fly computation" . Regardless the implementation the result should be upper limited to avoid `make us deaf` :).

derselbst:
Merging this one, follow up in #569.

Issue #571, opened 2019-10-19
Optimizing fluid_synth_write functions


carlo-bramini:
The rendering functions, like `fluid_synth_write_float()` for example, are implemented with a `for ()` cycle with inside some calculations. Inside this cycle, there is a block for ever sample:

```
if(cur >= synth->curmax)
{
    int blocksleft = (len - i + FLUID_BUFSIZE - 1) / FLUID_BUFSIZE;
    synth->curmax = FLUID_BUFSIZE * fluid_synth_render_blocks(synth, blocksleft);
    fluid_rvoice_mixer_get_bufs(synth->eventhandler->mixer, &left_in, &right_in);
    cur = 0;
}

```
In my opinion, this piece of code could be rewritten in a more efficient way.
I made two `do...while` cycles and I moved this block outside the loop with the samples, which was more speed critical. On my plaftorms, the improvements were very interesting.
It is true that we could use `FLUID_UNLIKELY` macro inside the 'if' statements, but in my opinion if we do not execute useless instructions, it would be better.
However, it does not seem to be a difficult and dangerous change to me.

What do you think?

EDIT: of course, the content of that 'if' statement is executed at multiple of FLUID_BUFSIZE, the improvement comes out by eliminating the execution costs of the test at every cycle.

derselbst:
> I made two do...while cycles and I moved this block outside the loop with the samples, which was more speed critical.

Moving that `if` out of the `for` loop sounds good. Since you've already done it I'd welcome a PR for that.

> On my plaftorms,

For curiosity, which platforms, pls.?

carlo-bramini:
> For curiosity, which platforms, pls.?

ARM Cortex M4, with single precision FPU, 120MHz, 8MB RAM.

derselbst:
> ARM Cortex M4,

Oh, right, I remember. That thing doesn't have branch prediction, right? In this case, you might also be interested in #569. Btw, do you have chorus activated when using fluidsynth on that platform?

carlo-bramini:
Yes, I deactivated many things.
Chorus and reverb are off.
Sample interpolation has been disabled at compile time by changing macro FLUID_INTERP_DEFAULT (perhaps having a setting initialized at FLUID_INTERP_DEFAULT when you call new_fluid_settings() would be better, because the user could change it before the creation of a channel)
I changed fluid_rvoice_get_sample() for using only the upper 16 bit of the sample.
I disabled the dithering.
[...]

PR #573, opened 2019-10-19
Speedup rendering functions

carlo-bramini:
See issue #571

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3801,7 +3801,7 @@ fluid_synth_write_float(fluid_synth_t *synth, int len,
                         void *lout, int loff, int lincr,
                         void *rout, int roff, int rincr)
 {
-    int n, j, k, cur;
+    int n, j, k, cur, size;
     float *left_out = (float *) lout;
     float *right_out = (float *) rout;
     fluid_real_t *left_in;
@@ -3819,14 +3819,14 @@ fluid_synth_write_float(fluid_synth_t *synth, int len,
     fluid_rvoice_mixer_get_bufs(synth->eventhandler->mixer, &left_in, &right_in);
 
     cur = synth->cur;
-    j = loff; k = roff;
+    j = loff; k = roff; size = len;
 
     do
     {
         /* fill up the buffers as needed */
         if(cur >= synth->curmax)
         {
-            int blocksleft = (len + FLUID_BUFSIZE - 1) / FLUID_BUFSIZE;
+            int blocksleft = (size + FLUID_BUFSIZE - 1) / FLUID_BUFSIZE;
             synth->curmax = FLUID_BUFSIZE * fluid_synth_render_blocks(synth, blocksleft);
             fluid_rvoice_mixer_get_bufs(synth->eventhandler->mixer, &left_in, &right_in);
             cur = 0;
@@ -3836,8 +3836,8 @@ fluid_synth_write_float(fluid_synth_t *synth, int len,
         n = synth->curmax - cur;
 
         /* keep track of emitted samples */
-        if (n > len) n = len;
-        len -= n;
+        if (n > size) n = size;
+        size -= n;
 
         /* update pointers to current position */
         left_in  += cur+n;
@@ -3859,7 +3859,7 @@ fluid_synth_write_float(fluid_synth_t *synth, int len,
         }
         while (++n < 0);
     }
-    while (len);
+    while (size);
 
     synth->cur = cur;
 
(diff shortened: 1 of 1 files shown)


src/synth/fluid_synth.c
```diff
@@ -3816,26 +3816,52 @@ fluid_synth_write_float(fluid_synth_t *synth, int len,
...
+        }
+        while (++n < 0);
     }
+    while (len);
 
-    synth->cur = l;
+    synth->cur = cur;
 
     time = fluid_utime() - time;
     cpu_load = 0.5 * (fluid_atomic_float_get(&synth->cpu_load) + time * synth->sample_rate / len / 10000.0);
```
derselbst: When control reaches here, `len` will be zero, causing a DIV by zero. Recompile with `cmake -Denable-trap-on-fpe=1` to get a SIGFPE.

carlo-bramini: Argh, I did not notice the use of len into cpu_load, I will fix it.

jjceresa:
Thanks for speed enhancement, however this new implementation doesn't behave as before. 
When  `fluid_synth_write_xxx()` is called with `len` to 0, the function will still call fluid_synth_render_blocks() and write in caller supplied buffer here:

*left_out = (float) left_in[n];
*right_out = (float) right_in[n];

This could lead with possible memory violation access. To fix that  the outer loop `do; {..} while(size)`
should be `while(size) {...}`.
***
Please, why did you use a reverse indexing ?. Shouldn't a classic indexing be more faster ? and easier to read, like this:
for (i=0,  left_in  += cur,   right_in += cur;   i < n;   i++)
{
            *left_out = (float) left_in[i];
            *right_out = (float) right_in[i];

            left_out += lincr;
            right_out += rincr;
}

jjceresa:
Thanks for speed enhancement, however this new implementation doesn't behave as before.
When `fluid_synth_write_xxx()` is called with `len` to 0, the function will still call fluid_synth_render_blocks() and write in caller supplied buffer here:

*left_out = (float) left_in[n];
*right_out = (float) right_in[n];

This could lead with possible memory violation access. To fix that the outer loop do; {..} while(size)
should be while(size) {...}.
****
Please, why did you use a reverse indexing ?. Shouldn't a classic indexing be more faster ? and easier to read, like this:
'''
for (i=0, left_in += cur, right_in += cur; i < n; i++)
{
    *left_out = (float) left_in[i];
    *right_out = (float) right_in[i];

    left_out += lincr;
    right_out += rincr;
}
'''

carlo-bramini:
I fixed the case when `len` is zero, thank you.

> Shouldn't a classic indexing be more faster

No, I don't think so. The readability of the code is quite subjective, but in my opinion the condition `(++n < 0)` is typically evaluated faster than `(++i < n)`.

derselbst:
> > Shouldn't a classic indexing be more faster
> 
> No, I don't think so.

Reversed indexing is fine for me. 

> condition (++n < 0) is typically evaluated faster than (++i < n).

I wouldn't nit-pick too much on those issues. After all, we are only talking about integers, and most of the time will be spent when accessing `left_in` and `right_in` and loading the data into the cache. That said, what you could try is traversing `left_in` and `right_in` in separate loops, to avoid potential cache conflicts (https://stackoverflow.com/q/8547778). And declaring `left_out`, `right_out`, `left_in` and `right_in` as `FLUID_RESTRICT`.

jjceresa:
>That said, what you could try is traversing left_in and right_in in separate loops, to avoid potential cache conflicts.

1)Interesting idea !.

2) Also to simplify a bit we could  replace `do; {..} while(size)` by  `while(size) {...}` this will allow to remove `if (FLUID_LIKELY((size = len) > 0))`

carlo-bramini:
> 1. Also to simplify a bit we could  replace `do; {..} while(size)` by  `while(size) {...}` this will allow to remove `if (FLUID_LIKELY((size = len) > 0))`

Actually, I did that intentionally instead of using a while(), it is true that the compiler could unroll the loop, but in my opinion it would be worth to help on this piece of code.

carlo-bramini:
> what you could try is traversing left_in and right_in in separate loops, to avoid potential cache conflicts

It is a good idea, although this seems to be platform dependant.
Actually, my platform performs better with a single loop rather than two splitted loops.

jjceresa:
>On my plaftorms, the improvements were very interesting.
 The improvement comes out by eliminating the execution costs of the test at every cycle.

I'm curious of the improvements you got. Would you please report these result ?

derselbst:
> Actually, my platform performs better with a single loop rather than two splitted loops.

Ok.

> > On my plaftorms, the improvements were very interesting.
> > The improvement comes out by eliminating the execution costs of the test at every cycle.
> 
> I'm curious of the improvements you got. Would you please report these result ?

Me too.

Unit test passed, will merge it now, because it's convenient for me. Thanks.

PR #575, opened 2019-10-21
Handle deprecation of GTimeVal

derselbst:
`GTimeVal` has been deprecated in glib 2.62 . While switching to `g_get_monotonic_time()`, I realized that we could simply reuse `fluid_utime()`  for that purpose.


derselbst:
> is now impacted by a float operation but I don't think this will cause problem.

I don't think so either, because the previous implementation also used the FPU:

```c
(unsigned int)((timeval.tv_sec - initial_seconds) * 1000.0 + timeval.tv_usec / 1000.0);
```

PR #576, opened 2019-10-21
Fix bugs in fluid_sample_set_sound_data with copy enabled

swesterfeld:
While developing liquidsfz, I noticed that clicks occurred at the end of samples but this was fixed if I passed false as copy argument. So I think there are two problems in fluid_sample_set_sound_data:

- memset should clear all the memory
- end position should be before the end of the memory

### DIFF
--- a/src/sfloader/fluid_sfont.c
+++ b/src/sfloader/fluid_sfont.c
@@ -609,7 +609,7 @@ fluid_sample_set_sound_data(fluid_sample_t *sample,
             goto error_rec;
         }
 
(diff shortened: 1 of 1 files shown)


derselbst:
You are right, thanks for the fix.

I'll slightly change your wording, then cherry-pick to `2.0.x` and merge back to `master`.

> end position should be before the end of ~the memory~ sample data

Otherwise it's a little confusing, because `storedNbFrames` actually is end of memory.

PR #577, opened 2019-10-23
move fluid_profile_ref_var()

jjceresa:


derselbst:
Ah, sry. I forgot that it's a declaration. Thx.

PR #579, opened 2019-10-24
Cleanup jack driver

derselbst:
Minor cleanup of the jack driver. Briefly tested, should work.


derselbst:
Something is broken here. No time to investigate. Deferred.

PR #582, opened 2019-10-26
Fix broken voice mixing

### LINKED ISSUE #580: Low frequency clipping at 96kHz
### FluidSynth version
2.0.x ~(and 1.1.x, IIRC)~

### Describe the bug
Low frequencies tend to clip when rendering at 96 kHz. Weirdly, it seems to matter which audio driver is used for playing. For me, clipping occurs with jack and sdl2, and when rendering via `fluid_synth_process()`, but not when using alsa. Investigation needed.

However: issue is not related to reverb. Reverb just makes the clipping more obvious, but it also occurs in dry sound.

**I would like to know:** Is there anybody able to reproduce the clippings for any audio driver?

### Expected behavior
Not clipping should occur.

### Steps to reproduce
Demo MIDI is attached below. A sample rendering of the clipping is included (clipping is quite evidently audible at the beginning of the file).

Using the following command:

```bash
fluidsynth -C0 -a sdl2 -jnil  -o synth.sample-rate=96000 -o synth.reverb.room-size=0.8 -o synth.reverb.damp=0.1 -o synth.reverb.width=0 -o synth.reverb.level=0.8 DP.sf2 DP_0xF.mid
```

### Additional context

[DP_0x1E.zip](https://github.com/FluidSynth/fluidsynth/files/3770858/DP_0x1E.zip)
[DP_0xF.zip](https://github.com/FluidSynth/fluidsynth/files/3768838/DP_0xF.zip)
[Soundfont to use](https://github.com/derselbst/ANMP-data/raw/dk64-defmod/soundfonts/N64/DP.sf2)

derselbst:
When rendering a voice, there are 3 cases to consider: silence, playing and finished. When optimizing away the `memset`, I incorrectly assumed that a voice cannot switch between playing and silence (like crazy) while rendering `FLUID_MIXER_MAX_BUFFERS_DEFAULT` . Apparently this does not hold true, esp. for percussion instrument while rendering at 96kHz. ~Thus, render one single block and immediately mix it into the destination buffer(s). (Rather than rendering all the `FLUID_MIXER_MAX_BUFFERS_DEFAULT` blocks and then do the mixing once).~

Fixes #580.

### DIFF
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -421,11 +421,11 @@ fluid_mixer_buffers_render_one(fluid_mixer_buffers_t *buffers,
 
     for(i = 0; i < blockcount; i++)
     {
-        int s = fluid_rvoice_write(rvoice, src_buf);
-
+        int s = fluid_rvoice_write(rvoice, &src_buf[FLUID_BUFSIZE * i]);
         if(s == -1)
         {
-            fluid_rvoice_buffers_mix(&rvoice->buffers, src_buf, last_block_mixed, unmixed_samples, dest_bufs, dest_bufcount);
+            // the voice is silent, mix back all the previously rendered sound
+            fluid_rvoice_buffers_mix(&rvoice->buffers, src_buf, last_block_mixed, total_samples - (last_block_mixed*FLUID_BUFSIZE), dest_bufs, dest_bufcount);
 
             unmixed_samples = 0;
             last_block_mixed = i+1;
@@ -442,7 +442,7 @@ fluid_mixer_buffers_render_one(fluid_mixer_buffers_t *buffers,
         }
     }
 
-    fluid_rvoice_buffers_mix(&rvoice->buffers, src_buf, last_block_mixed, unmixed_samples, dest_bufs, dest_bufcount);
+    fluid_rvoice_buffers_mix(&rvoice->buffers, src_buf, last_block_mixed, total_samples - (last_block_mixed*FLUID_BUFSIZE), dest_bufs, dest_bufcount);
 
     if(total_samples < blockcount * FLUID_BUFSIZE)
     {
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -364,25 +364,27 @@ get_dest_buf(fluid_rvoice_buffers_t *buffers, int index,
  *
  * @param buffers Destination buffer(s)
  * @param dsp_buf Mono sample source
- * @param offset Describes the block we are currently rendering to (offset is in samples though)
+ * @param start_block Describes the block we are currently rendering to
+ * @param sample_count number of samples to mix following \c start_block
  * @param dest_bufs Array of buffers to mixdown to
  * @param dest_bufcount Length of dest_bufs
  */
 static void
 fluid_rvoice_buffers_mix(fluid_rvoice_buffers_t *buffers,
                          const fluid_real_t *FLUID_RESTRICT dsp_buf,
-                         int offset,
+                         int start_block, int sample_count,
                          fluid_real_t **dest_bufs, int dest_bufcount)
 {
     int bufcount = buffers->count;
     int i, dsp_i;
 
-    if(dest_bufcount <= 0)
+    if(sample_count <= 0 || dest_bufcount <= 0)
     {
         return;
     }
 
     FLUID_ASSERT((uintptr_t)dsp_buf % FLUID_DEFAULT_ALIGNMENT == 0);
+    FLUID_ASSERT((uintptr_t)(&dsp_buf[start_block * FLUID_BUFSIZE]) % FLUID_DEFAULT_ALIGNMENT == 0);
 
     for(i = 0; i < bufcount; i++)
     {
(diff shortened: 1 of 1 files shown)


derselbst:
@jjceresa Pls. confirm the fix.

jjceresa:
This doesn't fix. Now clipping occurs at any sample rate. For example when playing `preset 5 (1 based) electric piano 1`, playing C4,D4,C4,D4,.... very quickly from C to D, the two notes are mixed and during the switch from one note to the other we hear clippings. Using `Preset 17 Electric organ 1` or `preset 18 electric organ 2` also shows  the issue.

jjceresa:
I think that `fluid_rvoice_buffers_mix()` should still receive `s` param which is the number of sample returned by `fluid_rvoice_write()`. Indeed, It is possible that `s < FLUID_BUFSIZE` and in this case `fluid_rvoice_buffers_mix()` should mix only `s` samples.
Actually `fluid_rvoice_buffers_mix()` always mixes `FLUID_BUFSIZE` samples (i.e one complete block). That means that `(FLUID_BUFSIZE - s)  invalid samples` from the `local buffer` will be mixed to the destination buffer. It seems that this local buffer  was not filled by 0.0, so I feel that this new parameter is still necessary.

jjceresa:
>I incorrectly assumed that a voice cannot switch between playing and silence (like crazy) while rendering FLUID_MIXER_MAX_BUFFERS_DEFAULT . Apparently this does not hold true...

Right, when a voice is started, amplitude envelope `DAHDSR  could be in `delay` stage at the beginning. During this time `fluid_rvoice_write()` return -1 to indicate that the voice is quiet.

rendering all the FLUID_MIXER_MAX_BUFFERS_DEFAULT blocks and then do the mixing once shouldn't be more efficient  ? even if  we keep `memset` to clear the local buffer ?.

jjceresa:
`>I incorrectly assumed that a voice cannot switch between playing and silence (like crazy) while rendering FLUID_MIXER_MAX_BUFFERS_DEFAULT . Apparently this does not hold true...

Right, when a voice is started, amplitude envelope `DAHDSR` could be in `delay`stage at the beginning. During this time `fluid_rvoice_write()` return -1 to indicate that the voice is quiet.

Rendering all the FLUID_MIXER_MAX_BUFFERS_DEFAULT blocks and then do the mixing once shouldn't be more efficient ? even if we keep `memset` to clear the local buffer ?.

jjceresa:
>I incorrectly assumed that a voice cannot switch between playing and silence (like crazy) while rendering FLUID_MIXER_MAX_BUFFERS_DEFAULT . Apparently this does not hold true...

Right, when a voice is started, amplitude envelope `DAHDSR` could be in `delay` stage at the beginning. During this delay `fluid_rvoice_write()` return -1 to indicate that the voice is quiet.

Rendering all the FLUID_MIXER_MAX_BUFFERS_DEFAULT blocks and then do the mixing once shouldn't be more efficient ? even if we keep `memset` to clear the local buffer ?.

derselbst:
> This doesn't fix. Now clipping occurs at any sample rate.

You are right. Believe me or not, when I tested it with jack and sdl2, it sounded absolutely fine. After listening to alsa, I heard what you mean.

At least now I know why there are so many differences between the driver: the `blockcount` varies dramatically. Alsa for example (at least on my system) always renders with `blockcount==1`. Jack and sdl2 however render many blocks.

> Rendering all the FLUID_MIXER_MAX_BUFFERS_DEFAULT blocks and then do the mixing once shouldn't be more efficient ?

Yes, that's how I implemented it now.

> even if we keep memset to clear the local buffer ?.

No. When I refactored `rvoice_mixer`, I also profiled it and `memset` was one of the most expensive things.

So, pls. have another try.

jjceresa:
>So, pls. have another try.

All is fine.
In last commit, I did simple parameter change, arithmetic simplification, and adding comments.

>Believe me or not, when I tested it with jack and sdl2, it sounded absolutely fine. After listening to alsa, I heard what you mean.

I believe you.

>At least now I know why there are so many differences between the driver: the blockcount varies dramatically.

Yes this blockcount isn't predictive and make debugging very difficult. Using 96kHz I had to change audio.period-size to get the driver working correctly. Probably most driver need to do sample rate conversion too.

derselbst:
Thx for testing.

> In last commit, I did simple parameter change, arithmetic simplification, and adding comments.

I prefer `buf[start_block * FLUID_BUFSIZE + dsp_i]`, because it tells the compiler that the start of accessing the `buf` will always be at the `FLUID_BUFSIZE*sizeof(fluid_real_t)` byte border. This matters, because I'm promising that `buf` itself is aligned to `FLUID_DEFAULT_ALIGNMENT`, which avoids the need of a peel loop in case the compiler tries to vectorize it. `buf[start_sample + dsp_i]` , however, can be somewhere in between, requiring a peel loop.

Will add a comment about that.

P.S.: pay attention to the base branch when merging: Base branch of `fix580` is `2.0.x`, not master. Will fix that as well.

jjceresa:
>I prefer buf[start_block * FLUID_BUFSIZE + dsp_i], because it tells the compiler that the start of accessing the buf..

Ok, I understand now why you did that. I forgot about alignment requirement.

>P.S.: pay attention to the base branch when merging: Base branch of fix580 is 2.0.x, not master. Will fix that as well.

Ok, sorry for this disorder.

jjceresa:
Pls, don't forget to propagate the fix to master. Thanks.

Issue #581, opened 2019-10-24
Output of reverb depends on sample rate used


derselbst:
### FluidSynth version 
\>=2.0.0

### Current behavior
_This is not really a bug, just a question to understand what's going on..._

I observed that the reverb output has a (somewhat strong) dependency on the sample rate being used for rendering.

Trying sample rates 22050, 48000 and 96000 Hz, to me personally 22050 Hz sounds best. 48000Hz has many more high frequencies, of course, which is still ok though. At 96000Hz, however, reverb "decays" much slower than at 48kHz. Also, the "wobble" is more evident for 96kHz. (In case you don't hear what I mean, try exaggerating the effect by hacking fluidsynth to render at 196kHz).

### Expected behavior
The difference between 48kHz and 96kHz should me "smaller" (?). Opinions?

Again, this is no request for change. Ofc. there must be a dependency to the sample rate. Just trying to understand things.

### Steps to reproduce
Using the MIDI and soundfont below with this command:

```bash
fluidsynth -C0 -a WhateverYouLike -jnil -o synth.sample-rate=96000 -o synth.reverb.room-size=0.8 -o synth.reverb.damp=0.1 -o synth.reverb.width=0 -o synth.reverb.level=0.8 DK64.sf2
```

### Other information

[DK64-JJ-Underground.zip](https://github.com/FluidSynth/fluidsynth/files/2080291/DK64-JJ-Underground.zip)
[Soundfont to use](https://github.com/derselbst/ANMP-data/raw/dk64-defmod/soundfonts/N64/DK64.sf2)

For testing purposes, you may also check out the example MIDI provided in #580, and hear how it correlates with sample rate.

@jjceresa I'm assigning you, as you are the expert for reverb related issues. Would be great if you could look into this.

jjceresa:
>Would be great if you could look into this.

I will do. First we must review and document how `hard` is reverb `dependency in regard to `sample rate`. I need time to review how was the behaviour of the previous reverb (`freeverb`) compared to the new one (`fdn-reverb`).

>At 96000Hz, however, reverb "decays" much slower than at 48kHz. Also, the "wobble" is more evident for 96kHz.

I always had difficulties for reproducing this "wobble". I need time to investigate using the supplied files (.mid and .sf2).

jjceresa:
[...]

PR #584, opened 2019-10-28
Make reverb output independent of sample rate

jjceresa:
This PR is the solution (3.2) discuted in issue https://github.com/FluidSynth/fluidsynth/issues/581#issuecomment-546996215.

### DIFF
--- a/src/rvoice/fluid_rev.c
+++ b/src/rvoice/fluid_rev.c
@@ -613,6 +613,16 @@ static int set_mod_delay_line(mod_delay_line *mdl,
 }
 
 /*-----------------------------------------------------------------------------
+ Return norminal delay length
+
+ @param mdl, pointer on modulated delay line.
+-----------------------------------------------------------------------------*/
+static int get_mod_delay_line_length(mod_delay_line *mdl)
+{
+    return (mdl->dl.size - mdl->mod_depth - INTERP_SAMPLES_NBR);
+}
+
+/*-----------------------------------------------------------------------------
  Reads the sample value out of the modulated delay line.
  @param mdl, pointer on modulated delay line.
  @return the sample value.
@@ -738,6 +748,7 @@ static void update_rev_time_damping(fluid_late *late,
 {
     int i;
     fluid_real_t sample_period = 1 / late->samplerate; /* Sampling period */
+    int delay_length;               /* delay length */
     fluid_real_t dc_rev_time;       /* Reverb time at 0 Hz (in seconds) */
 
     fluid_real_t alpha, alpha2;
@@ -756,8 +767,9 @@ static void update_rev_time_damping(fluid_late *late,
               Computes dc_rev_time
         ------------------------------------------*/
         dc_rev_time = GET_DC_REV_TIME(roomsize);
+        delay_length = get_mod_delay_line_length(&late->mod_delay_lines[NBR_DELAYS - 1]);
         /* computes gi_tmp from dc_rev_time using relation E2 */
-        gi_tmp = FLUID_POW(10, -3 * delay_length[NBR_DELAYS - 1] *
+        gi_tmp = FLUID_POW(10, -3 * delay_length *
                            sample_period / dc_rev_time); /* E2 */
 #else
         /*   roomsize parameters have the same response that Freeverb, that is:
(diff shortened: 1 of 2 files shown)


jjceresa:
Potential issue not related to the changes in this PR:

1)Note there is still a potential issue calling `fluid_revmodel_samplerate_change(). If the function fail, the caller have no possibility to know that the function fail because the function return void.
But the function could be changed to return a status (FLUID_FAILED, FLUID_OK).

2)Note that this potential issue already exist in the previous reverb (freeverb) in which the same function makes memory allocation but silently doesn't return an error if malloc fail.

3) Note that actually the caller is `DECLARE_FLUID_RVOICE_FUNCTION(fluid_rvoice_mixer_set_samplerate)`
- This caller silently doesn't check result when calling new_fluid_chorus(samplerate);
- Also, this caller silently doesn' tcheck result when calling fluid_revmodel_samplerate_change();
This potential issue exists when calling fluid_revmodel_samplerate_change(samplerate) and the  sample rate is augmented in which case malloc could fail.

I don't know is this caller could check the status and take appropriate action.

jjceresa:
Once the PR title is edited, I can't correct the title myself. Thanks to the brave soul would change french word "independant" to english one "independent" :)

derselbst:
Really impressive how you managed to get rid of that "degraded reverb tail", I have to say. I can't
This also eliminates the strange resonant frequencies I mentioned back ago [in this comment.](https://github.com/FluidSynth/fluidsynth/pull/380#issuecomment-392258182) I can barely tell the difference between 44.1kHz and 96kHz.

Additional thought: The only thing that might/should/could be dependent on the sample rate is `MOD_RATE`:

https://github.com/FluidSynth/fluidsynth/blob/cffc5b5d3a2b18cbb4ab97aeee24658c1678e03f/src/rvoice/fluid_rev.c#L579-L603

...at least to my ears, the modulation is audible at 44kHz, but almost gone at 96kHz. Does that make sense or is it only my ears?

jjceresa:
>Additional thought: The only thing that might/should/could be dependent on the sample rate is MOD_RATE:

Talking about the potential `side effect` of the modulator, the only thing audible I see is `MOD_DEPTH` (not MOD_RATE). If `mod_depth` is too high, we hear unwanted frequency modulation (similar to chorus effect). If `mod_depth` is too low this will no be sufficient to cancel the building of  high resonant frequencies (often called "ringing"). The comment above `#define MOD_DEPTH` explains a bit the appropriate range values to deal with.

`MOD_RATE` is an other thing. Consider this as a sub sampling rate at which the sinus modulator is computed. If 1, the sinus is  computed at every audio sample. If 50 (the actual value), the sinus is computed only every 50 audio samples which save a lot of cpu cycles.
Saving cpu cycles is the only reason of MOD_RATE presence. Using a value above 50 will produce a sinus not smooth enough and this leads with a distorted modulation which is audible.

>...at least to my ears, the modulation is audible at 44kHz, but almost gone at 96kHz. 

I dont' think that modulation is gone at 96 kHz, if for 96 KHz, you set MOD_DEPTH to 0 , you should hear that "ringing" come back . (As this isn't not obvious to hear, please use an head phone and a percussive (piano preset 2) (Christian's soundfont GUGSv1_47.sf2). Also the comparisons must be consecutive as soon as possible).

>Does that make sense or is it only my ears?

As far I understand, i see no relation with `modulation depth` (i.e MOD_DEPTH) and `sample rate`, so for me that doesn't make sense.

derselbst:
> As far I understand, i see no relation with modulation depth (i.e MOD_DEPTH) and sample rate, so for me that doesn't make sense.

Ok, no problem. Thanks for explaining.

jjceresa:
>The only thing that might/should/could be dependent on the sample rate is MOD_RATE:

Sorry I haven't answered completely. Actually `MOD_RATE and MOD_DEPTH` are expressed in `samples number` . So I don't see here a sample rate dependency. 

Of course the sinus `lfo frequency` is dependent of `sample rate` and MOD_RATE` (see set_mod_frequency() - line 984).

jjceresa:
For completeness of `fluid_revmodel_samplerate_change(rev,sample_rate)` function API, I would prefer add a return status (FLUID_OK, FLUID_FAILED), regardless fluidsynth would make use (or not) of the returned value.
If this reverb module is used in other application, this will allow `fluid_revmodel_samplerate_change(rev,sample_rate)`  behaviour to become coherent with
`new_fluid_revmodel(sample_rate) when a memory allocation error occurs. Opinion ?

jjceresa:
For completeness of `fluid_revmodel_samplerate_change(rev,sample_rate)` reverb function API, I would prefer add a return status (FLUID_OK, FLUID_FAILED), regardless fluidsynth would make use (or not) of the returned value.
If this reverb module is used in other applications, this will allow `fluid_revmodel_samplerate_change(rev,sample_rate)` function API behaviour to become coherent with `new_fluid_revmodel(sample_rate)` function API if a memory allocation error occurs. Opinion ?

jjceresa:
>As far I understand, i see no relation with modulation depth (i.e MOD_DEPTH) and sample rate

>Actually  MOD_DEPTH are expressed in samples number . So I don't see here a sample rate dependency.

Sorry I must correct me `twice` here. As  modulation depth is expressed in `samples number` (MOD_DEPTH)  for 44,1Khz) there is a true sample rate dependency. To be sure that modulation depth (in second) keeps constant for sample rate > 44,1Khz, it must be compensated. Recent commit `03c9efb` does this compensation and we can notice at 96Khz that few "ringing" and few "very low frequencies flutters " are diminished. (The comparison still remains difficult to do by ears - It is best to do the comparison trying the full range of  notes [A0..C8] if possible).

jjceresa:
Thanks for raising this bug. This revealed how important is the minimum length any delays lines should have. Also thanks for trying this. For now I am fine with this, It can be merged.

PR #586, opened 2019-11-03
Improve type-checking of fluid_log() format specifiers

derselbst:
Let gcc and clang know that `fluid_log()` receives a `printf` format string, allowing proper type checking, and fix all format warnings that were brought up.

@mawe42 I'm not sure if it's a good idea to "pollute" a public header with this `__attribute__`. But doing so would also allow type checks in user applications. Note that this `format __attribute__` is very old. [GCC 3.2 definitely supports](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html) it, and this [page claims that it can be found in version 2.3](https://www.ohse.de/uwe/articles/gcc-attributes.html#func-format) as well. So I won't do any version checking here.

Edit: ...Thinking about it, we already make quite a lot use of these attributes in public headers (visibility, deprecation).

### DIFF
--- a/include/fluidsynth/log.h
+++ b/include/fluidsynth/log.h
@@ -77,7 +77,11 @@ fluid_log_function_t fluid_set_log_function(int level, fluid_log_function_t fun,
 
 FLUIDSYNTH_API void fluid_default_log_function(int level, const char *message, void *data);
 
-FLUIDSYNTH_API int fluid_log(int level, const char *fmt, ...);
+FLUIDSYNTH_API int fluid_log(int level, const char *fmt, ...)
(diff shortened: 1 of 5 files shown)


mawe42:
> I'm not sure if it's a good idea to "pollute" a public header with this __attribute__. But doing so would also allow type checks in user applications. Note that this format __attribute__ is very old. GCC 3.2 definitely supports it, and this page claims that it can be found in version 2.3 as well. So I won't do any version checking here.

Sounds good to me!

derselbst:
Thanks.

PR #588, opened 2019-11-03
Remove unused macros from config.h

derselbst:


jjceresa:
I noticed that `MidiShare.h` is included by fluid_midishare.c.
Assuming MidiShare properly installed (include file + library file), I guess that MidiShare.h will be found by CMake even if `HAVE_MIDISHARE_H` macro isn't defined ?. Please, is it correct ?

derselbst:
> I guess that MidiShare.h will be found by CMake even if HAVE_MIDISHARE_H macro isn't defined ?

Correct.

jjceresa:
Thanks.

PR #589, opened 2019-11-07
FIXME: ensure that the decompressed OGG sample is 16-bit mono

derselbst:
Make the OGG vorbis sample decompression bulletproof.


mawe42:
Looks good! And btw: LGTM looks like a really cool service...

PR #593, opened 2019-11-13
AudioObjectGetPropertyDataSize returns the number of buffers

romanbsd:
I found it due to crash in address sanitizer for `num == 0`
Reference usage:
https://github.com/chromium/chromium/blob/master/media/audio/mac/audio_manager_mac.cc#L261

BTW, what happened to the 2.0.x branch?

### DIFF
--- a/src/drivers/fluid_coreaudio.c
+++ b/src/drivers/fluid_coreaudio.c
@@ -85,8 +85,7 @@ get_num_outputs(AudioDeviceID deviceID)
 
     if(OK(AudioObjectGetPropertyDataSize(deviceID, &pa, 0, 0, &size)))
     {
-        int num = size / (int) sizeof(AudioBufferList);
-        AudioBufferList bufList[num];
+        AudioBufferList bufList[size];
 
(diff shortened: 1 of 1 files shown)


derselbst:
What makes you think that `AudioObjectGetPropertyDataSize` returns the number of buffers? Fluidsynths implementation as well as your reference assume that `size` is the number of bytes. The only reason I see how `num` can get zero is if `size < sizeof(AudioBufferList)`. In this case, Chrome's implementation is broken as well, due to buffer overrun.

Unfortunately I cannot find proper documentation about `AudioObjectGetPropertyDataSize`  either.

The 2.0.x branch diverged with master, I thought 2.0.8 would be the last release, so I removed it.

romanbsd:
In any case, this definitely crashed on my MacOS Mojave. Returned size was 4.

romanbsd:
Doesn't crash now. But I reverted a94bc82a3c9054aa8c8ff773da79ebc537f2fee9 locally. Perhaps the reason for crash is related to a94bc82a3c9054aa8c8ff773da79ebc537f2fee9. Closing.

PR #609, opened 2020-01-18
Solve the sequencer client unregistering problem - dropped

derselbst:
The function [fluid_sequencer_register_fluidsynth](http://www.fluidsynth.org/api/seqbind_8h.html#ae18ee23fd409270c9da4f0cd5c557c3d), allocates an internal structure that glues the synth and the sequencer together. The pointer to that structure is kept in the sequencer internally. The user only receives a client id. When the sequencer is deleted, this structure should be freed again. In fluidsynth 1.1.x this was the case. However, under a certain condition, it could have led to a double free. At this time I saw no other option to fix the double free by making users to force unregistering any clients explicitly in their code. 

Today, we have a unit test `test_seqbind_unregister` which tests 3 ways to unregister. None of them ends in a double-free, but two of them leak the internal structure. This is not satisfying, which is I'm now bringing up two solutions which make sure that:

* Explicit client unregistering is not needed anymore (i.e. it is implicitly done when calling `delete_fluid_sequencer()`), and
* explicit client unregistering does not lead to a double free when calling `delete_fluid_sequencer()` afterwards.

This PR is the first proposal how to solve this. (The unit test `test_seq_event_queue_sort` currently fails, don't know why, will investigate if we decide to go with this solution.)

Explicit unregistering may be accomplished by:

1. Calling `fluid_sequencer_unregister_client()`, or
1. Sending a `FLUID_SEQ_UNREGISTERING` event via `fluid_sequencer_send_now()`

With this solution, the callback function passed to `fluid_sequencer_client_register()` must handle `FLUID_SEQ_UNREGISTERING` events by calling `fluid_sequencer_client_unregister()` (see case 2.). Although this only matters for clients that implement their custom callback.

Opinions?

### DIFF
--- a/src/midi/fluid_seq.c
+++ b/src/midi/fluid_seq.c
@@ -153,17 +153,6 @@ delete_fluid_sequencer(fluid_sequencer_t *seq)
 
     _fluid_seq_queue_end(seq);
 
-    /*	if (seq->clients) {
-    		fluid_list_t *tmp = seq->clients;
-    		while (tmp != NULL) {
-    			fluid_sequencer_client_t *client = (fluid_sequencer_client_t*)tmp->data;
-    			if (client->name) FLUID_FREE(client->name);
-    			tmp = tmp->next;
-    		}
-    		delete_fluid_list(seq->clients);
-    		seq->clients = NULL;
-    	}*/
-
     FLUID_FREE(seq);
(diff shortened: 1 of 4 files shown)


derselbst:
Superseded by #610

PR #610, opened 2020-01-18
Solve the sequencer client unregistering problem

derselbst:
This is the second solution to the problem of #609.

The difference is that responsibility for calling `fluid_sequencer_unregister_client()` in case of `FLUID_SEQ_UNREGISTERING` events has been moved to `fluid_sequencer_send_now()`. In other words, a `FLUID_SEQ_UNREGISTERING` event now really unregisters the client, no matter how the client's  callback function looks like. I'm afraid that technically this is a breaking change. This is also the reason why the unit test `test_seq_event_queue_sort` currently fails. (I'll fix the test if we decide to move forward with this approach.)

Opinions?

### DIFF
--- a/doc/fluidsynth-v20-devdoc.txt
+++ b/doc/fluidsynth-v20-devdoc.txt
@@ -67,6 +67,10 @@ What is FluidSynth?
 
 - FluidSynth is open source, in active development. For more details, take a look at http://www.fluidsynth.org
 
+\section NewIn2_1_1 What's new in 2.1.1?
+
+- requirements for explicit sequencer client unregistering have been relaxed: delete_fluid_sequencer() now correctly frees any registered sequencer clients (clients can still be explicitly unregistered)
+
 \section NewIn2_1_0 What's new in 2.1.0?
 
 - <span style="color:red">refrain from using fluid_synth_set_sample_rate()</span>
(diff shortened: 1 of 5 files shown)


derselbst:
Ok, thanks. I also prefer this solution.

> To solve the problem with custom callbacks not receiving the FLUID_SEQ_UNREGISTERING event anymore,

Misunderstanding (I guess the new if-clause in `fluid_sequencer_send_now()` confused you):

Clients will still receive the unregistering event (this is done in `fluid_sequencer_unregister_client()`). ...Just that no matter what clients do with that event, they will have been unregistered afterwards. Previously it was the responsibility of the clients callback to unregister itself when it received an unregistering event. To pacify my initial "breaking change" thought, you can also put it differently and say: It's actually a bug if clients are not being unregistered after they have received the event. This would make things easier, would you agree?

mawe42:
> I guess the new if-clause in fluid_sequencer_send_now() confused you

Yes, and I didn't notice that `fluid_sequencer_unregister_client` sends the event.

> To pacify my initial "breaking change" thought, you can also put it differently and say: It's actually a bug if clients are not being unregistered after they have received the event. This would make things easier, would you agree?

Yes, that sounds much better :-) And I totally agree that this is a bug that needs to be fixed, not just a changed feature/behaviour.

Issue #61, opened 2010-05-26
Stereo sample pairs should both use pitch generators from the right channel


derselbst:
Looking over the noteon method in fluid_defsfont.c I don't see any code handling stereo pairs. The [SoundFont](/apps/trac/fluidsynth/wiki/SoundFont) specification 2.04 section 7.10 says that stereo pairs should use the pitch generators from the right channel for both samples. This means that any [SoundFont](/apps/trac/fluidsynth/wiki/SoundFont) with non-identical pitch generators for stereo pairs will not synthesize correctly, leading to detuning/drift between the 2 channels.  

Reported by: elementgreen

Original Ticket: [fluidsynth/tickets/61](https://sourceforge.net/p/fluidsynth/tickets/61)

derselbst:
A small demo SF2 + MIDI would help understanding this one.

PR #614, opened 2020-01-24
Hardening fluid_sample_validate() against invalid flag combinations

derselbst:
Following up my thought in https://lists.nongnu.org/archive/html/fluid-dev/2020-01/msg00014.html, this hardens `fluid_sample_validate()` to reject samples that possibly use a possibly unknown compression. A unit test has been added to additionally test against other invalid `SfSampleType` flag combinations.

I've tested this change on my local machine and it doesn't seem to have any unexpected side-effects.


src/sfloader/fluid_sfont.c
```diff
@@ -700,13 +700,40 @@ int fluid_sample_set_pitch(fluid_sample_t *sample, int root_key, int fine_tune)
...
+    if(sample->sampletype & ~supported_flags)
+    {
+        FLUID_LOG(FLUID_WARN, "Sample '%s' has unknown flags, which might indicate that it uses a compression that it not supported by fluidsynth", sample->name);
+        return FLUID_FAILED;
+    }
+
+    if((sample->sampletype & EXCLUSIVE_FLAGS) & ((sample->sampletype & EXCLUSIVE_FLAGS) - 1))
+    {
+        FLUID_LOG(FLUID_WARN, "Sample '%s' must be either mono or left or right", sample->name);
+        return FLUID_FAILED;
```
mawe42: Is is really necessary to fail here? Maybe just warn the user and carry on?

src/sfloader/fluid_sfont.c
```diff
@@ -700,13 +700,40 @@ int fluid_sample_set_pitch(fluid_sample_t *sample, int root_key, int fine_tune)
...
+    }
+
+    if((sample->sampletype & FLUID_SAMPLETYPE_LINKED) && (sample->sampletype & EXCLUSIVE_FLAGS))
+    {
+        FLUID_LOG(FLUID_WARN, "Linked sample '%s' cannot be mono, left or right at the same time", sample->name);
+        return FLUID_FAILED;
```
mawe42: Same here: failure necessary?

src/sfloader/fluid_sfont.c
```diff
@@ -700,13 +700,40 @@ int fluid_sample_set_pitch(fluid_sample_t *sample, int root_key, int fine_tune)
...
     /* ROM samples are unusable for us by definition */
     if(sample->sampletype & FLUID_SAMPLETYPE_ROM)
     {
         FLUID_LOG(FLUID_WARN, "Sample '%s': ROM sample ignored", sample->name);
         return FLUID_FAILED;
     }
 
+    if(sample->sampletype & ~supported_flags)
+    {
+        FLUID_LOG(FLUID_WARN, "Sample '%s' has unknown flags, which might indicate that it uses a compression that it not supported by fluidsynth", sample->name);
```
mawe42: ...it *is* not supported...

src/sfloader/fluid_sfont.c
derselbst: Idk. Such a flag combination is not defined by the spec. But the spec also doesn't seem to specify how to treat samples with those invalid combinations. If we warn only, should we force to sample to MONO or simply leave `sample->sampletype` unchanged?

src/sfloader/fluid_sfont.c
derselbst: I'm open as well. But should we clear the LINKED flag, or force it to MONO or do nothing?

src/sfloader/fluid_sfont.c
mawe42: Not sure... I think it depends on if those flag combinations actually cause a real problem. If we rely on consistent flags and something breaks otherwise, we should correct. Otherwise leave as is?

src/sfloader/fluid_sfont.c
mawe42: I guess same as above: leave untouched if not a problem, otherwise force mono?

src/sfloader/fluid_sfont.c
mawe42: Actually, I've already seen a similar error message from fluidsynth when I tried to load a FLAC compressed soundfont with a libsndfile that was lacking FLAC support. So we're checking the flag, but not if the compressed data is actually OGG/Vorbis but simply accept all formats that libsndfile understands. Quite handy for my tests... but maybe not desired?

src/sfloader/fluid_sfont.c
derselbst: Ok. When looking through the code, I cannot find a place where the `sampletype` actually makes a difference. This may become relevant if we ever implement #61. So, let's leave it unchanged for now. For both cases. Will apply the changes.

mawe42: Sounds good!

src/sfloader/fluid_sfont.c
derselbst: > So we're checking the flag, but not if the compressed data is actually OGG/Vorbis but simply accept all formats that libsndfile understands.

I do not quite understand this sentence, sorry. Anyway:

* If the sample is FLAC, it (should) have the 0x20 flag set. In this case, I would expect fluidsynth to treat it's content as RAW PCM, which would probably sound very glitchy.
* If the sample is FLAC and (incorrectly) has the 0x10 flag (OGG) set, it would successfully be decoded by libsndfile, yes (assuming it was compiled with FLAC support).

But in any case, I would vote to drop the sample if we detect unknown flags. Until we decide to support FLAC compressed samples... if we decide. (_I'm going offline now_)

mawe42: >> So we're checking the flag, but not if the compressed data is actually OGG/Vorbis but simply accept all formats that libsndfile understands.
> I do not quite understand this sentence, sorry.

We check the sample type flag for valid values, like 0x10 for OGG. And then we simply pass the sample data on to libsndfile to be decoded. But we leave it up to libsndfile to determine the data format, and therefore accept all audio data that libsndfile can understand. Doesn't seem like it can cause harm, but as you propose to add warnings about inconsistencies in the Soundfont in this PR, I though it might be another place for a possible warning.

> Until we decide to support FLAC compressed samples...

We effectively already support FLAC compressed samples :-)

derselbst: > We effectively already support FLAC compressed samples :-)

Well... yes, but we should not. Neither have we checked, whether libsndfile has been compiled with FLAC support, nor are we checking the 0x20 flag anywhere. A sample having 0x10 set and containing anything but OGG is broken and IMO should not be used.

> but as you propose to add warnings about inconsistencies in the Soundfont in this PR, I though it might be another place for a possible warning.

Ok, so you want warn if `SF_INFO::format` is something else than `SF_FORMAT_OGG`. Only warn or also reject the sample? (I would vote for rejecting them, to avoid people using FLAC samples via a backdoor in fluidsynth.)

mawe42: Sorry about the late reply!

> Ok, so you want to warn if SF_INFO::format is something else than SF_FORMAT_OGG. Only warn or also reject the sample? (I would vote for rejecting them, to avoid people using FLAC samples via a backdoor in fluidsynth.)

After thinking about this more, maybe simply leave it as it is? To check if the sample data is really OGG encoded, we would need to add extra code. And there is really no pressing need to guard against this case. One, because libsndfile will check and throw errors if it can't parse the data. And second, because we're a synth and not a soundfont validator.

derselbst: > After thinking about this more, maybe simply leave it as it is? To check if the sample data is really OGG encoded, we would need to add extra code. And there is really no pressing need to guard against this case. One, because libsndfile will check and throw errors if it can't parse the data.

I don't see a pressing need either. It's just from my understanding of the (non-existent) SF3 spec, that samples marked as OGG should also contain OGG. Otherwise, we would support FLAC, opus, whatever via a backdoor. A backdoor only being available due to an implementation detail (because we've chosen libsndfile, which can cope with it somehow). So, if we are not rejecting it, I would at least warn, to indicate the broken- and unportableness of that soundfont.

> And second, because we're a synth and not a soundfont validator.

Pls. keep in mind that fluidsynth also checks for the correct order of instrument/preset generators. Doesn't really make any sense, but is required by the spec.

PR #616, opened 2020-01-31
CMakeLists.txt: make positional code (-fPIC) customizible.

atsushieno:
For some use cases it is necessary to specify -fPIC even if we build
static library e.g. building vst plugins (*.so) which may not load
shared libraries from outside the system paths (depends on DAWs).
For such environment we would like to build the final shared library
without depending on `libfluidsynth.so(.*)` but if we build libfluidsynth.a
it always comes without -fPIC. This change makes it optional.


atsushieno:
Usecase: https://github.com/atsushieno/juicysfplugin/commit/321a1279f94e7c3a4e21edbe06c5715b43efb356

derselbst:
Oh, that's a very special case, indeed. Thanks for explaining, makes sense to me. Allow me to change your last word "optional" to "adjustable", though.

PR #617, opened 2020-01-31
Fix linking against libinstpatch

realnc:
This fixes the case where linking fails if libinstpatch is not installed in a system/default location but instead needs to be found through `PKG_CONFIG_PATH`.


derselbst:
Ok, thanks!

PR #619, opened 2020-02-02
Make printing debug messages runtime configurable

derselbst:
Supplying --verbose to the fluidsynth executable now prints debug messages to stdout. Debug messages are still being printed by default when fluidsynth was compiled in debug mode.

### DIFF
--- a/src/fluidsynth.c
+++ b/src/fluidsynth.c
@@ -706,6 +706,7 @@ int main(int argc, char **argv)
 
         case 'v':
             fluid_settings_setint(settings, "synth.verbose", TRUE);
+            fluid_set_log_function(FLUID_DBG, fluid_default_log_function, NULL);
             break;
 
         case 'z':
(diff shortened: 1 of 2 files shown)


mawe42:
That's a very good idea! I would even go one step further and make this a new option to let the user choose the log level (error, warning, info, debug). Then `--verbose` could stay to control if the user wants to see MIDI messages as well, regardless of logging level?

derselbst:
Idk. I thought it would be overengineered to add a `--loglevel` option. I know it from FFMpeg and I always only use highest, default or quiet loglevel. Also keep in mind that we have `--quiet` which would then be somewhat equivalent to `--loglevel panic`.

> Then --verbose could stay to control if the user wants to see MIDI messages as well, regardless of logging level?

If the debug messages really get too annoying, one could simply pass `-o synth.verbose=1`.

PR #637, opened 2020-04-25
Properly handle overlapping notes when using fluid_event_note()

derselbst:
One issue kept me striking for years: There is `fluid_event_note()`, which provides a note event with a duration. Looks useful. But in the way it's currently implemented, it's not. It basically expands to `fluid_event_noteon()` and `fluid_event_noteoff()`. This naive implementation heavily breaks when notes overlap. That's because, the synth doesn't support overlapping notes, i.e. it kills any voices that are playing on the same key and channel as requested by the noteon. But the original noteoff event is still in the queue and not removed from there. Thus, overlapping notes will sound "chopped off".

This PR fixes that behaviour, by keeping track of the enqueued `fluid_event_note()`. And if an overlap is detected, the noteoff from the previous note causing the overlap is removed from the queue.

If the application decides to use separate `fluid_event_noteon()` and `fluid_event_noteoff()` events, behaviour remains unchanged.

I'm attaching a ZIP, containing such a broken MIDI file. Included is a rendering of channel 15 before and after the fix. You may use [this soundfont](https://github.com/derselbst/ANMP-data/raw/master/soundfonts/N64/JFG.sf2).

[overlapping-notes.zip](https://github.com/FluidSynth/fluidsynth/files/4533679/overlapping-notes.zip)


jjceresa:
>One issue kept me striking for years

Yes, `overlapping` notes is an old know issue and still potentially present in MIDI file. Also, an other potential issue in MIDI file is the absence of `noteoff `event corresponding to `noteon`.  That means that it would be nice that the MIDI file reader be capable to fix these issues.

derselbst:
> Also, an other potential issue in MIDI file is the absence of noteoff event corresponding to noteon

In this case, you'll have a hanging note. I see no way to fix that, sorry.

jjceresa:
>In this case, you'll have a hanging note. I see no way to fix that, sorry.

This is no way to fix that. Because this hanging note will quickly enters in dissonance with others note, the proper musical way to deal with is to avoid the note to be played!. Any way cancelling the  noteon message is welcome. In MIDI File this cancellation could be done by forcing the velocity of potential hanging notes to 0. Hopefully rare are MIDI File with hangings note, but if just only one exists, this is a disaster for our poor ears ! :)

derselbst:
> the proper musical way to deal with is to avoid the note to be played

From a musical point of view, you are probably right. From a technical point of view, however, this hides errors. So I would rather prefer a hanging note in this case and therefore letting the musician know that something is terribly wrong in this MIDI file, rather than not playing it at all.

doc/fluidsynth-v20-devdoc.txt
```diff
@@ -72,6 +72,7 @@ What is FluidSynth?
 
 \section NewIn2_2_0 What's new in 2.2.0?
 
+- #fluid_file_callbacks_t now uses <code>long long</code> as file-offset type (see #fluid_long_long_t). This is a breaking change, which allows to load SoundFonts bigger than 2GiB on Windows. This change required to bump fluidsynth's SOVERSION.
 - The sequencer has received a major revisal. For you that means:
   - The sequencer's queue no longer blocks the synthesizer thread, due to being busy arranging its events internally.
   - Eevents that share the same tick was given a new, documented order, see fluid_sequencer_send_at().
```
jjceresa: >Eevents ...

A typo.

>tick was given a new, documented order, see fluid_sequencer_send_at():

1)FLUID_SEQ_SYSTEMRESET events precede any other event type.
2)FLUID_SEQ_UNREGISTERING events succeed FLUID_SEQ_SYSTEMRESET and precede other event type.
3)FLUID_SEQ_NOTEON and FLUID_SEQ_NOTE events succeed any other event type.
4)Otherwise `the order is undefined`. 


Please, for case 4, it means the reader should understand that for other events `the order is left unchanged`. Is not it ?

derselbst: > Please, for case 4, it means the reader should understand that for other events the order is left unchanged. Is not it ?

No. It's undefined. Under the hood, we now use heap sort, which is not stable. Thus, the order will probably change. Marcus and I had a few discussions about that in #604.

Thx for pointing out the typo. Will fix it on master.

jjceresa: >Thus, the order will probably change.

Thx, for this useful precision.

PR #641, opened 2020-04-29
Fix a NULL deref in delete_rvoice_mixer_threads()

### LINKED ISSUE #640: Memory violation access if new_fluid_revmodel() return NULL
Just discover this bug, using the console application.
If `new_fluid_revmodel()` return NULL, the message `fluidsynth: error: Out of memory` is displayed followed by a `memory violation access`. I haven't yet investigated, I don't know if the `violation` occurs in the library or in the console application.

derselbst:
The function attempts to lock a mutex that might have not been created
yet, due to a previous error.

Fixes #640


derselbst:
@jjceresa FYI

jjceresa:
It look good. I saw no memory leaks. 
I tried it forçing `new_fluid_revmodel()` to return NULL, there is no more memory violation.
Thanks.

PR #642, opened 2020-04-29
Stop dsound properly.

jjceresa:
Stopping `dsound `before stopping `audio thread` avoid dsound playing transient
 glitches between the time audio task is stopped and dsound will be released.
 These short trailing glitches are particularly audible when captured by a reverb connected on output.


derselbst:
Thanks!

Issue #643, opened 2020-05-01
Call ipatch_close()


derselbst:
Fluidsynth should call `ipatch_close()` to avoid memory leaks. This could be done by keeping track of created `settings` xor `synth` instances: When all instances of either type have been destoryed, call `ipatch_close()`. When the first instance is created again, call `ipatch_init()`. Need to figure whether `settings` or `synth` is more appropriate.

derselbst:
... or make use of swami/libinstpatch#47 and always call `ipatch_init()` in `new_fluid_synth()` and `ipatch_close()` in `delete_fluid_synth()`

jjceresa:
>always call ipatch_init() in new_fluid_synth() and ipatch_close() in delete_fluid_synth()

This is the good solution and will work safely only if swami/libinstpatch#47 is applied. Otherwise Swami will fail during its deinitialization step  because libinstpatch was already deinitialized by fluidsynth.

PR #644, opened 2020-05-01
Make fluidsynth calling ipatch_close()

jjceresa:
This should address issue https://github.com/FluidSynth/fluidsynth/issues/643

~Warning: this works only if swami/libinstpatch#47 is applied, otherwise Swami will fail with memory violation.~


derselbst:
Technically, this change would have required bumping the minimum required libinstpatch version. To avoid that, I've made it backward compatible and added the missing call to `ipatch_init()` into `new_fluid_synth()`.

jjceresa:
> I've made it backward compatible.

Thanks !

jjceresa:
Just tried current fluidsynth master with current libinspatch and current swami. This works. Thanks.

PR #647, opened 2020-05-08
fix ordering of operations

jjceresa:
This should avoid an unlikely memory violation if `new_fluid_preset()` fails.


derselbst:
Nice spot, thanks!

Issue #652, opened 2020-06-07
Segfault when attempting to load RIFF SF2 soundbank


v1993:
### FluidSynth version
2.1.1

### Describe the bug
When I try to load specific SoundFont ([`Stgiga's HiDef Soundfont (2019-05-25).sf2`](https://mobp.foxette.eu/sf2-soundfonts.html), RIFF (little-endian) data SoundFont/Bank) fluidsynth crashes with error message and segfault:

```
v@v-home:~$ fluidsynth Downloads/Stgiga\'s\ HiDef\ Soundfont\ \(2019-05-25\).sf2
FluidSynth runtime version 2.1.1
Copyright (C) 2000-2020 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of E-mu Systems, Inc.

fluidsynth: error: File read failed
fluidsynth: error: Failed to read sample data
fluidsynth: error: Attempted to read 1411452394 words of sample data, but got -1 instead
fluidsynth: error: Unable to load all sample data

** (process:32741): CRITICAL **: 13:34:22.287: ipatch_container_init_iter: assertion 'IPATCH_IS_CONTAINER(container)' failed
Segmentation fault (core dumped)
```

### Expected behavior
Fluidsynth loads soundfont without any problems (or at least handles error gracefully and avoids segfault).

### Steps to reproduce
1. Get soundbank from link
2. Attempt to load it using fluidsynth

### Additional context
It seems to me some sort of regression as some time before (can't post exact version sadly) qsynth had no problems loading this exact soundfont. I'm usually using qsynth and have resorted to command line fluidsynth only after meeting with this issue. Due to segfault caused by this issue qsynth now crashes on startup.

mawe42:
I can confirm this behaviour with v2.1.1, v2.1.2 and v2.1.3 on Linux.

Current master does not show the problem and a quick bisect reveals that it has already been fixed by 9995fd88b2d9562948c26e0a0644417f107811a5 (the fix for #629). Which is interesting, as #629 was about a Windows-only problem.

derselbst:
Fluidsynth fails to load this soundfont, because we used to pass an integer to fread().

https://github.com/FluidSynth/fluidsynth/pull/629/files#diff-033efa405126e8aa738c6a502777cd61R48

[...]

PR #653, opened 2020-06-07
Avoid num_samples from becoming negative

derselbst:
For Soundfonts bigger 2GiB, `num_samples` becomes negative. When being passed to `safe_fread()` it's promoted to `long long` and when being passed to `fread()`, it's cast to `size_t`. Works fine in twos-complement, but still is not nice.

Addresses #652.


mawe42:
Looks good and makes the current behaviour a little cleaner, I think.

derselbst:
Thanks for reviewing!

PR #659, opened 2020-08-06
fix NULL permitted for out and fx pointer buffer in fluid_synth_process_LOCAL()

jjceresa:
This should avoid a segmentation fault if one pass NULL value for fx or out parameters.

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3699,6 +3699,7 @@ fluid_synth_process(fluid_synth_t *synth, int len, int nfx, float *fx[],
     return fluid_synth_process_LOCAL(synth, len, nfx, fx, nout, out, fluid_synth_render_blocks);
 }
 
+/* declared public (instead of static) for testing purpose */
 int
 fluid_synth_process_LOCAL(fluid_synth_t *synth, int len, int nfx, float *fx[],
                     int nout, float *out[], int (*block_render_func)(fluid_synth_t *, int))
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3654,22 +3654,21 @@ fx[ ((k * fluid_synth_count_effects_channels() + j) * 2 + 1) % nfx ]  = right_bu
  * @param len Count of audio frames to synthesize and store in every single buffer provided by \p out and \p fx.
  *  Zero value is permitted, the function does nothing and return FLUID_OK.
  * @param nfx Count of arrays in \c fx. Must be a multiple of 2 (because of stereo).
- * and in the range <code>0 <= nfx/2 <= (fluid_synth_count_effects_channels() * fluid_synth_count_effects_groups()).
-  Note that zero value is valid and allows to skip mixing effects in all fx output buffers.</code>.
+ * and in the range <code>0 <= nfx/2 <= (fluid_synth_count_effects_channels() * fluid_synth_count_effects_groups())</code>.
+  Note that zero value is valid and allows to skip mixing effects in all fx output buffers.
  * @param fx Array of buffers to store effects audio to. Buffers may
 alias with buffers of \c out. Individual NULL buffers are permitted and will cause to skip mixing any audio into that buffer.
  * @param nout Count of arrays in \c out. Must be a multiple of 2
-(because of stereo) and in the range <code>0 <= nout/2 <= fluid_synth_count_audio_channels().
- Note that zero value is valid and allows to skip mixing dry audio in all out output buffers.</code>.
+(because of stereo) and in the range <code>0 <= nout/2 <= fluid_synth_count_audio_channels()</code>.
+ Note that zero value is valid and allows to skip mixing dry audio in all out output buffers.
  * @param out Array of buffers to store (dry) audio to. Buffers may
 alias with buffers of \c fx. Individual NULL buffers are permitted and will cause to skip mixing any audio into that buffer.
  * @return #FLUID_OK on success,
  * #FLUID_FAILED otherwise,
- *  - fx NULL while nfx > 0, or out NULL while nout > 0.
- *  - nfx or nout not multiple of 2.
- *  - len < 0.
- *  - both nfx and nout set to 0 while len > 0
- *  - nfx or nout greater than respective internal mixer buffer count.
+ *  - <code>fx == NULL</code> while <code>nfx > 0</code>, or <code>out == NULL</code> while <code>nout > 0</code>.
+ *  - \c nfx or \c nout not multiple of 2.
+ *  - <code>len < 0</code>.
+ *  - \c nfx or \c nout exceed the range explained above.
  *
  * @parblock
  * @note The owner of the sample buffers must zero them out before calling this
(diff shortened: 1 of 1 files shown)


derselbst:
Passing NULL as buffer while at the same time specifying `nout>0` seems more like a bug (or an abuse) to me. This condition should better be handled as `fluid_return_val_if_fail` by returning FLUID_FAILED.

Apart from that, I really like the comments :)

jjceresa:
>Passing NULL as buffer while at the same time specifying nout>0 seems more like a bug (or an abuse) to me.

yes, it is more appropriate to return FLUID_FAILED to the caller.

jjceresa:
Please, because of the xml-like notation in fluid_synth_process's comment header, I'm not sure if comment addition were correctly placed. Could you look this ?

src/synth/fluid_synth.c
```diff
@@ -3704,45 +3714,84 @@ fluid_synth_process_LOCAL(fluid_synth_t *synth, int len, int nfx, float *fx[],
...
+    */
     fluid_return_val_if_fail(len >= 0, FLUID_FAILED);
     fluid_return_val_if_fail(len != 0, FLUID_OK); // to avoid raising FE_DIVBYZERO below
 
+    /*
+      Now len is > 0. This will lead in rendering in internal mixer buffer.
+      However this rendering must not occur if both nfx and nout are set to 0.
+      So now we check that both nfx and nout set to 0 while len > 0, is invalid.
+    */
+    fluid_return_val_if_fail(nfx || nout, FLUID_FAILED);
```
derselbst, line 3743: This change is incorrect and breaks the unit test. If `len>0` while all buffers are NULL, rvoice_mixer will still render `len `samples, but nothing will be mixed.

src/synth/fluid_synth.c
```diff
@@ -3704,45 +3714,84 @@ fluid_synth_process_LOCAL(fluid_synth_t *synth, int len, int nfx, float *fx[],
...
+
+    /* fx NULL while nfx > 0 is invalid */
+    fluid_return_val_if_fail((fx != NULL) || (nfx == 0), FLUID_FAILED);
+    /* nfx must be multiple of 2. Note that 0 value is valid and
+       allows to skip mixing in fx output buffers
+    */
     fluid_return_val_if_fail(nfx % 2 == 0, FLUID_FAILED);
+
+    /* out NULL while nout > 0 is invalid */
+    fluid_return_val_if_fail((fx != NULL) || (nfx == 0), FLUID_FAILED);
```
derselbst, line 3726: Copy paste mistake

src/synth/fluid_synth.c
```diff
@@ -3652,15 +3652,24 @@ fx[ ((k * fluid_synth_count_effects_channels() + j) * 2 + 1) % nfx ]  = right_bu
  *
  * @param synth FluidSynth instance
  * @param len Count of audio frames to synthesize and store in every single buffer provided by \p out and \p fx.
- * @param nfx Count of arrays in \c fx. Must be a multiple of 2 (because of stereo)
- * and in the range <code>0 <= nfx/2 <= (fluid_synth_count_effects_channels() * fluid_synth_count_effects_groups())</code>.
+ *  Zero value is permitted, the function does nothing and return FLUID_OK.
+ * @param nfx Count of arrays in \c fx. Must be a multiple of 2 (because of stereo).
+ * and in the range <code>0 <= nfx/2 <= (fluid_synth_count_effects_channels() * fluid_synth_count_effects_groups()).
+  Note that zero value is valid and allows to skip mixing effects in all fx output buffers.</code>.
```
derselbst, line 3658:
```suggestion
 * and in the range <code>0 <= nfx/2 <= (fluid_synth_count_effects_channels() * fluid_synth_count_effects_groups())</code>.
  Note that zero value is valid and allows to skip mixing effects in all fx output buffers.
```

src/synth/fluid_synth.c
```diff
@@ -3652,15 +3652,24 @@ fx[ ((k * fluid_synth_count_effects_channels() + j) * 2 + 1) % nfx ]  = right_bu
...
  * @param fx Array of buffers to store effects audio to. Buffers may
-alias with buffers of \c out. NULL buffers are permitted and will cause to skip mixing any audio into that buffer.
+alias with buffers of \c out. Individual NULL buffers are permitted and will cause to skip mixing any audio into that buffer.
  * @param nout Count of arrays in \c out. Must be a multiple of 2
-(because of stereo) and in the range <code>0 <= nout/2 <= fluid_synth_count_audio_channels()</code>.
+(because of stereo) and in the range <code>0 <= nout/2 <= fluid_synth_count_audio_channels().
+ Note that zero value is valid and allows to skip mixing dry audio in all out output buffers.</code>.
```
derselbst, line 3663:
```suggestion
(because of stereo) and in the range <code>0 <= nout/2 <= fluid_synth_count_audio_channels()</code>.
 Note that zero value is valid and allows to skip mixing dry audio in all out output buffers.
```

src/synth/fluid_synth.c
```diff
@@ -3652,15 +3652,24 @@ fx[ ((k * fluid_synth_count_effects_channels() + j) * 2 + 1) % nfx ]  = right_bu
...
-alias with buffers of \c fx. NULL buffers are permitted and will cause to skip mixing any audio into that buffer.
- * @return #FLUID_OK on success, #FLUID_FAILED otherwise.
+alias with buffers of \c fx. Individual NULL buffers are permitted and will cause to skip mixing any audio into that buffer.
+ * @return #FLUID_OK on success,
+ * #FLUID_FAILED otherwise,
+ *  - fx NULL while nfx > 0, or out NULL while nout > 0.
+ *  - nfx or nout not multiple of 2.
+ *  - len < 0.
+ *  - both nfx and nout set to 0 while len > 0
+ *  - nfx or nout greater than respective internal mixer buffer count.
```
derselbst, line 3672:
```suggestion
 *  - <code>fx == NULL</code> while <code>nfx > 0</code>, or <code>out == NULL</code> while <code>nout > 0</code>.
 *  - \c nfx or \c nout not multiple of 2.
 *  - <code>len < 0</code>.
 *  - \c nfx or \c nout exceed the range explained above.
```

src/synth/fluid_synth.c
jjceresa: Oups !, thanks

src/synth/fluid_synth.c
jjceresa: Ok.

src/synth/fluid_synth.c
jjceresa: Ok, thanks

src/synth/fluid_synth.c
jjceresa: Thanks. I assume that these xml-like notation will be read by a tool that produce formatted documentation in the Wiki, isn't it ?

derselbst: The tool is called doxygen and generates the [API doc](http://fluidsynth.org/api). Not to be confused with the [wiki](https://github.com/FluidSynth/fluidsynth/wiki).

derselbst:
Thanks!

src/synth/fluid_synth.c
jjceresa: >The tool is called doxygen and generates the API doc and supports embedding HTML.

Ok, thanks.

> Not to be confused with the wiki.

Yes. In fact the wiki is only manually "generated".

PR #666, opened 2020-08-31
Use a runtime check to detect version of libinstpatch

derselbst:
It could be that during runtime an older version of libinstpatch is used than the one fluidsynth was compiled against. In this case, libinstpatch will fail to load DLS fonts, because libinstpatch's initialization semantics don't match those compiled into fluidsynth.


jjceresa:
Good found !. This PR have the diabolic number 666 well suited for this kind of bug ;)

Issue #665, opened 2020-08-31
About audio driver "multi channels"


jjceresa:
Fluidsynth `mixer` allows to produce audio on multiple `stereo outputs` (using **synth.audio-groups** > 1).
This allows to map MIDI channel on separate output.
Fluidsynth can have more than one `Fx unit `(using **synth.effects-groups** > 1)
This allows to map MIDI channel on separate fx unit input.

Actually there is only jack driver that use multiple output capabilities.

Today many low cost audio device could have: 2,4,6 or 8 audio channels. So it is interesting to bring
the mixer capabilities to these audio devices.

Here, I added  multi channel capabilities on `dsound`and `waveout`driver that work well tested on 2 audio devices.
In short, this multi channel support is based on new functions (`fluid_synth_write_s16_channels()` and 
`fluid_synth_write_s16_channels()`) that use a large part of the code of actual **fluid_synth_write_s16()** and 
`fluid_synth_write_float()` functions. So simply, these new drivers call `fluid_synth_write_s16_channels()` instead of fluid_synth_write_s16() and `fluid_synth_write_float_channels()` instead of fluid_synth_write_float().
I will do a PR to propose this enhancement.

I think that `alsa`driver could be adapted similarly to call these new functions. Unfortunately I have no linux machine equipped with the required audio devices (4,6 or 8 channels).

If you are interested and if you have the required audio devices on linux, I can help to place the changes in **fluid_alsa.c**.

derselbst:
Sounds good, go ahead :+1:

PR #667, opened 2020-08-31
Add multi channels support for audio driver.

jjceresa:
This PR addresses https://github.com/FluidSynth/fluidsynth/issues/665?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDExMDcxMjEzNzY6MzI3ODEyOTQ%3D#issuecomment-683648045
1) Add new functions for multi channels support: `fluid_synth_write_float_channels()`, `fluid_synth_write_s16_channels()`
2) `dsound` and `waveout` driver make use of this support. tested on 2 audio devices: 
    - creative SB Live! (6 channels).
    - Realtek: ALC889A (8 channels).

### DIFF
--- a/src/drivers/fluid_dsound.c
+++ b/src/drivers/fluid_dsound.c
@@ -19,9 +19,11 @@
  */
 
 
+
 #include "fluid_synth.h"
 #include "fluid_adriver.h"
 #include "fluid_settings.h"
+
 
 #if DSOUND_SUPPORT
 
@@ -37,21 +39,60 @@ static DWORD WINAPI fluid_dsound_audio_run(LPVOID lpParameter);
 
 static char *fluid_win32_error(HRESULT hr);
 
+/**
+* The driver handle multiple channels.
+* Actually the number maximum of channels is limited to  2 * DSOUND_MAX_STEREO_CHANNELS.
+* The only reason of this limitation is because we dont know how to define the mapping
+* of speakers for stereo output number above DSOUND_MAX_STEREO_CHANNELS.
+*/
+/* Maximum number of stereo outputs */
+#define DSOUND_MAX_STEREO_CHANNELS 4
+/* speakers mapping */
+static DWORD channel_mask_speakers[DSOUND_MAX_STEREO_CHANNELS] =
+{
+    /* 1 stereo output */
+    {
+        SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT
+    },
+    /* 2 stereo outputs */
+    {
+        SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT |
+        SPEAKER_BACK_LEFT|SPEAKER_BACK_RIGHT
+    },
+    /* 3 stereo outputs */
+    {
+        SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT |
+        SPEAKER_FRONT_CENTER|SPEAKER_LOW_FREQUENCY |
+        SPEAKER_BACK_LEFT|SPEAKER_BACK_RIGHT
+    },
+    /* 4 stereo outputs */
+    {
+        SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT |
+        SPEAKER_FRONT_CENTER|SPEAKER_LOW_FREQUENCY |
+        SPEAKER_BACK_LEFT|SPEAKER_BACK_RIGHT |
+        SPEAKER_SIDE_LEFT|SPEAKER_SIDE_RIGHT
+    }
+};
+
 typedef struct
 {
     fluid_audio_driver_t driver;
-    LPDIRECTSOUND direct_sound;
-    LPDIRECTSOUNDBUFFER prim_buffer;
-    LPDIRECTSOUNDBUFFER sec_buffer;
-    HANDLE thread;
+    LPDIRECTSOUND direct_sound;      /* dsound instance */
+    LPDIRECTSOUNDBUFFER prim_buffer; /* dsound buffer*/
+    LPDIRECTSOUNDBUFFER sec_buffer;  /* dsound buffer */
+
+    HANDLE thread;        /* driver task */
     DWORD threadID;
-    fluid_synth_t *synth;
-    fluid_audio_callback_t write;
-    HANDLE quit_ev;
-    int   bytes_per_second;
-    DWORD buffer_byte_size;
-    DWORD queue_byte_size;
-    DWORD frame_size;
+    fluid_synth_t *synth; /* fluidsynth instance */
+    /* callback called by the task for audio rendering in dsound buffers */
+    fluid_audio_channels_callback_t write;
+    HANDLE quit_ev;       /* Event object to request the audio task to stop */
+
+    int   bytes_per_second; /* number of bytes per second */
+    DWORD buffer_byte_size; /* size of one buffer in bytes */
+    DWORD queue_byte_size;  /* total size of all buffers in bytes */
+    DWORD frame_size;       /* frame size in bytes */
+    int channels_count; /* number of channels in audio stream */
 } fluid_dsound_audio_driver_t;
 
 typedef struct
@@ -60,6 +101,7 @@ typedef struct
     char *devname;
 } fluid_dsound_devsel_t;
 
+/* enumeration callback to add "device name" option on setting "audio.dsound.device" */
 BOOL CALLBACK
 fluid_dsound_enum_callback(LPGUID guid, LPCTSTR description, LPCTSTR module, LPVOID context)
 {
@@ -69,6 +111,11 @@ fluid_dsound_enum_callback(LPGUID guid, LPCTSTR description, LPCTSTR module, LPV
     return TRUE;
 }
 
+/* enumeration callback to look if a certain device exists and return its GUID.
+   @context, (fluid_dsound_devsel_t *) context->devname provide the device name to look for.
+             (fluid_dsound_devsel_t *) context->devGUID pointer to return device GUID.
+   @return TRUE to continue enumeration, FALSE otherwise.
+*/
 BOOL CALLBACK
 fluid_dsound_enum_callback2(LPGUID guid, LPCTSTR description, LPCTSTR module, LPVOID context)
 {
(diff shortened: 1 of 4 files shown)


derselbst:
Before going down into a detailed review, I keep thinking about whether `fluid_synth_write_float_channels` and `fluid_synth_write_s16_channels` is the right way to go. Perhaps we should better use `fluid_synth_process()` and manually translate its planar buffers into interleaved PCM. The things I have in mind:

1. `fluid_synth_process` is ready for surround audio.
1. It should be possible to  remove the extra mixing currently done in `fluid_synth_process` by passing the user-provided buffers directly into `rvoice_mixer`.
1. The effects are always written to the first stereo buffer.
1. 200 new lines of code in `fluid_synth.c` vs. a functionality that (I think) we already have.
1. Having a possible "planar to interleaved PCM translation with optional dithering" might turn out as a functionality that we could add to our public API if requested; whereas I'd be cautious adding `fluid_synth_write_float_channels` or `fluid_synth_write_s16_channels` adding to the public API. (More rendering functions make it more confusing, IMO.)


For point 2. it would be helpful if we don't have more "generic" audio rendering interfaces that we have to care about.

For point 3. I've seen that you #668 would address this. However, I have always thought of this `fluid_rvoice_mixer_set_mix_fx()` more as a hack that we better should get rid of in the future. And IMO `fluid_synth_process` with its flexible buffer setup allows us to do this.

So, I'm not strongly against these two functions, I'm just thinking of alternatives. And my proposal would be to use `fluid_synth_process` and while performing the planar to interleaved PCM conversion, also perform the optional dithering. If performance turns out to be a bottleneck, we can still fall back to your provided solution.

What do you think?

jjceresa:
>Before going down into a detailed review, I keep thinking about whether fluid_synth_write_float_channels and fluid_synth_write_s16_channels is the right way to go. Perhaps we should better use fluid_synth_process() and manually translate its planar buffers into interleaved PCM.

Actually `fluid_synth_write_float()`, `fluid_synth_write_float() ` or similar `fluid_synth_write_xxxx_channels()` can be called by all audio drivers. These function accept a flexible buffer organisation (interleaved/non interleaved). This is required by the `host audio driver` that dictates how the buffers must be written (interleaved/not interleaved) by these functions. (for example alsa driver require interleaved for PCM and non interleaved for float sample). So, I see these functions as the simpler appropriate direct `interface `between the `mixer` and any `host device driver` to final load speakers.

>point 1,2

I see `fluid_synth_process()` as an appropriate generic `interface `between the `mixer `and any `application` that intends to do more processing. This interface exposes the main buffer and fx buffer. So the application can do what it wants with these buffers (map, process, mix).
fluid_synth_process() is probably the only one rendering API that an application should keep in mind.

>whereas I'd be cautious adding fluid_synth_write_float_channels or fluid_synth_write_s16_channels adding to the public API. (More rendering functions make it more confusing, IMO.)

Yes, and for now I don't think that `fluid_synth_write_xxxx_channels()` should be public API (IMO they are only useful for driver and should stay private).

>Point 3 and  PR 668

The intends was to propose an API that allows to
- map any MIDI channel to any dry buffer.
- map any MIDI channel to any fx unit input.
- map any fx output to any dry buffer.

The motivation is to get (for example):
- distinct MIDI instrument  played on distinct stereo speakers.
- distinct MIDI instrument having its own fx unit.
- instrument fx output not necessarily mixed on the same speakers that the one used for this instrument dry audio.

All these feature are possible if the mixer keeps its fx mix mode capability. 
Please see attached figure below.

>200 new lines of code in fluid_synth.c vs. a functionality that (I think) we already have.

fluid_synth_write_xxxx_channels() each have 20 lines of required new code.
fluid_synth_write_s16(), fluid_synth_write_float() each require 4 lines of new code because
they call respective fluid_synth_write_xxxx_channels() .
Total: 50 lines of new code. 150 lines of comments !.   ;)

jjceresa:
I have added the missing document in branch https://github.com/FluidSynth/fluidsynth/pull/668

jjceresa:
Thinking about points 2 and 5

>2 It should be possible to remove the extra mixing currently done in fluid_synth_process by passing the user-provided buffers directly into rvoice_mixer

Ok, that means that:
- the mixer (`fluid_rvoice_buffers_mix()`) should be able to put the dry audio `directly` in the caller-provided `dry buffers`.
- similarly `fluid_rvoice_mixer_process_fx()` should be able to put fx audio `directly` in the caller-provided `fx buffers`.
That means that before calling `fluid_synth_render_blocks()`, the mixer should have knowledge of  these `caller-provided buffers`.
2.1) So, from inside fluid_synth_process() an "internal API" is required to set the mixer with theses buffers. Up to the caller of 'fluid_synth_process()' to decide if the `fx buffers` should be mixed in the same buffer than `dry buffer`.

>5-Having a possible "planar to interleaved PCM translation with optional dithering" might turn out as a `functionality` that we could add to our public API if requested.

5.1) assuming `fluid_synth_process()` is called by an `audio driver`, it seems that this API `functionality` is required because before calling `fluid_synth_process()`, the `mixer` must  know **how** the buffer (possibly provided by the host device driver) should be written:
- (interleaved/not interleaved), 
- float or PCM  (with/without dithering).
- another option could be to request the mixer to force the mixing of `fx audio` into the dry buffer. This is useful for audio driver that intends fx mixed with dry audio. This avoid to provide and prepare fx buffer each time fluid_synth_process() is called.

That sounds as a mixer enhancement as skipping actual internal mixer buffers should be a bit faster.
Also this should simplify adding support for audio driver with multiple output.

All this means also that `fluid_synth_write_float()` and `fluid_synth_write_s16()` should be updated and may be
part of functionalities inside should be moved into the mixer. 

Both these "mixer API"  functionalities (2.1)  and (5.1) need to be designed carefully.
Now I have talked to much, I need to drink something.

derselbst:
Ok, I understand your points.

It means that internally, we are not using our existing rendering interfaces, we rather make old-existing ones more generic. At the same time, there seems to be a need for an audio rendering interface that supports interleaved rendering. Actually, [I do have the same need in my app, currently with manual planar to interleaved translation though](https://github.com/derselbst/ANMP/blob/ca873cb2692dae2781f155d36502592f3899fb3b/src/InputLibraryWrapper/FluidsynthWrapper.cpp#L873-L888).

On the other hand and without major changes to `rvoice_mixer`, my point 2.  is only possible with an interface like `fluid_synth_process` ...It's a dilemma...

I think that for now, I'll take your change. And if I ever get to implement my point 2. I'll also consolidate the rendering interfaces.

I'll give you a  more detailed review soon. And we need to find out why the unit tests are failing.

src/drivers/fluid_waveout.c
```diff
@@ -36,12 +36,48 @@
...
+* The driver handle multiple channels.
+* Actually the number maximum of channels is limited to  2 * WAVEOUT_MAX_STEREO_CHANNELS.
+* The only reason of this limitation is because we dont know how to define the mapping
+* of speakers for stereo output number above WAVEOUT_MAX_STEREO_CHANNELS.
+*/
+/* Maximum number of stereo outputs */
+#define WAVEOUT_MAX_STEREO_CHANNELS 4
+
+/* speakers mapping */
+static DWORD channel_mask_speakers[WAVEOUT_MAX_STEREO_CHANNELS] =
```
derselbst, line 49: This array is a duplicate from dsound. Would it make sense  to remove the `static` qualifier and simply reuse the `channel_mask` from the dsound?

src/drivers/fluid_waveout.c
```diff
@@ -201,19 +300,29 @@ new_fluid_waveout_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
...
-    wfx.nSamplesPerSec  = frequency;
-    wfx.nBlockAlign     = sample_size * wfx.nChannels;
-    wfx.nAvgBytesPerSec = frequency * wfx.nBlockAlign;
+    /* Initialize the format structure */
+    wfx.Format.nChannels  = synth->audio_channels * 2;
+    if(synth->audio_groups > WAVEOUT_MAX_STEREO_CHANNELS)
+    {
+        FLUID_LOG(FLUID_ERR, "Channels number %d exceed internal limit %d",
+                             wfx.Format.nChannels, WAVEOUT_MAX_STEREO_CHANNELS * 2);
+        return NULL;;
```
derselbst, line 309: `;;`

src/drivers/fluid_dsound.c
```diff
@@ -369,6 +476,37 @@ static DWORD WINAPI fluid_dsound_audio_run(LPVOID lpParameter)
...
+       dsound expects interleaved channels in a unique buffer.
+       For example 4 channels (c1, c2, c3, c4) and n samples:
+       { s1:c1, s1:c2, s1:c3, s1:c4,  s2:c1, s2:c2, s2:c3, s2:c4,...
+         sn:c1, sn:c2, sn:c3, sn:c4 }.
+
+       So, channels_off[], channnel_incr[] tables should initialized like this:
+         channels_off[0] = 0    channels_incr[0] = 4
+         channels_off[1] = 1    channels_incr[1] = 4
+         channels_off[2] = 2    channels_incr[2] = 4
+         channels_off[3] = 3    channels_incr[4] = 4
```
derselbst, line 495: It should be `channels_incr[3]`

src/drivers/fluid_waveout.c
```diff
@@ -66,6 +103,37 @@ static DWORD WINAPI fluid_waveout_synth_thread(void *data)
...
+       MME expects interleaved channels in a unique buffer.
+       For example 4 channels (c1, c2, c3, c4) and n samples:
+       { s1:c1, s1:c2, s1:c3, s1:c4,  s2:c1, s2:c2, s2:c3, s2:c4,...
+         sn:c1, sn:c2, sn:c3, sn:c4 }.
+
+       So, channels_off[], channnel_incr[] tables should initialized like this:
+         channels_off[0] = 0    channels_incr[0] = 4
+         channels_off[1] = 1    channels_incr[1] = 4
+         channels_off[2] = 2    channels_incr[2] = 4
+         channels_off[3] = 3    channels_incr[4] = 4
```
derselbst, line 123: same

src/drivers/fluid_waveout.c
jjceresa: Both dsound and waveout are independent cmake option.

derselbst: Oh, right. Ok then, never mind.

jjceresa:
>At the same time, there seems to be a need for an audio rendering interface that supports interleaved rendering.
Yes this audio rendering interface should support both` interleaved` (or non-interleaved) rendering.

>And we need to find out why the unit tests are failing.

Hope that commit e9a9d35  will fix that ?

derselbst:
Ok, seems to work now.

PR #668, opened 2020-08-31
Mapping of fx unit output to dry buffers in mix mode.

jjceresa:
Actually all fx unit output (in mix mode) are mapped to the `first buffer`.
This is not appropriate for synth.audio-groups > 1

This PR allows the mapping of fx output based on `fx unit index` and `synth.audio-groups` value.
This allows to get the `fx output `mixed to the respective  `buffer` on which a `MIDI channel` is mapped.
For example: with `synth.audio-groups = 3` and  `synth.effect-groups = 3`:
- MIDI chan 0 (dry + fx0) is mapped to buf 0
- MIDI chan 1 (dry + fx1) is mapped to buf 1
- MIDI chan 2 (dry + fx2) is mapped to buf 2


src/rvoice/fluid_rvoice_mixer.c
```diff
@@ -164,15 +166,22 @@ fluid_rvoice_mixer_process_fx(fluid_rvoice_mixer_t *mixer, int current_blockcoun
...
             int buf_idx = f * fx_channels_per_unit + SYNTH_REVERB_CHANNEL;
-            
-            for(i = 0; i < current_blockcount * FLUID_BUFSIZE; i += FLUID_BUFSIZE)
+            int samp_idx = buf_idx * FLUID_MIXER_MAX_BUFFERS_DEFAULT * FLUID_BUFSIZE;
+            int sample_count = current_blockcount * FLUID_BUFSIZE;
+
+            /* in mix mode, map fx out_rev at index f to a dry buffer at index dry_idx */
+            if(mixer->mix_fx_to_out)
+            {
+                /* dry buffer mapping, should be done more flexible in the future */
```
derselbst, line 175: Why do you think that "dry buffer mapping, should be done more flexible in the future"? IMO, if one wants a more flexible way, one should use `fluid_synth_process()`, in which case `mixer->mix_fx_to_out` will be false. And if we wanted to bring this "more flexible" mapping to the audio drivers as well (e.g. via an extra setting or so), we could/should/would use `fluid_synth_process()`.

derselbst (commented):
Overall this looks good to me, one code comment below. I also like the chart, but I would like to ask you for some minor changes:

- Could you add a legend (or similar) that explains those `/s` and `/m` symbols in your flow chart?
- `synth.fx-groups` should better be called `synth.effects-groups`
- Sometimes you use bounding boxes with a dotted line and sometimes with a solid line; I can imaging that e.g. the dotted around the ladspa box this mean "optional"; however, the dotted box around the "in1" ... "ink" boxes is not quite clear to me
- In the LADSPA box, the one "fx plugin" box is hard to read due to overlap
- "fx buf 1" and "fx buf k" arrows currently end in void; would it be worth to add a box "fluid_synth_process()" there?

mawe42:
Sorry, havent' got enough time to comment in details here. Just a quick suggestion: if we add useful charts like the one added in this PR, maybe we can also add the source file or at least an editable version (SVG?) as well? Otherwise those files can only be updated by completely recreating them (or by asking the original contributer to provide an update).

derselbst:
> maybe we can also add the source file or at least an editable version

Yep, good point. I guess it's .doc :)

src/rvoice/fluid_rvoice_mixer.c
jjceresa: This comment isn't sufficient and confusing. Please see comment below about  "dry buffer mapping, should be done more flexible in the future".

jjceresa:
>Could you add a legend (or similar) that explains those /s and /m symbols in your flow chart?

This chart is just a quick draft and incomplete. It should be part of a textual document that explains how the mixer behave with true sentences.

>Why do you think that "dry buffer mapping, should be done more flexible in the future"? 

I just intended to show that  the 3 mapping selectors (`MIDI chan-to-out,` `MIDI chan-to-fx`,  and `fx-to-out` ), are closely related and oriented MIDI channel. In others words when the musician think for a MIDI instrument mapping configuration, he decide of this 3 mapping at MIDI level:
- MIDI chan **x** to dry buf **i** (actually i = **x** % **audio-groups**)
- MIDI chan **x** to fx unit **j**  (actually **j** = x % **effects-groups**)
- fx **j**  to dry buf **k**         (actually **k** = **j** % audio-groups)
Actually these mapping are rigid an "should be done more flexible in the future" with the help of API.

>Just a quick suggestion: if we add useful charts like the one added in this PR, maybe we can also add the source file or at least an editable version (SVG?) as well?

Yes actually, I use powerpoint to do quick synoptic and functional chart. I will update this one and provide the source.

jjceresa:
>Yep, good point. I guess it's .doc :)


When making a document that need text + figure, synoptic... I usually use `Word` for text, but I dont use the build-in image editor in incorporated in Word. I prefer using `powerpoint `for fast simple figure editing which is pain less for me. Then I copy -paste the figure in the right place in the .doc. Once the figure is in the.doc, it is still editable by Word 's image editor.
the .doc.
If you are interested, I can provide a .doc with the image above to try this ?.
If you have "open office" you have also `powerpoint`.

jjceresa:
>Yep, good point. I guess it's .doc :)


When making a document that need text + figure, synoptic... I usually use `Word` for text, but I dont use the build-in image editor in incorporated in Word. I prefer using `powerpoint `for fast simple figure editing which is pain less for me. Then I copy -paste the figure in the right place in the .doc. Once the figure is in the.doc, it is still editable by Word 's image editor.
the .doc.
If you are interested, I can provide a .doc with the image above to try this ?.
If you have "open office" you have also `powerpoint`.

derselbst:
You can just provide the powerpoint, that's fine. We can write up a markdown file and just reference to the PDF.

derselbst:
Great! I extracted the chart from this PR and committed it separately.

PR #673, opened 2020-09-09
Fx unit api

jjceresa:
This PR addresses #669 point 2.1.
It proposes set/get API functions to change/read  fx unit parameters.
The deprecated shell commands are updated. Now the commands line have 2 parameters:
- first parameter is the fx unit index.
- second parameter is the value to apply to the fx unit.


jjceresa:
The following actual API: 
- fluid_synth_set_chorusXX(synth, param);
- fluid_synth_get_chorus_XX(synth, param);
- fluid_synth_set_reverbXX(synth, param);
- fluid_synth_get_reverbXX(synth, param);
becomes now redundant with the new one  called with init_idx set to -1:
- fluid_synth_set_chorusXX**2**(synth, **unit_idx**,param);
- fluid_synth_get_chorus_XX**2**(synth,**unit_idx**, param);
- fluid_synth_set_reverbXX**2**(synth, **unit_idx**, param);
- fluid_synth_get_reverb_XX**2**(synth, **unit_idx**, param);

That means that actual API should not be called by new application and should be deprecated sooner or later. Perhaps we could ask to the mailist if these deprecation will cause backward issues ?

jjceresa:
The following actual API: 
- fluid_synth_set_chorusXX(synth, param);
- fluid_synth_get_chorus_XX(synth, param);
- fluid_synth_set_reverbXX(synth, param);
- fluid_synth_get_reverbXX(synth, param);

becomes now redundant with the new one  called with **unit_idx** set to -1:

- fluid_synth_set_chorusXX**2**(synth, **unit_idx**,param);
- fluid_synth_get_chorus_XX**2**(synth,**unit_idx**, param);
- fluid_synth_set_reverbXX**2**(synth, **unit_idx**, param);
- fluid_synth_get_reverb_XX**2**(synth, **unit_idx**, param);

That means that actual API should not be called by new application (and perhaps should be deprecated sooner or later ?). Perhaps we could ask to the mail-list if these deprecation could cause any backward issues ?

src/synth/fluid_synth.c
```diff
@@ -5667,35 +5931,48 @@ fluid_synth_set_chorus_full(fluid_synth_t *synth, int set, int nr, double level,
...
+
+        if(set & FLUID_CHORUS_SET_TYPE)
+        {
+            synth->chorus_type = type;
+        }
     }
 
-    param[0].i = set;
+    /* put fx unit index in bits [b15..b8] and set mask in bits [b7..b0] */
+    param[0].i = (fxunit_idx << 8) | (set & FLUID_CHORUS_SET_ALL) ;
```
derselbst, line 5975: Be careful with signed shifts: https://sonarcloud.io/project/issues?id=FluidSynth_fluidsynth&open=AXVgLmC8i5dl9RN5Jo-q&pullRequest=673&resolved=false&types=BUG

I would prefer to store both parameters separately.

jjceresa:
>Be careful with signed shifts

There is no problem with the signed shift because the negative value of `fxunit_idx` is -1 so there is no lost of sign. But I must admit it will be more readable to store both parameters separately. This will require to augment MAX_EVENT_PARAMS to 7.
I will do that change in future commit.

derselbst:
> There is no problem with the signed shift because the negative value of fxunit_idx is -1 so there is no lost of sign.

Left signed shifts are undefined. That means _anything_ can happen. Admittedly very unlikely for twos-complement. But we shouldn't rely on that.

> This will require to augment MAX_EVENT_PARAMS to 7. I will do that change in future commit.

Ok

derselbst:
Thanks. I'll try to find some time during the weekend to go over this in more detail.

src/synth/fluid_synth.c
```diff
@@ -5405,54 +5404,121 @@ fluid_synth_set_reverb_preset(fluid_synth_t *synth, unsigned int num)
...
+ *  must be in the range [-1..synth->effects_groups[. If -1 the changes are
+ *  applied to all fx units (same as fluid_synth_set_reverb()).
+ * @param roomsize Reverb room size value (0.0-1.0)
+ * @param damping Reverb damping value (0.0-1.0)
+ * @param width Reverb width value (0.0-100.0)
+ * @param level Reverb level value (0.0-1.0)
+ * @return #FLUID_OK on success, #FLUID_FAILED otherwise
+ *
+ * @note Not realtime safe and therefore should not be called from synthesis
+ * context at the risk of stalling audio output.
```
derselbst, line 5470: We need to check whether this non-realtime safeness is actually true.

src/bindings/fluid_cmd.c
```diff
@@ -1086,275 +1086,367 @@ fluid_handle_reverbpreset(void *data, int ac, char **av, fluid_ostream_t out)
...
 
-    if(ac < 1)
+    /* name and min/max values table */
+    static struct value values[NBR_REVERB_CDE] =
     {
-        fluid_ostream_printf(out, "rev_setdamp: too few arguments.\n");
-        return FLUID_FAILED;
-    }
+        {"room size", 0.0F, 1.0F}, {"damp", 0.0F, 1.0F},
+        {"width", 0.0F, 100.0F},   {"level", 0.0F, 1.0F}
```
derselbst, line 1197: The ranges should not be hardcoded. Instead, they should be retrieved with `fluid_settings_getnum_range()` and friends.

derselbst (commented):
A few general thoughts:

1. Do we really want to add all those `fluid_synth_set_reverb_group_roomsize|damp|with|level()` overloads? Thinking about it, I would prefer to have a single setter function with pointers to the arguments. Pointers can be NULL, if the user doesn't want to set the respective properties:
`fluid_synth_set_reverb_group(fluid_synth_t *synth, int fxunit_idx, const double* roomsize, const double* damping, const double* width, const double* level);` Likewise for chorus. @mawe42 What do you think?

2. For the getter functions I'm not sure, but I tend to say that here it would make sense to keep the separate getter functions.

jjceresa:
* @note Not realtime safe and therefore should not be called from synthesis
* context at the risk of stalling audio output.
>We need to check whether this non-realtime safeness is actually true

Assuming the "synthesis context" means "the application task that intend to call any rendering API", I see no race condition issue here because this API (like any other fx API) talks via the ring buffer.

jjceresa:
>Do we really want to add all those fluid_synth_set_reverb_group_roomsize|damp|with|level() overloads? Thinking about it, I would prefer to have a single setter function with pointers to the arguments. Pointers can be NULL, if the user doesn't want to set the respective properties:
fluid_synth_set_reverb_group(fluid_synth_t *synth, int fxunit_idx, const double* roomsize, const double* damping, const double* width, const double* level)

I understand your point, but theses setter functions with possible pointer NULL have the following inconvenient.
- still will require a little overload inside to build the parameters to pass to fluid_synth_set_reverb_full().
- is more complicated to use when the user want to change only one parameter. (And  the overload is moved to the application side).
- We lose the symmetry with respective getter functions.

mawe42:
> 1. Do we really want to add all those `fluid_synth_set_reverb_group_roomsize|damp|with|level()` overloads? Thinking about it, I would prefer to have a single setter function with pointers to the arguments. Pointers can be NULL, if the user doesn't want to set the respective properties:
>    `fluid_synth_set_reverb_group(fluid_synth_t *synth, int fxunit_idx, const double* roomsize, const double* damping, const double* width, const double* level);` Likewise for chorus. @mawe42 What do you think?
> 2. For the getter functions I'm not sure, but I tend to say that here it would make sense to keep the separate getter functions.

I'm not sure. It feels weird to pass in pointers to setter functions. And if we want separate getters, then I would vote to have separate setters as well, for consistency.

Another approach could be to use a param enum. Something like this:
```
enum fluid_reverb_param
{
    FLUID_REVERB_ROOMSIZE,
    FLUID_REVERB_DAMP,
    FLUID_REVERB_LEVEL,
    FLUID_REVERB_WIDTH,
};

FLUIDSYNTH_API int fluid_synth_set_reverb_group(fluid_synth_t *synth, int fxunit_idx, int param, double value);
FLUIDSYNTH_API double fluid_synth_get_reverb_group(fluid_synth_t *synth, int fxunit_idx, int param);
```

mawe42:
Some nit-picking on the naming of the functions and parameters:

In a function called `fluid_synth_set_reverb_group()`, I would expect at least one parameter to be named something with `group`. 

So if we want to call the different fx instances "groups", then we should do so consistently everywhere. Likewise if they should be called "fx units", then we should use that name in all places. So I would vote for 
```
fluid_synth_set_reverb_unit(fluid_synth_t *synth, int fxunit, ...)
```

Or if we want to call them "groups", then
```
fluid_synth_set_reverb_group(fluid_synth_t *synth, int group, ...)
```

Having two names for the same concept feels wrong and will make the API and documentation more difficult to understand.

jjceresa:
>Another approach could be to use a param enum.

Nice proposal.

***
>In a function called fluid_synth_set_reverb_group(), I would expect at least one parameter to be named something with group.

yes , of course this should be corrected.

***
I just realize that to get an internal fxunit(or fxgroup) (reverb or chorus) replaced by an external LADSPA effect, each internal effect unit need to have its own switch on/off. Actually and in this PR, there are only one switch on/off. We need to add the relevant API:

fluid_synth_set_reverb_group_on(fluid_synth_t *synth, int group, int on)
fluid_synth_set_chorus_group_on(fluid_synth_t *synth, int group, int on)

derselbst:
> Another approach could be to use a param enum.

Good idea. It would also allow to add new parameters without deprecating any existing API function.

> Actually and in this PR, there are only one switch on/off. We need to add the relevant API

Good point. Let's do it.

derselbst:
> > Another approach could be to use a param enum.
> 
> Good idea. It would also allow to add new parameters without deprecating any existing API function.

On the other hand, we have parameters of  different types, i.e. most of them are double, but chorus nr. is int. Should we promote it to double as well?

If yes, I'd vote to change the getter function to something like this:

`FLUIDSYNTH_API int fluid_synth_get_reverb_group(fluid_synth_t *synth, int fxunit_idx, int param, double* value);`

I.e. returning FLUID_OK|FAILED and returning the parameter in its own buffer.

jjceresa:
>>Another approach could be to use a param enum.
>>FLUIDSYNTH_API int fluid_synth_set_reverb_group(fluid_synth_t *synth, int fxunit_idx, int param, double value);
>Good idea. It would also allow to add new parameters without deprecating any existing API function.

Ah!, we have a name conflict with the existing API that set all parameters at the same time.
`fluid_synth_set_reverb_group(fluid_synth_t *synth, int fxunit_idx, double roomsize, double damping, double width, double level)`.
May be should we rename the API that set only one parameter something like that ?.
`FLUIDSYNTH_API int fluid_synth_set_reverb_group_param(fluid_synth_t *synth, int group, int param, double value)`

derselbst:
> we have a name conflict with the existing API that set all parameters at the same time.

With the enum approach, there is no need for a function that sets all parameters at once. This would be redundant.

mawe42:
Having the getter return FLUID_OK or FLUID_FAILED sounds good, as it would now be possible to specify an invalid parameter number.

And thinking about the function signature, I think the following names would be the best fit:
```
int fluid_synth_reverb_group_set_param(fluid_synth_t *synth, int group, int param, double value);
int fluid_synth_reverb_group_get_param(fluid_synth_t *synth, int group, int param, double *value);
```

I agree with Tom that the function with all parameters in the signature would be redundant. And including it would remove the possibility of simply adding more parameters to the effects without changing API function signatures. Or do we need a way to set all parameters at once?

mawe42:
Or if we wanted a shorter name, we could also drop the "_group" and simply use
```
int fluid_synth_reverb_set_param(fluid_synth_t *synth, int fx_group, int param, double value);
int fluid_synth_reverb_get_param(fluid_synth_t *synth, int fx_group, int param, double *value);
```

Also renames the group param to fx_group, to make it even clearer what number is expected here. After all, the corresponding setting is called "effects-groups".

derselbst:
> Or do we need a way to set all parameters at once?

No, we don't. The params are to be set upon synth initialization, before rendering. If params need to be changed afterwards, they need to be changed from synth context anyway.

> Or if we wanted a shorter name, we could also drop the "_group" and simply use

Sounds good to me.

src/synth/fluid_synth.c
```diff
@@ -5600,190 +5753,305 @@ fluid_synth_set_chorus_on(fluid_synth_t *synth, int on)
...
+        FLUID_API_RETURN(FLUID_FAILED);
     }
 
-    if(set & FLUID_CHORUS_SET_LEVEL)
+    /* check if chorus value is in max min range */
+    if(param == FLUID_CHORUS_TYPE) /* integer*/
     {
-        synth->chorus_level = level;
+        f_min = (double)FLUID_CHORUS_MOD_SINE;
+        f_max = (double)FLUID_CHORUS_MOD_TRIANGLE;
```
derselbst, line 5871: Casting an enum to double really strikes me. Counter proposal: Instead of simply passing `int on` to `fluid_synth_chorus_on()`, we provide it with the `enum fluid_chorus_mod`. Additionally, we'll need an extra value that defines `FLUID_CHORUS_OFF`.

jjceresa:
>Instead of simply passing int on to `fluid_synth_chorus_on()`, we provide it with the enum `fluid_chorus_mod`. Additionally, we'll need an extra value that defines FLUID_CHORUS_OFF.

You probably made a confusion. enum `fluid_chorus_mod` has nothing to do with `fluid_synth_chorus_on()`.
 
> Instead of simply passing int on....we'll need an extra value that defines FLUID_CHORUS_OFF.

In fact  `fluid_synth_chorus_on(), fluid_synth_reverb_on()` require a boolean `on` to enable/disable respective effects. So it seems that `int on` parameter is appropriate isn't ?. 
However perhaps could we rename these API  like this`fluid_synth_chorus_enable(fluid_synth_t *synth, int enable)` which is more 'naming consistent' with `fluid_rvoice_mixer_chorus_enable() (idem for reverb)?.

derselbst:
> > Instead of simply passing int on to `fluid_synth_chorus_on()`, we provide it with the enum `fluid_chorus_mod`. Additionally, we'll need an extra value that defines FLUID_CHORUS_OFF.
> 
> You probably made a confusion. [...] In fact `fluid_synth_chorus_on(), fluid_synth_reverb_on()` require a boolean `on` to enable/disable respective effects

No confusion. My point was: why does `fluid_synth_reverb_on` require a boolean?  Wouldn't it be better to supply the `enum fluid_chorus_mod`? For example if you wanted to use triangle chorus, you would currently need to 

```
fluid_synth_chorus_on(synth, 1, true );
fluid_synth_reverb_set_param(synth, 1, FLUID_CHORUS_TYPE, FLUID_CHORUS_MOD_TRIANGLE );
```

With my propsal this could be simplified to:

```
fluid_synth_chorus_on(synth, 1, FLUID_CHORUS_MOD_TRIANGLE );
```

derselbst:
> However perhaps could we rename these API like thisfluid_synth_chorus_enable(fluid_synth_t *synth, int enable) which is more 'naming consistent' with `fluid_rvoice_mixer_chorus_enable() (idem for reverb)?.

`fluid_rvoice_mixer_chorus_enable` is not public. I prefer to keep "on", it's shorter.

mawe42:
How about `fluid_synth_chrous_bypass`? Would be consistent with user interfaces of effects in DAWs and other programs. You instantiate the effect (synth.effects-groups), it's enabled by default and you can bypass it with a switch.

jjceresa:
>No confusion. My point was: why does fluid_synth_reverb_on require a boolean? Wouldn't it be better to supply the enum >fluid_chorus_mod? For example if you wanted to use triangle chorus, you would currently need to
>..With my propsal this could be simplified to:

fluid_synth_chorus_on(synth, 1, FLUID_CHORUS_MOD_TRIANGLE ); /* chorus triangle on */
fluid_synth_chorus_on(synth, 1, FLUID_CHORUS_MOD_SINE); /* chorus sinus on */
fluid_synth_chorus_on(synth, 1, FLUID_CHORUS_OFF); /* chorus off */

Ah, yes this is tricky.

jjceresa:
>How about fluid_synth_chrous_bypass? Would be consistent with user interfaces of effects in DAWs and other programs. You instantiate the effect (synth.effects-groups), it's enabled by default and you can bypass it with a switch.

The "by_pass" you refer  for those daws and other programs seems to work as a "short cut" functionality which means that the input signal is forwarded to the output and remains dry when the effect is bypassed and wet when the effect isn't bypassed.

In fluidsynth as a synthesizer the reverb/chorus effect are intended to  `add` little chorus or little reverb to dry acoustic instrument which have these natural effect (more or less). That means that the reverb or chorus path should be `off` to keep the instrument dry. Using "bypass" doesn't reflect internal fluidsynth's signal path topology.

derselbst:
> Ah, yes this is tricky.

Tricky in a sense of "hard-to-understand" or "good idea"?

> Using "bypass" doesn't reflect internal fluidsynth's signal path topology.

I agreee.

mawe42:
> Using "bypass" doesn't reflect internal fluidsynth's signal path topology.

Good points. You are right, it doesn't really fit.

jjceresa:
>Tricky in a sense of "hard-to-understand" or "good idea"?

As far I understand, the goal of your proposal is to simplify the user life by making (chorus) automatically set to "on" each time the user set a parameter (FLUID_CHORUS_MOD_SINE or FLUID_CHORUS_MOD_TRIANGLE). In order to be "simple-to-understand"  perhaps this  "good idea" should be applied to all chorus parameters. 
I mean, each time the user set any chorus parameter using `fluid_synth_chorus_set_param()(...,int  param,....)`  the chorus is automatically set to "on" except when enum `param` is FLUID_CHORUS_ONOFF for which chorus is set to "off" if `value` is FALSE or "on" otherwise.
This way we need only a single `fluid_synth_chorus_set_param() API` to handle chorus unit. This leads to get rid of fluid_synth_chorus_on() API and  fluid_synth_chorus_get_param(...,FLUID_CHORUS_ONOFF,....) can now return the state on/off of any chorus unit ......Same things for reverb.

derselbst:
> the chorus is automatically set to "on" except when enum param is FLUID_CHORUS_ONOFF for which chorus is set to "off" if value is FALSE

`fluid_synth_chorus_set_param` is receiving a `value` param of type double. You are suggesting to pass a boolean value as type double via this interface. This is a bad idea. I don't see a benefit in squashing the `fluid_synth_chorus_on` function into the `fluid_synth_chorus_set_param`. I don't think this interface should set any effect group to "on" in an automatic implicit way.

mawe42:
> each time the user set any chorus parameter using fluid_synth_chorus_set_param()(...,int param,....) the chorus is automatically set to "on" except when enum param is FLUID_CHORUS_ONOFF for which chorus is set to "off" if value is FALSE or "on" otherwise.

I think that would be a bad idea. It's an unexpected side-effect and would make life of application programmers really difficult. Consider a UI where you can set chorus params and have a separate on/off switch. Turning the level dial would automatically switch the effect on. That might not be what the app programmer wants.

mawe42:
> I don't see a benefit in squashing the fluid_synth_chorus_on function into the fluid_synth_chorus_set_param.

I agree. I think setting parameters of an effect and switching it on/off are two separate "levels" of control. So it makes sense to have two distinct functions here.

jjceresa:
> I don't see a benefit in squashing the fluid_synth_chorus_on function into the fluid_synth_chorus_set_param.

>I agree. I think setting parameters of an effect and switching it on/off are two separate "levels" of control. So it makes sense to have two distinct functions here.

>And I following this argument, I think the _on function should not control the chorus type but only switch it on or off.

Ok, so the actual on/off interface seems good enough.

derselbst:
> If we don't like casting enums to double, then we could cast the value param based on the param type instead

I don't quite understand what you mean. Are you suggesting to pass `values` as `void*`?

jjceresa:
>If we don't like casting enums to double, then we could cast the value param based on the param type instead
>>I don't quite understand what you mean.

Me too.

mawe42:
> > If we don't like casting enums to double, then we could cast the value param based on the param type instead
> > > I don't quite understand what you mean.
> 
> Me too.

That was in response to Toms comment here: https://github.com/FluidSynth/fluidsynth/pull/673#discussion_r518680654
But as that bit of code is now outdated, I guess it doesn't matter anymore.

derselbst:
> That was in response to Toms comment here: [#673 (comment)](https://github.com/FluidSynth/fluidsynth/pull/673#discussion_r518680654)
> But as that bit of code is now outdated, I guess it doesn't matter anymore.

It still matters because the user is still forced to pass the `enum fluid_chorus_mod` as `double` parameter via `fluid_synth_chorus_set_param`. This is not a clean interface. So again, I would vote for using

`fluid_synth_chorus_on(fluid_synth_t *synth, int fx_group, enum fluid_chorus_mod);`

If you don't like the name we could think of a different one:

* `fluid_synth_chorus_mode`
* `fluid_synth_chorus_type`
* `fluid_synth_chorus_state`

Since the enum is called "mod", my preference is to use `_mode`.

[If we decide for this change, we should do the same for the reverb function.]

mawe42:
I'm undecided... Would compilers complain about the following? (i'm really not sure)
```
fluid_synth_chorus_set_param(synth, -1, FLUID_CHORUS_TYPE, FLUID_MOD_TRIANGLE);
```

If not, then I think this is very similar to
```
fuid_synth_chrous_set_param(synth, -1, FLUID_CHORUS_NR, 50);
```
and while it's not squeaky clean, I still think it's ok.

If we decide that implicit casts from int for the setter, and requiring explicit cast back to int for the getter is unacceptable, then we should abandon the whole param enum idea and go back to individual getters and setters. Having some params use separate functions and other params use the generic one will be confusing, in my opinion. The separate _on functions are ok, because they control the effect on a different level. But nr and type are semantically the same as level, speed and depth, so they should be controllable in the same fashion.

jjceresa:
>The separate _on functions are ok, because they control the effect on a different level. But nr and type are semantically the same as level, speed and depth, so they should be controllable in the same fashion.

Yes, `_on` functions control the mixer, and `_set_param`  functions control individual fx unit (inside the mixer). For user it seems that theses separate API reflect well internal level of functionalities.

>f we decide that implicit casts from int for the setter, and requiring explicit cast back to int for the getter is unacceptable, then we should abandon the whole param enum idea and go back to individual getters and setters.

A compromise is possible by keeping the whole `param enum` idea for `double` and adding another function API param enum for `int`. (i.e fluid_synth_chorus_set_param_num( ..,int param, double value) , fluid_synth_chorus_set_param_int(..,int param, int value)). Of course, splitting the param enum interface this way will complicate a bit internal interface ?.

derselbst:
> Would compilers complain about the following?

No. From a purely technical point of view, I see no reason why casting an enum to double and back would not work. That doesn't mean that it will work fine in all scenarios. Keep in mind that fluidsynth is used by many programming languages. There is no guarantee, that there will a concellation where this does not work.

Apart from that, I feel ok with promoting a single int to double and cast it back. An enum, however, only has a few distinct values. Casting that to double feels like really poor design to me. An enum could also represent bitmasks. Would this survive an intermediate cast to double? I don't know.

Also, this approach would lead to redundant client code:

```c
fluid_synth_chorus_on(synth, 1, true);
fluid_synth_reverb_set_param(synth, 1, FLUID_CHORUS_TYPE, FLUID_CHORUS_MOD_TRIANGLE);
```

rather than

```c
fluid_synth_chorus_on(synth, 1, FLUID_CHORUS_MOD_TRIANGLE);
```

I think of it as two separate functions: one that affects the _mode_ of the effect, the other function affecting its effect _parameters_.

Chorus can be in three different modes: Sinus, Triangle, OFF.

Likewise for reverb: `FLUID_REVERB_MOD_ON`, `FLUID_REVERB_MOD_OFF`. If we ever wanted to add another type of reverb it would nicely integrate into this "mode" API function. (We had this discussion in the past, remember: Freeverb vs. FDN reverb)


---

So, if we can't agree on this, we should fall back to separate setter functions. Without the need for an "all-in-one" setter function:

```c
int fluid_synth_set_chorus_group_nr(fluid_synth_t *synth, int fxunit_idx, int nr);
int fluid_synth_set_chorus_group_level(fluid_synth_t *synth, int fxunit_idx, double level);
int fluid_synth_set_chorus_group_speed(fluid_synth_t *synth, int fxunit_idx, double speed);
int fluid_synth_set_chorus_group_depth(fluid_synth_t *synth, int fxunit_idx, double depth_ms);
int fluid_synth_set_chorus_group_type(fluid_synth_t *synth, int fxunit_idx, int type);
```

mawe42:
> Keep in mind that fluidsynth is used by many programming languages. There is no guarantee, that there will a concellation where this does not work.

So you are saying that there could be platforms or programming languages where a double/float can not accurately represent smaller (<= 16 bit) integer numbers? That would be very unusual, I think... But proving that something does not exist is hard, so I can understand if you want to be on the absolute safe side here. I guess we should move back to separate getters/setters for both effects then.

> Also, this approach would lead to redundant client code:
>```
>fluid_synth_chorus_on(synth, 1, true);
>fluid_synth_reverb_set_param(synth, 1, FLUID_CHORUS_TYPE, FLUID_CHORUS_MOD_TRIANGLE);
>```

I don't agree that this example contains redundant code. I think it's perfectly fine and much cleaner than the alternative. Consider the case when you want to configure a disabled chorus unit: you would have to ensure to remember the mode for later, when the user actually wants to switch it on. Or consider the case when a user wants to switch a chorus unit off and later on again. Before switching it off, you need to save the mode param somewhere only to be able to set it back to the original value.

mawe42:
One more comment about the double casting: LADSPA does the same, actually. All values, even control values for things that the plugin internally uses as an integer, are passed as floats. Now you could argue that LADSPA is not the nicest interface, but it's still widely used and I quite like it for it's simplicity.

derselbst:
> Or consider the case when a user wants to switch a chorus unit off and later on again. Before switching it off, you need to save the mode param somewhere only to be able to set it back to the original value.

Ok.

> LADSPA does the same, actually

I don't want to compare to other interfaces, because I don't know the historical background or the motivation of their developers back then.

Here, we are designing a new interface. And I think it's hard to make decisions here, because it's hard to think of a practical or musical sense for this new interface. I have to admit, that I still only see the pure technical need for this API. I'm still not convinced that it makes sense from a musical perspective, sry.

mawe42:
> I don't want to compare to other interfaces, because I don't know the historical background or the motivation of their developers back then.

Understood, and I agree. I just wanted to point out that there is a precedence for "float-only" interfaces. And it is an interface that I myself am very comfortable with, including the "float-only" aspect.

> I have to admit, that I still only see the pure technical need for this API.

Yes, I feel the same, I'm not invested in this feature. If it were up to me, then I think we should hide the fact that there are multiple effect instances used internally (or rather: keep it hidden).

But JJC said he really needs the real-time multi-channel mixing feature and also that he needs to be able to control the effect groups separately. And even though I don't understand why and why in this way, I still think if somebody has a real need, then we should be open to change. But every time I think about both of these changes more generally, I switch from "let's do it" to "I'm really not sure" and back again.

Fact is: we are currently changing or deprecating existing APIs that are used by a lot of people, to satisfy a use-case that so far only one person has expressed an interest in (or even understands why there is a need for it). I really hope it's worth it and that JJC will make good and frequent use of this feature :-)

jjceresa:
>Here, we are designing a new interface. And I think it's hard to make decisions here, because it's hard to think of a practical or musical sense for this new interface. 

>Yes, I feel the same, I'm not invested in this feature.

To me, having independent control fx reverb/chorus parameter on independent MIDI instruments (each on its MIDI channel) has a real musical sense. This is audible in many music.

>I have to admit, that I still only see the pure technical need for this API. I'm still not convinced that it makes sense from a musical perspective,sry

>we are currently changing or deprecating existing APIs that are used by a lot of people, to satisfy a use-case that so far only one person has expressed an interest in (or even understands why there is a need for it).

The "lot of people" your refer are impossible to number and  never gave input when questioned on the mailing list. I hope they are happy  when they change the reverb room-size and this impacts all MIDI instruments at the same time !.

I wouldn't wanna be the only person who breaks habits of  lot of people. So far I would prefer to abandon this PR.

mawe42:
> To me, having independent control fx reverb/chorus parameter on independent MIDI instruments (each on its MIDI channel) has a real musical sense. This is audible in many music.

I very much agree with this, using different effects on different MIDI instruments is common, normal, makes musical sense. What I don't quite understand is why you need the solution you are proposing. Why the many other solutions that already exist are not suitable. Even solutions already existing in fluidsynth: instead of the internal reverb and chorus, you could simply use multiple external LADSPA effects and control them interactively via the API to achieve the same effect. Then use a "unity gain amp" plugin to mix your buffers any way you like.
 
> > we are currently changing or deprecating existing APIs that are used by a lot of people, to satisfy a use-case that so far only one person has expressed an interest in (or even understands why there is a need for it).
> 
> The "lot of people" your refer are impossible to number and never gave input when questioned on the mailing list. I hope they are happy when they change the reverb room-size and this impacts all MIDI instruments at the same time !.

Of course, lot of people is just a guess. But I do think that the reverb and chorus APIs are the more well known parts of fluidsynth. And I could probably name a few high profile programs that make use of it.
 
> I wouldn't wanna be the only person who breaks habits of lot of people. So far I would prefer to abandon this PR.

Don't get me wrong... I'm not saying abandon the idea. It's not up to me anyway. I'm just saying I hope your need for this feature is real. That you will use it frequently and that it is not just a "this would make sense" idea. And these thoughts come into my head because I still (i know... I repeat myself) haven't understood where you are coming from. As Windows has only recently gained multi-channel support (thanks to you!), the use-case must be fairly new. So I'm wondering much of it is based in the real world and how much of it is a solution looking for a problem.

jjceresa:
>very much agree with this, using different effects on different MIDI instruments is common, normal, makes musical sense. What I don't quite understand is why you need the solution you are proposing. Why the many other solutions that already exist are not suitable. Even solutions already existing in fluidsynth: instead of the internal reverb and chorus, you could simply use multiple external LADSPA effects and control them interactively via the API to achieve the same effect. Then use a "unity gain amp" plugin to mix your buffers any way you like.

The solution you propose is to get rid of internal functionalities that already exist and replace these by external effects. For now I would prefer application the more simple as possible. This gives me more time as I prefer playing on piano rather then computer keyboard.

>As Windows has only recently gained multi-channel support (thanks to you!), the use-case must be fairly new.

This PR has nothing to do with audio multi-channels, and corresponds only to the real need already expressed in the first message.

mawe42:
> This PR has nothing to do with audio multi-channels, and corresponds only to the real need already expressed in the first message.

Sorry, but the first message does not express a need, it simply states what you want to change... not why. The only thing that could be taken as a motivation for this PR is the link to #669. And that issue is about the multi-channel real-time control you proposed. So I assumed that this PR here is simply the ground work for #669.

Ok, so you are saying that you need to control separate fx per MIDI channel even if the real-time multi-channel feature does not make it into FluidSynth. I hadn't understood that until now.

jjceresa:
>So I assumed that this PR here is simply the ground work for #669.

This PR is the ground work for #669 `point 2.1` only.   #669 point 1  is addressed by PR 672.
For clarity, I should have to split this issue 669 in two. Also 669 issue title (that I cannot update anymore) doesn't reflect point 2. Sorry.

jjceresa:
>So, if we can't agree on this, we should fall back to separate setter functions. Without the need for an "all-in-one" setter function

The all-in-one setter (param enum) function is really a nice idea as it allows to add parameters in the future without requiring new API. For example the reverb unit actually has only a "late stage" but we could add the missing "early stage" controlled by new enum values.

To avoid `double/int values` casting, I proposed a compromise to split the interface like this:
- fluid_synth_xxxx_set_param_num( ..,int param, double value) 
- fluid_synth_xxxx_set_param_int(..,int param, int value)).
- fluid_synth_xxxx_get_param_num( ..,int param, double value) 
- fluid_synth_xxxx_get_param_int(..,int param, int value)).
Opinions ?.

mawe42:
Thanks JJC, also a good approach. And similar to the way we currently handle the synth settings.

Even though Tom already said choices by other people are of no concern to us, I think its still interesting to see how other people have approached this problem.
- LADSPA: uses floats for all parameters, see `fluid_ladspa_effect_set_control()`.
- VST2(and 3?): also only use floats, but with a twist: here it is always 32-bit float params in the range 0.0 - 1.0.
- LV2: uses unions containing many different types, including public factory methods to create the correct union instance from a basic type.

But of course they have to use something generic because they can't know what people might want as parameters in their plugins. We know exactly which parameters our two effects accept.

So, my preference would be (in order):
1. generic setters/getters with floats
2. JJC new suggestion with typed generic setters/getters
3. The current style of specific setters/getters and one all-in-one getter/setter.

Options 2. and 3. can easily switch places for me, though. It depends on how likely it is that we will add or change the parameters of the internal effects. If it's not likely, then 3. would switch place with 2.

derselbst:
My problem with

```c
fluid_synth_xxxx_set_param_num( ..,int param, double value)
fluid_synth_xxxx_set_param_int(..,int param, int value)).
fluid_synth_xxxx_get_param_num( ..,int param, double value)
fluid_synth_xxxx_get_param_int(..,int param, int value)).
```

is that people need to look up the enum param and see which typed function to use. I, personally, consider this to be uncomfortable. However, if we really wanted to do this, we could also use a variable argument list:

```c
int fluid_synth_xxxx_set_param(synth, int param, ...);
```

Depending on `param`, variable argument number one will be treated as `int` or `double`. libtiff actually uses that approach: [TIFFGetField()](http://www.libtiff.org/man/TIFFGetField.3t.html)

jjceresa:
>My problem with
fluid_synth_xxxx_set_param_num( ..,int param, double value)
fluid_synth_xxxx_set_param_int(..,int param, int value)).
is that people need to look up the enum param and see which typed function to use.

No because, I forgot to mention that there would be also 2 distincts param enum type.
enum fluid_xxxx_param_num
{
....
};

enum fluid_xxxx_param_int
{
....
};

Anyway, the  `enum param idea` requires the user to look for the `function name first` then to the `typed enum`, which is still a bit uncomfortable compared to actual `specific function for each param`. I think that it worth to:
1)come back using `actual specific function for each param` which is  easier to use (less parameters) and clean API.
2)Internally we can keep the "enum param- all-in-one" setter function (not public ofc), because,  (a) it allows to check param values range,  (b) it requires little overload when called from public API, (c) It makes quite easy the addition of future `specific function for new param`.

derselbst:
> I think that it worth to come back using actual specific function for each param

Ok.

derselbst (approved):
Should be good now. Waiting for the CI to complete. Thanks!

jjceresa:
>Waiting for the CI to complete. Thanks!

Thanks for reviewing.

PR #677, opened 2020-09-21
Make winmidi driver multi devices capable.

jjceresa:
This PR allows `winmidi` driver to open multiple `midi devices` and forward MIDI messages on distinct consecutive MIDI channels to the output.
For example, this allows to use 2 MIDI controllers connected to a single `synth instance` with `synth.midi-channels` setting set to 32. 
The `synth` instance will receive messages from the `first MIDI controller` on MIDI channel 0 to 15,
and messages from the `second MIDI controller` on MIDI channels 16 to 31.

Please see more information in fluid_winmidi.c header.

### DIFF
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -77,10 +77,10 @@
 #define MIDI_SYSEX_MAX_SIZE     512
 #define MIDI_SYSEX_BUF_COUNT    16
 
-typedef struct fluid_winmidi_driver_s fluid_winmidi_driver_t;
+typedef struct fluid_winmidi_driver fluid_winmidi_driver_t;
 
 /* device infos structure for only one midi device */
-typedef struct device_infos_t
+typedef struct device_infos
 {
     fluid_winmidi_driver_t *dev; /* driver structure*/
     unsigned char midi_num;      /* device order number */
@@ -91,10 +91,10 @@ typedef struct device_infos_t
     MIDIHDR sysExHdrs[MIDI_SYSEX_BUF_COUNT];
     /* Sysex data buffer */
     unsigned char sysExBuf[MIDI_SYSEX_BUF_COUNT * MIDI_SYSEX_MAX_SIZE];
-}device_infos;
+}device_infos_t;
 
 /* driver structure */
-struct fluid_winmidi_driver_s
+struct fluid_winmidi_driver
 {
     fluid_midi_driver_t driver;
 
@@ -104,7 +104,7 @@ struct fluid_winmidi_driver_s
 
     /* devices informations table */
     int dev_count;   /* device informations count in dev_infos[] table */
-    device_infos dev_infos[1];
+    device_infos_t dev_infos[1];
 };
 
 #define msg_type(_m)  ((unsigned char)(_m & 0xf0))
@@ -136,7 +136,7 @@ static void CALLBACK
 fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
                        DWORD_PTR dwParam1, DWORD_PTR dwParam2)
 {
-    device_infos *dev_infos = (device_infos *) dwInstance;
+    device_infos_t *dev_infos = (device_infos_t *) dwInstance;
     fluid_winmidi_driver_t *dev = dev_infos->dev;
     fluid_midi_event_t event;
     LPMIDIHDR pMidiHdr;
@@ -386,7 +386,7 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
     }
 
     /* allocation of driver structure size dependant of max_devices */
-    i = sizeof(fluid_winmidi_driver_t) + (max_devices - 1) * sizeof(device_infos);
+    i = sizeof(fluid_winmidi_driver_t) + (max_devices - 1) * sizeof(device_infos_t);
     dev = FLUID_MALLOC(i);
 
     if(dev == NULL)
@@ -405,7 +405,7 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
     }
 
     /* look if the device name contains a semicolon (;) */
-    if(FLUID_STRCHR(dev_name, '\;'))
+    if(FLUID_STRCHR(dev_name, ';'))
     {
         /* multi devices name "x;[y;..]". parse devices index: x;y;..
           Each ascii index are separated by semicolon caracter.
@@ -496,7 +496,7 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
     /* try opening the devices */
     for(i = 0; i < dev->dev_count; i++)
     {
-        device_infos *dev_infos = &dev->dev_infos[i];
+        device_infos_t *dev_infos = &dev->dev_infos[i];
         dev_infos->dev = dev;            /* driver structure */
         dev_infos->midi_num = i;         /* device order number */
         dev_infos->channel_map = i * 16; /* map from input to output */
@@ -600,7 +600,7 @@ delete_fluid_winmidi_driver(fluid_midi_driver_t *p)
     /* stop MIDI in devices and free allocated buffers */
     for(i = 0; i < dev->dev_count; i++)
     {
-        device_infos *dev_infos = &dev->dev_infos[i];
+        device_infos_t *dev_infos = &dev->dev_infos[i];
         if(dev_infos->hmidiin != NULL)
         {
             /* stop the device and mark any pending data blocks as being done */
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -63,6 +63,9 @@
 
 #if WINMIDI_SUPPORT
 
+/* uncomment this macro to enable printf displayed */
+#define PRINTF_MSG
+
 #include "fluid_midi.h"
 #include "fluid_mdriver.h"
 #include "fluid_settings.h"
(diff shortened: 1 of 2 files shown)


derselbst:
That's a nice feature. But I won't have time to take a close look at it in the next days.

However, rather than fixing the broken behaviour of `atoi()` with `fluid_is_uint()`, please have a look at whether [g-ascii-strtoll](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-ascii-strtoll) is  a suitable replacement.

jjceresa:
>That's a nice feature. But I won't have time to take a close look at it in the next days.

No problem. For information this has been tested using fluidsynth console application and 2 MIDI controllers connected to 2 MIDI in devices.This works as expected. (We should just paid attention to set `synth.midi-channels` to 32 because the `router` limits the channel value to synth.midi-channels - 1).

src/drivers/fluid_winmidi.c
```diff
@@ -27,35 +27,85 @@
...
+    fluid_winmidi_driver_t *dev; /* driver structure*/
+    unsigned char midi_num;      /* device order number */
+    unsigned char channel_map;   /* MIDI channel mapping from input to output */
+    UINT dev_idx;                /* device index */
+    HMIDIIN hmidiin;             /* device handle */
     /* MIDI HDR for SYSEX buffer */
     MIDIHDR sysExHdrs[MIDI_SYSEX_BUF_COUNT];
+    /* Sysex data buffer */
+    unsigned char sysExBuf[MIDI_SYSEX_BUF_COUNT * MIDI_SYSEX_MAX_SIZE];
+}device_infos;
```
derselbst, line 94: The name of this typedef should be `device_infos_t`. Likewise, the name of the struct should not end with `_t`.

src/drivers/fluid_winmidi.c
```diff
@@ -97,8 +153,12 @@ fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
 
     case MIM_DATA:
         event.type = msg_type(msg_param);
-        event.channel = msg_chan(msg_param);
+        event.channel = msg_chan(msg_param) + dev_infos->channel_map;
 
+#ifdef PRINTF_MSG
+        printf("\ndev_idx:%d, channel in:%d,out: %d\n",
```
derselbst, line 159: I think it would be worth to convert those `printf`s to `FLUID_LOG(FLUID_DBG)` messages + a little more user-friendly message of what's going on. Otherwise, I don't see a benefit of having those `printf`s in the code.

src/drivers/fluid_winmidi.c
```diff
@@ -145,27 +209,107 @@ fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
...
+ midi.winmidi.maxdevices setting is the maximum number of devices opened
+ by the driver.
+*/
 void fluid_winmidi_midi_driver_settings(fluid_settings_t *settings)
 {
     MMRESULT res;
     MIDIINCAPS in_caps;
     UINT i, num;
+    /* maximum MIDI devices that the driver must handle */
+    fluid_settings_register_int(settings, "midi.winmidi.maxdevices", 1, 1, 16, 0);
```
derselbst, line 282: This new setting needs to be documented in /doc/fluidsettings.xml

src/drivers/fluid_winmidi.c
```diff
@@ -145,27 +209,107 @@ fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
...
+ MIDI devices names are enumerated and added to midi.winmidi.device setting
+ options. Example:
+ 0:Port MIDI SB Live! [CE00], 1:SB PCI External MIDI, default, multi:x[ y z ..]
+
+ Devices name prefixed by index (i.e 1:SB PCI External MIDI) are real devices.
+ "default" name is the default device.
+ "multi:x[ y z ..]" is the multi device naming. Its purpose is to indicate to
+ the user how he must specify a multi device name in the setting.
+ A multi devices name must begin with the prefix 'multi:' followed by the list
+ of real devices index separated by space(s). Example: "multi:5 3 0"
```
derselbst, line 271: So the purpose of this "multi:" is to let the user specify to which devices fluidsynth should connect to, right? In this case, why do we need that "multi:" prefix? How about a simple semicolon-separated list `1;2;3;4`.

src/drivers/fluid_winmidi.c
```diff
@@ -224,115 +370,178 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
...
+        int dev_idx;               /* device index */
+        char *cur_idx, *next_idx;  /* current and next ascii index pointer */
+        next_idx = &dev_name[MULTI_DEV_PREFIX_LEN];
+        do
         {
-            res = midiInGetDevCaps(i, &in_caps, sizeof(MIDIINCAPS));
+            /* try to convert next ascii index */
+            cur_idx = next_idx;
+            dev_idx = g_ascii_strtoll(cur_idx, &next_idx, 10);
+            if (cur_idx == next_idx                /* not an integer number */
```
derselbst, line 422: If there are two spaces in the string the parsing would fail, wouldn't it? To make it more robust, how about using `fluid_strtok()`?

src/drivers/fluid_winmidi.c
```diff
@@ -224,115 +370,178 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
...
+                || dev_idx < 0                     /* invalid device index */
+                || dev_idx >= num                  /* invalid device index */
+                || (dev->dev_count >= max_devices) /* exceed max allowed */
+               )
+            {
+                dev->dev_count = 0; /* error, end of parsing */
```
derselbst, line 428: I think it would be worth to `FLUID_LOG(FLUID_ERR)`  if there was an error during parsing.

derselbst (commented):
Very nice! Only some minor issues below.

src/drivers/fluid_winmidi.c
jjceresa: Ok.

src/drivers/fluid_winmidi.c
jjceresa: This could be done. I will see that.

src/drivers/fluid_winmidi.c
jjceresa: >How about a simple semicolon-separated list 1;2;3;4

This could be done, in the hope that any single `name device` will never contain a semicolon !?. If later it appears that a single device name contains a semicolon we could revert to using a prefix naming which offers a more robust way to distinguish between `single device naming` and 'multi device naming'.

derselbst: > name device will never contain a semicolon !?

I think it's highly unlikely that a single device name contains a `;`. And it's even more unlikely that this device name consists of only numbers and `;`. That said, you could implement some kind of fallback: If there is a token in the sequence that cannot be parsed (e.g. `1;a;b`) then you could treat it as pure device name instead.

src/drivers/fluid_winmidi.c
jjceresa: In fact this new setting could be removed and replaced by a function that parses the `device name` and returns the number
of devices. I will do that in next commits.

src/drivers/fluid_winmidi.c
jjceresa: Ok.

src/drivers/fluid_winmidi.c
jjceresa: Ok.

src/drivers/fluid_winmidi.c
jjceresa: Ok

src/drivers/fluid_winmidi.c
```diff
@@ -145,27 +199,101 @@ fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
...
+                    break;
+                }
+                fluid_settings_add_option(settings, "midi.winmidi.device",
+                                          new_dev_name);
+                FLUID_FREE(new_dev_name);
             }
         }
+
+        /* add multi device name template among other real devices names */
+        fluid_settings_add_option(settings, "midi.winmidi.device", multi_dev_name);
```
derselbst, line 296: I think it's confusing to add this device name template as option to the setting. Instead, this template should be documented in fluidsettings.xml.

src/drivers/fluid_winmidi.c
```diff
@@ -231,108 +478,81 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
...
-    /* Prepare and add SYSEX buffers */
-    for(i = 0; i < MIDI_SYSEX_BUF_COUNT; i++)
-    {
-        hdr = &dev->sysExHdrs[i];
+        device_infos_t *dev_infos = &dev->dev_infos[i];
+        dev_infos->dev = dev;            /* driver structure */
+        dev_infos->midi_num = i;         /* device order number */
+        dev_infos->channel_map = i * 16; /* map from input to output */
+        FLUID_LOG(FLUID_DBG, "opening device at index %d", dev_infos->dev_idx);
+        printf("opening device at index %d\n", dev_infos->dev_idx);
```
derselbst, line 515: Duplicate printf

derselbst:
Perfect, thank you!

jjceresa:
Thanks for reviewing.

PR #687, opened 2020-10-13
Ensure waveout driver compatible with Win9X/NT versions

jjceresa:
This makes `waveout `driver working as before on Win9X/NT.
Note: As far the new drivers build successfully on Win9X/NT, this PR should fixes issue 679.


derselbst:
Cool, thanks!

Issue #693, opened 2020-10-22
--version don't work on windows (2.1.5)


yw662:
### FluidSynth version
_Execute `fluidsynth --version` and provide the output._
```
.\fluidsynth.exe --version
Unknown switch '-'
Usage: fluidsynth [options] [soundfonts]
Try -h for help.
```
(while actually)
```
.\fluidsynth.exe -V
FluidSynth runtime version 2.1.5
Copyright (C) 2000-2020 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of E-mu Systems, Inc.

FluidSynth executable version 2.1.5
Sample type=float
```
### Describe the bug
_Provide a clear and concise description of the current situation, e.g. how the bug manifests._

### Expected behavior
_Provide a clear and concise description of what you expected to happen._
(As shown above)
.\fluidsynth.exe --version should show the same output as .\fluidsynth.exe -V

jjceresa:
The bug you noticed is because you probably use a binary compiled without getopt support.
In this case all long options aren't parsed by fluidsynth application.

jjceresa:
@yw662  Please ignore PR #694 that isn't appropriate. The best would be you compile fluidsynth by adding getopt support.

derselbst:
If fluidsynth has been compiled without getopt support, I'd prefer to simply print a hint message, stating that "This version of fluidsynth was compiled without getopt support. Thus the long options are not supported." And then I'll need to go through the issue templates and replace `--version` by `-V`.

jjceresa:
>If fluidsynth has been compiled without getopt support, I'd prefer to simply print a hint message

Did you mean print this hint message in the usage string like that ?:

>fluidsynth -h
FluidSynth runtime version 2.1.5
Copyright (C) 2000-2020 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of E-mu Systems, Inc.
Usage:
  This version of fluidsynth was compiled without getopt support. Thus the long options are not supported.
  fluidsynth [options] [soundfonts] [midifiles]
 Possible options:
 -a, --audio-driver=[label]
    The name of the audio driver to use.
    Valid values: dsound, file, waveout
 .....
 .....

jjceresa:
Should this PR https://github.com/FluidSynth/fluidsynth/pull/697 fixes this issue ?

yw662:
[...]

PR #694, opened 2020-10-23
fix help message when getopt not supported

jjceresa:
This PR should fixes issue https://github.com/FluidSynth/fluidsynth/issues/693.


derselbst (changes requested):
Sorry, but duplicating the entire help documentation is not an appropriate way to fix this issue.

jjceresa:
Of course you are right. So please ignore this PR.

PR #697, opened 2020-10-26
Add hint message when compiled without getopt support

jjceresa:


derselbst:
Ok, I've made a cosmetic change. Thanks!

PR #699, opened 2020-10-26
Fix two small gcc warnings

mawe42:
Removes two gcc warnings about an unused variable and an uninitialized use of another variable.

### DIFF
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -126,7 +126,7 @@ fluid_rvoice_mixer_process_fx(fluid_rvoice_mixer_t *mixer, int current_blockcoun
     const int fx_channels_per_unit = mixer->buffers.fx_buf_count / mixer->fx_units;
     int i, f;
     int dry_count = mixer->buffers.buf_count; /* dry buffers count */
-    int dry_idx; /* dry buffer index */
+    int dry_idx = 0; /* dry buffer index */
 
     void (*reverb_process_func)(fluid_revmodel_t *rev, const fluid_real_t *in, fluid_real_t *left_out, fluid_real_t *right_out);
     void (*chorus_process_func)(fluid_chorus_t *chorus, const fluid_real_t *in, fluid_real_t *left_out, fluid_real_t *right_out);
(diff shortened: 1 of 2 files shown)


jjceresa:
For `7fc5656`, I am the guilty :).

For `c8fea77` I see the variable correctly initialized !.

mawe42:
> For `7fc5656`, I am the guilty :).

I think we've all done it from time to time :)

> For `c8fea77` I see the variable correctly initialized !.

`dry_idx` only got initialized if `mixer->mix_fx_to_out` was set.

jjceresa:
>dry_idx only got initialized if mixer->mix_fx_to_out was set.

Yes, and dry_idx got unused if mixer->mix_fx_to_out isn't set. So I wonder how gcc could see a possible issue ?

jjceresa:
>Yes, and dry_idx got unused if mixer->mix_fx_to_out isn't set. So I wonder how gcc could see a possible issue ?

The possible issue exists, in the case of `race condition` between line 173
            if(mixer->mix_fx_to_out)
            {
                /* dry buffer mapping, should be done more flexible in the future */
                dry_idx = (f % dry_count) * FLUID_MIXER_MAX_BUFFERS_DEFAULT * FLUID_BUFSIZE;
            }

and line 183
                                    mixer->mix_fx_to_out ? &out_rev_l[dry_idx + i] : &out_rev_l[samp_idx],


, does gcc be able to do this `race condition` assumption ?

mawe42:
Ahh, I see what you mean now. I completely overlooked the `mixer->mix_fx_to_out ?` in lines 183 and 184! And it seems GCC overlooks them as well ;-)

mawe42:
I guess it makes sense. The initialization and usage of `dry_idx` is based on an external variable, which could change at any time during execution. So gcc can't guarantee that it's been initialized before use and outputs the `warning: ‘dry_idx’ may be used uninitialized in this function`.

jjceresa:
>And it seems GCC overlooks them as well ;-)

I have done a fix to any possible race condition in commit 42819d3 . I would like to verify if ggc still complains if we remove
dry_idx initialization. Please could you remind me which log CI did you read about this gcc warnings (may be CI travis ?) ?

mawe42:
I just checked it with your fix and removal of the `dry_idx` initialization and gcc still complains. I noticed the warning when compiling locally. This is the gcc on my machine:
```
$ gcc --version
gcc (Ubuntu 8.4.0-1ubuntu1~18.04) 8.4.0
```

The warnings go away if I pull the usage of dry_idx (i.e. the whole for-loop) into the same `if` branch as the initialization. So it seems gcc doesn't like the fact that it's initialized in one `if` block and used in another.

jjceresa:
>I just checked it with your fix and removal of the dry_idx initialization and gcc still complains.

Many thanks for your quick try!. This helps a lot because it is far more longer for me to check this through CI Travis log.

>So it seems gcc doesn't like the fact that it's initialized in one if block and used in another.

This is what I have noticed, earlier in other situation. That means that ggc warnings are still useful, but in some situations we could safely ignore these.

derselbst (approved):
LGTM, I'll take it.

PR #702, opened 2020-11-02
New style for FluidSynth Settings XML and other small cleanup

mawe42:
This is an attempt to create a clearer and more accessible style for the documentation of the FluidSynth settings. It also contains a few fixes for the settings documentation and other smaller cleanup and rewording commits.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -120,7 +120,7 @@ Developers: Settings can be deprecated by adding: <deprecated>SOME TEXT</depreca
             <def>2</def>
             <min>2</min>
             <max>2</max>
-            <desc>Specifies the number of effects per group. Currently there only are two effects (i.e. reverb and chorus).</desc>
+            <desc>Specifies the number of effects per effects group. Currently this value can not be changed so there are always two effects per group available (reverb and chorus).</desc>
         </setting>
         <setting>
             <name>effects-groups</name>
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -165,12 +165,12 @@ Developers: Settings can be deprecated by adding: <deprecated>SOME TEXT</depreca
             <name>midi-bank-select</name>
             <type>str</type>
             <def>gs</def>
-            <vals>gm, gs, xg, mma</vals>
+            <vals>gs, gm, xg, mma</vals>
             <desc>
                 This setting defines how the synthesizer interprets Bank Select messages.
                 <ul>
-                    <li>gm: ignores CC0 and CC32 messages.</li>
                     <li>gs: (default) CC0 becomes the bank number, CC32 is ignored.</li>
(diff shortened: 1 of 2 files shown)


mawe42:
For reference (and in case it doesn't look like this on your machines/browsers), it is intended to look like this:
settings_screenshot

derselbst:
Wow, you've got talent! This looks modern and professional. Looking back at my approach now, I realize some sort of 90s kitsch.

On my machine, it looks good. @jjceresa You have an older browser. Can you pls. check how it looks like for you?

mawe42:
> Wow, you've got talent! This looks modern and professional. Looking back at my approach now, I realize some sort of 90s kitsch.

Thanks! :-) Not sure about modern, I seem to be stuck in the early 2000 with regard to my design skills. And I also liked the... GeoCities feeling... of the old styling ;) But I think this way we get a bit more overview and make the documentation easier to read.

jjceresa:
@jjceresa You have an older browser. Can you pls. check how it looks like for you?

Here the browser is Firefox 52.9.0 (32-bit) which is the last update available for windows XP (they have abandoned XP for future update). Unfortunately the changes done doesn't display here.
Please, after pulling the new branch settings-stylesheet, I noticed that only `fluidsettings.xml` file have a new modification date (2/11/2020) but `fluidsettings.xls` date is still older (14/07/2018). I guess that Marcus's change relies mainly on `.xls` update as well?. is it right ?

jjceresa:
>For reference (and in case it doesn't look like this on your machines/browsers), it is intended to look like this:

I guess that the `left panel` works like a table of content (TOC) and the `right panel` is displaying the current item page selected in the TOC isn't ?.

jjceresa:
>But I think this way we get a bit more overview and make the documentation easier to read.

Talking about electronic `hyper text` technical documentation, the best thing I saw and prefer to use is the `book like` style. This look like as table of content displayed to the left and the page viewer to the right. A great future is when the user is reading a page and selects an `hyper text` (so called "link") in the right page,  this produces an automatic selection of the title in the TOC. This great simple and efficient future allows the user to quickly learn the documentation organization regardless of the frequent change update done in the book.

jjceresa:
>For reference (and in case it doesn't look like this on your machines/browsers), it is intended to look like this:

>Unfortunately the changes done doesn't display here.
Please, after pulling the new branch settings-stylesheet, I noticed that only fluidsettings.xml file have a new modification date (2/11/2020) but fluidsettings.xls date is still older (14/07/2018).

I just pulled settings-stylesheet branch a second time and now fluidsettings.xml,xls are correctly updated. 
Now Marcus's change are displayed using firefox Firefox 52.9.0 (and also on older I.E V8). Good look and easier to use looking for setting information. Thanks to Marcus's good job!

mawe42:
Great, thanks for testing it, JJC!

PR #703, opened 2020-11-12
API documentation cleanup and improvements

mawe42:
Motivation: I always thought it is a shame that the huge amount of documentation that we have in our code-base is not as easily accessible as it could be. So I decided to rework the doxygen generated API docs. My main goals were:
 - give readers a good overview over the different parts of the FluidSynth API
 - make the documentation of functions and other API elements more consistent, especially with regard to the order in which items appear on the page
 - update the styling to remove some of the doxygen '90 look and feel
 - include the fluidsettings into the API documentation again, to enable easy and direct linking to the settings

The work isn't quite finished yet, I've got some things I would like to improve. But I wanted to bring this into the open now and ask for initial feedback, as the changes also contain quite a lot of code churn. Mainly in the comments, of course. But also some moving of functions to a different place in the same file.

Things still on my list:
  - [x] improve usability on mobile / small screens
  - [x] further improve the styling
  - [x] document which fluidsettings are real-time capable
  - [x] split the large devdocs file into separate and more manageable files
  - ~~[ ] add an overview graphic showing all pars of the FluidSynth API and how they connect, with links to individual API reference pages~~ (will do this later, in a separate PR)
  - [x] decide if using manual ordering of functions / members would be better than the alphabetic sorting
 - [x] update static Doxyfile for generation of API for fluidsynth.org
 - [x] write style guide for consistent doxygen comments and new @startlifecycle/@endlifecycle commands

Looking forward to your feedback! In case you hate the new styling, you can simply revert 921c0e08c00ead8ac416ac0bee73601765687cdf, 5a70df1cb79dbf7fd1138714c7fd00118defeba6 and 1d9d7e9042d9b5ed855399a5d9509956fb88fef8, then you will see the page in the original doxygen styling again.


mawe42:
And for reference, this is how the new styling is supposed to look (at least in my browser.... :-)
new_doc_styling

jjceresa:
A great initiative Marcus !

midi-pascal:
Hi,

Really nice update!

The previous documentation layout was a labyrinth in which I have often 
been lost ;-)

Thumbs up.

derselbst (commented):
First of all: Really nice work!

I've found a few minor issues:

* What's the idea behind the "click to disable panel synchorization" button? Is it useful to disable it at all?
* We need to update the pregenerated `Doxyfile` as well. It's used by the cron job on `fluidsynth.org` directly.

I have a slight preference to stick to the old manual ordering of functions. In alphabetical order, the `delete_*` function are first while the `new_*` functions are always last.

mawe42:
> * What's the idea behind the "click to disable panel synchorization" button? Is it useful to disable it at all?

Actually, I have no idea :-) That tree sidebar widget is generated by doxygen itself and sadly cannot be changed easily...

> * We need to update the pregenerated `Doxyfile` as well. It's used by the cron job on `fluidsynth.org` directly.

Ah! I wondered what that is used for. Ok, I'll add it to the list.

> I have a slight preference to stick to the old manual ordering of functions. In alphabetical order, the `delete_*` function are first while the `new_*` functions are always last.

Same here. While I really like the alphabetical ordering for all other functions, it is weird to have the `new_*` functions at the bottom and so far away from the `delete_*` functions. I thought it might work to use named subgroups on each page, like I already do in the Synth / MIDI Channel Setup page. But I just noticed that doxygen doesn't sort functions in named groups anyway. :-/

mawe42:
While battling doxygens over who knows best what documentation should look like, I started to think that maybe it would be easier to use doxygen only to analyse the source-code and have it output an xml file with all the infos. Then either use an existing tool like 
[breathe](https://breathe.readthedocs.io/en/latest/) to feed it into sphinx, or even create a custom tool to generate exactly the HTML structure and javascript we want and need.

What do you think: would it be acceptable to add another tool into the mix, provided we keep doxygen as the source analysis tool?

derselbst:
> What do you think: would it be acceptable to add another tool into the mix, provided we keep doxygen as the source analysis tool?

Keep in mind that all those nice tools would have to be installed and set up correctly on `fluidsynth.org`. Only Josh himself can do it. So for now, I'd prefer to stick to doxygen only. I think the change of this PR is already a huge improvement to the previous situation.

mawe42:
> Keep in mind that all those nice tools would have to be installed and set up correctly on fluidsynth.org. Only Josh himself can do it.

Ah, I forgot how the API docs on fluidsynth.org get generated. Hm... let's hope he has xsltproc installed.

> So for now, I'd prefer to stick to doxygen only. I think the change of this PR is already a huge improvement to the previous situation.

That's fine by me. Probably best to do incremental changes to these things.

mawe42:
Regarding the sort ordering and the problem of `new_*` and `delete_*` positions, how about this:
Screenshot from 2020-11-13 21-41-40


To achieve that, I've added two new command aliases:
```
ALIASES += startlifecycle="\name Lifecycle Functions\n@{"
ALIASES += endlifecycle="@}"
```

That means we can mark all `new_*` and `delete_*` functions with the following:
```
/** @startlifecycle */
int new_fluid_...
int delete_fluid_...
int delete_fluid_...
/** endlifecycle */

**Comment by mawe42** [comment]
The different membergroup names need to be unique, otherwise doxygen gets confused and adds all lifecycle functions from all sibling groups to the first group. So it's now:
```
/** @startlifecycle{Some Name} */
```
which results in the header "Lifecycle Functions for Some Name".

Those changes are now committed. I really like this and that solves the sorting problem, in my opinion. Inside the lifecycle block, the sorting order is manual... though I don'tknow if that is due to a bug in doxygen or by design. But if it's a bug and gets fixed at some point, it wouldn't matter much, I think.

**Comment by mawe42** [comment]
> We need to update the pregenerated Doxyfile as well. It's used by the cron job on fluidsynth.org directly.

I was in the middle of updating the static doc/Doxyfile when I noticed differences in the INPUT configuration. And then I finally understood: the API doc for fluidsynth.org is created by directly executing doxygen in the doc/ source directory!

That will be problematic with these changes, as I'm using cmake to find xsltproc and render the fluidsettings.xml into a doxygen file during the build.

I see different options here:
1. Ask Josh to manually add the xsltproc command into the cron job
2. As Josh to change the API build to cmake instead
3. Create a build job for the API docs on travis, github actions or other and ask Josh to change the cronjob from a local build to downoading a tarball and extracting it into a local dir
4. Commit the generated doc/fluidsettings.txt into the repo
5. Remove the integration of fluidsettings.xml into the API docs

My preference would actually be option 3. That would give us much more control of the API doc build in the future and
would remove the dependencies on stuff installed on his server. But I don't know how approachable he is. And it's probably also the option causing the most work.

Option 4. would be the simplest. But it would mean committing yet another special (and generated) file for the build on fluidsettings.org. Yet, this would be my second choice.

Option 5 would be a shame. The tight integration of the settings reference is great because the settings are so important for fluidsynth, in my opinion.

@derselbst what do you think?

**Comment by derselbst** [comment]
> Hm... let's hope he has xsltproc installed.

The server doesn't run cmake. We might have to version control the `fluidsetting.txt`.

> the API doc for fluidsynth.org is created by directly executing doxygen in the doc/ source directory!

:-)

The lifecycle stuff looks good!

Generating the API docs with cmake is technically the only right way to do it. Because without cmake the `version.h` file is missing. So I tend to go for option 3 as well. But Josh needs to download the tarball from a fixed location. Currently, I don't know how to achieve this. Build artifacts are specific to one build job. And the url to this job keeps changing. Need to think about that.

**Comment by mawe42** [comment]
> Generating the API docs with cmake is technically the only right way to do it. Because without cmake the version.h file is missing. So I tend to go for option 3 as well. But Josh needs to download the tarball from a fixed location. 

Maybe we can create a GitHub action that reacts to a release, builds the API doc and uploads the generated tarball as an additional asset to the release using the following API:
https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#upload-a-release-asset

Then Josh can simply link to FluidSynth/fluidsynth/releases/latest/assets/apidoc.zip to update the website.

But I would suggest to leave this for later. If you are ok with committing the fluidsettings.txt file, then we can concentrate on getting the doc to look good and worry about the proper build afterwards.

**Comment by mawe42** [comment]
Hm... another idea: I assume you currently create releases manually. So in the release process, you could simply build the API docs, zip them up and attach them as a new asset to the latest release. So basically what I propose to automate using the GitHub Actions, but manually.

And then work out how to do it automatically later on.

**Comment by derselbst** [comment]
I don't want to bind the publishing of API docs to GitHub releases. API docs should be continuously deployed as it is currently the case. If we improve the docs today, people should benefit from it right away rather than waiting for the next release to come.

> If you are ok with committing the fluidsettings.txt file

I'm ok, knowingly that nothing lasts longer than a temporary ;)

**Comment by mawe42** [comment]
Ok, I think I'm finished for now :-)

I would welcome another review, especially for the recent changes.

File: doc/examples/fluidsynth_simple.c [hunk 1]
```diff
@@ -17,6 +17,7 @@
 
 int main(int argc, char **argv)
 {
+    asdf
```
derselbst, line 20: jklö

mawe42: Argh! :-)

derselbst:
Found this one:

Screenshot_20201114_173405

While we are on it: Perhaps we should address the doxygen warnings as well:
```
warning: Tag `SHOW_DIRECTORIES' at line 71 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `HTML_ALIGN_MEMBERS' at line 135 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `XML_SCHEMA' at line 202 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `XML_DTD' at line 203 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: doxygen no longer ships with the FreeSans font.
You may want to clear or change DOT_FONTNAME.
Otherwise you run the risk that the wrong font is being used for dot generated graphs.

```

derselbst:
* Why is the `fluidsettings.xsl` in the `doc/` and `doc/doxygen` folder?
* The `fluid_mixer` scheme graphic is now a png and you cannot click on it anymore. Is that intended? I used an ugly looking small resolution jpeg to make people click on it. (omg, it's so 90's style)
* Following up our discussion about deprecation, we should add something like
```
/** @page deprecated Deprecated Functions
 * This page contains functions that have been marked obsolete.
 * Functions listed here will be removed in the next major release.
 * It is therefore not wise to use them in new code.
 */
```

mawe42:
> Found this one:
> Screenshot_20201114_173405

Hm... that doesn't happen here. Which doxygen version are you using? I'm on 1.8.13.

> While we are on it: Perhaps we should address the doxygen warnings as well:

Good idea, fixed in recent commit.

mawe42:
> * Why is the `fluidsettings.xsl` in the `doc/` and `doc/doxygen` folder?

doc/fluidsettings.xsl is the style to render fluidsettings.xml if you view it directly in the browser. doc/doxygen/fluidsettings.xsl is the style to render a doxygen page to be included in the API docs.

I guess we could remove doc/fluidsettings.xsl and change all links that point directly to fluidsynth.org/api/fluidsettings.xml to point to the API docs instead.

> * The `fluid_mixer` scheme graphic is now a png and you cannot click on it anymore. Is that intended? I used an ugly looking small resolution jpeg to make people click on it. (omg, it's so 90's style)

I didn't understand that the uglyness was intended :-) Well, I personally don't need a link on that image. But we can change it back to link to the PDF, if you want. I would keep the better looking default image, though.

> * Following up our discussion about deprecation, we should add something like

Good idea, I've added it to the devdocs. This also changes the order, moving the "Deprecated Functions" page before recent changes. But I think thats ok...

derselbst:
> Which doxygen version are you using?

1.8.14




> > * The `fluid_mixer` scheme graphic is now a png and you cannot click on it anymore. Is that intended? I used an ugly looking small resolution jpeg to make people click on it. (omg, it's so 90's style)
> 
> I didn't understand that the uglyness was intended :-) Well, I personally don't need a link on that image. But we can change it back to link to the PDF, if you want. I would keep the better looking default image, though.

I prefer to have vector graphics.




> > * Why is the `fluidsettings.xsl` in the `doc/` and `doc/doxygen` folder?
> 
> doc/fluidsettings.xsl is the style to render fluidsettings.xml if you view it directly in the browser. doc/doxygen/fluidsettings.xsl is the style to render a doxygen page to be included in the API docs.

Do we need to have two different styles? Or would it be possible to combine them (somehow)?

mawe42:
> I prefer to have vector graphics.

Ok, how about simply using SVG as in the most recent commit?

mawe42:
Can you check the "Lifecycle Functions for Audio Driver" header problem again? I've committed a fix which I think solves this issue. It's a little dirty as it is using a doxygen internal command. But there doesn't seem to be a good way to specify linebreaks that work in multiple doxygen versions otherwise...

derselbst:
> Can you check the "Lifecycle Functions for Audio Driver" header problem again?

Looks good now.

> Ok, how about simply using SVG as in the most recent commit?

Ok.

mawe42:
> Do we need to have two different styles? Or would it be possible to combine them (somehow)?

I don't think it's possible to render two different versions from a single stylesheet. So if we want to keep the standalone settings reference, then two styles are necessary.

I guess we could try to combine the two versions into a single output, rendering html that can also be understood by doxygen. But I would have to research if it is even possible to create \ref targets in the doxygen html subset. A quich search only brough to the ability to define anchors, and I assume those can't be referenced directly in the rest of the API doc.

What is bothering you about having two stylesheets for a single XML file?

mawe42:
> Do we still need ALIASES += startlifecycle="\name Lifecycle Functions\n@{" in Doxyfile?

No, we don't. I will remove it.

mawe42:
BTW: I did some more research on how we could generate the API docs in a GitHub Action and store the results somewhere accessible.
An easy way would be to create a small script that returns the URL to the zipped artifacts of the latest successful build. The GutHub API is really comfortable and easy to use, so it would be just a few lines of shell script, mainly curl and a bit of json parsing with jq or even grep/awk. So instead of Josh calling wget or similar to download the file, we could supply him with a shell script that downloads the latest build instead. It could even keep track of the current version and only re-download and unpack if something has actually changed.

derselbst:
> if we want to keep the standalone settings reference, then two styles are necessary.

Ok, let's keep it for now.

> What is bothering you about having two stylesheets for a single XML file?

I favor the original style a bit more, so I was just wondering. Nvm.

> An easy way would be to create a small script that returns the URL to the zipped artifacts of the latest successful build

Ok, sounds like a plan. Feel free to do it, if you can spare the time. You seem to be more into this topic anyway. I'd suggest doing the  GitHub Actions automation in another PR.

So, after you've removed the unused alias and updated the `Doxyfile` with the linebreak issue, I think we're ready to complete this PR. Should we merge or squash it?

mawe42:
> I favor the original style a bit more, so I was just wondering.

I do too, actually. I just couldn't find an easy way to recreate it with doxygen commands and thought it's probably not worth it. But if you feel the same, then I'll try again tonight.

> Ok, sounds like a plan. Feel free to do it, if you can spare the time. You seem to be more into this topic anyway. I'd suggest doing the GitHub Actions automation in another PR.

Will do.

> So, after you've removed the unused alias and updated the Doxyfile with the linebreak issue, I think we're ready to complete this PR. Should we merge or squash it?

I tend towards a merge, that way the git log will give a little more information about these changes in the future.

mawe42:
The API doc build is already enabled (thanks Tom!). So what is left is a tool to download the last successful build artifacts. I decided to write a little Python 3 tool for that, as that reduces the number of requirements (only Python 3 required) and adds some nice little features like artifact caching and robust error handling: https://github.com/mawe42/github-artifact-download

I guess we should also write a small shell script using this tool and updating a target directory with the contents of the downloaded file, to make Joshs life even easier...

mawe42:
The settings reference looks a little more like the fluidsettings.xml style and is more consistent with the rest of the API reference pages now. It now looks like this:
Screenshot from 2020-11-16 22-30-17

I think I will leave it here. So from my perspective, this is ready to go.

mawe42:
Oh, and if you would rather not merge all these individual commits, I could clean up the history and combine the ones that fit together.

derselbst:
> I think I will leave it here. So from my perspective, this is ready to go.

Ok, cool.

> Oh, and if you would rather not merge all these individual commits, I could clean up the history and combine the ones that fit together.

If you still have the breath to rebase it, you're welcome. Otherwise, I would take it as is. Just let me know.

In any case, thanks again for this tremendous work!

mawe42:
> If you still have the breath to rebase it, you're welcome. Otherwise, I would take it as is. Just let me know.

I just had a look through the commits and had a hard time finding ones that could easily be combined. So I guess I'm out of breath after all :-)

> In any case, thanks again for this tremendous work!

Very welcome. It was great fun, actually... something to do during the lockdown ;) And I think our documentation deserves a better presentation. Will hopefully also make life easier in the future, if people have easier access to the docs.

jjceresa:
>Very welcome. It was great fun

Thanks Marcus. 
I have not understood  all the details about the process. It seems that all the process is mainly based on the use of doxygen tool. Have you relied on others tools or/and basics shell command to achieve your goal ?

mawe42:
It's mostly cleanup work in doxygen comments of the source code as well as the additional documentation files from devdocs. I think the hardest part was a review of all API elements and understanding where each element should go in the new API modules structure, so that related bits are acutally together on a page. Finally a custom layout and custom CSS styles for the doxygen-generated pages, of course.

New is only the XSL transform of the fluidsettings.xml to a doxygen-style document via xsltproc. It gets called automatically during the cmake build of the docs. 

Still in discussion is the build of the API docs via GitHub actions (already enabled) and asking Josh to download the generated docs from GitHub instead of building them locally on the fluidsynth.org server. This is still work in progress.

jjceresa:
Thanks for these useful general explanations. This help to better understand which target documentation should be generated and where these target docs should be installed. As far I understand the process resumes in 3 steps.
1)cmake prepare build commands. These commands should probably then executed manually to build and install the docs.
     somewhere in the /doc folder.
2)Then the GitHub API docs should be automatically generated from the result of step (1) using those "GitHub action". Thus the resulting docs are exposed on Github html page.
3)Then having a way to replicate the docs from GitHub to fluidsynth.org.

jjceresa:
Tested http://www.fluidsynth.org/api/  using old firefox and old opera browsers on Windows XP. This displays well.
Thanks to the API organization in functional parts with the related table of contents. Great and excellent job.

mawe42:
That's good to know, thanks for testing it, JJC!

mawe42:
Ah, I just noticed that I didn't update the static fluidsettings.txt with the changes to the rendered settings reference. @derselbst would it be ok if I commit the changed file to master quickly?

derselbst:
Sure

mawe42:
Thanks, done.

PR #706, opened 2020-11-25
Change owner of the SoundFont registered trademark.

Some1NamedNate:
As of the time of this PR, the SoundFont registered trademark is owned by Creative Technology Ltd.
( http://tmsearch.uspto.gov/bin/showfield?f=doc&state=4803:rj74xq.2.1 )


derselbst:
Ok, thanks.

PR #708, opened 2020-11-28
Handle GS SysEx messages for setting whether a channel is used for rhythm part.

chirs241097:
Some MIDI files that uses the GS standard uses channels other than channel 10 as percussion channel. Currently FluidSynth ignores the messages setting that up, causing notes meant to be played with a drum instrument played with a melodic instrument or vice versa. This patch will partially fix the issue.

Currently the implementation in this patch doesn't cover a specific "quirk" in Roland GS Modules: they seem to remember the last used instrument in the selected mode. This patch simply sets the instrument number to 0 instead.

A test file is attached. If played correctly (with `-o synth.device-id=16`) no out of place drum or piano sounds should be heard.

[wikipedia_MIDI_sample_gstest.mid.gz](https://github.com/FluidSynth/fluidsynth/files/5610727/wikipedia_MIDI_sample_gstest.mid.gz)

### DIFF
--- a/src/midi/fluid_midi.h
+++ b/src/midi/fluid_midi.h
@@ -187,6 +187,7 @@ enum midi_sysex_manuf
 /* SYSEX sub-ID #1 which follows device ID */
 #define MIDI_SYSEX_MIDI_TUNING_ID       0x08    /**< Sysex sub-ID #1 for MIDI tuning messages */
 #define MIDI_SYSEX_GM_ID                0x09    /**< Sysex sub-ID #1 for General MIDI messages */
+#define MIDI_SYSEX_GS_ID                0x42    /**< Model ID (GS) serving as sub-ID #1 for GS messages*/
 
 /**
  * SYSEX tuning message IDs.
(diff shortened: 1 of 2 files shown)


derselbst:
Aha, interesting. Thank you very much!

> If played correctly (with -o synth.device-id=16)

This  is an important detail, that, I think, we need to document more explicitly. @mawe42 Where / howto document this? I'd say we need to figure out which SysEx messages fluidsynth currently supports and document that in [FluidFeatures](https://github.com/FluidSynth/fluidsynth/wiki/FluidFeatures). Additional we need a general link to that sysx ID somewhere in the UserManual? What do you say?

chirs241097:
Current default value for `synth.device-id` is 0. This doesn't match the default value for real hardware GS synths (they use 16 by default, see p.76 of SC-55mkII's manual or p.53 of SC-55's manual, their values starts from 1). Also the range doesn't match (Roland uses 0-31 for their GS modules, and 16-31 for later GM2 modules), although this is less of a problem: the only problem I could think of is the synth instance with `device-id=0x7f` cannot be targeted individually because that id is reserved for broadcasting.

mawe42:
> This is an important detail, that, I think, we need to document more explicitly. @mawe42 Where / howto document this? I'd say we need to figure out which SysEx messages fluidsynth currently supports and document that in [FluidFeatures](https://github.com/FluidSynth/fluidsynth/wiki/FluidFeatures). Additional we need a general link to that sysx ID somewhere in the UserManual? What do you say?

Documenting the supported SysEx messages on FluidFeatures sounds like a very good idea. And we should probably add a little more detail to the documentation of the `synth.device-id` setting  in the settings reference.

Issue #707, opened 2020-11-25
Tempo set with fluidsynth_player_set_bpm() is reset by fluidsynth


glm35:
### FluidSynth version

2.1.1

### Describe the bug

I'd like to be able to play midi files at my own tempo. I'm aware of the well documented limitation that "Tempo change events contained in the MIDI file can override the specified BPM at any time". So I created midi files without "set tempo" meta message.

But I observed several similar issues:

1) when I set the tempo with fluid_player_set_bpm() before calling fluid_player_play(), the midi file is always played at 120 bpm, whatever the value I configure.

2) when I play the file in a loop: I can change the tempo if I call fluid_player_set_bpm() after fluid_player_play(). But when the file is looped, the tempo is reset to 120 bpm.

3) it is the same thing if I add several midi files to the playlist with fluid_player_add(): when the next midi file starts playing, the tempo is reset to 120 bpm.

### Expected behavior

I expect that:

- midi files without "set tempo" meta message are played at the tempo set with fluid_player_set_bpm()
- the tempo configured with fluid_player_set_bpm() is preserved when the file is looped.
- when several files without "set tempo" meta messages are added to the playlist, a single call to fluid_player_set_bpm() will configure the tempo for all the files

### Additional context

Looking at fluidsynth code in fluid_midi.c: I understand that fluid_player_reset() is called just before a midi file is played or looped, and fluid_player_reset() resets the tempo to 120 bpm (player->miditempo = 500000).

Commenting out the "player->miditempo = 500000" line in fluid_player_reset() fixed the issue for me.

jjceresa:
>Commenting out the "player->miditempo = 500000" line in fluid_player_reset() fixed the issue for me.

This fixed your issue because the MIDI file you are using doesn't contains no tempo-change meta event. This fix doesn't works with MIDI file with tempo-change meta message because any tempo-change meta message played will override the current tempo this is a normal behaviour (not a bug).

Did you try to change the state of `player.reset-synth` setting to FALSE ?

For each of your 3 expectations to be possible (regardless if the MIDI file contains tempo-change messages or not), we should change the default behaviour of the player so that the `fluid_player_set_bpm()` API should have precedence over any potential tempo-changes events during the playing. Not sure that `player.reset-synth` setting will solve this.

jjceresa:
>Did you try to change the state of player.reset-synth setting to FALSE ?

After looking in the code "player.reset-synth" seems not related to tempo and cannot solve your issue.

[...]

PR #711, opened 2020-12-05
Fix MIDI player tempo reset issues

jjceresa:
This PR addresses issue https://github.com/FluidSynth/fluidsynth/issues/707 according to the last discussion.
@glm35, please, your are welcome to pull this branch `fix-player-tempo` an test the new function `fluid_player_set_tempo()`, and `fluid_player_get_tempo()`.


jjceresa:
Please, as I am not familiar with doxygen formatting rule, could you have a check , in particular for `fluid_player_tempo_type` enum in `midi.h` file ?.

include/fluidsynth/midi.h
```diff
@@ -193,7 +193,7 @@ FLUIDSYNTH_API void delete_fluid_midi_driver(fluid_midi_driver_t *driver);
 /* @} */
 
 /**
- * @defgroup midi_player MIDI Player
+ * @defgroup midi_player MIDI file Player
```
mawe42, line 196: Nitpick: for places where we use title case (in the name of groups, for example), we should ensure to use it consistently. So use "MIDI File Player" here, not "MIDI file Player".

derselbst:
Looks good to me. I'll keep this open for a few days to give @glm35 a chance to test this.

jjceresa:
>Looks good to me....

Ok. Thanks for reviewing and fixing API docs, and grammar.

glm35:
Thanks for this PR!

I did some basic functional tests:
- midi file with tempo and midi file without tempo
- with and without calling set_tempo(FLUID_PLAYER_TEMPO_BPM flavor) before playback start
- file played one time, two times in a playlist, three times in a loop

All good for the moment, and this solves all the problems I reported in issue #707.

I intend to do more tests: set tempo after playback start, get_tempo(), midi file with tempo change, FLUID_PLAYER_TEMPO_MIDI, FLUID_PLAYER_TEMPO_DEFAULT and FLUID_PLAYER_TEMPO_RELATIVE. It will probably take me the rest of the week to find some time to do that.

derselbst:
Great, thanks for having a first try! And no problem, take your time.

glm35:
I did a few more tests today:
- set tempo 3s after playback start, with midi file with and without "set tempo"
- midi file with tempo change

Still all good :-)

... but test plan not completed yet...

derselbst:
> ... but test plan not completed yet...

@glm35 Have you had opportunity to complete you test plan successfully?

glm35:
No, I haven't.

I took a side step to build a small player GUI that will also help me for the tempo API tests (https://github.com/glm35/pyfluidsynth3/blob/gl-devel-new-tempo-api/examples/player_gui.py).

I will have more time to work on this this week, so if you can wait a litter more...

derselbst:
> I will have more time to work on this this week, so if you can wait a litter more...

Sure, no problem.

glm35:
An update about my testing of the new tempo API...

I tested fluid_player_set_tempo() with FLUID_PLAYER_TEMPO_DEFAULT, and it is fine for me. When the midi file contains "set tempo" messages, the playback continues with the last value seen in the file, which is nice. When the midi file does not contain any "set tempo", the playback continues with fluidplayer default tempo (120 bpm), which what I expected.

glm35:
I also tested fluid_player_get_tempo(). For me, it is working as documented. However, I find this new API a bit complex to understand and to use.

I think that the main use case for this API is to get the current tempo used by the player when the tempo values come from the the midi file. If you want to know this, you need to call get_tempo() with FLUID_PLAYER_TEMPO_DEFAULT, and I find this confusing because from me the default tempo value is the value used by the player when no tempo is specified by the user or in the midi file (120 bpm). Maybe it would be clearer to rename FLUID_PLAYER_TEMPO_DEFAULT to FLUID_PLAYER_TEMPO_MIDI_FILE? Also, you will only get the tempo value in "midi tempo" format (us/quarter note). If you want it in bpm, you need to compute it.

Calling get_tempo() with FLUID_PLAYER_TEMPO_BPM or with FLUID_PLAYER_TEMPO_MIDI will give you the value configured by the user with set_tempo() if any, else the default value used by the player (120 bpm). This is consistent with set_tempo(), but not very useful IMO. I would prefer that calling set_tempo() with those tempo types would return the current tempo value in the desired format (bpm or midi). I think this would make the API more intuitive to use, and also it would allow the user to get the current tempo in bpm.

To conclude, beyond troubleshooting, I personally don't have a real use case for get_tempo(). But I think that it might be enough to have a simpler API that would just return the current tempo used by the player in the desired format, ie that would only accept the tempo types FLUID_PLAYER_TEMPO_BPM ans FLUID_PLAYER_TEMPO_MIDI, and that would drop the sync_mode parameter. But that is just my point of view.

src/midi/fluid_midi.c
```diff
@@ -2232,23 +2243,142 @@ int fluid_player_set_loop(fluid_player_t *player, int loop)
...
+ * change or controlled by external tempo expressed in BPM or in micro seconds
+ * per quarter note.
+ *
+ * @param player MIDI player instance. Must be a valid pointer.
+ * @param tempo_type Must a be value of #fluid_player_tempo_type and indicates the
+ *  meaning of tempo value and how the player will be controlled, see below.
+ * @param tempo Tempo value or multiplier.
+ * 
+ *  #FLUID_PLAYER_TEMPO_DEFAULT, the player will be controlled by internal MIDI
+ *  file tempo changes. The @c tempo parameter will be ignored.
```
glm35, line 2291: Could you precise that if no tempo is given in the midi file, a default value of 120 bpm will be used?

src/midi/fluid_midi.c
```diff
@@ -2232,23 +2243,142 @@ int fluid_player_set_loop(fluid_player_t *player, int loop)
...
+ *
+ *  #FLUID_PLAYER_TEMPO_BPM, the player will be controlled by the external tempo
+ *  value provided  by the tempo parameter in bpm (i.e in quarter notes per minute)
+ *  which must be in the range (1 to 60000000).
```
glm35, line 2295: 60000000 bpm, really?

src/midi/fluid_midi.c
```diff
@@ -1608,7 +1609,9 @@ fluid_track_send_events(fluid_track_t *track,
 
         if(event->type == MIDI_SET_TEMPO && player != NULL)
         {
-            fluid_player_set_midi_tempo(player, event->param1);
+            /* memorize the tempo change value comming from the MIDI file */
```
glm35, line 1612: typo: comming -> coming
(can also be found in other places, I only report this one)

src/midi/fluid_midi.c
```diff
@@ -2317,10 +2448,98 @@ int fluid_player_get_total_ticks(fluid_player_t *player)
...
+ * per quarter note.
+ *
+ * @param player MIDI player instance. Must be a valid pointer.
+ * @param tempo_type A value of #fluid_player_tempo_type that indicates the
+ * tempo information to retrieve, see below.
+ *
+ * @param tempo Pointer to the value the returned information will be written to.
+ *  This pointer can be NULL to ignore the tempo information
+ *
+ * @param sync_mode Pointer an int for returning the tempo mode the player
```
glm35, line 2464: typo: "Pointer an int" -> "Pointer to an int"

jjceresa:
>Maybe it would be clearer to rename FLUID_PLAYER_TEMPO_DEFAULT to FLUID_PLAYER_TEMPO_MIDI_FILE? 

Basically you right. This name is confusing. The enum name should be FLUID_PLAYER_TEMPO_MIDIFILE or (FLUID_PLAYER_TEMPO_INTERNAL) to express that xx_set_tempo() is a function that set the player to follow internal tempo change coming from the MIDI file .

The other enum value ( FLUID_PLAYER_TEMPO_BPM ans FLUID_PLAYER_TEMPO_MIDI) set the player to follow external tempo value provided by xx_set_tempo(). This name should be renamed FLUID_PLAYER_TEMPO_EXTERNAL_BPM and 
FLUID_PLAYER_TEMPO_EXTERNAL_MIDIFILE.

jjceresa:
>Calling get_tempo() with FLUID_PLAYER_TEMPO_BPM or with FLUID_PLAYER_TEMPO_MIDI will give you the value configured by the user with set_tempo() if any, else the default value used by the player (120 bpm). This is consistent with set_tempo(), but not very useful IMO.
But I think that it might be enough to have a simpler API that would just return the current tempo used by the player in the desired format.

To keep consistency with 'xx_set_tempo(),  `xx_get_tempo()` should be able to return external tempo value (FLUID_PLAYER_TEMPO_EXTERNAL_BPM, FLUID_PLAYER_TEMPO_EXTERNAL_MIDIFILE) set by `xx_set_tempo`.
Also `xx_get_tempo()` should be able to return if the player is in internal mode or external mode.

Actually  `xx_get_tempo()` does not return  the current tempo used by the player. This could be added using the enum value
FLUID_PLAYER_TEMPO_CURRENT_BPM and FLUID_PLAYER_TEMPO_CURRENT_MIDIFILE. This will return the tempo actually used by the player in bpm or us/quarter note .

So for xx_get_tempo, I propose the following enum type value:
- FLUID_PLAYER_TEMPO_EXTERNAL_BPM, FLUID_PLAYER_TEMPO_EXTERNAL_MIDIFILE, return external tempo in bpm or us/quarter note.
- FLUID_PLAYER_TEMPO_BPM, FLUID_PLAYER_TEMPO_MIDIFILE, return current tempo used by the player in bpm or us/quarter note.

Opinions ?

jjceresa:
>Could you precise that if no tempo is given in the midi file, a default value of 120 bpm will be used?

Yes, when no tempo change is given in a midi file, a default value of 120 bpm should be used. This is a MIDI file specification.
It will be useful to add this mention in the API comment.

***
>60000000 bpm, really?

Yes, internally valid tempo value should be in the range [1..600000000] regardless if the value is in bpm or in us/quarter note.

jjceresa:
>But I think that it might be enough to have a simpler API that would just return the current tempo used by the player in the desired format.

This is done in recent commit 0836a52 . One just need to call fluid_player_get_tempo() with FLUID_PLAYER_GET_TEMPO_USED_BPM or FLUID_PLAYER_GET_TEMPO_USED_MIDI.

derselbst:
First of all, I'm glad to see that everything is working as expected. Thanks for testing @glm35!

Regarding the `get_tempo()` function: I have a similar feeling as @glm35: It is too overengineered. A client app won't need this. The client app knows whether it has set an absolute tempo or a relative one. Splitting up the enum into a `fluid_player_get_tempo_type` and `fluid_player_set_tempo_type` doesn't really make this simpler. Thinking about it I would rather say to drop `fluid_player_get_tempo()` entirely, due to a missing use case at the present. Instead, the existing functions `fluid_player_get_bpm()` and `fluid_player_get_midi_tempo()` should be continued to use. Those functions obtain the tempo the player is using right now. This might be

* the tempo given by the MIDI file multiplied by the relative factor, or
*  the absolute tempo dictated by the client app.

jjceresa:
>Thinking about it I would rather say to drop fluid_player_get_tempo() entirely, due to a missing use case at the present. Instead, the existing functions fluid_player_get_bpm() and fluid_player_get_midi_tempo() should be continued to use. Those functions obtain the tempo the player is using right now.

Ok.

jjceresa:
Now, the `tempo multiplier` factor set by `fluid_player_set_tempo(..,FLUID_PLAYER_TEMPO_RELATIVE, multiplier)` is only used when the player follow the internal tempo changes. So it seems that we should do the following changes:
1)pass this multiplier value in the `tempo` parameter when calling `fluid_player_set_tempo(..,FLUID_PLAYER_TEMPO_INTERNAL, multiplier)`
2)Get rid of `FLUID_PLAYER_TEMPO_RELATIVE` enum value.

These simplifications allows the caller to set the player in `internal tempo` mode and supplying the `multiplier` factor using only one API function call.
Opinions ?

derselbst:
> Now, the tempo multiplier factor set by fluid_player_set_tempo(..,FLUID_PLAYER_TEMPO_RELATIVE, multiplier) is only used when the player follow the internal tempo changes. So it seems that we should do the following changes:

So, rather than ignoring the `tempo` parameter with `FLUID_PLAYER_TEMPO_INTERNAL`, we use it to pass the `multiplier` directly. Yes, that sounds good to me.

derselbst (approved):
Looks good now. I'll wait for the CI and then squash it. Thanks!

PR #713, opened 2020-12-06
Add shell commands to the MIDI File Player

jjceresa:
Commands are:
- player_start                     Start playing from the beginning of current song
- player_stop                     Stop playing
- player_cont                     Continue playing
- player_step num              Move forward/backward in current song to +/-num ticks
- player_next                     Move to next song
- player_loop num              Set loop number to num (-1 = loop forever)
- player_tempo_bmp num   Set tempo to num beats per minute


jjceresa:
Please note that `player_tempo_bpm` command call actual bugged `fluid_player_set_bpm()`. Once `fix-player-tempo` branch (https://github.com/FluidSynth/fluidsynth/pull/711) will be merged into master, actual branch (this PR) will require to be updated and completed before merging it to master.

derselbst:
We have some issues here:

* You cannot add additional arguements to the existing public API functions `new_fluid_cmd_handler` and `new_fluid_server`. Use recently introduced `new_fluid_cmd_handler2` for this purpose and add a `new_fluid_server2` function.
* You've changed the order of processing in `fluidsynth.c`: You're now executing the command file before having loaded a default soundfont. This will be problematic if the user relies on the default soundfont. If the user has `noteon` commands in his command file, those commands will now fail because no preset has been assigned yet.

src/midi/fluid_midi.c
```diff
@@ -2261,6 +2261,11 @@ int fluid_player_set_midi_tempo(fluid_player_t *player, int tempo)
  */
 int fluid_player_set_bpm(fluid_player_t *player, int bpm)
 {
+    if(bpm <= 0)
+    {
+        return FLUID_OK; /* to avoid a division by 0 */
```
derselbst, line 2266: `return FLUID_FAILED` ?

derselbst:
I just gave it a test and it works very well. Thanks!

jjceresa:
>I just gave it a test and it works very well.

Thanks for reviewing.

PR #717, opened 2020-12-15
Support for FLUID_STRTOL()

carlo-bramini:
Since Windows supports up to 32 MIDI devices, I think that FLUID_STRTOL() is more than enough.
In my opinion, it would be also a good idea to not use directly a GLib function, since the source code already wraps all calls to library functions with macros.

### DIFF
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -359,7 +359,7 @@ fluid_winmidi_parse_device_name(fluid_winmidi_driver_t *dev, char *dev_name)
     {
         /* try to convert current ascii index */
         char *end_idx = cur_idx;
-        dev_idx = g_ascii_strtoll(cur_idx, &end_idx, 10);
+        dev_idx = FLUID_STRTOL(cur_idx, &end_idx, 10);
 
         if(cur_idx == end_idx      /* not an integer number */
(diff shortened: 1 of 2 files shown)


derselbst:
`g_ascii_strtoll` uses the C locale, whereas `strtol` uses the current locale. However, since we are only dealing with integers this shouldn't matter. @jjceresa Was there a specific reason to use `g_ascii_strtoll` which I'm missing? Or do you think it's safe to replace `g_ascii_strtoll` with `strtol`?

jjceresa:
The use of to use `g_ascii_strtoll`  comes from the second comment in https://github.com/FluidSynth/fluidsynth/pull/677. As far `strtol` does the same job than `g_ascii_strtoll`  the replacement suggested by Carlo is preferable. I will check this in next days

jjceresa:
>Was there a specific reason to use g_ascii_strtoll which I'm missing?

No, I see no reason to use it. Also, I have checked `strtol' that works similarly. Thanks to Carlo for pointing this useful replacement !.

derselbst:
> The use of to use g_ascii_strtoll comes from the second comment in #677.

Oh yeah, now I remember.

> > Was there a specific reason to use g_ascii_strtoll which I'm missing?
> 
> No, I see no reason to use it

Ok then, thanks!

PR #722, opened 2020-12-27
Fix the fluid_synth_set_sample_rate() change problem for jack driver

### LINKED ISSUE #608: Solve the sample-rate-change problem
The synth exposes the function `fluid_synth_set_sample_rate()` which allows to change the synthesizers sample-rate in real-time. The actual change is not executed immediately. Instead, it is deferred to the next rendering call.

With the implementation of #380 and #548, memory allocation in the reverb and chorus is now required when changing the sample-rate. This is still done while rendering. If the allocation fails, there is no way for the caller to know.

---

How is `fluid_synth_set_sample_rate()` used downstream?

In fluidsynth 1.1.x, `fluid_synth_set_sample_rate()` had the following hint:

http://www.fluidsynth.org/api-1.x/synth_8h.html#ab614bfeafc8b8b632cf1433f8dfce85a

And this is exactly what I find downstream: Create a synth with an incorrect sample-rate, followed by a call to `fluid_synth_set_sample_rate()` to fix the sample-rate. The real-time capability of `fluid_synth_set_sample_rate()` is not used. AFAIK it has never been documented as "real-time capable" either.

Because of the misusage downstream and its broken concept (returning void, no error handling possible), JJC and I have deprecated `fluid_synth_set_sample_rate()`. Chaning the sample-rate of the synth in real-time is not what we consider to be a valid use-case.

Yet, fluidsynth's jack audio driver may require to change the sample-rate of a synth on-the-fly. That's why we need an internal solution. Two approaches are coming to my mind:

1. Implement an internal sample-rate change function, that may only be called, when no other thread is using the synth's API (not even rendering takes place). Else the behavior is undefined.
2. When creating the effects units initially, allocate as much memory needed for the highest sample-rate we support.

Solution 1. would exactly fit the puropse required for the jack driver.

I'm not sure if 2. is possible. In any case, it would be a pessimization for people who use fluidsynth on low-resource-hardware. Although this proposal sounds charming, as it might serve as rescue for the public `fluid_synth_set_sample_rate()` function, keep in mind that an error handling is required, but `fluid_synth_set_sample_rate()` returns void.

Any opinion how to solve that sample-rate-change problem is highly welcome!

derselbst:
This PR adds a new sample rate change function, which executes the change immediately. For internal use only.

Resolves #608.

### DIFF
--- a/src/drivers/fluid_jack.c
+++ b/src/drivers/fluid_jack.c
@@ -501,9 +501,7 @@ fluid_jack_client_register_ports(void *driver, int isaudio, jack_client_t *clien
         {
             FLUID_LOG(FLUID_INFO, "Jack sample rate mismatch, adjusting."
                   " (synth.sample-rate=%lu, jackd=%lu)", (unsigned long)sample_rate, jack_srate);
-            fluid_synth_set_sample_rate(synth, jack_srate);
-            /* Changing sample rate is non RT, so make sure we process it and/or other things now */
-            fluid_synth_process_event_queue(synth);
+            fluid_synth_set_sample_rate_immediately(synth, jack_srate);
         }
         else
         {
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3252,17 +3252,17 @@ fluid_synth_update_presets(fluid_synth_t *synth)
 }
 
 static void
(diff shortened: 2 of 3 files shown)


derselbst (commented):
`fluid_clip(sample_rate, 8000.0f, 96000.0f);` should be applied on top level

Issue #724, opened 2020-12-28
Jack thread terminates if synth.effects-channels and/or synth.audio-channels are used within a config file


mchehab:
### FluidSynth version

```
$ ./src/fluidsynth --version
FluidSynth runtime version 2.1.5
Copyright (C) 2000-2020 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.1.5
Sample type=double
```

### Describe the bug

When either one of the options bellow:
```
set synth.audio-channels 6
set synth.effects-channels 6
```
Are used within a config file, the Jack Kthread terminates, causing `fluidsynth` to silently crash. When this bug happens, it is not possible anymore to manually adjust the connection graph with `qjackctl`, or to use `fluidsynth`.

This bug also affects `qsynth`, making impossible to use it with more than 2 stereo channel outputs.

### Expected behavior

`Fluidsynth` should be able to accept changing the number of audio and effect channels via configuration settings. 

### Steps to reproduce

Create a `config` file with:

```
set audio.jack.multi true
set audio.jack.autoconnect true
set synth.audio-channels 6
set synth.effects-channels 6
```

Call `fluidsynth -v -f config`. The result is:
```
FluidSynth runtime version 2.1.5
Copyright (C) 2000-2020 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of E-mu Systems, Inc.

fluidsynth: debug: Using 'alsa_seq' midi driver
Warning: 'audio.jack.multi' is not a realtime setting, changes won't take effect.
Warning: 'audio.jack.autoconnect' is not a realtime setting, changes won't take effect.
Warning: 'synth.audio-channels' is not a realtime setting, changes won't take effect.
fluidsynth: error: requested set value for setting 'synth.effects-channels' out of range
set: Value out of range. Try 'info synth.effects-channels' for valid ranges
Warning: 'synth.effects-channels' is not a realtime setting, changes won't take effect.
fluidsynth: Received EOF while reading commands, exiting the shell.
Failed to execute user provided command configuration file 'config'
fluidsynth: debug: Using 'jack' audio driver
Jack: JackClient::SetupDriverSync driver sem in flush mode
Jack: JackLinuxFutex::Connect name = jack_sem.1000_default_fluidsynth
Jack: Clock source : system clock via clock_gettime
[...]

PR #725, opened 2020-12-28
fluid_jack: prevent using a different setting at synth and jack

mchehab:
When Jack plugin is set to use multiple stereo outputs, if
set commands (even at the config file) is changing the number
of audio and/or effect effect channels/groups, it may happen
that the settings used by fluid_synth to be different than the
ones used by fluid_jack driver.

This cause errors at the fluid_synth_process() function,
making it return FLUID_FAILED, causing the Jack thread
to stop.

When this happens, the entire jack plugin stops working, which
prevents connecting the fluidsynth outputs. So no audio
goes to the outputs.

The real fix would be to add a runtime update that would be
changing the fluid_synth settings when this happens. However,
a change like that is not trivial and might require some
changes at the API.

So, instead, check if the settings will cause problems
in advance, not leting the plugin initialize on such cases
and printing an error message.

### DIFF
--- a/src/drivers/fluid_jack.c
+++ b/src/drivers/fluid_jack.c
@@ -327,12 +327,15 @@ static int
 fluid_jack_client_register_ports(void *driver, int isaudio, jack_client_t *client,
                                  fluid_settings_t *settings)
 {
+    fluid_synth_t* synth = NULL;
     fluid_jack_audio_driver_t *dev;
     char name[64];
     int multi;
     int i;
     unsigned long jack_srate;
     double sample_rate;
+
+    fluid_jack_obtain_synth(settings, &synth);
 
     if(!isaudio)
     {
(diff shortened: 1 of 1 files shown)


src/drivers/fluid_jack.c
```diff
@@ -445,9 +455,26 @@ fluid_jack_client_register_ports(void *driver, int isaudio, jack_client_t *clien
         }
 
         fluid_settings_getint(settings, "synth.effects-channels", &dev->num_fx_ports);
+
+        if(synth->effects_channels != dev->num_fx_ports)
```
derselbst, line 459: Possible NULL deref on `synth`

src/drivers/fluid_jack.c
```diff
@@ -445,9 +455,26 @@ fluid_jack_client_register_ports(void *driver, int isaudio, jack_client_t *clien
...
+                      synth->effects_channels * synth->effects_groups,
+                      dev->num_fx_ports);
+            goto error_recovery;
+        }
+
         fluid_settings_getint(settings, "synth.effects-groups", &i);
 
         dev->num_fx_ports *= i;
+
+        if(synth->effects_groups != i)
```
derselbst, line 471: Possible NULL deref on `synth`

derselbst (commented):
Thanks. Two issues below.

src/drivers/fluid_jack.c
mchehab: I updated the PR with the above changes. Yet, if you prefer, I can try to come up with a different solution for issue #724, as discussed there.

mchehab:
This is superseeded by PR #737. Please ignore it if #737 gets merged.

derselbst:
Superseded by #739.

Issue #727, opened 2020-12-29
SoundFonts cannot be unloaded if polyphony is ever exceeded


dwhinham:
### FluidSynth version
v2.1.5

### Describe the bug
If polyphony is exceeded and FluidSynth has to allocate a voice by calling `fluid_synth_free_voice_by_kill_LOCAL()`, two problems occur:
- The value returned by `fluid_synth_get_active_voice_count()` becomes inaccurate - the active voice count never returns back to 0 because the counter is not decremented in the usual way.
- SoundFont samples are not unref'd, and therefore if an attempt is made to unload the SoundFont, the deferred unload timer is started, and `fluid_synth_sfunload_callback()` unsuccessfully tries to unload the SoundFont forever.

### Expected behavior
- `fluid_synth_get_active_voice_count()` should provide an accurate value even if polyphony was exceeded during the session.
- The SoundFont should successfully unload, either immediately or eventually (via the deferred timer mechanism), if polyphony was exceeded during the session.

### Steps to reproduce
- Start FluidSynth and load a SoundFont.
- Set polyphony to a low value (for ease of testing), e.g. 2.
- Play more notes than the polyphony permits.
- Attempt to unload the SoundFont.
- Set a breakpoint inside `fluid_synth_sfunload_callback()` and notice that it continues to be called repeatedly long after the request was made, even though all voices may be idle.

### Additional context

As a quick hack to monitor the active voice count, I added the following code just before the call to `fluid_istream_readline()`, near line 510 of `fluid_cmd.c`:

```c++
printf("%d\n", fluid_synth_get_active_voice_count(shell->handler->synth));
```

This is just so I could hit enter and see the result. After polyphony is exceeded, this value never returns to zero.

Thanks for your hard work on FluidSynth!

derselbst:
Yeah, I know. That bug already exists since at least fluidsynth 1.1.6. The problem is that the call to `fluid_voice_off()` in `fluid_synth_free_voice_by_kill_LOCAL()` triggers a delayed `fluid_voice_stop()` which postpones it until the next rendering call is made, where we're being told that the rvoice has indeed finished, so we can safely call `fluid_voice_stop()` now (in which all the sample unref is performed). However, `fluid_synth_alloc_voice_LOCAL()` doesn't wait for the _stop_, it just calls `fluid_voice_init()` right away. Within there, this "overflow rvoice"  is  supposed to handle the premature off, but it seems broken.

[...]

PR #729, opened 2020-12-30
Fix bug reported in issue 727

jjceresa:
This PR addresses issues reported in #727

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -1017,7 +1017,6 @@ delete_fluid_synth(fluid_synth_t *synth)
     fluid_return_if_fail(synth != NULL);
 
     fluid_profiling_print();
-
     /* turn off all voices, needed to unload SoundFont data */
     if(synth->voice != NULL)
     {
@@ -4629,6 +4628,9 @@ fluid_synth_check_finished_voices(fluid_synth_t *synth)
             else if(synth->voice[j]->overflow_rvoice == fv)
             {
                 fluid_voice_overflow_rvoice_finished(synth->voice[j]);
+
+                /* Decrement voice count */
+                synth->active_voice_count--;
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


src/synth/fluid_voice.c
```diff
@@ -321,12 +321,13 @@ fluid_voice_init(fluid_voice_t *voice, fluid_sample_t *sample,
...
 
-    /* Increment the reference count of the sample to prevent the
-       unloading of the soundfont while this voice is playing,
-       once for us and once for the rvoice. */
+    /*
+       We increment the reference count of the sample to indicate that this
+       sample is about to be owned by the rvoice. This will prevent the
+       unloading of the soundfont while this rvoice is playing.
+    */
     fluid_sample_incr_ref(sample);
```
derselbst, line 329: Where is the matching `fluid_voice_sample_unref` for this one? You've removed one below and I can't find another one.

src/synth/fluid_synth.c
```diff
@@ -4629,6 +4628,9 @@ fluid_synth_check_finished_voices(fluid_synth_t *synth)
             else if(synth->voice[j]->overflow_rvoice == fv)
             {
                 fluid_voice_overflow_rvoice_finished(synth->voice[j]);
+
+                /* Decrement voice count */
+                synth->active_voice_count--;
```
derselbst, line 4633: For consistency I think it would be better to move this call into `fluid_voice_overflow_rvoice_finished()`.

derselbst (commented):
That looks surprisingly simple. I haven't tested it yet, I may have an opportunity later to do so.

src/synth/fluid_voice.c
derselbst: Sry, scratch that. I got confused. Ofc `fluid_voice_sample_unref(&voice->rvoice->dsp.sample);` is the matching call.

src/synth/fluid_synth.c
derselbst: But I see you did it earlier and the unit tests failed. Perhaps you could just add a small comment to explain why the decrement must be put here.

derselbst:
Closed in favor of #736 which is based on `2.1.x` branch.

PR #737, opened 2021-01-03
Handle early the "set" commands from the configuration file.

mchehab:
As discussed at issue #724, if no-realtime set commands are used inside
a configuration file, fluidsynth can be on an invalid state, where the synth will be using a different setting than the modules that are initialized after parsing the command.

In order to solve it, parse earlier the set commands.

There are 4 patches on this series.

Patch 1 changes the function which handles the `set` command: it will now verify if the value actually changed. That prevents printing realtime warnings when the value was not changed;

Patch 2 adds a new CMD parser that handles only `set` commands, and is called before initializing the synth;

Patch 3 avoid prints a warning when the commands from config files are parsed.

Patch 4 ensures that the new early parsing logic is only used for settings that can't be changed in real time. I'm not sure if this is needed, but it should prevent backward-compatibility issues, if some config file was using `set` for realtime parameters, and are relying on them being processed only after other commands. Probably overkill, but it shouldn't hurt to add this one to the series.

This solves issue #724. This replaces the approach taken at PR #725.

### DIFF
--- a/include/fluidsynth/shell.h
+++ b/include/fluidsynth/shell.h
@@ -67,6 +67,9 @@ void delete_fluid_cmd_handler(fluid_cmd_handler_t *handler);
 /** @endlifecycle */
 
 FLUIDSYNTH_API
+int parse_fluid_cmd_set(fluid_settings_t *settings, char *config_file);
+
+FLUIDSYNTH_API
 void fluid_cmd_handler_set_synth(fluid_cmd_handler_t *handler, fluid_synth_t *synth);
 
 FLUIDSYNTH_API
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -384,6 +384,15 @@ static const fluid_cmd_t fluid_commands[] =
 #endif
 };
 
+/** the table of the commands that are parsed before creating the synth */
+
+static const fluid_cmd_t pre_fluid_commands[] =
+{
+    {
+        "set", "settings", fluid_handle_set, ""
+    },
+};
+
 /**
  * Process a string command.
  *
@@ -1995,7 +2004,7 @@ fluid_handle_set(void *data, int ac, char **av, fluid_ostream_t out)
         fluid_ostream_printf(out, "set: Value out of range. Try 'info %s' for valid ranges\n", av[0]);
     }
 
-    if(!fluid_settings_is_realtime(handler->synth->settings, av[0]))
+    if(handler->router && !fluid_settings_is_realtime(handler->synth->settings, av[0]))
     {
         fluid_ostream_printf(out, "Warning: '%s' is not a realtime setting, changes won't take effect.\n", av[0]);
     }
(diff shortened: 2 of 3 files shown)


derselbst:
Ok, thanks. Patches 1 and 3 are fine. Patch 4 won't work, because `fluid_settings_is_realtime` only returns true after the `settings` object has been used for creating a `synth`. In your case the settings object has been freshly created, no settings have set a callback, hence it will always return false. Also, I don't see a reason why we would need patch 4 either.

And I don't quite like patch 2 because

* of its hackish `fluid_synth_t synth = { .settings = settings };` approach, and
* it shadows the existing API around `fluid_source` and `new_fluid_cmd_handler`.

I'll make a counter proposal for it upcoming weekend (hopefully).

mchehab:
> Ok, thanks. Patches 1 and 3 are fine. 

Good!

> Patch 4 won't work, because `fluid_settings_is_realtime` only returns true after the `settings` object has been used for creating a `synth`. In your case the settings object has been freshly created, no settings have set a callback, hence it will always return false. Also, I don't see a reason why we would need patch 4 either.

Ok. Let's drop it.

> And I don't quite like patch 2 because
> 
>     * of its hackish `fluid_synth_t synth = { .settings = settings };` approach, and
> 
>     * it shadows the existing API around `fluid_source` and `new_fluid_cmd_handler`.

I opted for this strategy due to a few reasons:

* It made the patch less intrusive, in the sense that the code won't change much, and that the already-existing logic will be called. So, no code duplication.
* As the new cmd handler will be called just to parse files (and not to be used on interactive sections), from API PoV, it seemed better to place everything (file open, file parse and free resource logic) altogether.
* From the PoV of the API callers, the arguments that makes more sense are `fluid_settings_t *settings` and `char *config_file`;
* Currently, `fluid_source()` expects that settings is inside `synth` struct. It is possible to add a `settings` struct at `fluid_cmd_handler_t`, in order to keep using `fluid_source()` without creating first `fluid_synth_t`, but this could break ABI compatibility if not done properly.
* From API PoV, I was not sure if `fluid_cmd_handler_t` is an opaque type or not. So, I opted to be more conservative, in the sense that the existing APIs won't be affected.

> I'll make a counter proposal for it upcoming weekend (hopefully).

Ok. Whatever works best.

derselbst:
I've implemented my original idea in #739. Briefly tested it, seems to work. Pls. have another test.

> Currently, fluid_source() expects that settings is inside synth struct. It is possible to add a settings struct at fluid_cmd_handler_t, in order to keep using fluid_source() without creating first fluid_synth_t, but this could break ABI compatibility if not done properly.
From API PoV, I was not sure if fluid_cmd_handler_t is an opaque type or not. So, I opted to be more conservative, in the sense that the existing APIs won't be affected.

`fluid_cmd_handler_t` is defined in `fluid_cmd.c` itself. I don't see a possibility to break ABI here.

mchehab:
> I've implemented my original idea in #739. Briefly tested it, seems to work. Pls. have another test.

Worked fine, but I had to add this patch:

https://github.com/FluidSynth/fluidsynth/commit/8c49121cf12953f7a4b5d33755cdefbf3cfa1637

More details on my comment to PR #739.

> > Currently, fluid_source() expects that settings is inside synth struct. It is possible to add a settings struct at fluid_cmd_handler_t, in order to keep using fluid_source() without creating first fluid_synth_t, but this could break ABI compatibility if not done properly.
> > From API PoV, I was not sure if fluid_cmd_handler_t is an opaque type or not. So, I opted to be more conservative, in the sense that the existing APIs won't be affected.
> 
> `fluid_cmd_handler_t` is defined in `fluid_cmd.c` itself. I don't see a possibility to break ABI here.

OK!

Yeah, this solution is indeed more elegant than this original PR.

derselbst:
> Worked fine, but I had to add this patch:

Ok, I see. This problem is not limited to settings commands. It could happen with router commands as well. I fixed that.

> More details on my comment to PR #739.

Sorry, but there is no comment.

mchehab:
> > Worked fine, but I had to add this patch:
> 
> Ok, I see. This problem is not limited to settings commands. It could happen with router commands as well. I fixed that.
> 
> > More details on my comment to PR #739.
> 
> Sorry, but there is no comment.

Sorry! I forgot to click at the "submit review" button. Just submitted it.

derselbst:
Superseded by #739.

PR #739, opened 2021-01-08
Handle settings-related commands in user command file early

derselbst:
An alternative approach to #737.

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -2002,7 +2002,7 @@ fluid_handle_set(void *data, int ac, char **av, fluid_ostream_t out)
         fluid_ostream_printf(out, "set: Value out of range. Try 'info %s' for valid ranges\n", av[0]);
     }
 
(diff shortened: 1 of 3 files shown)


src/bindings/fluid_cmd.c
```diff
@@ -4398,13 +4402,23 @@ fluid_cmd_handler_handle(void *data, int ac, char **av, fluid_ostream_t out)
...
+    }
+    else
+    {
+        fluid_ostream_printf(out, "unknown command: %s (try help)\n", av[0]);
+        return FLUID_FAILED;
     }
-
-    fluid_ostream_printf(out, "unknown command: %s (try help)\n", av[0]);
-    return FLUID_FAILED;
 }
```
mchehab, line 4423: Ok, this patch solved nicely the issue with non-set commands.

On my tests, this patch series is working just fine!

mchehab (approved):

src/bindings/fluid_cmd.c
derselbst: Great, thanks for testing!

PR #745, opened 2021-01-19
Add support for new_fluid_audio_driver2 to two Windows-specific drivers.

chirs241097:
This patch is largely based around the ideas in fluid_pulse.c: no fx buffers for the callback, `new_fluid_audio_driver` becomes a call to `new_fluid_audio_driver2`, etc.

I haven't done win32 programming for years... so there could be hidden issues in these patches. Tested on Wine (64-bit wine prefix), Windows 7 and Windows 10 VMs, plus an old Windows 7 box, all using a mingw build (x86_64-w64-mingw32).

### DIFF
--- a/src/drivers/fluid_waveout.c
+++ b/src/drivers/fluid_waveout.c
@@ -267,7 +267,7 @@ void fluid_waveout_audio_driver_settings(fluid_settings_t *settings)
 fluid_audio_driver_t *
 new_fluid_waveout_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
 {
-    return new_fluid_waveout_audio_driver2(settings, NULL, synth);
+    return new_fluid_waveout_audio_driver2(settings, (fluid_audio_func_t)fluid_synth_process, synth);
 }
 
 fluid_audio_driver_t *
@@ -395,11 +395,6 @@ new_fluid_waveout_audio_driver2(fluid_settings_t *settings, fluid_audio_func_t f
             delete_fluid_waveout_audio_driver(&dev->driver);
             return NULL;
         }
-    }
-    else
-    {
-        dev->lbuf = NULL;
-        dev->rbuf = NULL;
     }
 
     /* get the selected device name. if none is specified, use default device. */
@@ -543,15 +538,8 @@ void delete_fluid_waveout_audio_driver(fluid_audio_driver_t *d)
         CloseHandle(dev->hQuit);
     }
 
-    if(dev->lbuf != NULL)
-    {
-        FLUID_FREE(dev->lbuf);
-    }
-
-    if(dev->rbuf != NULL)
-    {
-        FLUID_FREE(dev->rbuf);
-    }
+    FLUID_FREE(dev->lbuf);
+    FLUID_FREE(dev->rbuf);
 
     HeapFree(GetProcessHeap(), 0, dev);
 }
@@ -567,8 +555,8 @@ static int fluid_waveout_write_processed_channels(fluid_synth_t *data, int len,
     float *out[2] = {drv->lbuf, drv->rbuf};
     float *lptr;
     float *rptr;
-    memset(drv->lbuf, 0, len * sizeof(float));
-    memset(drv->rbuf, 0, len * sizeof(float));
+    FLUID_MEMSET(drv->lbuf, 0, len * sizeof(float));
+    FLUID_MEMSET(drv->rbuf, 0, len * sizeof(float));
     ret = drv->func(drv->synth, len, 0, NULL, 2, out);
     lptr = (float*)channels_out[0] + channels_off[0];
     rptr = (float*)channels_out[1] + channels_off[1];
--- a/src/drivers/fluid_waveout.c
+++ b/src/drivers/fluid_waveout.c
@@ -88,8 +88,8 @@ typedef struct
     void *synth;
     fluid_audio_func_t func;
     fluid_audio_channels_callback_t write_ptr;
-    float *lbuf;
-    float *rbuf;
+    float **drybuf;
+    float **efxbuf;
 
     HWAVEOUT hWaveOut;
     WAVEHDR  waveHeader[NB_SOUND_BUFFERS];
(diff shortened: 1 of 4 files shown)


src/drivers/fluid_waveout.c
```diff
@@ -507,9 +543,45 @@ void delete_fluid_waveout_audio_driver(fluid_audio_driver_t *d)
...
+                               int channels_incr[])
+{
+    int i;
+    int ret;
+    fluid_waveout_audio_driver_t *drv = (fluid_waveout_audio_driver_t*) data;
+    float *out[2] = {drv->lbuf, drv->rbuf};
+    float *lptr;
+    float *rptr;
+    memset(drv->lbuf, 0, len * sizeof(float));
+    memset(drv->rbuf, 0, len * sizeof(float));
```
derselbst, line 571:
```suggestion
    FLUID_MEMSET(drv->lbuf, 0, len * sizeof(float));
    FLUID_MEMSET(drv->rbuf, 0, len * sizeof(float));
```

derselbst (commented):
Looks promising. Can't see any "hidden issues" either. Thanks!

Is there a reason for not providing fx buffers? No effects will ever be audible. You could just write

```c
drv->func(drv->synth, len, 2, out, 2, out);
```

and the fx would be mixed into the regular out buffers.

Some cosmetic issues below.

chirs241097:
Passing `fluid_synth_process` instead of `NULL` to the callback does work for both drivers. I only tested this on Wine though.

As for the fx buffers, quoting the documentation for `fluid_audio_func_t`:
> Whereas fluid_synth_process() allows aliasing buffers, there is the guarentee that out and fx buffers provided by fluidsynth's audio drivers never alias.

Writing `drv->func(drv->synth, len, 2, out, 2, out);` would break this guarantee. I wasn't providing the fx buffers simply because I was doing whatever the pulseaudio driver does. The alsa driver doesn't provide an fx buffer either. The only driver I have referenced that does provide the fx buffers was the jack driver, which does have separate buffers for the fx and output buffers.

However not being able to render the effects definitely is a problem. I would push a fix to this branch later today. I might start a separate pull request fixing it for the pulseaudio and alsa drivers as well.

chirs241097:
The latest commit should fix the effects buffer issue. Also multiple stereo outputs should be supported now. The code has been tested on all platforms formerly tested, plus a Windows XP x64 VM.
Can't test the multi-output feature right now because I haven't got any code that makes use of the feature.

derselbst:
> As for the fx buffers, quoting the documentation for `fluid_audio_func_t`:
> 
> > Whereas fluid_synth_process() allows aliasing buffers, there is the guarentee that out and fx buffers provided by fluidsynth's audio drivers never alias.

Right, I forgot, sorry. As the docs say, this prevents the user from erroneously applying custom effects multiple times.

Thinking this further, calling the callback function as you originally did

`func(drv->synth, len, 0, NULL, 2, out);`

would still allow the user render in effects by calling `fluid_synth_process(synth, len 2, out, 2, out);`. I think that's also the reason why it's done that way in Pulse and Alsa drivers.

It is still hard for me to judge which way is the right one to go here:

1. Giving the audio driver an extra set of fx buffers (like you did now), or
2. keeping only the minimum set of audio buffers in the driver itself, and let the user's callback function decide whether or not to mix in fx.

Since you seem to have a real use-case for the `new_fluid_audio_driver2()` thing, I would leave that decission to you.

Pls. note that if you'd vote for option 1. there is a problem with your provided implementation: You are allocating a total of `audio_channels` stereo fx channels. But `fluid_synth_process` can only handle a maximum number of
 `synth.effects-channels *  synth.effects-groups` stereo fx channels, which is `2` by default. So, whenever `audio_channels` is greater `2`, calling
 `fluidsynth_process(synth, len, drv->channels_count, drv->efxbuf, drv->channels_count, drv->drybuf);`
would fail, because more fx buffers were provided than the synth has been configured with.

You'll also notice, that if you fix this issue, by allocating the correct number of fx stereo channels, you will have the problem that you cannot easily do the dry+fx mixture in `fluid_waveout_write_processed_channels()`:

`*optr[ch] = drv->drybuf[ch][i] + drv->efxbuf[ch][i];`

Which (unless I've missed smth.) would makes us fall back to option 2.: 
`drv->func(drv->synth, len, 0, NULL, audio_channels, out);`

Sorry for the confusion. But the multi-output feature you added is still nice!

chirs241097:
Uh... thanks for letting me know that. Well, guess I'll scrap the last commit then. It seems that only the jack driver provides the effect buffers anyway (which is also hidden behind the flag `audio.jack.multi`). So the most sensible choice right now is indeed opting for the second option.

My particular use case of `new_fluid_audio_driver2` is very simple -- just doing some monitoring stuff. Since I've just realized that my callback function should mix the effects with the dry output, I've modified my callback function to something like this:
```C++
        [](void *t, int l, int nfx, float *fx[], int nout, float *out[])->int
        {
            FluidSynthDevice *self = static_cast<FluidSynthDevice*>(t);
            float *fxb[4] = {out[0], out[1], out[0], out[1]};
            int ret = fluid_synth_process(self->synth, l, nout * 2, fxb, nout, out);
            // unrelated processing stuff that accesses the output buffer read-only ...
            return ret;
        }
```
However, something seems a bit off: the chorus effect doesn't render, while reverb renders fine (testing was done while using the pulseaudio driver). Am I missing something here?

Also since now `fluid_xxxx_write_processed_channels` is always used instead of `fluid_synth_write_float_channels`, the former of which is currently incapable of doing multiple outputs, this patch would introduce a feature regression. I'd like to hear what you think should be done here. Personally I would just allocate 2 * `audio_channels` buffers and pass them to the callback (while still not providing any buffers for effects).

Finally, I really appreciate you helping me with these details because I'm also working on a WASAPI driver for fluidsynth. Learning these will definitely save me a lot of time!

derselbst:
> Well, guess I'll scrap the last commit then.

No, just scrapping the fx buffers should do. I.e. like you said "I would just allocate 2 * audio_channels buffers and pass them to the callback (while still not providing any buffers for effects)."

> It seems that only the jack driver provides the effect buffers anyway

Yes, I'll elaborate the documentation of `fluid_audio_func_t` a bit on that.


> However, something seems a bit off: the chorus effect doesn't render, while reverb renders fine

Pls. double-check that chorus is on (`fluid_synth_chorus_on()`) and that CC 93 is >0 . Perhaps play around with the chorus parameter a bit (the defaults might be too quiet to hear). Also, in your callback function it should be 

```c
int ret = fluid_synth_process(self->synth, l, sizeof(fxb) / sizeof(fxb[0]) /*=4*/, fxb, nout, out);
```

> Finally, I really appreciate you helping me with these details because I'm also working on a WASAPI driver for fluidsynth

Aha, issue 504, very nice. Looking forward to that!

chirs241097:
Done. Now these two drivers will not provide effect buffers to the callback, just like every other driver except jack.

I think I also figured out why chorus wasn't rendering. See you in another PR ;)

derselbst:
Great, thank you!

(I've squashed your last two commits and merged them to master, but Github is not smart enough to recognize the merge.)

PR #747, opened 2021-01-19
Update Oboe driver

### LINKED ISSUE #740: Oboe driver can have lower latency if Oboe sample rate conversion used
Thanks for adding support for Oboe!

I notice that you sometimes set the sample rate explicitly.
If AudioFlinger does the sample rate conversion then it can add 80+ msec of latency.
So we added an SRC in Oboe that only adds a few msec.

To enable it:

    builder.setSampleRateConversionQuality(SampleRateConversionQuality::Medium);

Docs at:

https://google.github.io/oboe/reference/classoboe_1_1_audio_stream_builder.html#af7d24a9ec975d430732151e5ee0d1027

### LINKED ISSUE #741: Oboe driver should use std:shared_ptr for stream
I notice that fluid_oboe.cpp declares a pointer to AudioStream.
https://github.com/FluidSynth/fluidsynth/blob/b5da68393c14a116cc4f1aca72cbd2936dce1f51/src/drivers/fluid_oboe.cpp#L55

That could lead to occasional use after free race conditions.
So we now recommend using std::shared_ptr. Example at: https://github.com/google/oboe/releases/tag/1.4.2

You will also need to change the

    delete dev->stream;

to 

    dev->stream.reset();

derselbst:
This incorporates Phil's suggestions around the oboe driver:

* The quality of the sample-rate conversion now depends on the performance mode.
* The `stream` uses a `std::shared_ptr` to avoid use-after-free.
* Oboe has been updated to 1.5.0 in our Makefile.

Resolves #740.
Resolves #741.


derselbst:
@atsushieno Perhaps you would like to give this a try before we merge it.

atsushieno:
@derselbst Thanks for the heads up. I haven't built and ran locally, but figured that the changes makes the code much better (especially in that it is not just C code with .cpp extension). My ongoing projects that use fluidsynth don't use audio driver, and I have to go through non-trivial steps to set up my old projects, so please go ahead without my local verification. I will report back if anything happens.

Mapiarz:
Thanks @derselbst, I really appreciate your continued work on this and FS in general.

If I may chip in:
1. I'd leave number of frames per callback unspecified, as discussed in #740. Moreover, `setFramesPerCallback` has been deprecated in favor of `setFramesPerDataCallback` (https://google.github.io/oboe/reference/classoboe_1_1_audio_stream_builder.html#a3f397821f61eabaeedaf31064c859a54)
2. I would not couple performance mode setting with the SRC quality. I think it's worth introducing a separate, oboe driver specific, setting to control SRC quality. Some users, myself included, would benefit from a fine grained control over SRC. Plus, it's extremely useful to disable SRC altogether when debugging (while still using Oboe's `PerformanceMode::LowLatency`). Docs for Oboe only mention `None`, `Fastest` and `Best`. Nonetheless, looking at source code it seems that all values defined here: https://github.com/google/oboe/blob/master/include/oboe/Definitions.h#L235 are used, e.g. here: https://github.com/google/oboe/blob/master/src/flowgraph/resampler/MultiChannelResampler.cpp#L55. I'd expose all of the enum values defined in `Definitions.h` and set the default to something that would provie a good middle ground that would benefit most users, even if they don't set the correct sample rate for their output device, e.g. `Medium`.
3. I really think it's worth documenting somewhere what kind of impact `synth.sample-rate` has when using Oboe on Android. Do you know where it could be done? Maybe the above setting could mention that and explain that SRC will take place if `synth.sample-rate` is not set to a devices native output sample rate? 
4. Not so much a suggestion but rather a question: is there a way to access the Oboe audio stream that the driver creates? That would also be useful, even if just for debugging.

I could try and contribute but I will only be able to do so starting next week. Not sure how timely this is but if not super timely, I could help.

derselbst:
@Mapiarz I've implemented points 1. und 3.

For point 2. I would introduce a new setting `audio.oboe.sample-rate-conversion-quality`. But should it be a string setting (like the already existing `audio.oboe.performance-mode` and `audio.oboe.sharing-mode` settings)? Or should it be an int setting, which just receives the pure enum value? And if it should be an int setting, should we break the API and change the other to settings to int as well? (fluidsynth 2.2.0 won't be ABI compatible anyway, this might be a good time to do it)

derselbst:
... on the other hand, fluidsynth + oboe are most likely used by Java guys. For those strings might be more convenient to use, rather than using the enum values from `oboe.h`...

Mapiarz:
Thank you for the update.

As for point 2. I don't really have a strong opinion or preference. Depending on the use case I can see either as being 'better'. As long as the settings are consistent, either all are strings or all are ints, I think it's fine.

derselbst:
This should be ready now. I'll merge it upcoming weekend. Thanks for the feedback!

philburk:
Changes look good!

PR #752, opened 2021-01-23
Remove dead code from dsound and waveout drivers

derselbst:
Follow up for #745:

- remove dead code now that `callback` is never NULL
- fix a possible NULL deref in both drivers
- make sure fx are audible in the special case of `callback == fluid_synth_process`


derselbst:
Seems like we are still going to need that code: #760

Issue #504, opened 2019-02-01
Add WASAPI support.


carlo-bramini:
The Windows Audio Session API (WASAPI) is the successor of the DirectSound for all applications starting from Windows Vista. It is able to provide excellent realtime performance. When configured in exclusive mode, it makes the same results of kernel streaming, but without the need to program in kernel mode, since it is an user mode component.

I think it will be supported one day ;)

jjceresa:
>The Windows Audio Session API (WASAPI) is the successor of the DirectSound for all applications starting from Windows Vista. It is able to provide excellent realtime performance.

Nice proposal. Thanks Carlo.

carlo-bramini:
I'm having some troubles with this implementation.
The new driver tries to create the interfaces that it needs, but it can fail if it runs on a version of Windows without WASAPI support included, like XP for example.
When this happens, FluidSynth exits nicely with the message:

`Failed to create the audio driver`

and it does not try to run other included drivers that may be supported.
Actually, it seems to me that the current implementation does not allow to fall down to other drivers if this condition will happen.
Perhaps, an user can decide rewrite his code by using `fluid_audio_driver_register()` and select just one driver at time, but the current code in fluidsynth.c does not seem to be able to bypass this issue.
What do you suggest to me?

derselbst:
> and it does not try to run other included drivers that may be supported.

The user decides which driver to start by setting `audio.driver` or `-a` flag. The user gets what he wants and if it fails fluidsynth quits. That's intended for the fluidsynth executable.

A user application is free to implement a fallback for audio drivers. E.g. based on the available options for `audio.driver`. Can't recommend messing around with `fluid_audio_driver_register()` in this case. Don't see why it's needed anyway.

> The new driver tries to create the interfaces that it needs, but it can fail if it runs on a version of Windows without WASAPI support included, like XP for example.

Perhaps not making WASAPI the default driver on windows should be fine then? If the user wants it, he may specify it explicitly.

carlo-bramini:
Excuse me, is there a particular reason because FluidSynth must fail when the first driver found in the list cannot initialize correctly? If I can say my opinion, it won't be a bad idea to continue and try to initialize the next driver in the list, when `audio.driver` is set to `"default"`.
And what about allowing `audio.driver` to handle values like `"wasapi;directsound;waveout"`?
It is fully backward compatible with current implementation when you specify a single driver and it can eventually handle more of them if requested by the user.

[...]

PR #754, opened 2021-01-25
Add WASAPI driver.

chirs241097:
See #504.
Please review my code with vigorous scrutiny because I haven't really worked on audio stuff for a while (last time I did this I was using OpenAL, which is not as low-level as WASAPI either). Please pay extra attention to:
 - Memory leaks
 - Misuse of the WIN32 API (I'm not a Windows person... unfortunately)
 - Threading stuff

This driver is currently tested and verified to work on:
 - Windows Vista x64 VM
 - Windows 7 x64 VM
 - Windows 10 1909 x64 VM
 - Windows 10 21296 x64 on a ThinkPad X1 Yoga 1st Gen with 3 different sound cards (Conexant CX20753/4, Scarlett Solo Gen 2, Aureon 7.1 USB)
The test build is built with a x86_64-w64-mingw32 toolchain.

This driver is capable of reaching very low latency in exclusive mode (~6ms on Scarlett Solo with 48kHz).

I'm aware that this driver is kind of spamming the logs in comparison to other drivers... So if you think those are too much, I'll definitely reduce the log spam.

Other stuff to discuss includes the driver preference for Windows. Currently I just simply listed the WASAPI driver between the DirectSound driver and the WaveOut driver alphabetically. However I do think this needs changing. A method to apply conversion to integer samples with dithering in audio drivers should also be considered. At least one driver other than the WASAPI driver (pulseaudio driver) would benefit from it.


src/drivers/fluid_wasapi.c
```diff
@@ -0,0 +1,811 @@
...
+    return new_fluid_wasapi_audio_driver2(settings, (fluid_audio_func_t)fluid_synth_process, synth);
+}
+
+fluid_audio_driver_t *new_fluid_wasapi_audio_driver2(fluid_settings_t *settings, fluid_audio_func_t func, void *data)
+{
+    fluid_wasapi_audio_driver_t *dev = NULL;
+    double sample_rate;
+    int periods, period_size;
+    long long buffer_duration_reftime;
+    long long periods_reftime;
```
derselbst, line 154: Pls. use `fluid_long_long_t` everywhere. `long long` is a C99 type and MSVC has just started a few months ago (or so) to fully support it...

src/drivers/fluid_wasapi.c
```diff
@@ -0,0 +1,811 @@
...
+    HRESULT ret;
+    OSVERSIONINFOEXW vi = {sizeof(vi), 6, 0, 0, 0, {0}, 0};
+
+    if(!VerifyVersionInfoW(&vi, VER_MAJORVERSION|VER_MINORVERSION|VER_SERVICEPACKMAJOR,
+        VerSetConditionMask(VerSetConditionMask(VerSetConditionMask(0,
+            VER_MAJORVERSION,VER_GREATER_EQUAL),
+            VER_MINORVERSION,VER_GREATER_EQUAL),
+            VER_SERVICEPACKMAJOR, VER_GREATER_EQUAL)))
+    {
+        FLUID_LOG(FLUID_ERR, "wasapi: this driver requires Windows Vista or newer.");
```
derselbst, line 176: Do we need to have this "runtime" check? I would have expected cmake to complain (or silently not build wasapi) if you'd try to build it on e.g. XP.

src/drivers/fluid_wasapi.c
```diff
@@ -0,0 +1,811 @@
...
+
+    ret = IAudioClient_IsFormatSupported(dev->aucl, share_mode, (const WAVEFORMATEX*)&wfx, (WAVEFORMATEX**)&rwfx);
+    if(FAILED(ret))
+    {
+        FLUID_LOG(FLUID_ERR, "wasapi: device doesn't support the mode we want. %x", (unsigned)ret);
+        goto cleanup;
+    }
+    else if(ret == S_FALSE)
+    {
+        FLUID_LOG(FLUID_INFO, "wasapi: requested mode cannot be fully satisfied.");
```
derselbst, line 311: This log level should be kept at `FLUID_INFO`. If it is supposed to occur very rarely, perhaps even promote it to `FLUID_WARN`.

src/drivers/fluid_wasapi.c
```diff
@@ -0,0 +1,811 @@
...
+        if(rwfx->Format.nSamplesPerSec != wfx.Format.nSamplesPerSec) // needs resampling
```
derselbst, line 312: An alternative to resampling would be changing the sample-rate of the synth. (Similar to how the jack driver does it.) I have no preference for either way. But if you said you'd rather change the sample-rate, I could help you to fetch the `synth` pointer from the `settings`.

src/drivers/fluid_wasapi.c
```diff
@@ -0,0 +1,811 @@
...
+    {
+        ret = IMMDevice_GetId(dev, &id);
+        if(FAILED(ret))
+        {
+            FLUID_LOG(FLUID_ERR, "wasapi: cannot get id of device. %x", (unsigned)ret);
+            goto cleanup;
+        }
+        nsz = wcslen(id);
+        d->id = FLUID_ARRAY(wchar_t, nsz + 1);
+        memcpy(d->id, id, sizeof(wchar_t) * (nsz + 1));
```
derselbst, line 730: `FLUID_MEMCPY`

derselbst (commented):
Looks very good already! I only had a brief look over it. I'll probably have more time towards the end of the week (to haunt some more memory leaks ;) ).

Also, I'm not so much of a Windows programmer either. But @carlo-bramini might be interested in this as well.

Some general notes:

- Pls. add `WASAPI_SUPPORT` to [cmake's report summary](https://github.com/FluidSynth/fluidsynth/blob/70abf1953f83ccd791839b75a257a36181aeea8e/cmake_admin/report.cmake#L81), so that we can see more easily when it's being built.
- Need to figure out what makes the unit tests segfault.
- Is there a need or benefit in supporting `int16` samples?
- When you log out errors, you're currently only logging the integer. Pls additionally log a textual representation of the error (given by `GetLastError()`). You can directly use `"%s", fluid_win32_error(hr)` for this.
- I think that most of the `FLUID_INFO` logs can be safely downgraded to `FLUID_DBG` level.

Some other minor issues below.

chirs241097:
No clue why half of the unit tests failed. I use a cross build setup so I can't just run `make check` to run the unit tests unfortunately. Also path to the test soundfont is determined at build time, which doesn't help either.

Anyway I just ran some of the failed unit tests on the Windows 10 VM manually (with `TEST_SOUNDFONT` redefined), but none of them failed. Now I'm even more clueless why those tests failed on CI. There's definitely something wrong though, since the CI mingw builds also failed those tests... I'll look into this again after fixing those memory leaks.

Integer samples is very much needed for this driver because the sample format must match the native sample format of the device in exclusive mode. None of the sound cards I've got have support for 32-bit float samples. Dropping integer samples will essentially make exclusive mode useless for most consumer sound cards.

As for error messages, since WASAPI uses the COM style error handling, so `GetLastError()` wouldn't work. We probably need another whole chunk of `static char *fluid_win32_error(HRESULT hr)` for that.

src/drivers/fluid_wasapi.c
```diff
@@ -0,0 +1,811 @@
...
+
+    for(i = 0; i < cnt; ++i)
+    {
+        IMMDevice *dev = NULL;
+        
+        ret = IMMDeviceCollection_Item(dcoll, i, &dev);
+        if(FAILED(ret))
+        {
+            FLUID_LOG(FLUID_ERR, "wasapi: cannot get device #%u. %x", i, (unsigned)ret);
+            continue;
```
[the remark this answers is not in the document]
chirs241097: If `IMMDeviceCollection_Item` fails, `dev` would be `NULL`: https://docs.microsoft.com/en-us/windows/win32/api/mmdeviceapi/nf-mmdeviceapi-immdevicecollection-item

src/drivers/fluid_wasapi.c
chirs241097: WASAPI uses COM, and therefore only needs the headers to build. Mingw always ship those headers, so I'd imagine it's possible to build this driver on Windows XP. If this driver is used on Windows XP or lower, `CoCreateInstance` would return `REGDB_E_CLASSNOTREG` and could be sort of confusing for end users.

src/drivers/fluid_wasapi.c
chirs241097: Oops, almost forgot msvc has terrible C standards support until recent versions!

src/drivers/fluid_wasapi.c
chirs241097: Personally I'd respect the values from the settings. If we automatically change the sample rate of the synth according to the audio driver, should we also consider automatically falling back to shared mode if exclusive mode is selected but isn't available?

src/drivers/fluid_wasapi.c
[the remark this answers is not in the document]
derselbst: Right, sry.

src/drivers/fluid_wasapi.c
derselbst: Ok.

src/drivers/fluid_wasapi.c
derselbst: Ok, then let's respect the `settings` and fail if the mode is not available (as you already do).

derselbst:
> Integer samples is very much needed for this driver because the sample format must match the native sample format of the device in exclusive mode. None of the sound cards I've got have support for 32-bit float samples. Dropping integer samples will essentially make exclusive mode useless for most consumer sound cards.

Ok, then let's keep it. Do you know whether this also applies to WaveOut and Dsound drivers?

> No clue why half of the unit tests failed.

I think there is a problem when the CI makes a call to `fluid_wasapi_audio_driver_settings`.  This would explain why all the test which create a `settings` fail. Unfortunately, I cannot reproduce this on my local Win VM either. I'll further investigate later this week.

> As for error messages, since WASAPI uses the COM style error handling, so GetLastError() wouldn't work

And it doesn't look like `FormatMessage()` would work for COM error either, I guess. Ok, let's ditch this idea for now.

chirs241097:
Just tested a build with msys2 on my Windows 10 VM. Only one unit test failed (test_utf8_open). I'm guessing it could have something to do with their CI environment having no audio devices available?

DirectSound and WaveOut doesn't have exclusive access to sound devices as far as I know, so not having integer sample support doesn't really matter. If the WASAPI is the only driver that needs integer sample support, I'm OK with simply yanking the dithering code from fluid_synth.c.

Error handling in COM is a mess... That's probably why even Microsoft themselves have given up (Windows only gives the hex value of the `HRESULT` in a lot of places, for example Windows Update). I personally use www.hresult.info for debugging, which is unfortunately comprehensive either (they don't have `AUDCLNT_E_INVALID_DEVICE_PERIOD` there).

derselbst (commented):
Thanks for fixing the unit tests. I had an opportunity to give this a test on Win10. Unfortunately `IAudioClient_IsFormatSupported` keeps failing with `AUDCLNT_E_UNSUPPORTED_FORMAT`. No matter whether I use int16, float, shared or exclusive mode. When I changed the synth's sample-rate to 48kHz, it works fine, at least in shared mode.

Three other nit-picks below.

chirs241097:
> IAudioClient_IsFormatSupported keeps failing with AUDCLNT_E_UNSUPPORTED_FORMAT

In exclusive mode, this is expected. The selected audio format must be supported by the device, otherwise this call would fail.

However I'm not sure why that happens to shared mode as well. Shared mode should support a wide range of formats (if the requested format differs from the mixing engine's format, this call should return `S_FALSE` and take the "needs resampling" route). In my testing setup (VirtualBox 6.1.16, Windows 10 1909 x64), the only supported hardware format is 16-bit 44100Hz PCM, the mixing engine also runs at 44100Hz and would resample anything other than that (I've tested 22050Hz, 32000Hz and 48000Hz) to 44100Hz.

Please check the default format of the shared mode mixing engine (you can find it in the advanced tab of the speakers properties dialog, which could be opened by clicking the device icon in the classic volume mixer `sndvol.exe`). In that tab you will also find a checkbox for exclusive mode, which must be checked for exclusive mode to work.

Would you mind also giving me the specs of your testing setup (if it's a virtual machine, the host software; if it's not, the sound card being used. And the operating system version info) so that I can investigate this issue further?

chirs241097:
Just wrote a quick test for WASAPI devices... The file attached contains:
 - `fluid_wasapi_test.c`: the source code. Build with `gcc fluid_wasapi_test.c -o fluid_wasapi_test.exe -lfluidsynth` (or equivalent, plus any other required arguments). Simply run the executable and take stdout as the test result.
 - `w10i.txt`: test results on the laptop mentioned above. OS is Windows 10 x64 build 21296.
 - `wine.txt`: test results on the same laptop, but on Gentoo Linux with wine-vanilla 5.22 using the pulseaudio driver. Wine's pulseaudio driver doesn't implement the exclusive mode at all.
 - `alc888_ich10hda.txt`: test results on a Windows 7 SP1 x64 box. This is the oldest machine with Windows on it that I can test on. Sound card is Intel HDA (ICH10) with the Realtek ALC888 chip.
 - `vboxw10vm.txt`: results for the Windows 10 VM running within VirtualBox 6.1.16.
[fluid_wasapi_test.zip](https://github.com/FluidSynth/fluidsynth/files/5880939/fluid_wasapi_test.zip)

All devices I've tested support all listed sample rates in shared mode (mostly of them requires resampling, which is indicated by `OK(I)` in the output). So I'm really curious why only 48kHz is supported by your setup.

derselbst:
I used a Laptop with Win10 x64 build 18363
[myLaptop.txt](https://github.com/FluidSynth/fluidsynth/files/5882003/myLaptop.txt)

Three devices here:
-  B326HUL (the screen loudspeakers via HDMI)
- (USB) Headset Earphone
- Speakers/Headphones (onboard audio)

When the headset it plugged in, it's the default. If it isn't, onboard audio is default.

Your demo program is really useful! I tested a few more configurations listed as OK and they do work! I think it would be worth to add something like as well. I'm thinking of a new commandline option for the fluidsynth exec that prints this list. Something like

`fluidsynth.exe --query-soundcards` or `--query-audio-formats` or `--query-audio-devices`.

We don't have to decide about that in this PR. But I'll take this on the agenda an look into this later.

chirs241097:
Interesting. So sound cards that don't support all sample rates in shared do exist. Just out of curiosity, may I know the codec chip of your integrated sound card? You can find out that in `alsamixer` by pressing `F6` and select the integrated sound card (usually "HDA Intel PCH" on an Intel machine, not sure about others). The chip name is shown in the second line.

Glad to hear my test program is useful! I did notice that a few drivers (notably the pulseaudio and alsa drivers) of fluidsynth doesn't support device enumeration (without using routines outside of fluidsynth). Should we do something about it?

derselbst:
It's a pure Windows laptop, so no alsa here. But the technical specs say it's a `Realtek ALC3281` controller.

> I did notice that a few drivers (notably the pulseaudio and alsa drivers) of fluidsynth doesn't support device enumeration

Idk. At least I myself don't have a particular need for this. But if you have one, I'm open to suggestions.

chirs241097:
Ah, I see. I don't have a use case for device enumeration either, just think it's a nice thing to have for all drivers frequently used with command line usage if we decide to add the `--query-*` option.

derselbst:
Having this query option on windows is reasonable because
- wasapi rejects certain modes (which may be incomprehensible to certain users), and
- to find (the best) working mode for low latency kernel streaming in exclusive mode.

On Linux, I have never seen alsa or pulseaudio rejecting a particular configuration. And if you want low latency here, you'd just use jack.

So, since we both don't have a use-case for this, I tend to say: "You ain't gonna need it."

derselbst (approved):
Great! I've formatted the code, gave it a final test and it works fine. Thank you!

Once the CI has completed, I'll squash merge this.

chirs241097:
Thanks a bunch for testing! I'll test my application with this in a couple of days. I'll look out for ways to improve the performance of the audio thread loop as well (I believe there's an event-based mechanism to feed data to the audio client which might give even better performance). Also feel free to mention me if there's an issue with this driver. I'll do my best to come up with a fix.

PR #755, opened 2021-01-26
Allow the MIDI file player restart on-demand

rncbc:
Resets the default sample timer properly, allowing the internal
MIDI file player to restart its playlist on any other time but
the initial first.

Lets Qsynth play any MIDI files that are drag-n-dropped, anytime
after the first, following synth engine initialization.


derselbst (approved):
Ok, thanks. Will be part of 2.1.7 which I'll be releasing upcoming weekend.

derselbst:
Cherry-picked to `2.1.x` branch

Issue #758, opened 2021-01-28
Windows waveout driver crash


jjceresa:
This is issue has been introduced by commit 772f62fc91c06f070791fea9cd81e9629d7ff71f.

This commit has introduced the use of `audio.periods` and `audio.period-size` settings
The segmentation fault occurs for low value of audio.periods, audio.periods-size, for example
- audio.periods=4, audio.periods-size=64 or 128.
This produces a segmentation when the driver is started, or when quitting the application.
Will fix that in a PR.

derselbst:
Fixed by #759

PR #759, opened 2021-01-28
Fixes waveout driver crash

jjceresa:
This PR addresses issue https://github.com/FluidSynth/fluidsynth/issues/758.

It ensures that the `internal buffer size` used by waveout device driver and the `extra buffers size` required by fluid_synth_process() are both coherent (i.e they should be of same size). To ensure this, both kind of buffers are now dependent of`audio.period` and `audio.period-size settings`.


derselbst (approved):
Ok, thanks!

PR #762, opened 2021-01-30
fix player ordering of operations in fluidynth console application

jjceresa:
This change ensure that seeking player's commands in configuration file are taking account correctly.
Hope this will not break ordering of operations for others components.


derselbst (changes requested):
Sorry, but those `msleep` delays are not acceptable. Esp. because they won't help anything when user calls the rendering API manually.

Also, why are you moving the `fluid_source()` call after the creation of the audio driver? If the user has any prog or bank changes in the config file, there is no guarantee that the first note will be started with the desired preset.

Pls. what exactly is your motivation for this change?

jjceresa:
>Sorry, but those msleep delays are not acceptable. Esp. because they won't help anything when user calls the rendering API manually.
Also, why are you moving the fluid_source() call after the creation of the audio driver?
Pls. what exactly is your motivation for this change?

This change ensures that seeking player commands in configuration file are taking account correctly.
Without the change, any seeking commands (player_next, player_step) in configuration file will be ignored due to the fact that:
- The player API are executed only when the internal timer is already running. When this timer is the sample timer, that means that the synthesis thread (audio driver) must be running first, before any call to the player API functions.
- player `seek` API is asynchronous. So, there is not guarantee that 2 or more successive call succeed. for example any call `immediately` after the first seek call will fail. This is the reason of the delays. I know that it isn't perfect.

>If the user has any prog or bank changes in the config file, there is no guarantee that the first note will be started with the desired preset.

You are right, I missed this. So we can ignore this PR for now. We just need to note that the player API doesn't behave correctly and player_next, player_step command in command file won't work.

jjceresa:
>If the user has any prog or bank changes in the config file, there is no guarantee that the first note will be started with the desired preset.

This issue is fixed by last commit 8c80422.

jjceresa:
I think that we can close this PR and take https://github.com/FluidSynth/fluidsynth/pull/765 instead.

Issue #763, opened 2021-01-31
The MIDI file player starts playing immediately at soon it is created


jjceresa:
On `master`, the player starts playing immediately at soon it is created (`new_fluid_player()`) and a MIDI file is added `fluid_player_add()`.

expected behaviour: The player should stay in "stop" state after its creation.

Steps to reproduce:
1) comment line 994 in fluidsynth.c: `//        fluid_player_play(player);`
2) fluidsynth anysf2.sf2 anymidifile.mid

derselbst:
Should be fixed by #765

jjceresa:
>Should be fixed by #765

Yes, it is fixed by #765.

PR #764, opened 2021-01-31
fix intial player state

jjceresa:
This PR addresses issue https://github.com/FluidSynth/fluidsynth/issues/763.

### DIFF
--- a/include/fluidsynth/midi.h
+++ b/include/fluidsynth/midi.h
@@ -209,7 +209,7 @@ enum fluid_player_status
 {
     FLUID_PLAYER_READY,           /**< Player is ready */
     FLUID_PLAYER_PLAYING,         /**< Player is currently playing */
-    FLUID_PLAYER_DONE             /**< Player is finished playing */
+    FLUID_PLAYER_DONE             /**< Player is stopped (or finished playing) */
 };
 
 /**
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -1651,7 +1651,7 @@ new_fluid_player(fluid_synth_t *synth)
     }
 
     /* set the player in stopped state */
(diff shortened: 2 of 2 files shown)


derselbst (commented):
This would mean that `FLUID_PLAYER_READY` becomes unused. But the status is still exported to clients via `fluid_player_get_status` and they might depend on this state in some way.

jjceresa:
>This would mean that FLUID_PLAYER_READY becomes unused.

yes.

>But the status is still exported to clients via fluid_player_get_status and they might depend on this state in some way.

Yes, but as this FLUID_PLAYER_READY value has actually an incorrect meaning for client, perhaps this value should be deprecated ?. Also perhaps the value FLUID_PLAYER_DONE should be renamed FLUID_PLAYER_STOPPED which is more understandable ?.

derselbst:
> Also perhaps the value FLUID_PLAYER_DONE should be renamed FLUID_PLAYER_STOPPED which is more understandable ?.

Maybe. But this change would unnecessarily break client code. I would rather vote to improve the documentation.

> Yes, but as this FLUID_PLAYER_READY value has actually an incorrect meaning for client, perhaps this value should be deprecated ?.

Same. Also I think it's important to retain an enum that describes the "initial state" of the player, see my other PR.

jjceresa:
>Maybe. But this change would unnecessarily break client code. I would rather vote to improve the documentation.

Ok.

>Same. Also I think it's important to retain an enum that describes the "initial state" of the player. see my other PR.

Ok. Please which PR ?

jjceresa:
>Ok. Please which PR ?

I think that we can close this PR and take your solution instead.

jjceresa:
>I think that we can close this PR and take your solution instead.

I think that we can close https://github.com/FluidSynth/fluidsynth/pull/762 and take your solution instead.

derselbst:
Superseded by #765

Issue #757, opened 2021-01-28
add a "skip n bars" feature


juanlucasrey:
useful when doing rapid cycle of writing/listening using the command line only.

so for example 

>fluidsynth -ni file.sz2 generated_midi.mid --skip 8

to avoid listening to the first 8 bars again and again.

I can of course write and generate a second midi file skipping whatever bars I need, but I assume an option like that would not be too hard to add (I might be wrong)

derselbst:
What do you mean by "bar"?

jjceresa:
>I can of course write and generate a second midi file skipping whatever bars I need, but I assume an option like that would not be too hard to add (I might be wrong)

You probably expect that the player is able to handle bar:beat:unit pointer time isn't ?

juanlucasrey:
right midi probably does not have that information, but I am not sure. regardless any kind of measure that would be recognisable by the player, (like seconds for example) would be as useful.

I'm just trying to replicate the functionality of a full fledged GUI (like Sibelius or others) where I am able to play what I write starting form a certain point, and not the beginning every time.

jjceresa:
>right midi probably does not have that information, but I am not sure. regardless any kind of measure that would be recognisable by the player, (like seconds for example) would be as useful.

MIDI file specs provide time signature events, that allows any application to express time in bar:beat:unit which is the relative musical time any musician are familiar with. This kind of time is always relative to a quarter note regardless the absolute duration (i.e in second) of a quarter note.  Actually the simple fluidsynth player ignore time signature events. And this isn't easy to implement. However, the player know the tick time which is a fraction of a quarter note.
Using recent master you have a player command `player_step num` that allows to move the current position forward/backward. You should try it using fluidsynth application console.

juanlucasrey:
I might test once it is released. when would that be more or less?

juanlucasrey:
oh I see it is only available on the console. ok will test there. I guess then it would be a matter of adding an option on the terminal call driving that.

jjceresa:
>I'm just trying to replicate the functionality of a full fledged GUI.

Did you read the Wiki about the player API ?

juanlucasrey:
yes not fully but I am aware of it. I try to avoid though because my "iteration" involves another application. here is what I do after some writing, hopefully it's self explanatory:

abc2midi ./some_song.abc -o some_song.mid && fluidsynth -ni some_soung.sf2 ./some_song.mid

juanlucasrey:
[...]

PR #765, opened 2021-02-01
Fix player_seek command in user command file

### LINKED ISSUE #763: The MIDI file player starts playing immediately at soon it is created

derselbst:
@jjceresa I was more thinking of a solution like that, to fix the `player_step`. I can't think of a use-case to support `player_next` in the command file. Because if the user wanted to skip the first `n` files, he could have just left them out on the commandline in the first place.

P.S.: I would vote to rename `player_step` to `player_seek` to be more consistent with the API.

Addresses #757.
Resolves #763.

### DIFF
--- a/src/bindings/fluid_cmd.c
+++ b/src/bindings/fluid_cmd.c
@@ -372,11 +372,11 @@ static const fluid_cmd_t fluid_commands[] =
     },
     {
         "player_stop", "player", fluid_handle_player_stop,
-        "player_stop                Stop playing"
+        "player_stop                Stop playing (cannot be executed in a user command file)"
     },
     {
         "player_cont", "player", fluid_handle_player_continue,
-        "player_cont                Continue playing"
+        "player_cont                Continue playing (cannot be executed in a user command file)"
     },
     {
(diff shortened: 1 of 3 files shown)


jjceresa:
>I was more thinking of a solution like that, to fix the player_step

This solution works. We just need to document somewhere that player_next, and player_stop  won't work

>I would vote to rename player_step to player_seek to be more consistent with the API.

Good idea. The user just needs to keep in mind that the position argument +/-num  is relative to the current position.

jjceresa:
PS. Just for info. Using the command player_step -num may requires polyphony settings to be augmented as suggested by the warning messages:
fluidsynth: warning: Ringbuffer full, try increasing polyphony!

PR #773, opened 2021-02-07
Add Chris' soundcard enumeration tool

### LINKED ISSUE #761: Add device enumeration tool for new WASAPI driver
Integrate [Chris' device enumeration tool](https://github.com/FluidSynth/fluidsynth/pull/754#issuecomment-768345124) the the fluidsynth executable.

derselbst:
The fluidsynth executable now supports querying for supported sample-rates and sample-formats via the ` -D, --query-audio-devices` commandline options.

The tool itself relies on eating the log levels the WASAPI driver gave out. This is a bit suboptimal, but for now I think it's good enough. Alternatively one would have to place all the code in `fluid_wasapi.c` itself, which is also questionable and much more effort.

I'm open to a better naming of the option.

Resolves #761.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -379,7 +379,6 @@ if ( WASAPI_SUPPORT )
   set ( fluidsynth_SOURCES ${fluidsynth_SOURCES} fluid_wasapi_device_enumerate.c )
 endif ( WASAPI_SUPPORT )
 
-
 add_executable ( fluidsynth
     ${fluidsynth_SOURCES}
 )
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -375,6 +375,11 @@ target_link_libraries ( libfluidsynth
 
 set ( fluidsynth_SOURCES fluidsynth.c )
 
+if ( WASAPI_SUPPORT )
+  set ( fluidsynth_SOURCES ${fluidsynth_SOURCES} fluid_wasapi_device_enumerate.c )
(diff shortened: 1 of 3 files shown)


chirs241097:
While I find this option useful for certain cases, I do have a couple of suggestions:

1) To prevent license shenanigans like that one happened several years ago (sox effect processing code causing issues for Debian IIRC), maybe consider adding a license header to the file? I will license my original code (`fluid_wasapi_test.c` as in #754) under the terms of LGPLv2.1 or later (same as fluidsynth).
2) The two other Windows-specific drivers also supports device enumeration. For sake of consistency, maybe modify the code a little bit so that it can handle those drivers as well (enumerate devices handled by the driver specified by `-a`)?

I had extra thoughts on this option... but couldn't remember them off the top of my head. I'll post those here if I can recall them.

derselbst:
Thanks for your feedback.

1. I absolutely agree, will do that later.
2. I see no benefit in supporting the other drivers as well. The core idea of this feature is to assist the user in which modes are (natively) supported by the soundcard. This becomes important, when there is only one supported mode (like in my case), or when the user wants to find modes with least overhead for best performance. I don't think those results would differ when using different drivers.

chirs241097:
Well, that makes sense. I thought DirectSound sometimes cuts off names of sound cards, so it would be useful to add enumeration for that driver as well; but turns out `fluidsynth -o help` would do the job, so it's not really that necessary.

I've got no more issues with this PR to report then.

PR #774, opened 2021-02-07
wait for seek position if a seek is in progress

jjceresa:
When `fluid_player_get_current_tick()` is called immediately after fluid_player_seek(), the position returned by fluid_player_get_current_tick() is the `position` before calling fluid_player_seek().
This PR makes fluid_player_get_current_tick() the opportunity to return the new `seek position`.

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -2419,10 +2419,24 @@ fluid_player_join(fluid_player_t *player)
  * Get the number of tempo ticks passed.
  * @param player MIDI player instance
  * @return The number of tempo ticks passed
+ * Note: if a seek is in progress, the function waits for the seek position
+ *       before returning.
  * @since 1.1.7
  */
 int fluid_player_get_current_tick(fluid_player_t *player)
 {
+    int time_out = 0;
+
+    /* wait a limited time (500 ms max) if a seek is in progress */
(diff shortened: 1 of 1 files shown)


derselbst (changes requested):
Sorry, I disagree with this change:

1. If a client has made a successful call to `fluid_player_seek()` the client knows what the player's next position will be. I.e. there is no need to make another call to `fluid_player_get_current_tick()`.
1. `fluid_player_get_current_tick()` only returns accurate values when called from synth context. Otherwise those values might be dirty.
1. Imagine  a client has registered a custom callback via `fluid_player_set_playback_callback()`. If he would call `fluid_player_get_current_tick()` within this callback, your proposal would unnecessarily block the synth context for 500ms, when the client has made a simultaneous call to `fluid_player_seek()`.

jjceresa:
Ok, no problem. I understand all points. That means that the proper way for a client using the player is to keep track of the `seek position` passed to `fluid_player_seek()`. Than if the client wants to display the `player position` he must use this previous `seek position` if it exists otherwise the client must display the value returned by `fluid_player_get_current_tick()`.
An example of a such client is the function `fluid_handle_player_cde()` which must be fixed to display a correct `player position` when calling `player_print_position()`.

PR #780, opened 2021-02-23
Add optional per tick callback to player

arcln:
This pull request introduces an optional per-tick callback for the midi player. It solves the problem described in #777.

From a user point of view, it adds the `fluid_player_set_tick_callback()` function, that works in a very similar way than `fluid_player_set_playback_callback()`. 

The callback is then invoked from `fluid_player_callback()` every time `player->cur_ticks` changes.

### DIFF
--- a/include/fluidsynth/midi.h
+++ b/include/fluidsynth/midi.h
@@ -64,6 +64,32 @@ typedef int (*handle_midi_event_func_t)(void *data, fluid_midi_event_t *event);
  * @param data User defined data pointer
  * @param tick The current (zero-based) tick, which triggered the callback
  * @return Should return #FLUID_OK on success, #FLUID_FAILED otherwise
+ *
+ * This callback is fired at a constant rate depending on the current BPM and PPQ.
+ * e.g. for PPQ = 192 and BPM = 140 the callback is fired 192 * 140 times per minute (448/sec).
+ *
+ * It can be used to sync external elements with the beat,
+ * or stop / loop the song on a given tick.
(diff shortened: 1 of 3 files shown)


include/fluidsynth/midi.h
```diff
@@ -58,6 +58,15 @@ extern "C" {
...
 /* @} */
 
+/**
+ * Generic callback function fired once by MIDI tick change.
+ *
+ * @param data User defined data pointer
+ * @param tick The current (zero-based) tick, which triggered the callback
+ * @return Should return #FLUID_OK on success, #FLUID_FAILED otherwise
+ */
+typedef int (*handle_midi_tick_func_t)(void *data, int tick);
```
derselbst, line 68: Please elaborate a little bit on this documentation by giving an example, why/when this tick-callback could be useful. Doing might help users to better understand when to use which callback type.

derselbst (commented):
Ok, let's add another callback function, I don't mind.

include/fluidsynth/midi.h
derselbst: And it would avoid that some future day a maintainer would say: "Hey, what's that? We don't need that, strip it away." ;)

arcln: done :)

derselbst:
Actually, I found the code quite useful. So unless there's a technical error in it, I would take it (and proper `@code @endcode` doxygen markers)

derselbst:
Thanks!

Issue #781, opened 2021-02-23
Expose fluid_player_reset() API


arcln:
### Is your feature request related to a problem?
Using the player, I'd like to discard the current playing MIDI and start playing another without re-allocating a new fluid_player.

### Describe the solution you'd like
The `fluid_player_reset()` does exactly what I am looking for.

### Describe alternatives you've considered
Reallocating a player, but it is more complex and probably slower.

### Additional context
I PR'd this in #782.

arcln:
I had misunderstood the `fluid_player_reset()` and it doesn't actually clear the playlist.

Allocating another player is the only way to solve this for now.

PR #782, opened 2021-02-23
expose fluid_player_reset()

arcln:


derselbst (changes requested):
Making this function public is dangerous because it might delete tracks that are still being used by the synth callback. Returning FLUID_FAILED when the player is still playing doesn't help, because `fluid_player_stop` only sets an atomic int but doesn't wait until any pending synth callback has been done. A client application could simply do
```c
fluid_player_stop();
fluid_player_join();
fluid_player_reset();
```
from _outside_ the synth context and the application would crash whenever the synth callback is still running, due to use-after-free.

But even if the `fluid_player_reset` would be called from _within_ synth context (i.e. [from here](https://github.com/FluidSynth/fluidsynth/blob/04c30be0b95e9382332c05341d6434b56c1a87af/src/midi/fluid_midi.c#L1606)) the application would still crash when controlflow reaches the `fluid_track_next_event` a few lines below.

arcln:
How to solve #781 then? By creating another function like `fluid_player_clear()` or `fluid_player_delete_queue()` ?

derselbst:
One could introduce a new state `FLUID_PLAYER_STOPPING`, which will be set by `fluid_player_stop()`. If this state has been set, the synth callback would be responsible for transitioning to `FLUID_PLAYER_DONE` when it runs the next time.

`fluid_player_join` would still wait for `FLUID_PLAYER_DONE`, but now we can be sure that no synth callback will be messing around with tracks anymore after the join has returned.

And in `fluid_player_reset()` you could change the safety check to
`if state != FLUID_PLAYER_DONE && state != FLUID_PLAYER_READY then return FLUID_FAILED`. But the acutal _reset_ should be performed in a `fluid_player_reset_LOCAL()`. That way, `fluid_player_playlist_load()` can perform the reset while bypassing this safety check.

arcln:
I've got this working but it doesn't solve the problem because `fluid_player_reset()` does not actually clear the playlist.
The playlist is only cleared in `fluid_player_delete()`.

Considering all the problems it creates, I guess I'll just allocate another player to solve #781.

PR #783, opened 2021-02-25
fix race condition in fluid_player_callback

### LINKED ISSUE #784: Possible deadlock in fluid_player_join because of race condition
### FluidSynth version
git 04c30be (latest)

### Describe the bug
`fluid_player_join()` might block indefinitely or return without the player being actually stopped, leading to various crashes.

### Expected behavior
I expect to be able to `delete()` a player after `join()` has returned.

### Steps to reproduce
Consider the following order of execution:

-> the `fluid_player_callback()` already passed the status check at the beginning, but did not return yet
-> `stop()` sets the state to DONE from another thread
-> the callback finishes and sets the status back to PLAYING

### Additional context
Solved by #783

arcln:
closes #784

### DIFF
--- a/include/fluidsynth/midi.h
+++ b/include/fluidsynth/midi.h
@@ -209,6 +209,7 @@ enum fluid_player_status
 {
     FLUID_PLAYER_READY,           /**< Player is ready */
     FLUID_PLAYER_PLAYING,         /**< Player is currently playing */
+    FLUID_PLAYER_STOPPING,        /**< Player is stopping, but hasn't finished yet */
     FLUID_PLAYER_DONE             /**< Player is finished playing */
 };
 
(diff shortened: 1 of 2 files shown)


arcln:
The failing test does not uses the player code, so I'll assume this PR is not responsible for the failing checks.

derselbst (approved):
LGTM. I'm just thinking whether this should be applied on the `2.1.x` branch so that it becomes part of `2.1.8`. It's a bug, yes, but it exists for so long and nobody has complained... Is it worth to introduce a new player state for an EOL release?

arcln:
Considering it creates a (very little) user facing change, I think it's fine to wait for 2.2.

Issue #788, opened 2021-03-04
Audio driver breaks on cho_set_nr 100


derselbst:
### FluidSynth version
2.2 b1 and recent master

### Describe the bug
Commanding `cho_set_nr 100` makes audio go weird (silence on linux, hangs on windows)

### Expected behavior
Reporting out-of-range error

PR #789, opened 2021-03-04
fix return in fluid_synth_chorus_set_param(), fluid_synth_reverb_set_param()

jjceresa:
This PR should fix issue https://github.com/FluidSynth/fluidsynth/issues/788


derselbst:
Yes, it does. Thanks!

Issue #787, opened 2021-03-04
Failed to early-execute command configuration file


derselbst:
Recent master keeps complaining that it could not execute non-existing command files. Annoying, fix that.

PR #792, opened 2021-03-04
add fluid_fexist() function

jjceresa:
This should fix issue https://github.com/FluidSynth/fluidsynth/issues/787


derselbst (changes requested):
No, sorry. `fopen` may fail for various reasons. Also, changing the behaviour of the public functions `fluid_get_sysconf` and `fluid_get_userconf` is not appropriate. I've addressed this issue in #794.

jjceresa:
>I've addressed this issue in #794.

Ok

PR #795, opened 2021-03-06
Improvements to LADSPA subsystem

### LINKED ISSUE #793: Increase maximum number of LADSPA nodes
Following up on https://lists.nongnu.org/archive/html/fluid-dev/2021-03/msg00000.html

The maximum number of nodes in the LADSPA system is currently hard-wired to `100`. When using multi-channel outputs and/or many effects or buffers, this limit can be easily reached.

A quick solution would be to simply increase FLUID_LADSPA_MAX_NODES to a higher number, like 500 or 1000. That would increase static memory consumption of FluidSynth by `400 * sizeof(void*)` or `900 * sizeof(void*)`.

Alternatively, we could change the way the nodes are being allocated and move to dynamic allocation and linked lists instead. I would be fine either way, leaning a bit towards simply increasing the limit. Any other thoughts?

mawe42:
This PR fixes #793 by removing the static limit on the number of nodes and effects by using dynamic allocation / linked lists instead. It also contains some cleanup work to make the LADSPA code (hopefully) easier to understand.

I've tested it locally and it seems to work as intended.

### DIFF
--- a/src/bindings/fluid_ladspa.c
+++ b/src/bindings/fluid_ladspa.c
@@ -108,13 +108,9 @@ struct _fluid_ladspa_fx_t
     /* The buffer size for all audio buffers (in samples) */
     int buffer_size;
 
-    fluid_list_t *nodes;
-
-    /* Pointers to host nodes in nodes list */
     fluid_list_t *host_nodes;
 
-    /* Pointers to user audio nodes in nodes list */
-    fluid_list_t *audio_nodes;
+    fluid_list_t *user_nodes;
 
     fluid_list_t *effects;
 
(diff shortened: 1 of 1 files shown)


src/bindings/fluid_ladspa.c
```diff
@@ -240,15 +231,18 @@ fluid_ladspa_fx_t *new_fluid_ladspa_fx(fluid_real_t sample_rate, int buffer_size
...
 
-    /* clear the remaining input or output nodes */
-    for(i = 0; i < fx->num_nodes; i++)
+    /* Delete the remaining host nodes */
+    for(list = fx->host_nodes; list; list = fluid_list_next(list))
     {
-        delete_fluid_ladspa_node(fx->nodes[i]);
+        node = (fluid_ladspa_node_t *) fluid_list_get(list);
+        delete_fluid_ladspa_node(node);
     }
```
derselbst, line 246: I think you're missing a call to `delete_fluid_list(fx->host_nodes);` here.

derselbst (commented):
LGTM, only one issue :)

src/bindings/fluid_ladspa.c
mawe42: Good catch, thanks! Fixed with latest commit.

derselbst (approved):
Perfect, thanks!

PR #796, opened 2021-03-07
Update URLs to HTTPS

TotalCaesar659:
P.S. Please, update this URL too.


derselbst (commented):
There are more URLs in other files that could be updated as well. Could you do this, while you're on it?

TotalCaesar659:
Updated more URLs (except for PDF and similar files).

derselbst (approved):
Perfect, thank you!

PR #804, opened 2021-03-13
Rename generated table files from .c to .inc.h

dreamer:
These two files are not ordinary C files, and are being compiled
through #include's in other C source files, not through invoking
compiler on generated files. This might confuse both developers and
automated systems.

For longer explanation see discussion in #800.

### DIFF
--- a/src/gentables/make_tables.c
+++ b/src/gentables/make_tables.c
@@ -72,11 +72,11 @@ int main (int argc, char *argv[])
     if (argc < 2)
         return -1;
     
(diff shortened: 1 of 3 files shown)


derselbst:
I keep merging `2.1.x` into master, so I don't need this. But this way we at least see that CI is green on master :) 

P.S.: 2.1.8 should be ready by tomorrow.

dreamer:
@derselbst Superb, thanks! :)

PR #805, opened 2021-03-13
Rename generated table files from .c to .inc.h

dreamer:
These two files are not ordinary C files, and are being compiled
through #include's in other C source files, not through invoking
compiler on generated files. This might confuse both developers and
automated systems.

For longer explanation see discussion in #800.


dreamer:
This is a backport of #804 targetting branch 2.1.x.

derselbst (approved):
Ok, Thanks!

PR #810, opened 2021-03-14
Invalid generators were not removed from zone list

### LINKED ISSUE #808: fluidsynth crashes when loading malformed sf2 file
version: master(v.2.1.1), ubuntu18.04(v1.1.9), ubuntu20.04(v.2.1.1), ....

https://github.com/FluidSynth/fluidsynth/blob/master/src/sfloader/fluid_sffile.c#L1952
It says Gen_SampleId is the last gen, and then set `level` to 3 and break.
```c
else if(genid == Gen_SampleId)
{
    /* sample is last gen */
    level = 3;
    READW(sf, genval.uword);
    ((SFZone *)(p2->data))->instsamp = FLUID_INT_TO_POINTER(genval.uword + 1);
    break; /* break out of generator loop */
}
```

but if a malformed sf2 doesn't contain Gen_SampleId, it will finally goto here:
https://github.com/FluidSynth/fluidsynth/blob/master/src/sfloader/fluid_sffile.c#L2041

```c
if(level == 3) // level == 2 because no Gen_SampleId found
{
    SLADVREM(z->gen, p3);    /* zone has sample? */
}
else // enter here
{
    /* its a global zone */
    if(!gzone) // not the first time enter here, so gzone == TRUE, skip
    {
        gzone = TRUE;

        /* if global zone is not 1st zone, relocate */
        if(*hz != p2)
        {
            void *save = p2->data;
            FLUID_LOG(FLUID_WARN, "Instrument '%s': Global zone is not first zone",
                      ((SFPreset *)(p->data))->name);
            SLADVREM(*hz, p2);
            *hz = fluid_list_prepend(*hz, save);
            continue;
        }
    }
    else // finally enter here and free it, but actually it's gzone. so it will cause an use-after-free later ...
    {
        /* previous global zone exists, discard */
        FLUID_LOG(FLUID_WARN, "Instrument '%s': Discarding invalid global zone",
                  ((SFInst *)(p->data))->name);
        *hz = fluid_list_remove(*hz, p2->data);
        delete_zone((SFZone *)fluid_list_get(p2));
    }
}
```

it will be freed again at: https://github.com/FluidSynth/fluidsynth/blob/master/src/sfloader/fluid_sffile.c#L2293
```c
entry = zone->gen;

while(entry)
{
    FLUID_FREE(fluid_list_get(entry)); // free at here
    entry = fluid_list_next(entry);
}
```
fluid_synth_sfload() -> fluid_defsfloader_load() -> fluid_defsfont_load() -> fluid_sffile_close() -> delete_inst() -> delete_zone()

------

**HERE** is an example that trigger this vuln:  [**vuln.zip**](https://github.com/FluidSynth/fluidsynth/files/6136193/vuln.zip)

image

image

Programs like VLC that use this library are affected by this vulnerability:

image

derselbst:
`fluid_list_remove()` should receive the beginning of a list, so it can adjust the predecessor of the element to be removed. Otherwise, the element would remain in the list, which in this case led to a use-after-free afterwards.

Resolves #808

### DIFF
--- a/src/sfloader/fluid_sffile.c
+++ b/src/sfloader/fluid_sffile.c
@@ -1355,7 +1355,7 @@ static int load_pmod(SFData *sf, int size)
  * ------------------------------------------------------------------- */
 static int load_pgen(SFData *sf, int size)
 {
-    fluid_list_t *p, *p2, *p3, *dup, **hz = NULL;
+    fluid_list_t *p, *p2, *p3, *dup, **hz = NULL, *start_of_zone_list;
     SFZone *z;
     SFGen *g;
     SFGenAmount genval;
@@ -1369,7 +1369,7 @@ static int load_pgen(SFData *sf, int size)
         /* traverse through all presets */
         gzone = FALSE;
         discarded = FALSE;
-        p2 = ((SFPreset *)(p->data))->zone;
+        start_of_zone_list = p2 = ((SFPreset *)(p->data))->zone;
 
         if(p2)
         {
@@ -1516,11 +1516,12 @@ static int load_pgen(SFData *sf, int size)
                 }
                 else
                 {
+                    SFZone * pzone = fluid_list_get(p2);
                     /* previous global zone exists, discard */
                     FLUID_LOG(FLUID_WARN, "Preset '%s': Discarding invalid global zone",
                               ((SFPreset *)(p->data))->name);
-                    *hz = fluid_list_remove(*hz, p2->data);
-                    delete_zone((SFZone *)fluid_list_get(p2));
+                    *hz = fluid_list_remove(start_of_zone_list, pzone);
+                    delete_zone(pzone);
                 }
             }
 
@@ -1864,7 +1865,7 @@ static int load_imod(SFData *sf, int size)
 /* load instrument generators (see load_pgen for loading rules) */
 static int load_igen(SFData *sf, int size)
 {
-    fluid_list_t *p, *p2, *p3, *dup, **hz = NULL;
+    fluid_list_t *p, *p2, *p3, *dup, **hz = NULL, *start_of_zone_list;
     SFZone *z;
     SFGen *g;
     SFGenAmount genval;
@@ -1878,7 +1879,7 @@ static int load_igen(SFData *sf, int size)
         /* traverse through all instruments */
         gzone = FALSE;
         discarded = FALSE;
-        p2 = ((SFInst *)(p->data))->zone;
+        start_of_zone_list = p2 = ((SFInst *)(p->data))->zone;
 
         if(p2)
         {
@@ -2024,11 +2025,12 @@ static int load_igen(SFData *sf, int size)
                 }
                 else
                 {
+                    SFZone * izone = fluid_list_get(p2);
                     /* previous global zone exists, discard */
                     FLUID_LOG(FLUID_WARN, "Instrument '%s': Discarding invalid global zone",
                               ((SFInst *)(p->data))->name);
-                    *hz = fluid_list_remove(*hz, p2->data);
-                    delete_zone((SFZone *)fluid_list_get(p2));
+                    *hz = fluid_list_remove(start_of_zone_list, izone);
+                    delete_zone(izone);
                 }
             }
 
--- a/src/sfloader/fluid_sffile.c
+++ b/src/sfloader/fluid_sffile.c
@@ -1516,12 +1516,13 @@ static int load_pgen(SFData *sf, int size)
                 }
                 else
                 {
-                    SFZone * pzone = fluid_list_get(p2);
+                    p2 = fluid_list_next(p2); /* advance to next zone before deleting the current list element */
                     /* previous global zone exists, discard */
                     FLUID_LOG(FLUID_WARN, "Preset '%s': Discarding invalid global zone",
                               ((SFPreset *)(p->data))->name);
-                    *hz = fluid_list_remove(start_of_zone_list, pzone);
-                    delete_zone(pzone);
+                    fluid_list_remove(start_of_zone_list, z);
+                    delete_zone(z);
+                    continue;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


derselbst:
@veritas501 You may give it a test.

veritas501:
Okay. It seems that the old bug has been fixed. I'm still testing if there are more bugs. @derselbst

mawe42:
I must admit I don't understand why this fixes the bug. `*hz` is a pointer to the first element in the zone list. So at the start it's effective identical to the new `start_of_zone_list`. And `*hz` only gets changed if we encounter a global zone that is not first in list. In this case, the global zone is prepended to `*hz` and `*hz` is updated accordingly.

derselbst:
> *hz is a pointer to the first element in the zone list.

No, `hz` is a pointer to pointer. `*hz` is (usually) equal to `p2`. So, it would have been like putting
 `fluid_list_remove(p2, p2->data);`

mawe42:
Ah, of course! Hm... then I have a hard time understanding the purpose of `hz` in the first place, and when it gets updated. I guess I need more time going through this again. That code reads a little like it's been deliberately obfuscated :-) I once had the impression that I understand what's going on, but that was a long time ago...

derselbst:
The purpose of `hz` is not quite clear to me as well. It could be a leftover when they passed `hz` to a separate function that removed the zone:

https://github.com/FluidSynth/fluidsynth/blob/62e375c71c815ca12f859b8722db5dff12f0c897/fluidsynth/src/sfloader/fluid_defsfont.c#L2443

The core logic around it already exists since "Initial Revision", so I'm a bit cautious to touch it. It would probably be a good idea to write a bunch of test cases, before changing it.

mawe42:
Looking into `load_pgen()` some more, I think the logic behind that `hz` pointer is completely broken... and not just in the place where this PR introduces a fix. It's purpose is to be able to move a global zones that is not the first zone in a preset to the beginning of the zone list. But as `*hz` points to the local `p2` variable (i.e. the current entry in the zone list), the `preset->zone` list is not updated correctly.

mawe42:
I don't see how the global zone relocation code will ever have run. As far as I understand it, the following check will always be false, because `*hz` and `p2` point to the same memory location:
https://github.com/FluidSynth/fluidsynth/blob/b8fb6c81e1ca27c0bba2f6a0168832214f91d497/src/sfloader/fluid_sffile.c#L1517

And looking back through the history of this code, I arrive at the initial commit which has the same behaviour. So I guess we could either remove the relocation code altogether, or fix try to fix it.

derselbst:
> It's purpose is to be able to move a global zones that is not the first zone in a preset to the beginning of the zone list.

Yes, I also had the feeling it has to do with moving around zones. I also don't see how `(*hz != p2)` can ever evaluate to true.

> And looking back through the history of this code, I arrive at the initial commit which has the same behaviour

I arrived at iiwusynth: https://cvs.savannah.nongnu.org/viewvc/iiwusynth/iiwusynth/src/iiwu_defsfont.c?revision=1.1&view=markup

We could ask Peter, I'm sure he will remeber what he did 19 years ago :D

> So I guess we could either remove the relocation code altogether, or fix try to fix it.

I think we should fix it. Ideally by creating test cases. But this will take time. I would prefer to merge this and release 2.1.8 afterwards (even if it's not perfectly correct).

src/sfloader/fluid_sffile.c
```diff
@@ -1516,11 +1516,12 @@ static int load_pgen(SFData *sf, int size)
                 }
                 else
                 {
+                    SFZone * pzone = fluid_list_get(p2);
                     /* previous global zone exists, discard */
                     FLUID_LOG(FLUID_WARN, "Preset '%s': Discarding invalid global zone",
                               ((SFPreset *)(p->data))->name);
-                    *hz = fluid_list_remove(*hz, p2->data);
```
mawe42, line 1522: I don't think this fix introduces unintended behaviour. As `*hz` is an alias for `p2`, you this line updates `p2` back to preset->zone, effectively starting the whole zone iteration from the beginning. The old code did not have that effect.

src/sfloader/fluid_sffile.c
```diff
@@ -1516,11 +1516,12 @@ static int load_pgen(SFData *sf, int size)
...
-                    delete_zone((SFZone *)fluid_list_get(p2));
+                    *hz = fluid_list_remove(start_of_zone_list, pzone);
```
mawe42, line 1523: I meant this line with the previous comment. Got confused with the side-by-side diff view...

mawe42: I think simply removing the assignment to `*hz` would be ok here. Same in `load_iggen()`.

derselbst: You're right. I think it should have been

```suggestion
                    start_of_zone = fluid_list_remove(start_of_zone_list, pzone);
```

And then we have to save `p2->next`, advance `p2`, and `continue` the loop. Also, `p3` in the `while(p3)` statement below could still point to the zone we've just deleted, couldn't it?

derselbst: Nope, we cannot remove the assignment. We need it in case we are dealing with the first zone to adjust the beginning of the list.

mawe42: We will never deal with the first zone here, because the prerequisite ist that there is a previous global zone.

derselbst: Ofc, sry.

mawe42: But you are right, need also need to continue here. So we should be safe to do the following:
```
fluid_list_remove(start_of_zone_list, pzone);
delete_zone(pzone);
p2 = fluid_list_next(p2);
continue; /* zone iteration */
```

mawe42:
Yes, or like that :-) That should do it, at least until we cleanup this code properly.

derselbst:
Great! Then I will complete this tomorrow, to give @veritas501  the chance to report back.

PR #812, opened 2021-03-14
Parallelize SF2 loading

derselbst:
The loading of SF2 samples can be parallelized as well, at least for bigger soundfonts like the Stgiga 4GB monster.

The following test is performed with hot-caches:

`time src/fluidsynth -i -a alsa Stgiga\'s\ HiDef\ Soundfont\ \(2019-05-25\).sf2`

Serial version:
```
real    0m15,460s
user    0m14,163s
sys     0m1,192s
```

Parallelized version proposed by this PR:
```
real    0m5,851s
user    0m14,089s
sys     0m1,194s
```

### DIFF
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -422,11 +422,14 @@ int fluid_defsfont_load_all_sampledata(fluid_defsfont_t *defsfont, SFData *sfdat
         }
         else
         {
-            /* Data pointers of SF2 samples point to large sample data block loaded above */
-            sample->data = defsfont->sampledata;
-            sample->data24 = defsfont->sample24data;
-            fluid_sample_sanitize_loop(sample, defsfont->samplesize);
-            fluid_voice_optimize_sample(sample);
+            #pragma omp task firstprivate(sample, defsfont) default(none)
+            {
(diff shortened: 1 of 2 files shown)


mawe42:
Looks good, and the speedup is great!

I wonder if we could clean up the `fluid_defsfont_load_all_sampledata` function at the same time. Currently there are two places where we differentiate between sf2 and sf3... and one is even in the loop. Maybe could we make the function consist of two large blocks, one for sf2 and the other for sf3 (both with their own loop over the samples). Or maybe even clearer, use two separate funcions, `fluid_defsfont_load_all_sf2_sampledata` and `fluid_defsfont_load_all_sf3_sampledata`?

derselbst:
Not sure. Yes, one could out-source the logic into two separate functions. Then one would have two loops and two openMP parallel sections. I would tend to keep it as is.

mawe42:
It's not a huge deal, just an idea that might make the code easier to read. Files are not going to change their type from SF2 to SF3 during load, so I think having two separate loops for this case would be clearer. But I do admit that I haven't got a clue about openMP, so I don't know what the downsides of two parallel sections would be.

derselbst:
Ok, let's take it as it is.

Issue #821, opened 2021-03-27
SoundFont loader does not properly validate preset generators


mawe42:
The default SoundFont loader attempts to validate the generators in preset zones against a list of invalid generators in [valid_preset_genid()](https://github.com/FluidSynth/fluidsynth/blob/661a06b5045b7004c3f6f8d7c5c5aea6e94e6440/src/sfloader/fluid_sffile.c#L2393). The list of invalid preset generators has a sentinel value of 0 which marks the end of the list. The problem is that the first entry in this list is the generator `Gen_StartAddrOfs`, which has the id 0. This effectively prevents genid checks against any other entries in this list.

Affects at least current and recent versions of FluidSynth, but probably much older versions as well.

I'm currently preparing a code cleanup branch for fluid_sffile, so I would address this issue in there if you don't mind.

jjceresa:
>The list of invalid preset generators has a sentinel value of 0

Please note that the list of invalid inst generator `invalid_inst_gen[]` has also 0 which marks the end of the list. I am aware that this is not an issue for `valid_inst_genid()`. If you have in mind to change the end of list marker value for `invalid_preset_gen[]`, then for consistency  perhaps it worth to change this for  `invalid_inst_gen[]` as well ?.

mawe42:
Yes, good point. And I've already made both list behaviours consistent and without sentinal. Will push a refactor-branch tonight or tomorrow, then you can review those proposed changes.

Issue #813, opened 2021-03-14
Fix preset/instrument zone validation logic


derselbst:
### FluidSynth version
_any_ version back to iiwusynth

### Describe the bug
This is a follow-up task to #810. It was discovered that the logic in `load_pgen()` and `load_igen()` around `*hz` is broken.

Implement a bunch of test cases that validate the current behaviour and afterwards fix the zone moving logic in both functions.

mawe42:
As preparation I'm currently preparing a PR that simply renames some variables. I think all those `p3`, `p2` and `p` vars make the code really hard to understand and reason about.

derselbst:
I think it's ok if we fix this after 2.2.0. For now, I would like to publish a release candidate, so that we can release 2.2.0 in early April.

mawe42:
Yes, I think so as well. I've experimented with file format fuzzing using [AFL](https://github.com/google/AFL). Before the fix in #810, it quickly identified some crashes. With current master, it's now at > 5 million permutations and hasn't found a crash or timeout so far. I'll leave it running at bit longer, but I think at least our loader code is quite robust against broken soundfonts now.

derselbst:
Interesting, haven't heard of AFL before.

In the meantime, I've started implementing a test on [this branch](https://github.com/FluidSynth/fluidsynth/tree/zone-validation-test). The idea is to set up a rough structure of a `SFData` (e.g. one preset, one instrument, each having some zones with some generators assigned), and then we use the file-callbacks to read from a custom buffer that contains the malformed (or correct) data.

mawe42:
Looks good, although I wonder if it would be better to test this on a higher level. Maybe we could create some fairly simple sf2 files with and without defects, then check that the loaded structure of the file is correct. That might make the test more "future proof", and would test even more code paths at the same time.

derselbst:
So, you vote for an integration test. This probably would be more future proof. But also harder to create. And harder to debug in case something breaks.

I had in mind to concentrate on this validation logic and create as many tests as necessary to cover any branch in two functions at least once. This will be already complicated enough. Setting up an entire SF2 file for all those cases... oh dear, I don't know.

mawe42:
I had something like this in mind:
 - Create utility functions to turn a `fluid_defsfont_t` into a text representation, maybe best to use XML (using libxml2) or another well-defined format. So this would dump all relevant info about samples, presets, zones in presets, instruments and zones in instruments to this representation.
[...]

PR #823, opened 2021-03-27
Refactor of sffile and defsfont code, including some bugfixes

mawe42:
This PR makes a few refactoring changes to the the fluid_sffile and fluid_defsfont code and also includes fixes for #821 and #813.

The most important refactorings are:
 - avoid redefining an enum for the generator IDs and use the enum from gen.h instead
 - get rid of the hack around the `SFZone->inststamp` pointer by treating the GEN_INSTRUMENT and GEN_SAMPLEID generators like any other generators and resolving the indices in the defsfont import functions instead

I developed it on top of the defsfont-integration-test branch to make sure it doesn't introduce new behaviour. So to test this PR, I would suggest to merge it locally with defsfont-integration-test. I could also rebase it onto that branch, if it would help with review.

With the integration test, I'm fairly certain that this PR doesn't introduce any unwanted effects or new behaviour, but as it's touching such a core part of FluidSynth I would really welcome a very thorough review from both of you. Both about the code changes themselves, but also about the refactorings as a whole. As they touch very old and often used code code paths, the risk of breaking stuff is obviously quite high, even with the integration test as a safety net.


jjceresa:
Looking after `Fix preset generator validity checks` c4d38a7: look  good.
Note: We have duplicate code in valid_inst_genid() and valid_preset_genid().

jjceresa:
Suggestion: in 16d2f43 struct _SFPhdr, perhaps could we add a little comment to indicate that library, genre and morphology aren't not used ?.

jjceresa:
About change in `Remove instsamp hack` dfbef11, I guess this will not augment loading time ?

mawe42:
> Note: We have duplicate code in valid_inst_genid() and valid_preset_genid().

I don't see duplicate code in those two functions. What do you mean exactly?

> Suggestion: in 16d2f43 struct _SFPhdr, perhaps could we add a little comment to indicate that library, genre and morphology aren't not used ?

We could do. I'm just wondering what the purpose would be. There is a comment in sffile where we skip those thields, but why would we (or somebody working on FS in the future) need an additional comment in SFPhdr? i think the statement in the SF2 specs about "preserving" those three values means that software that loads and then writes SF2 files should keep the content of those fields untouched. But we will never write SF2 files...

> About change in Remove instsamp hack dfbef11, I guess this will not augment loading time ?

It does increase loading time a tiny bit. Here are the timings for loading 10 instances of the FluidR3 soundfont on my machine:
```
master branch:
real	0m0,740s
user	0m1,064s
sys	0m0,144s

refactor branch:
real	0m0,761s
user	0m1,088s
sys	0m0,134s
```

If we think that slowdown is too much, then I think we could easily find some places for optimiziation. Like using `fluid_list_prepend` instead of `fluid_list_append` for the sample lists in sffile and defsfont. With that change, the timings are:
```
real	0m0,668s
user	0m0,978s
sys	0m0,154s
```
Or, if we didn't want to change the order or samples in the list, we could simply keep track of the added last_list element when appending to lists in sffile, reducing the O(n) complexity everywhere to O(1). That would probably give an even greater speedup.

mawe42:
> Or, if we didn't want to change the order or samples in the list

Wrong thinking on my part... using fluid_list_prepend in both places preserves the original order, of course. So that is an easy optimisation without side-effects. I've added the change to this PR.

jjceresa:
>I don't see duplicate code in those two functions.

Please don't mind about that. Ignore my comment.

>We could do. I'm just wondering what the purpose would be. There is a comment in sffile where we skip those fields

Ah yes. That is sufficient. Please don't mind.

>It does increase loading time a tiny bit.

This augmentation seems negligible. Thanks for all the cleanup.

derselbst (commented):
Looks pretty good to me as well. I am getting a bit cold feet when looking at the changes in `load_pgen` and `load_igen` though. So I would suggest the following procedure:

1. Release 2.2.0 on easter
2. Get Marcus' integration test ready
3. Get my `zone-validation-test` ready
4. Test this PR against both tests
4. Merge this PR

mawe42:
Sounds good to me. But I would add another point to the list:
 - Run fuzzing tests with AFL on this pR (which I have done already, but better be safe than sorry)

In case you haven't written the zone unit tests yet: I do have some small (1.5k) soundfonts created during the fuzzing that trigger the invalid zone handling. So we could simply use those as test cases and check them against the expected defsfont state with the dump_sfont tool.

derselbst:
> In case you haven't written the zone unit tests yet

I have a few tests, also the broken soundfont from 808 has a test case. It's quite a lot copy pasted. Unfortunately I don't see how to make it much better without loosing too much flexibility while putting in too much effort.

derselbst:
When the integration test is ready, I would merge master back in and resolve conflicts.

derselbst (approved):
Great! As expected, the only detected behaviour change is that `GEN_INSTRUMENT` is no longer allowed in `load_igen`. Once CI is green, we can merge this as well.

mawe42:
> As expected, the only detected behaviour change is that GEN_INSTRUMENT is no longer allowed in load_igen.

Yeah... I wondered if that is actually a good thing to add. Yet another SoundFont validity check that FluidSynth doesn't need (because it can cope with this defect). But maybe it will help to improve other SoundFont implementations out there... ;-)

derselbst:
> > As expected, the only detected behaviour change is that GEN_INSTRUMENT is no longer allowed in load_igen.
> 
> Yeah... I wondered if that is actually a good thing to add.

I'm very confident that it won't hurt. Whereas fluidsynth previously added a useless instrument generator it now simply skips that generator. Loading is not affected. The user doesn't even see a message, that it had been skipped. That's just fine. It's the correct behaviour.

mawe42:
> The user doesn't even see a message, that it had been skipped.

If a generator is skipped, then we warn that "Some invalid generators were discarded". But of course you are right... it's only a warning, not a hard failure.

derselbst:
Ah, sry, I was only looking at the unit test before and afterwards and didn't recognize it.

mawe42:
> Once CI is green, we can merge this as well.

Looks good, so I would merge this in the next hour (unless you shout stop before :-)

PR #825, opened 2021-04-03
Collect coverage during test execution

derselbst:
Use `gcovr` to collect line and branch coverage information during test execution. Executing `make coverage` will run unit tests and generate a HTML coverage report and if possible also a generic SonarQube report recognizable by SonarCloud. Currently, this is only possible when cmake was configured with `-Denable-coverage=1` and if GCC is used as compiler.


mawe42:
That's great! Having code coverage info for our tests will be very helpful!

Issue #803, opened 2021-03-12
Failed to create the audio driver


jedker999:
Im using macOS big sur and have installed fluidsynth v 2.1.7. When I want to start it I get an error:

fluidsynth: error: Error setting the audio callback. Status=-66748
Failed to create the audio driver

any ideas how to fix it?

derselbst:
Same issue as https://lists.nongnu.org/archive/html/fluid-dev/2021-02/msg00003.html

No idea how to fix it. Pls. verify that when using the portaudio driver `fluidsynth -a portaudio` it fails as well. If it does fail, it's not a fluidsynth issue.

jedker999:
> Same issue as https://lists.nongnu.org/archive/html/fluid-dev/2021-02/msg00003.html
> 
> No idea how to fix it. Pls. verify that when using the portaudio driver `fluidsynth -a portaudio` it fails as well. If it does fail, it's not a fluidsynth issue.

unfortunately throws error as well:
||PaMacCore (AUHAL)|| Error on line 1316: err='-66748', msg=Unknown Error
fluidsynth: error: Error opening PortAudio stream: Internal PortAudio error
Failed to create the audio driver

derselbst:
This error is printed when the call to `AudioUnitSetProperty` fails. According to www.osstatus.com the returned error is `kAudioComponentErr_NotPermitted`. You'll probably be more lucky trying to reach out in a Mac Forum.

jedker999:
> This error is printed when the call to `AudioUnitSetProperty` fails. According to [www.osstatus.com](http://www.osstatus.com) the returned error is `kAudioComponentErr_NotPermitted`. You'll probably be more lucky trying to reach out in a Mac Forum.

Thanks for response.

giona69:
@jedker999 any luck on solving this problem? I have it too ...

mawe42:
Just an idea (coming from somebody who doesn't use Macs): According to [this](https://github.com/bruce0505/ios7frameworks/blob/master/AudioUnit/AUComponent.h#L790), the `kAudioComponentErr_NotPermitted` error might point to insufficient privileges: 
> @constant   kAudioComponentErr_NotPermitted
>      app needs "inter-app-audio" entitlement or host app needs "audio" in its UIBackgroundModes. Or app is trying to register a component not declared in its Info.plist.

Could you try giving fluidsynth more permissions under System Preferences / Security & Privacy / Privacy? The following might be interesting:
 - Microphone
[...]

PR #828, opened 2021-04-07
Fix failure to create the audio driver on macOS

ringoz:
This PR addresses issues #803
The fix is straightforward, just replace the deprecated API usages.


derselbst:
I'll backport this to 2.1.x later.

nmestanza:
Hello,

How Can I install the version that fixes this issue on MACOS?

PR #839, opened 2021-04-10
WASAPI driver reorganization

pedrolcl:
Avoid initializing COM in the caller's thread context.

See also: #833

### DIFF
--- a/src/drivers/fluid_wasapi.c
+++ b/src/drivers/fluid_wasapi.c
@@ -102,6 +102,13 @@ static const IID   _IID_IAudioRenderClient =
 
 #define FLUID_WASAPI_MAX_OUTPUTS 1
 
+/* Please do not change the order of the HANDLE array */
+#define FLUID_WASAPI_START 0
+#define FLUID_WASAPI_THREAD 1
+#define FLUID_WASAPI_QUIT 2
+#define FLUID_WASAPI_LAST_HANDLE FLUID_WASAPI_QUIT
+#define FLUID_WASAPI_MAX_HANDLES FLUID_WASAPI_LAST_HANDLE + 1
+
 typedef void(*fluid_wasapi_devenum_callback_t)(IMMDevice *, void *);
 
 static DWORD WINAPI fluid_wasapi_audio_run(void *p);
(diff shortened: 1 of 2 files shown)


src/drivers/fluid_wasapi.c
```diff
@@ -672,7 +680,15 @@ static void fluid_wasapi_foreach_device(fluid_wasapi_devenum_callback_t callback
         IMMDevice_Release(dev);
     }
 
-cleanup:
+
+cleanup1:
+
+    if (!com_was_initialized)
+    {
+        CoUninitialize();
```
derselbst, line 688: `CoUninitialize` must be called at the very end. When you call it here, other COM objects are still alive, which causes a segfault.

derselbst (commented):
Ok, this looks pretty good already. Thank you! I have two points:

1. The unit tests are failing. For a possible reason, see below.
2. `new_fluid_wasapi_audio_driver2` should wait until the synth thread has finished basic initialization, before returning the audio driver to the caller. On my laptop, WASAPI always fails because the default sample rate 44.1 khz is not supported by my soundcard. Previously, fluidsynth was exiting correctly with error. Now, it just stays open doing nothing. So, if the synth thread fails _before_ it enters the `for(;;)` loop, `new_fluid_wasapi_audio_driver2` should return NULL.

src/drivers/fluid_wasapi.c
```diff
@@ -181,60 +176,182 @@ fluid_audio_driver_t *new_fluid_wasapi_audio_driver2(fluid_settings_t *settings,
...
+
+    dev->thread = CreateThread(NULL, 0, fluid_wasapi_audio_run, (void *)dev, 0, &dev->thread_id);
+
+    if(dev->thread == NULL)
+    {
+        FLUID_LOG(FLUID_ERR, "wasapi: failed to create audio thread.");
+        goto cleanup;
+    }
+
+    if (WaitForMultipleObjects(2, &dev->start_ev, FALSE, 200) == WAIT_OBJECT_0) {
```
derselbst, line 252: Please do not rely on the order of struct members. If we add a member after `start_ev` this will horribly break. Pls. use an explicit array. Also, I'd vote to increase the timeout to 1000 msec. Ideally, you could add a
`FLUID_LOG(FLUID_WARN, "wasapi: initialization timeout!")` if `WAIT_TIMEOUT` was returned.

derselbst (commented):
I've tested it and it works as expected. Thanks! I have one point though:

derselbst (approved):
I made some tiny changes. Sorry for the confusion. I'll wait for the CI to complete and then I'll take it. Thanks!

PR #843, opened 2021-04-10
Auto select audio driver

mawe42:
As mentioned in #838, this PR adds auto-selection of audio and MIDI drivers.

A new default value of "auto" is used for `midi.driver` and `audio.driver`. If a driver is set to "auto", then all available drivers are tried in the order that they appear in the `fluid_audio_drivers` and `fluid_midi_drivers` lists.

In order prevent flooding the user with lots of error messages during the probing, a new `int flags` parameter was added to the MIDI and audio driver creation functions. The flag field is set to `FLUID_(MDRIVER|AUDRIVER)_PROBE` during probing. Drivers can then change the log level or error messages accordingly (I've modified all drivers to honor this, but can't test all drivers locally).

The flag is also used to add additional info from the driver definitions list. The jack driver is currently the only driver using this to switch between simply connecting and trying to start a new server.

Still needs some more documentation and lots of testing on other platforms. And obviously feedback about the approach in general.

### DIFF
--- a/doc/fluidsynth.1
+++ b/doc/fluidsynth.1
@@ -55,6 +55,8 @@ want to deactivate the use of the shell, start FluidSynth with the \-i
 option: 'fluidsynth \-ni soundfont.sf2 midifile1.mid midifile2.mid'.
 .PP
 Run fluidsynth with the \-\-help option to check for changes in the list of options.
+.PP
+Since version 2.2.1 fluidsynth automatically probes for an available audio and midi driver, if none was explicitly specified. During auto-probing, error messages from the drivers will not be printed, unless \-\-verbose was given. Error messages will be restored, when explicitly specifying a driver with the \-a or \-m options respectively.
 .SH OPTIONS
 \fBfluidsynth\fP accepts the following options:
 
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -256,7 +256,7 @@ create_fluid_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth,
 {
     unsigned int i;
     char *valid_options;
-    char *selected_option;
+    char *selected_option = NULL;
     fluid_audio_driver_t *driver = NULL;
     const fluid_audriver_definition_t *def;
     int auto_select = fluid_settings_str_equal(settings, "audio.driver", "auto");
@@ -327,11 +327,18 @@ create_fluid_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth,
     else
     {
         fluid_settings_dupstr(settings, "audio.driver", &selected_option);
-        FLUID_LOG(FLUID_ERR, "Couldn't find the requested audio driver '%s'.",
-                selected_option ? selected_option : "NULL");
+        if (fluid_settings_option_is_valid(settings, "audio.driver", selected_option))
+        {
+            FLUID_LOG(FLUID_ERR, "Couldn't start the requested audio driver '%s'.",
+                    selected_option ? selected_option : "NULL");
+        }
+        else
+        {
+            FLUID_LOG(FLUID_ERR, "Invalid audio driver '%s'.",
+                    selected_option ? selected_option : "NULL");
+            FLUID_LOG(FLUID_INFO, "Valid audio drivers are: %s", valid_options);
+        }
         FLUID_FREE(selected_option);
-
-        FLUID_LOG(FLUID_INFO, "Valid audio drivers are: %s", valid_options);
     }
 
     FLUID_FREE(valid_options);
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -251,47 +251,91 @@ void fluid_audio_driver_settings(fluid_settings_t *settings)
     }
 }
 
-static const fluid_audriver_definition_t *
-find_fluid_audio_driver(fluid_settings_t *settings)
+static fluid_audio_driver_t *
+create_fluid_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth,
+        fluid_audio_func_t func, void *data)
 {
     unsigned int i;
-    char *name;
-    char *allnames;
+    char *valid_options;
+    char *selected_option;
+    fluid_audio_driver_t *driver = NULL;
+    const fluid_audriver_definition_t *def;
+    int auto_select = fluid_settings_str_equal(settings, "audio.driver", "auto");
 
     for(i = 0; i < FLUID_N_ELEMENTS(fluid_audio_drivers) - 1; i++)
     {
-        /* If this driver is de-activated, just ignore it */
         if(!IS_AUDIO_DRIVER_ENABLED(fluid_adriver_disable_mask, i))
         {
             continue;
         }
 
-        if(fluid_settings_str_equal(settings, "audio.driver", fluid_audio_drivers[i].name))
+        def = &fluid_audio_drivers[i];
+
+        if (!auto_select && !fluid_settings_str_equal(settings, "audio.driver", def->name))
         {
-            FLUID_LOG(FLUID_DBG, "Using '%s' audio driver", fluid_audio_drivers[i].name);
-            return &fluid_audio_drivers[i];
+            continue;
         }
-    }
 
-    fluid_settings_dupstr(settings, "audio.driver", &name);        /* ++ alloc name */
-    FLUID_LOG(FLUID_ERR, "Couldn't find the requested audio driver '%s'.", name ? name : "NULL");
+        FLUID_LOG(FLUID_DBG, "Trying '%s' audio driver", def->name);
 
-    allnames = fluid_settings_option_concat(settings, "audio.driver", NULL);
-    if(allnames != NULL)
-    {
-        if(allnames[0] != '\0')
+        if (func == NULL)
         {
-            FLUID_LOG(FLUID_INFO, "Valid drivers are: %s", allnames);
+            driver = (*def->new)(settings, synth);
         }
         else
         {
-            FLUID_LOG(FLUID_INFO, "No audio drivers available.");
+            if (def->new2 == NULL)
+            {
+                FLUID_LOG(FLUID_DBG, "'%s' does not support callback mode", def->name);
+                driver = NULL;
+            }
+            else
+            {
+                driver = (*def->new2)(settings, func, data);
+            }
         }
 
-        FLUID_FREE(allnames);
+        if (driver)
+        {
+            driver->define = def;
+            return driver;
+        }
+
+        FLUID_LOG(FLUID_DBG, "'%s' failed to start", def->name);
+
+        if (auto_select)
+        {
+            continue;
+        }
+
+        break;
     }
 
-    FLUID_FREE(name);
+    valid_options = fluid_settings_option_concat(settings, "audio.driver", NULL);
+    if (valid_options == NULL || valid_options[0] == '\0')
+    {
+        FLUID_FREE(valid_options);
+        FLUID_LOG(FLUID_ERR, "No audio drivers available.");
+        return NULL;
+    }
+
+    if (auto_select)
+    {
+        FLUID_LOG(FLUID_ERR,
+                "Couldn't auto-select an audio driver, tried %s",
+                valid_options);
+    }
+    else
+    {
+        fluid_settings_dupstr(settings, "audio.driver", &selected_option);
+        FLUID_LOG(FLUID_ERR, "Couldn't find the requested audio driver '%s'.",
+                selected_option ? selected_option : "NULL");
+        FLUID_FREE(selected_option);
+
+        FLUID_LOG(FLUID_INFO, "Valid drivers are: %s", valid_options);
+    }
+
+    FLUID_FREE(valid_options);
 
     return NULL;
 }
@@ -318,21 +362,7 @@ find_fluid_audio_driver(fluid_settings_t *settings)
 fluid_audio_driver_t *
 new_fluid_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
 {
-    const fluid_audriver_definition_t *def = find_fluid_audio_driver(settings);
-
-    if(def)
-    {
-        fluid_audio_driver_t *driver = (*def->new)(settings, synth);
-
-        if(driver)
-        {
-            driver->define = def;
-        }
-
-        return driver;
-    }
-
-    return NULL;
+    return create_fluid_audio_driver(settings, synth, NULL, NULL);
 }
 
 /**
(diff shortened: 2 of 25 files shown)


mawe42:
The `usbsan` failure is due to a memory leak in `fluid_settings_register_str` (see #844)

jjceresa:
Tested on Windows and works as expected:

```
fluidsynth: Trying to auto-select a MIDI driver
fluidsynth: Using 'winmidi' MIDI driver
fluidsynth: Trying to auto-select an audio driver
fluidsynth: Using 'dsound' audio driver
```

jjceresa:
>In order prevent flooding the user with lots of error messages during the probing....

Now the error level has been changed from FLUID_ERR to FLUID_DBG. That means that if a problem occurs during driver initialization, the user will  see the message only for debug build but never for release build. As far there is only one error message per driver, I think It will be be preferable that the user see this error message for each tried drivers.

mawe42:
> Tested on Windows and works as expected:
> 
> ```
> fluidsynth: Trying to auto-select a MIDI driver
> fluidsynth: Using 'winmidi' MIDI driver
> fluidsynth: Trying to auto-select an audio driver
> fluidsynth: Using 'dsound' audio driver
> ```

Excellent, thanks for testing it! Just to be sure: the two "Trying to auto-select... Using ..." block come from two invocations of FluidSynth, correct?

mawe42:
> Now the error level has been changed from FLUID_ERR to FLUID_DBG. That means that if a problem occurs during driver initialization, the user will see the message only for debug build but never for release build.

Please note that the error level is only reduced to FLUID_DBG during auto probing. And you can increase the error level by specifying `-v` on the command line. As far as I know, the error level has nothing to do with the build type.

jjceresa:
>Please note that the error level is only reduced to FLUID_DBG during auto probing

I still think that it is a bad behaviour to mask any error message that informs the user during auto probing. Also changing the semantic of the message from ERR to DBG seems wrong.

>And you can increase the error level by specifying -v on the command line.

This only works for fluidsynth application. This doesn't work when calling new_fluid_midi_driver() or new_fluid_audio_driver() for which the error message is now not displayed.

jjceresa:
The option "auto" is missing in the list of options for both settings audio.driver and midi.driver.

derselbst:
> > Please note that the error level is only reduced to FLUID_DBG during auto probing
> 
> I still think that it is a bad behaviour to mask any error message that informs the user during auto probing. Also changing the semantic of the message from ERR to DBG seems wrong.

I tend to go with Marcus' arguing. When the drivers are being probed, error messages shouldn't matter too much. On the other hand, `auto` will become the new default. Hence probing will become the new default. Maybe, as a compromise, we should demote ERR to INFO instead? (I must admit that I had no opportunity to give this a serious test. Maybe wait a little longer, so I can make up my mind.)

> > And you can increase the error level by specifying -v on the command line.
> 
> This only works for fluidsynth application. This doesn't work when calling new_fluid_midi_driver() or new_fluid_audio_driver() for which the error message is now not displayed.

Technically, the application is free to display debug messages as well via `fluid_set_log_function()`.


> The option "auto" is missing in the list of options for both settings audio.driver and midi.driver.

I think this is intended, see `fluid_settings_option_is_valid()`

mawe42:
> > I still think that it is a bad behaviour to mask any error message that informs the user during auto probing. Also changing the semantic of the message from ERR to DBG seems wrong.
> 
> I tend to go with Marcus' arguing. When the drivers are being probed, error messages shouldn't matter too much. On the other hand, `auto` will become the new default. Hence probing will become the new default.

I thought about this feature from the perspective of a `fluidsynth` command-line user. And in this use-case I think if we're using auto probing and the probe finds a working audio driver, then failures and error messages of failed drivers don't really matter. Auto-probing for me is essentially: "don't care, use what works". So the most important info here is: which driver eventually succeeded? If users want to see the error messages, they can either specify it explicitly with `-a <driver>` or choose to see the debug output from the probing with `-v`. But maybe we should also add which drivers were actually tried (I thought I did that, but apparently not...)

What I want to avoid is output like this:
```
$ ./fluidsynth /usr/share/sounds/sf2/FluidR3_GM.sf2 -i
FluidSynth runtime version 2.2.0

fluidsynth: Trying to auto-select a MIDI driver
fluidsynth: jack: Cannot connect to server socket err = No such file or directory
fluidsynth: jack: Cannot connect to server request channel
fluidsynth: jack: jack server is not running or cannot be started
fluidsynth: jack: JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
fluidsynth: jack: JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
fluidsynth: Failed to connect to Jack server.
fluidsynth: Using 'alsa_seq' MIDI driver
fluidsynth: Trying to auto-select an audio driver
fluidsynth: jack: Cannot connect to server socket err = No such file or directory
fluidsynth: jack: Cannot connect to server request channel
fluidsynth: jack: jack server is not running or cannot be started
fluidsynth: jack: JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
fluidsynth: jack: JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
fluidsynth: Failed to connect to Jack server.
fluidsynth: Using 'alsa' audio driver
```

I think output like this is extremely intimidating to users. Especially if all it's really saying is: "jack midi and audio drivers unavailable, ALSA works."

I think the following is much nicer:
```
$ ./fluidsynth /usr/share/sounds/sf2/FluidR3_GM.sf2 -i
FluidSynth runtime version 2.2.0

fluidsynth: Trying to auto-select a MIDI driver
fluidsynth: Using 'alsa_seq' MIDI driver
fluidsynth: Trying to auto-select an audio driver
fluidsynth: Using 'alsa' audio driver
```

But maybe it should be more like this:
```
$ ./fluidsynth /usr/share/sounds/sf2/FluidR3_GM.sf2 -i
FluidSynth runtime version 2.2.0

fluidsynth: Starting MIDI driver auto-selection
fluidsynth: Trying 'jack_client' MIDI driver
fluidsynth: 'jack_client' MIDI driver failed to start
fluidsynth: Trying 'alsa_seq' MIDI driver
fluidsynth: Using 'alsa_seq' MIDI driver
fluidsynth: Starting audio driver auto-selection
fluidsynth: Trying 'jack_client' audio driver
fluidsynth: 'jack_client' audio driver failed to start
fluidsynth: Trying 'alsa' audio driver
fluidsynth: Using 'alsa' audio driver
```

Like I said, I think it's mainly useful for our command-line program. For the API I could imagine having a different default... maybe even change existing behaviour and require API users to explicitly choose which audio-driver to use (including "auto")? After all... why should the FluidSynth library decide for the API user which audio-driver to try first? When is this actually useful?

> Maybe, as a compromise, we should demote ERR to INFO instead?

`FLUID_INFO` has the same effect for the command-line program as `FLUID_ERR` or `FLUID_WARN`: the messages are displayed on start of the program. Maybe we need more granular log level control for our binary?

> > The option "auto" is missing in the list of options for both settings audio.driver and midi.driver.
> 
> I think this is intended, see `fluid_settings_option_is_valid()`

Yes, that was intended, but probably not the best choice. Would be good to see the "auto" option when probing possible values via the API.

I'm not overly invested in this feature, to be honest. My main aim is to prevent FluidSynth from starting a Jack server by default. That could be achieved with a much smaller PR. However, I liked the idea of auto-probing and wanted to see if and how it could be implemented, so it's not a feature I need. But I strongly think that if we want auto-probing, then we need to do it in a way that doesn't spam the user with error messages from the probing.

jjceresa:
>I liked the idea of auto-probing and ...... But I strongly think that if we want auto-probing, then we need to do it in a way that doesn't spam the user with error messages from the probing.

Yes, I understood all your points. We could notice that Jack is the only one that spam user with a lot of error message (as described in  previous example). Other driver like dsound, waveout, winmidi,... doesn't spam user and print only one message when they fail.

>I thought about this feature from the perspective of a fluidsynth command-line user.

Yes. But If a user want to see the error messages related to a driver, he can choose to see the debug output from the probing with -v. and in this case the user is flooded by a lot of message not related to a driver. This is a dilemma !

So, may be the compromise would be in the driver code on a case by case basis. For example, we could consider that such driver that doesn't spam user should ignore the probing flag and print this unique ERR message before they return ?.

jjceresa:
>On the other hand, auto will become the new default. Hence probing will become the new default.

Does "auto" should be the new default  for both audio and midi driver simultaneously ?.

We could also consider that the default driver should stay the actual one. In this case if a user want to use the "auto select" future for a particular driver, then he choose this with -o  in the command line:
-o audio.driver=auto  #only the audio driver will be auto selected. The MIDI driver in use will be the OS's related default
-o midi.driver=auto    #only the MIDI driver will be auto selected. The audio driver in use will be the OS's related default
-o audio.driver=auto -o midi.driver=auto  #both audio and midi driver will be auto selected

derselbst (commented):
I had a few chances to test this and it works very well. Also, I think it's correct to use debug log level while auto probing. If everything fails the user sees something like:

```
fluidsynth: Trying to auto-select a MIDI driver
fluidsynth: error: not enough MIDI in devices found. Expected:1 found:0
fluidsynth: error: Couldn't auto-select a MIDI driver, tried winmidi
Failed to create the MIDI thread; no MIDI input
```

Which I think is good enough. If the user needs more output, simply specifying `-m winmidi` will give details of the errors.

So I would take this. My only question: Can we release this with 2.2.1? Because it's a new feature and technically doesn't fit into the versioning scheme.

jjceresa:
>Also, I think it's correct to use debug log level while auto probing. If everything fails the user sees something like....
fluidsynth: error: not enough MIDI in devices found. Expected:1 found:0
Which I think is good enough.

Please, be aware that this ERROR message is displayed by fluid_winmidi_parse_device_name() that ignores that the driver is in auto probing context. If we agree with this behaviour, then for consistency this behaviour should be adopted for audio drivers as well (at least for windows audio driver).

>Can we release this with 2.2.1? Because it's a new feature and technically doesn't fit into the versioning scheme.
In this case the following documentations should be updated:
- doc of audio.driver, midi.driver settings should be updated to keep the user aware of this magic auto probing. Particularly if `auto` isn't part of the list of options.
- doc of fluidsynth console application should be updated to inform that the user must now specify -m xxxxx -a yyyyy to get  more details.

derselbst:
> Please, be aware that this ERROR message is displayed by fluid_winmidi_parse_device_name() that ignores that the driver is in auto probing context.

Technically, you are correct, yes.

> doc of audio.driver, midi.driver settings should be updated to keep the user aware of this magic auto probing. Particularly if auto isn't part of the list of options

The auto probing feature should be internal to fluidsynth. I.e. it should not be exported via API. So I think it's ok that "auto" is not visible via API.

> doc of fluidsynth console application should be updated to inform that the user must now specify -m xxxxx -a yyyyy to get more details.

The only place I can think of for this kind of documentation is the man page. What do you think?

jjceresa:
>The only place I can think of for this kind of documentation is the man page. What do you think?

Indeed this is the only place.

mawe42:
>> doc of fluidsynth console application should be updated to inform that the user must now specify -m xxxxx -a yyyyy to get more details.

> The only place I can think of for this kind of documentation is the man page. What do you think?

I think the fluidsynth binary should also mention this when it failed to find a working audio/midi driver.

mawe42:
> > Please, be aware that this ERROR message is displayed by fluid_winmidi_parse_device_name() that ignores that the driver is in auto probing context.
> 
> Technically, you are correct, yes.

That is actually an oversight on my part. To keep things consistent with all other drivers, the following line should not be displayed, in my opinion:
```
fluidsynth: error: not enough MIDI in devices found. Expected:1 found:0
```
> The auto probing feature should be internal to fluidsynth. I.e. it should not be exported via API. So I think it's ok that "auto" is not visible via API.

Well, "auto" is not visible, but it could still be used via the API. In fact, with the current implementation it is actually the default when you don't specify an audio driver in the `fluid_settings`. (see 6426722dc0140138f32f29beb0817cf21b80a7c3)

So if we wanted the API usage to remain unchanged, then we would have to revert that commit and make the "auto" default happen only in `fluidsynth.c` instead. Sounds like a good idea too, in my opinion.

mawe42:
> So I would take this. My only question: Can we release this with 2.2.1? Because it's a new feature and technically doesn't fit into the versioning scheme.

Not sure... as it is currently implemented, it is not only a new feature, but also quite a large change in behaviour. So in my opinion much to invasive for a patch-level release.

I'm still not really happy with the way this is implemented. I agree with JJC that `FLUID_DBG` is the wrong logging level for the error messages during probing. `FLUID_INFO` would be much more appropriate. But for that to work the way I think it should, the fluidsynth binary would need a new option that would allow control of the logging level. And it should default to "warning", in my opinion.

mawe42:
> the fluidsynth binary would need a new option that would allow control of the logging level

And to clarify this point: I think the `--verbose` option is really much too verbose. I think it would be great to get more information about driver failures without also seeing all MIDI messages on the console. So I would propose the following:
1. Close this PR without merge
2. Introduce more granular logging for the fluidsynth binary and make MIDI message dumping a separate switch.
3. Extend the driver loading API so that the flags like "auto-probing" or "don't auto-start a jack server" could be sent to drivers before they are created. Change drivers so that they output FLUID_INFO if auto-probing flag is set.
4. Introduce the auto-probing code into `fluidsynth.c` and not into the FS library itself.

mawe42:
> So I would propose the following

Or maybe take a step back and think this over again...

My initial motivation for this change was the fact that FS defaults to `jack` on Linux and automatically creates a jack server if none is available (which messes up my PulseAudio setup, see #838). The easiest solution to this problem would be to move `alsa` in front of `jack` in the audio driver list. That small change would make FS also more consistent, in my opinion, because the MIDI driver list already has `alsa_seq` and `alsa_raw` before `jack`.

And the more I think about this PR, the more I wonder if auto-probing drivers is a good idea. Which problem does this solve, and for which target audience?

derselbst:
This idea came up in https://github.com/FluidSynth/fluidsynth/issues/504#issuecomment-462129661 . I thought it makes fluidsynth a little more user-friendly. But I'm ok with not merging it. I would however keep that branch in case we see a need for it again.

We could also simply change the order of audio driver as you suggested.

mawe42:
> This idea came up in [#504 (comment)](https://github.com/FluidSynth/fluidsynth/issues/504#issuecomment-462129661) . I thought it makes fluidsynth a little more user-friendly.

Hah... and I was of the same opinion in 2019, it seems :-)

Ok... so we don't require users to explicitly choose their audio driver. We know which drivers we could try on their platform, we even have a preferred order, but we fail as soon as the first driver in that list doesn't work. That does sound like a good problem to fix with auto-loading.

I really like @carlo-bramini idea in https://github.com/FluidSynth/fluidsynth/issues/504#issuecomment-463137454: extend the `audio.driver` (and `midi.driver`) options to allow for a list of drivers to choose. Only do auto-probing if the setting contains more than one driver. Then change the fluidsynth executable to collect all drivers available on the current platform and concat them as the default value for audio and midi drivers, but let the user override those options, of course.

jjceresa:
>We could also simply change the order of audio driver as you suggested.

It seems that it is the most important feature of this PR (with the jack stuff as well).

>And the more I think about this PR, the more I wonder if auto-probing drivers is a good idea.Which problem does this solve, and for which target audience?

Auto-probing is user friendly because it shows the driver's name in use. That said, without auto-probing it is `unlikely` that a driver selection will fail as far the first driver in the definition list is the usual OS's driver: (i.e alsa for linux, dsound for Windows).
At driver creation, we could simply display the driver's name in use without any auto-probing logic, and if the creation fails we could simply show the list of driver options.

mawe42:
Hm... so what do we do? Simply make ALSA the default on Linux?

@carlo-bramini do you have an opinion on this?

derselbst:
> Hm... so what do we do? Simply make ALSA the default on Linux?

Let's do it. I would like to release 2.2.1 this weekend. We better do it now than late.

mawe42:
Ok, i'll prepare a PR with just that change tomorrow.

mawe42:
The much smaller solution is now in #878. I'll close this PR.

PR #844, opened 2021-04-10
Fix a memory leak if a setting is registered multiple times

mawe42:
If a string setting is registered multiple times, the previous default value wasn't freed, leading to a memory leak.


mawe42:
Have I already mentioned that I think our CI infrastructure is great? :)

derselbst:
> Have I already mentioned that I think our CI infrastructure is great? :)

Only one or two times :)

Thanks!

PR #845, opened 2021-04-11
Fix incomplete chorus reverb help strings

jjceresa:


src/bindings/fluid_cmd.c
```diff
@@ -214,7 +214,7 @@ static const fluid_cmd_t fluid_commands[] =
     },
     {
         "reverb", "reverb", fluid_handle_reverb,
-        "reverb [0|1|on|off]         Turn all reverb groups on or off"
+        "reverb [group] [0|1|on|off] Turn all or one reverb group on or off"
```
mawe42, line 217: I think the square brackes around `[0|1|on|off]` are misleading, because so far I understood them as meaning "optional". But that param is not optional. So maybe change it to `(0|1|on|off)`? (Same below for chrous)

mawe42: Thanks!

PR #848, opened 2021-04-13
Make Windows CRT linkage user-selectable

### LINKED ISSUE #840: New option to enable/disable linking to the static C runtime library (Windows, MSVC compiler)
Right now, the cmake build system forces linking all targets to the static C runtime library when building with MSVC. Applications linking dynamically to the CRT and the fluidsynth library may experience problems like the described in this ticket:
https://github.com/rncbc/qsynth/issues/39

This is a well known problem, as explained here:
https://docs.microsoft.com/en-us/answers/questions/240332/assert-when-using-mtd-dynamic-library.html

A new cmake option "enable linking the static C runtime library when compiling with MSVC", enabled by default.
When disabled, all targets would be dynamically linked to the C runtime DLL.

This has impacted qsynth, a fluidsynth GUI, which is a Qt application. The alternative of rebuilding each version of the Qt libraries was discarded because it is very time consuming.

derselbst:
This allows the user to select the CRT linkage via `CMAKE_MSVC_RUNTIME_LIBRARY`. If not specified, dynamic linkage becomes the default again.

Documentation of `fluid_free()` has been updated.

Resolves #840.
Supersedes and closes #842.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -56,7 +56,7 @@ jobs:
         SET "PATH=d:\deps\bin;%PATH%"
         pkg-config --list-all
         mkdir build && cd build || exit -1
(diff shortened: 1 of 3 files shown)


derselbst:
@pedrolcl FYI

Issue #836, opened 2021-04-10
Support reverb with the WASAPI audio driver


jimhen3ry:
### Related discussion on the mailing list
WASAPI and reverb #831

### Is your feature request related to a problem?
FluidSynth reverb setting apparently become ineffective when the Windows WASAPI audio driver is used.

### Describe the solution you'd like
Support reverb with the WASAPI audio driver. If exclusive versus non-exclusive WASAPI is an issue, allow a choice between exclusive and non-exclusive modes.

### Describe alternatives you've considered
Using the PortAudio WASAPI driver.

### Additional context
Reverb is absolutely necessary for virtual organs and WASAPI nearly so. It would be a big help for the Miditzer to be able to allow users to use the WASAPI audio driver without giving up reverb.

jimhen3ry:
Related discussion on the mailing list:
WASAPI and reverb #831

Is your feature request related to a problem?
FluidSynth reverb setting apparently become ineffective when the Windows WASAPI audio driver is used.

Describe the solution you'd like:
Support reverb with the WASAPI audio driver. If exclusive versus non-exclusive WASAPI is an issue, allow a choice between exclusive and non-exclusive modes.

Describe alternatives you've considered:
Using the PortAudio WASAPI driver.

Additional context:
Reverb is absolutely necessary for virtual organs and WASAPI nearly so. It would be a big help for the Miditzer to be able to allow users to use the WASAPI audio driver without giving up reverb.

jjceresa:
>Using the PortAudio WASAPI driver.

Have you considered using PortAudio WDM/KS as alternative ?

jimhen3ry:
> Have you considered using PortAudio WDM/KS as alternative ?

Yes. 

A design goal for the Miditzer is to be very easy for a complete novice to get up and running. We believe that if someone isn't up and running within minutes of downloading, without support or reading documentation, then it is likely they will delete the program and be lost to us. To date, we make dsound the default audio driver because we know it works 99.99% of the time.

When users add a MIDI keyboard the latency of dsound becomes an issue and a lot of support time is spent trying to coach users through tuning their system for reduced latency. We have debated WASAPI versus WDM/KS for many years. The big stumbling block is the lack of understanding of how many users would be negatively affected by using WASAPI or WDM/KS. 

[...]

PR #849, opened 2021-04-16
A few fixes for the WASAPI driver

chirs241097:
* Explicitly use period size provided by the driver in shared mode.
* Ensure `time_to_sleep` is always computed correctly.
* Render effects if no custom audio processing function is supplied (#836).


derselbst (approved):
Perfectly correct. Thanks!

jimhen3ry:
I just ran a quick test using the latest commits and I can verify that reverb is working with WASAPI. Thanks for the quick work Chris! 

We will work to get a beta version of the Miditzer out that includes the latest commits in FluidSynth to get more testing of everything. Unfortunately, we won't be a fast as you Chris.

PR #854, opened 2021-04-20
[DSOUND] Remove useless braces.

carlo-bramini:
It is not really clear to me the reason because braces are written into the initialization of `channel_mask_speakers[]`, since their presence causes these messages from the compiler:

`fluid_dsound.c:63:5: warning: braces around scalar initializer`

In my opinion, it is better to remove them.


jjceresa:
>In my opinion, it is better to remove them.

Your are right those braces aren't necessary. I usually add braces for structure field but here channel_mask_speakers[] is an array, so braces are useless. I have compiled this using Visual Studio 2010 on Window XP without getting any warning.

Please, feel free to remove those braces in waveout.c as well.

derselbst (approved):
Ok, thanks.

PR #866, opened 2021-04-26
[MinGW-x64] Fix some warnings for MinGW.

carlo-bramini:
These corrections fix this warnings on MinGW for x86_64:
`warning: ‘static’ is not at beginning of declaration [-Wold-style-declaration]`


derselbst:
Sry, duplicate of #861.

PR #868, opened 2021-04-28
[WINDOWS] Fix declaration of GUIDs (try 2).

carlo-bramini:
This PR superseeds #867.
Tested with MinGW-w64 and MSVC-2019


jjceresa:
Tested successfully with Visual Studio 2010 (that is the the only one MSVC that can be installed on Windows XP).

derselbst:
Thanks!

PR #871, opened 2021-05-01
Define _WIN32_WINNT during compilation

### LINKED ISSUE #869: Windows "load time linking" failure because inet_ntop is missing in dll WS2_32.dll
This issue is introduced by commit c8c3966586df5202cdb5234c62d83c3d29077e6b.

The goal of the commit is to make cmake finding inet_ntop.
cmake and compilation using Visual studio 2010 is successful. 
When executing fluidsynth console there is a load time linking failure message that says that `"the inet_ntop function entry point isn't found in the WS2_32.dll"`.

derselbst:
Seems like it's necessary to specify the Windows version you want to compile for. Esp. when compiling for an older Windows.

Since the CMake build system itself cannot know, for which Windows version we want to compile for, I suggest manipulating the `CFLAGS` environment variable directly in the build pipeline.

+Minor unrelated cleanup.

Resolves #869.


jjceresa:
>Since the CMake build system itself cannot know, for which Windows version we want to compile for

Could we think of a new option for CMake so that if the user choose  `cmake -Dwindows-version=xxxx` than CMake will set HAVE_INETNTOP only if xxxx is above 0x501 ?

jjceresa:
Please, could you help me to point on the result of the CI Build binary for XP. I will dot a try ?

derselbst:
Here's an overview of the produced artifacts: https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=4567&view=artifacts&pathAsName=false&type=publishedArtifacts

The fluidsynth-win32 is the correct one for you: https://dev.azure.com/tommbrt/d3638885-de4a-4ce7-afe7-f237ae461c07/_apis/build/builds/4567/artifacts?artifactName=fluidsynth-Win32&api-version=6.0&%24format=zip

derselbst:
> Could we think of a new option for CMake so that if the user choose cmake -Dwindows-version=xxxx than CMake will set HAVE_INETNTOP only if xxxx is above 0x501 ?

I'm not sure. Technically it's a bit redundant, because in this case you could also set the `-Denable-ipv6=0`. I will think about it.

jjceresa:
>I'm not sure. Technically it's a bit redundant, because in this case you could also set -Denable-ipv6=0 to disable inet-ntop.

Sorry, this doesn't work. If the user  set -Denable-ipv6=0, cmake still set HAVE_INETNTOP and later the compiler will still use inet_ntop() function.

jjceresa:
>The fluidsynth-win32 is the correct one for you: https://dev.azure.com/tommbrt/d3638885-de4a-4ce7-afe7-f237ae461c07/_apis/build/builds/4567/artifacts?artifactName=fluidsynth-Win32&api-version=6.0&%24format=zip

Thanks. It runs well on Windows XP.

derselbst:
> > I'm not sure. Technically it's a bit redundant, because in this case you could also set -Denable-ipv6=0 to disable inet-ntop.
> 
> Sorry, this doesn't work. If the user set -Denable-ipv6=0, cmake still set HAVE_INETNTOP and later the compiler will still use inet_ntop() function.

Oh, indeed. `enable-network=0` will do the trick. But that's a poor solution admittedly.

+1 for your `-Dwindows-version=xxxx` proposal. It would be good if we find a reasonable good default value for this cmake option, so that you don't have to specify it every time when compiling fluidsynth.
`-Dwindows-version=xxxx` can still be provided explicitly, when compiling e.g. in a build pipeline. I have an idea how to do this. I'll make a proposal in a few days or so.

jjceresa:
>It would be good if we find a reasonable good default value for this cmake option, so that you don't have to specify it every time when compiling fluidsynth.

This should not be a high problem because cmake memorizes user's options in its cache. So, on next invocation cmake will use most recent user's options.

Note: Also, the `-Dwindows-version=xxxx` proposal will  allows potentially to set a preprocessor definition like `_WIN32_WINNT=xxxx`.
This way, any Windows specific code source can know the Windows version target to do specific stuff pertinent to this version.

derselbst:
@jjceresa CMake now uses the windows version you are running by default. No additional argument is required. Please give it a test.

jjceresa:
>CMake now uses the windows version you are running by default. Please give it a test.

Running on Windows XP, CMake Configure output is:
```
Targetting Windows Version 0x0501
...
Looking for inet_ntop
Looking for inet_ntop - not found
...
Configuring done
Generating done
```
Then Compilation + link success. Execution: success.
All works as you expected. Thanks!

PR #873, opened 2021-05-03
[WAVEOUT] Dynamic alloc of WAVEHDR

carlo-bramini:
It seems that the addition of multichannel output has broken the WaveOut driver.
If you try to run FluidSynth with -a waveout, you will get this message:

`fluidsynth: audio.periods 8 exceeds internal limit 4`

Actually, the default value for period is set to 8, so it will never work unless you change that value before opening the driver. Rather than lowering the default period value or rising the limit of the number of buffers, in my opinion it would be better to free the driver from this limitation, by allocating the needed amount of memory for WAVEHDR too.
Attached patch fixes this bug.

### DIFF
--- a/src/drivers/fluid_waveout.c
+++ b/src/drivers/fluid_waveout.c
@@ -32,9 +32,6 @@
 #include <ks.h>
 #include <ksmedia.h>
 
-/* Number of driver buffers in the chain */
-#define NB_SOUND_BUFFERS    4
-
 /**
 * The driver handle multiple channels.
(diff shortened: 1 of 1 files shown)


jjceresa:
>in my opinion it would be better to free the driver from this limitation, by allocating the needed amount of memory for WAVEHDR too.

Thanks Carlo, Indeed, it is better.

derselbst (approved):
Thanks!

PR #875, opened 2021-05-06
Rewrite Android CI Pipeline

### LINKED ISSUE #689: Provide Andoird Binaries with DLS support
With the current binaries, DLS files cannot be loaded, #686.

Android CI build has to build with support for libinstpatch. ~~But before this can be done, the AndroidCI build first needs to be fixed #688.~~

derselbst:
Original TODO of this PR: Compile fluidsynth Android binaries with libinstpatch (DLS) support (#689). This would have required to add some "recipe" to Cerbero. And I ended up with rewriting the entire pipeline. Why?

1. Because, I don't really like Cerbero:

    * I cannot use it on my maschine, because my version of openSUSE Leap is not supported by Cerbero.
    * Cerbero is a black box: You have no influence on which versions should be compiled. You don't know how it's being built either. E.g. libffi is compiled with a custom meson build system. Why? 
    * Sometimes cerbero compiles against android-16 sometimes against android-21.
    * Cerbero doesn't use software that is already installed on the system: E.g. the Android NDK is already present, yet Cerbero downloads it again.

2. I don't like CircleCI either: You need to sign in just for viewing the CI logs... seriously?

Hence I'm porting to Azure Pipelines, because

* I'm very familiar with its syntax, and
* it allows our Azure release pipeline to consume the android binary artifacts and automatically attach them to the GH release (just like we already do for the Windows binaries).

At the same time I'm rewriting the build pipeline, so that we

* have more control over the build itself,
* can lower the minimum Android API to 24 (rather than 30 before)
* add DLS support via libinstpatch, and
* drastically reduce the build time by using caching: now it will be completed in 2 minutes rather than 20 to 30 mintues before.

The pipeline is heavily based on my original do-it-yourself-cross-compilation-shell-script idea [posted on Gist](https://gist.github.com/derselbst/885641b86b8bd0e5fd1b1d846d4b286b).

I made an attempt to use meson to compile glib. But I completely failed to do anything with meson, it kept complaining about:

```
meson.build:1:0: ERROR: Unable to detect GNU compiler type:


A full log can be found at /home/vsts/work/1/s/android-build-root/glib-2.62.6/_builddir/meson-logs/meson-log.txt
```

And when you look into `meson-log.txt` there's no additional info in there. I'm done with python stuff, so I kept using autotools with an older glib version.

One TODO might be to figure out how to execute fluidsynth's unit tests (perhaps by using the android-emulator somehow??).

Resolves #689.


derselbst:
@TimeLimitExceeded You originally requested fluidsynth Android binaries with DLS support. Feel free to test it:
https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=4845&view=artifacts&pathAsName=false&type=publishedArtifacts

atsushieno:
Hi, I just noticed that you had been rewriting the Android build scripts, which is a great work. Thanks and sorry I didn't really have time for that. I didn't like the cerbero workflow either, and was thinking to rewrite it now (and noticed this).

At this stage I don't really think the old build pipeline is needed anymore. I'm all for removing it from the top README.

derselbst:
Thanks for the feedback! And no problem, it was quite fun to rewrite it and the Cerbero pipeline was a good template.

I'll probably remove the Cerbero pipeline from the README in the future. I also want to merge the two Windows pipelines, let's see when this will be done.

There's still an open TODO about executing the unit tests in the Android Pipeline. In case you have an idea how to do that you are very welcome :)

atsushieno:
I cannot promise but I would be happy to help if I can! But is it possible to run Azure pipelines scripts locally?

derselbst:
> But is it possible to run Azure pipelines scripts locally?

Not with ease, unfortunately. They once had a [local run feature](https://stackoverflow.com/a/53047000), but it seems it was not completed. One could still register a build agent in Azure Devops (i.e. your own computer) and then pin the pipeline to that agent. But our computer would then need a [similar environment as they use in ADops](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md).

The best offer I can make is to invite you to our organization so that you can just make changes on your own branch and then see how the pipeline behaves.

atsushieno:
Okay, thanks anyways, I guess I should rather create a standalone shell script that runs locally.

atsushieno:
And thanks for invitation. I'm not sure if I would really need or prefer using it, so please feel free to deactivate me at any time, less build previledge is safe and better :-)

PR #876, opened 2021-05-06
Restore old behaviour of fluid_player_join()

### LINKED ISSUE #872: Calling fluid_player_join causes an infinite loop after fluid_player_stop
### FluidSynth version
Since 676923757c4f96c5c34931da5566d5b5be3e15eb. v2.2.0.rc1 fails, v2.2.0 fails, 0ea101721aa96a504c51c8f697733a28046b35f5 fails. All compiled on Debian Bullseyes.

### Describe the bug
Calling `fluid_player_join` causes an infinite loop after `fluid_player_stop`.

### Expected behavior
That `fluid_player_join` does not cause an infinite loop in this situation. It did work in at least v2.1.8 and v2.2.0.beta1.

### Steps to reproduce
Attached is a small example program that shows the behavior. [fluidsynth_infinite_loop.zip](https://github.com/FluidSynth/fluidsynth/files/6411267/fluidsynth_infinite_loop.zip)

### Additional context
The regression is caused by #783 adding a `STOPPING` state, and in this flow that never transitions to `STOPPED` so `fluid_player_join` keeps waiting forever.

derselbst:
This basically reverts #783 because it causes a deadlock in existing client code.

Resolves #872.

### DIFF
--- a/include/fluidsynth/midi.h
+++ b/include/fluidsynth/midi.h
@@ -246,7 +246,7 @@ enum fluid_player_status
 {
     FLUID_PLAYER_READY,           /**< Player is ready */
     FLUID_PLAYER_PLAYING,         /**< Player is currently playing */
-    FLUID_PLAYER_STOPPING,        /**< Player is stopping, but hasn't finished yet */
+    FLUID_PLAYER_STOPPING,        /**< Player is stopping, but hasn't finished yet (currently unused) */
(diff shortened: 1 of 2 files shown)


mawe42:
Looks good to me, thanks! Should we maybe also add an explicit warning to `fluid_player_join()` and explain that it's not meant as means to check if a player can be deleted?

mawe42:
Great, thanks a lot!

PR #896, opened 2021-05-24
Include header files in Android Artifacts

### LINKED ISSUE #894: Header files are missing from Android release zip
Currently Android build artifacts on the Releases page are created from the new Azure pipelines. I noticed that it is missing headers and when we build apps that references fluidsynth API in our code we will have to get those headers from somewhere separately.

Windows release binary zip seems to contain the headers. I think it is the right way for Android too.

- when we build apps that references fluidsynth API in our code we will have to get those headers from somewhere separately, otherwise.

- Add `fluidsynth.h` in the archive too, probably within `include/fluidsynth` directory (assuming it is platform independent).

- I could download headers manually.

### Additional context

Not sure if it is worth mentioning, but I make use of fluidsynth for Android in this app (which I think has build setup in closer mindset to the latest Android builds here rather than my original Cerbero builds), https://github.com/atsushieno/aap-lv2-fluidsynth (It is still not an intuitive use case and not for testing header inclusion at all.)

Checked release v2.2.1. It might be different from the latest azure pipeline settings.

derselbst:
Include header files and remove symbolic links to `so.x.y`. I had to change the folder structure a bit, it should be self-explaining though.

Resolves #894.

### DIFF
--- a/.azure/azure-pipelines-android.yml
+++ b/.azure/azure-pipelines-android.yml
@@ -601,11 +601,14 @@ jobs:
 
     - script: |
         set -ex
-        mkdir -p $(Build.ArtifactStagingDirectory)/$(ARCH)
-        cd $(Build.ArtifactStagingDirectory)/$(ARCH)
+        mkdir -p $(Build.ArtifactStagingDirectory)/lib/$(ARCH)
+        cd $(Build.ArtifactStagingDirectory)/lib/$(ARCH)
         cp -a $(PREFIX)/lib/* .
         ls -Rg .
         rm -rf *.dll *.alias gettext/ libtextstyle.* *.a *.la
+        mkdir -p $(Build.ArtifactStagingDirectory)/include
+        cd $(Build.ArtifactStagingDirectory)/include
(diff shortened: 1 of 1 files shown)


derselbst:
Example artifact can be found here: https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=5280&view=artifacts&pathAsName=false&type=publishedArtifacts

derselbst:
@atsushieno FYI

atsushieno:
Yay it looks fixed!

Sorry one last thing, is there any reason you prefer "i686", "aarch64", and "arm" as subfolders? If they were "x86", "arm64-v8a", and "armeabi-v7a", then we can simply copy the archive contents into `src/main/jniLibs` directly, without separate steps for copying contents for each arch.

derselbst:
> If they were "x86", "arm64-v8a", and "armeabi-v7a", then we can simply copy the archive contents into

Good hint! I'm now using `$ANDROID_ABI_CMAKE`.

PR #905, opened 2021-06-04
Keep pedaling (sustain or sostenuto) effective when the same note is played more than once.

jjceresa:
When playing the same note more than once the previous note is forced in release stage except if pedaling (sostenuto or sustain) is active. This makes it sound more like a real piano. See https://lists.nongnu.org/archive/html/fluid-dev/2021-06/msg00001.html

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -6819,8 +6819,9 @@ fluid_synth_release_voice_on_same_note_LOCAL(fluid_synth_t *synth, int chan,
                 synth->storeid = fluid_voice_get_id(voice);
             }
 
-            /* Force the voice into release stage (pedaling is ignored) */
-            fluid_voice_release(voice);
+            /* Force the voice into release stage except if pedaling
+               (sostenuto or sustain) is active */
+            fluid_voice_noteoff(voice);
         }
     }
 }


jjceresa:
@mrbumpy409 This PR should fix the behaviour you started on the mailing list (https://lists.nongnu.org/archive/html/fluid-dev/2021-06/msg00001.html).
Please, could you have a try on your side to confirm that it works as expected ?.

derselbst (commented):
Looks correct to me. But I would like to test this behaviour with overlapping notes in the next days. Can you remember why you wanted pedaling to be ignored back then?

jjceresa:
>Can you remember why you wanted pedaling to be ignored back then?

If I remember well, I noticed that when playing the same note more than once while pedaling this result in sound growing louder. I was thinking that this wasn't natural. I was wrong because in fact this is very natural on real acoustic piano for which keeping the previous note still alive allows a fluid legato playing with following notes played at different dynamics (i.e velocity).

jjceresa:
>But I would like to test this behaviour with overlapping notes in the next days.

For overlapping same note, new voices (with respective velocity) will be started for each following noteon. On noteoff these voices will be sustained by the active pedal. Then when the pedal become inactive, all voices are released at the same time.

derselbst (approved):
Works well, thanks!

Issue #912, opened 2021-06-12
Add Android testing setup


atsushieno:
context: https://github.com/FluidSynth/fluidsynth/pull/875#issuecomment-846898475 / https://github.com/FluidSynth/fluidsynth/issues/897#issuecomment-859687282

Current Android build setup on Azure DevOps misses testing on Android:

https://github.com/FluidSynth/fluidsynth/blob/f81caf37f192a518f24c460481ddeb8f1cfb8382/.azure/azure-pipelines-android.yml#L586

Since the tests that are based on ctest are not simply runnable on Android, we first have to convert existing tests into runnable style in Android manner. A dedicated test runner application would suffice.

As commented on the issue linked above, I came up with such a solution (test generator + runnable setup). A PR will follow and be linked here. So far it lacks the actual Azure DevOps setup, but having a locally runnable test app would be a good starter.

derselbst:
### Interim report

The main problem is that the include and library paths are hardcoded in the CMakeLists.txt:

https://github.com/FluidSynth/fluidsynth/blob/92089e0a0f4c1d1d20dee21c88e7bad7f3a88084/test-android/app/src/main/cpp/CMakeLists.txt#L66-L70

When we build in Azure, the paths are different. To circumvent this, we could use the build-scripts in test-android/build-scripts/. Although this would work, it would still mean to have hard-coded paths and increased maintenance effort, as they always need to be kept in sync with the rest of the build.

Instead, I would like to integrate the android CMake part into our main CMake build system. I've already started this on the `android-emulator` branch, but it's far from being complete. The idea is that CMake should detect, whether it's compiled for ANDROID or not. If complied for ANDROID, it should directly compile all unit tests into a single .so file, which could then be directly consumed by gradle.

I'll keep working on this from time to time. But it will take some...

syntaro:
Hello, May I cut in ?

This is my CMakeLists.txt for Android project.
And I have an question with it.

Which version FluidSynth.so is not require glib2.0. Is it in the future?
I am importing glib2.0 and std++ both on it. ...

```
cmake_minimum_required(VERSION 3.4.1)

project(mylib
        VERSION 1.0
        DESCRIPTION "mylib project"
        LANGUAGES CXX
)

add_library(native-lib SHARED src/main/cpp/native-lib.cpp)

# For more information about using CMake with Android Studio, read the
# documentation: https://d.android.com/studio/projects/add-native-code.html

[...]

PR #913, opened 2021-06-12
Add Android testing setup

atsushieno:
context: https://github.com/FluidSynth/fluidsynth/issues/912

This adds basic testing setup along with custom build scripts and test source generator scripts.

The rationale behind the app and how-to documentation to set up and run both locally and on Azure DevOps are described on `test-android/README.md`.

Changes are rebased off of master HEAD and can be either packed in merge-commit or merged as is (some extraneous reverts involved though).


atsushieno:
Some defects I'm aware of:

- Azure DevOps setup is not done. Maybe it can be done based on existing information like [this](https://medium.com/genetec-tech/android-ui-testing-in-azure-devops-81bbe7cea9fd). I'm still unsure if setting up emulator costs too much build time and/or network resources and have no idea how the team can afford to accept the cost, so I would leave it to @derselbst for now, but feel free to delegate it to me if it was not easy, it would certainly require some Android development background.
- The package versions are hard-coded on build-scripts/build-env.sh as they were just copied from Azure DevOps setup. Ideally they had better be retrieved from somewhere, but that would affect how you manage the builds. Considering that it would not be too version dependent, at this state I leave them as is.

derselbst:
Works like a charm! I only added a missing `mkdir` and relaxed the CMake version check. I'll keep the commit history when merging this, as it could be helpful in the future to understand the trouble you've been through :)

One question: Currently, a single `gradlew build` command handles all four arch. How does gradle know that it should handle those four archs? I couldn't find that place. I'm asking because in ADops we will only be able to handle a single arch.

atsushieno:
awesome!

> One question: Currently, a single `gradlew build` command handles all four arch. How does gradle know that it should handle those four archs? I couldn't find that place. I'm asking because in ADops we will only be able to handle a single arch.

Looks like Azure DevOps has awkward limitation.. I just added an customizible build property. Now you can run gradle like: `./gradlew connectedCheck -PcustomAbiFilters=x86`

derselbst:
Perfect, thank you! I'm merging this and in the next days / weeks will work in the ADops integration.

> Looks like Azure DevOps has awkward limitation

It's intentional: You get up to 10 parallel build slots in Azure, but each virtual machine only gets a dual-core CPU. Building all four archs on a single VM takes way too long (for my taste). But building each architecture in a different slot reduces overall build time by factor four ;)

PR #914, opened 2021-06-12
[GLIB] Replace g_shell_parse_argv.

carlo-bramini:
With reference to discussion #847 this is my change for replacing g_shell_parse_argv().


derselbst:
I currently don't see that we'll get rid of glib soon. Therefore I'm closing this. I may come back though if / when we're ready for it.

PR #927, opened 2021-06-30
autoconnect all available effects ports

albedozero:
Step through all available effects outputs and connect them to jack inputs, and wrap around if there are fewer input ports than effects outputs.


derselbst (approved):
Well spotted, thanks!

PR #929, opened 2021-07-02
wasapi: increase initialization timeout to 2 seconds, same as shutdow…

pedrolcl:
…n timeout


derselbst:
Do you have a particular reason, or do you just want to keep the timeouts consistent?

pedrolcl:
> Do you have a particular reason, or do you just want to keep the timeouts consistent?

Yes, the 2 seconds timeout was chosen for consistency, but the reason to increase the initialization timeout is that one of my test devices (Asus T101HA, Windows 10) fails to initialize with a timeout error most of the times.

This patch has been used already in the Windows setup package of VMPK 0.8.4 and dmidiplayer 1.4, and will be included as well in the next Qsynth Windows release.

derselbst (approved):
Ok, good to know. Thanks!

PR #934, opened 2021-07-05
only send all_notes_off when stopping player

albedozero:
Repurposes the unused `FLUID_PLAYER_STOPPING` state as a flag to tell the player callback to send `fluid_synth_all_notes_off` only once after `fluid_player_stop` is called.

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -2088,7 +2088,11 @@ fluid_player_callback(void *data, unsigned int msec)
 
     if(fluid_player_get_status(player) != FLUID_PLAYER_PLAYING)
     {
-        fluid_synth_all_notes_off(synth, -1);
+        if (fluid_player_get_status(player) == FLUID_PLAYER_STOPPING)
+        {
+            fluid_synth_all_notes_off(synth, -1);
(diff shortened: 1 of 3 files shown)


albedozero:
The STOPPING state was introduced in #783 so that `fluid_player_join` would exit only when it was safe to delete a player, and reverted to unused in #876 because it would cause `fluid_player_join` to loop endlessly if used to check if a player is finished while audio isn't being rendered, but it seems `fluid_player_join` isn't meant to be used in either of these ways. Looks like client code is being patched to reflect this (e.g. [#9214](https://github.com/OpenTTD/OpenTTD/pull/9214))

derselbst (changes requested):
Your issue is valid. But your proposed change would again require that a synth context is running. Otherwise there will be a deadlock again. OpenTTD's has been fixed, yes. But that's no guarantee that other clients won't run into the same issue. We've always recommended the calling convention

```
fluid_player_join();
delete_fluid_player();
```

which doesn't make any assumptions if a running synth context or not.

Please solve it by using a custom flag.

albedozero:
Thanks! Hopefully it's fine to add an atomic to fluid_player_t like this?

derselbst (approved):
LGTM, thanks!

PR #937, opened 2021-07-09
Update license header for fluid_synth_monopoly.c

### LINKED ISSUE #936: src/synth/fluid_synth_monopoly.c license clarification
The license header at the top of `src/synth/fluid_synth_monopoly.c` says it's under "GNU Library General Public License" version 2, but the rest of the project is "GNU Lesser General Public License" version 2.1.

I'm guessing this was just a file that happened to be missed when the license was changed. Can you clarify if this was intentional or not? I'm happy to submit a PR to fix it if the intention was to also update it to LGPL-2.1.

Thanks!

cghawthorne:
Fixes #936


derselbst (approved):
Thanks!

Issue #928, opened 2021-07-01
If PulseAudio restarts, Fluidsynth stays running but no audio is produced


DestyNova:
### FluidSynth version
```
FluidSynth runtime version 2.1.1
Copyright (C) 2000-2020 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of E-mu Systems, Inc.

FluidSynth executable version 2.1.1
Sample type=double
```
(installed in Ubuntu 20.04 via apt)

### Describe the bug
FluidSynth is running as a daemon using `systemctl`, using the PulseAudio driver. Everything works fine until the PulseAudio service is restarted (I'm not sure why PulseAudio crashes, but it's almost certainly nothing to do with FluidSynth).
Afterwards, FluidSynth produces an error message on stdout, but keeps running and never produces audio again until I manually restart it.

### Expected behaviour
FluidSynth should terminate immediately with a non-zero exit code if the output driver fails. This would allow `systemctl` (or other daemon scripts) to restart FluidSynth and start a connection to the new instance of PulseAudio (if it's running).

### Actual behaviour
The following message is written to stdout:
```
fluidsynth: error: Error writing to PulseAudio connection.
```
FluidSynth stays running but produces no more audio until manually killed and restarted.

### Steps to reproduce
1. Start PulseAudio (e.g. with `systemctl --user start pulseaudio` or `pulseaudio --start`)
1. Start FluidSynth with the `pulseaudio` output driver (e.g. in my case `fluidsynth -is -g 2 -a pulseaudio -m alsa_seq -r 48000 /usr/share/sounds/sf3/MuseScore_General_Lite.sf3`)
1. Kill and restart PulseAudio.
1. Watch the FluidSynth stdout for an error, and try playing a MIDI file (e.g. `aplaymidi -p129:0 test.mid` assuming FluidSynth is using that MIDI port address).

Thanks for the great work on FluidSynth, by the way.

derselbst:
[...]

PR #940, opened 2021-07-13
Implement Android Oboe audio error recovery mode.

atsushieno:
Context: https://github.com/FluidSynth/fluidsynth/discussions/931

There is a new settings "audio.oboe.audio-error-recovery-mode" which has
string value of "Reconnect" (default) or "Stop".

Under `Reconnect` mode, it automatically recreate AudioStream for the
same audio device ID (which is the default = valid device, unless a
specific ID is specified). The behavior is the same as OpenSLES.

In the future Fluidsynth might want to provide consistent error handling
mode for audio device unplugged state, but so far this change makes apps
behave not too weird.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -591,6 +591,15 @@ and commit the results. Refresh with the following command:
             </desc>
         </setting>
         <setting>
+            <name>oboe.error-recovery-mode</name>
+            <type>str</type>
+            <def>Reconnect</def>
+            <vals>Reconnect, Stop</vals>
+            <desc>
+                Sets the error recovery mode when audio device error such as earphone disconnection occurred. It reconnects by default (same as OpenSLES behavior), but can be stopped if Stop is specified.
+            </desc>
+        </setting>
+        <setting>
             <name>oss.device</name>
             <type>str</type>
             <def>/dev/dsp</def>
--- a/src/drivers/fluid_oboe.cpp
+++ b/src/drivers/fluid_oboe.cpp
@@ -116,7 +116,7 @@ constexpr char OBOE_ID[] = "audio.oboe.id";
 constexpr char SHARING_MODE[] = "audio.oboe.sharing-mode";
 constexpr char PERF_MODE[] = "audio.oboe.performance-mode";
 constexpr char SRCQ_SET[] = "audio.oboe.sample-rate-conversion-quality";
-constexpr char RECOVERY_MODE[] = "audio.oboe.audio-error-recovery-mode";
+constexpr char RECOVERY_MODE[] = "audio.oboe.error-recovery-mode";
 
 void fluid_oboe_audio_driver_settings(fluid_settings_t *settings)
 {
--- a/src/drivers/fluid_oboe.cpp
+++ b/src/drivers/fluid_oboe.cpp
@@ -43,6 +43,7 @@ using namespace oboe;
 constexpr int NUM_CHANNELS = 2;
 
 class OboeAudioStreamCallback;
+class OboeAudioStreamErrorCallback;
 
 /** fluid_oboe_audio_driver_t
  *
@@ -55,7 +56,16 @@ typedef struct
     fluid_synth_t *synth = nullptr;
     bool cont = false;
     std::unique_ptr<OboeAudioStreamCallback> oboe_callback;
+    std::unique_ptr<OboeAudioStreamErrorCallback> oboe_error_callback;
     std::shared_ptr<AudioStream> stream;
+
+    double sample_rate;
+    int is_sample_format_float;
+    int device_id;
+    int sharing_mode; // 0: Shared, 1: Exclusive
+    int performance_mode; // 0: None, 1: PowerSaving, 2: LowLatency
+    oboe::SampleRateConversionQuality srate_conversion_quality;
+    int error_recovery_mode; // 0: Reconnect, 1: Stop
 } fluid_oboe_audio_driver_t;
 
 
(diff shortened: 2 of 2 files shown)


src/drivers/fluid_oboe.cpp
```diff
@@ -115,6 +138,9 @@ void fluid_oboe_audio_driver_settings(fluid_settings_t *settings)
     fluid_settings_add_option(settings,   SRCQ_SET, "Medium");
     fluid_settings_add_option(settings,   SRCQ_SET, "High");
     fluid_settings_add_option(settings,   SRCQ_SET, "Best");
+
+    fluid_settings_register_str(settings, RECOVERY_MODE, "Reconnect", 0);
+    fluid_settings_add_option(settings, RECOVERY_MODE, "Stop");
```
derselbst, line 143: Missing `fluid_settings_add_option("Reconnect");`

src/drivers/fluid_oboe.cpp
```diff
@@ -269,5 +297,33 @@ void delete_fluid_oboe_audio_driver(fluid_audio_driver_t *p)
     delete dev;
 }
 
+
+void
+OboeAudioStreamErrorCallback::onErrorAfterClose(AudioStream *stream, Result result)
+{
+    if (dev->error_recovery_mode == 1) { // Stop
+        FLUID_LOG(FLUID_INFO, "Oboe driver encountered an error (such as earphone unplugged). Stopped.");
```
derselbst, line 305: `FLUID_ERR`

src/drivers/fluid_oboe.cpp
```diff
@@ -269,5 +297,33 @@ void delete_fluid_oboe_audio_driver(fluid_audio_driver_t *p)
...
+        return;
+    }
+    else
+        FLUID_LOG(FLUID_INFO, "Oboe driver encountered an error (such as earphone unplugged). Recovering...");
```
derselbst, line 309: `FLUID_WARN`

src/drivers/fluid_oboe.cpp
```diff
@@ -269,5 +297,33 @@ void delete_fluid_oboe_audio_driver(fluid_audio_driver_t *p)
     delete dev;
 }
 
+
+void
+OboeAudioStreamErrorCallback::onErrorAfterClose(AudioStream *stream, Result result)
+{
+    if (dev->error_recovery_mode == 1) { // Stop
+        FLUID_LOG(FLUID_INFO, "Oboe driver encountered an error (such as earphone unplugged). Stopped.");
+        return;
```
derselbst, line 306: In "Stop" mode you return here. The stream remains set but is now closed. Is it safe when the stream will be stopped() and closed() again in `delete_fluid_oboe_audio_driver` later on?

doc/fluidsettings.xml
```diff
@@ -590,6 +590,15 @@ and commit the results. Refresh with the following command:
                 Sets the performance mode as pointed out by Oboe's documentation.
             </desc>
         </setting>
+        <setting>
+            <name>oboe.error-recovery-mode</name>
```
derselbst, line 594: There is a mismatch: The setting is called `audio.oboe.audio-error-recovery-mode` in the code.

On the other hand, I don't think a driver specific setting will be necessary. Maybe we should promote this to a general setting right away. Something like `audio.error-recovery-mode` sounds good to me. This way it could be used by any driver.

In case you insist on a driver specific setting, I would vote for calling it `audio.oboe.error-recovery-mode`.

derselbst (commented):
Nice job, thanks! Some minor issues / questions below.

doc/fluidsettings.xml
atsushieno: Thanks for the review. I fixed all other issues, and I am all for making error recovery generic.

Though I think it should rather be done in common API level, not just in Oboe driver (implementation on those unsupported platforms can be left blank). Automatic reconnection is kind of hack and app developers, not driver developers (us), would rather like to get notified so that they can control the situation in their own way, not just by choosing a configuration option.

Would it sound good if we add a new error callback setter on `fluid_audio_driver_t` ? like:

```
typedef void (*fluid_audio_driver_error_callback_t)(fluid_audio_driver_t*, void*);
fluid_audio_driver_set_error_callback(fluid_audio_driver_t* driver, fluid_audio_driver_error_callback_t func, void* data);
```

Current settings-based error handler may be still useful, but at the sane time I'm not very inclined to move "Reconnect" behavior onto driver-agnostic plate, as (1) it is a hacky behavior, and (2) it requires driver-specific implementation. I'm just not opinionated and it can be implemented, left as is though.

atsushieno: Some investigation I made on device change callbacks:

- Apple https://developer.apple.com/library/archive/documentation/Audio/Conceptual/AudioSessionProgrammingGuide/HandlingAudioHardwareRouteChanges/HandlingAudioHardwareRouteChanges.html
- cannot find anything on win32. On wasapi I found this relevant https://social.msdn.microsoft.com/Forums/en-US/c145b1ce-187f-4b08-9a8b-b4baacf6591d/change-playback-device-of-an-existing-wasapi-audiosession
- UWP https://docs.microsoft.com/en-us/uwp/api/windows.media.devices.mediadevice.defaultaudiorenderdevicechanged?view=winrt-20348
- not sure about alsa/jack/pulseaudio etc. either.
- JUCE has error callback support in `juce_audio_devices` API https://docs.juce.com/master/classAudioIODeviceCallback.html#a7768d640daf7d5d56399ced8188b2b75
- there is an ongoing discussion at miniaudio https://github.com/mackron/miniaudio/issues/325

src/drivers/fluid_oboe.cpp
atsushieno: You're right, it is not safe to leave it assigned. Not it is reset.

doc/fluidsettings.xml
derselbst: I'm struggling with the callback function. Yes, it seems like a good idea and goes hand in hand with an idea I had in #928, but it also raises many questions:
* In which cases do you get an error callback?
  * Device disconnected?
  * Audio Server executable crashed? (think of pulseaudio or jack)
  * Buffer underrun??
* In which state will the driver be after an error occurred?
  *  Will it be a recoverable state?
  * If yes, how should the client app recover from it? The client app doesn't know anything about the internals of the driver.

This all is actually quite beyond the scope of this PR. So, I'd say to keep it as driver specific setting
 `audio.oboe.error-recovery-mode` for now. Because changing a driver specific setting in the future should be less painful than a general setting.

atsushieno: Agreed, let's make this PR simpler. It is now renamed in code.

derselbst (approved):
Thank you!

PR #942, opened 2021-07-18
Emscripten: Fix CMake issues

Ghabry:
Two small fixes to the CMakeLists to make fluidsynth build when using emscripten (WASM).

---

Not included (obviously):

Another major problem is glib because glib does not compile when using emscripten (it needs custom patches), so I patched out glib. With glib removed I can confirm that it works.


derselbst:
Ok, thanks! You're welcome to say a few words about how much you like glib and about how the alternative  C++11 approach could work for you: #847

PR #947, opened 2021-07-27
Restore support for Mac OS X 10.5 and earlier

### LINKED ISSUE #946: Fluidsynth does not build on macOS 10.5 and earlier
With #828, Fluidsynth no longer builds on MacOS 10.5 Leopard and earlier. The AudioComponent symbols added in that PR were introduced in 10.6, and so targeting earlier OSes will produce a compile-time error.

There has been discussion over at MacPorts on how to work around this issue (see https://trac.macports.org/ticket/63230 and https://github.com/macports/macports-ports/pull/11706). The consensus has been to check to see if fluidsynth is interested in supporting these OS versions, or if the project would prefer to leave the problem to downstream package maintainers.

In practical terms, the issue can likely be worked around with a few #ifdefs, e.g.

```
#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060
    ComponentDescription desc;
#else
    AudioComponentDescription desc;
#endif
```

etc.

Is this a feasible change from the project's perspective?

evanmiller:
Expands on #803 and closes #946.

Tested on macOS 11.4 as well as 10.4.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -152,3 +152,4 @@ Marcus Weseloh
 Jean-jacques Ceresa
 Vladimir Davidovich
 Tamás Korodi
+Evan Miller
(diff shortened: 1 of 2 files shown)


derselbst (approved):
Thanks!

PR #961, opened 2021-08-23
[LADSPA] handle lib not found when deleting effect

albedozero:
If a LADSPA library is not found, delete_fluid_ladspa_effect just frees the effect and returns, since nothing else in the struct has been allocated.


albedozero:
Fixes #962

derselbst (changes requested):
Thanks for the report! The problem is a NULL dereference which happens when looking up `effect->desc->PortCount`, where `desc` is NULL. Your proposed change doesn't fix this, unfortunately. If e.g. `get_plugin_descriptor()` returned NULL, it would still crash with your change, because `lib != NULL` and `desc == NULL`. I've proposed different fix, pls. have a look: #963

albedozero:
That makes more sense to me now, thanks! #963 does it.

PR #966, opened 2021-08-28
Fix build with old CMake

derselbst:


derselbst:
@mawe42 @pedrolcl  We have a tiny problem: The change #943 broke compilation with old CMake older 3.11 with this error:

> Bad COMPATIBILITY value used for WRITE_BASIC_CONFIG_VERSION_FILE():   "SameMinorVersion"

To fix it, we have the following options:

1. Add a `CMAKE_VERSION` guard to skip that section if CMake is too old
1. Raise the minimum required CMake version to 3.13.3
1. Use a different compatibility (e.g. `SameMajorVersion`)

Option 1. has the drawback that depending on the CMake version, some files may or may not be installed. This will be troublesome for packagers as they will have to put a bunch of `%if` clauses in their spec files to conditionally package those files or not.

Option 2. will sooner or later come anyway. Probably for fluidsynth 2.3.0, maybe 2.2.4, but surely not for the upcoming 2.2.3. Doing so also means to loose out-of-the-box support for some linux distros (see the failed ones [here](https://build.opensuse.org/package/show/home:derselbst:home:derselbst:anmp:PR-966/fluidsynth)).

Option 3. is not correct, technically, and might be confusing.

I tend to go with option 2., this would also mean to defer releasing #943 into some future release of fluidsynth. I'm open whether this would be a 2.2.4 or 2.3.0.

Do you have any opinions?

pedrolcl:
In my opinion, #943 does not need to be in master right now. It may be offered as a patch for developers/packagers wanting to use it, but in addition to the need to raise the CMake version to 3.13 that you've found, there is also the problem of over-linking, which would be exported with the library target to other applications using the FS library imported target, not only the CL client.

So, my proposal would be:

* revert the merge on master
* create a new branch for cmake modernization
* apply #943 to the new branch. Now we can start applying some additional changes to this branch like:
    + raise cmake_minimum_required ( VERSION 3.13 )
    + fix the over-linking
    + apply other work planned for this modernizing endeavor on the same branch

When the branch is ready for merge on master, party! :champagne:

mawe42:
I think raising the cmake version in a later release would be a good way forward. And I like the idea of a build-process-enhancement-branch that collects some more cmake changes and cleanup to justify the bump even more.

derselbst:
Ok, sounds good! However, I would really like to rebase master rather than reverting a merge commit, if you don't mind.

mawe42:
> I would really like to rebase master

Rebasing master is quite unusual and probably *very* unexpected for a lot of people. Is that really necessary?

pedrolcl:
> I would really like to rebase master rather than reverting a merge commit, if you don't mind.

I have no intention of giving git advice to anyone. Do whatever you prefer.

derselbst:
> very unexpected for a lot of people. Is that really necessary?

Ok, usual reverting then :) will do it later.

derselbst:
Changes reverted on master, new PR drafted. Closing this in favor of #969.

Issue #411, opened 2018-07-05
Prefer CMAKE_INSTALL_LIBDIR over LIB_SUFFIX


t-chaik:
### FluidSynth version 
1.1.11

### Current behavior
Fluidsynth CMake scripts uses a custom `LIB_SUFFIX` variable for managing `libdir`. 

### Expected behavior
CMake's GNUInstallDirs module defines `CMAKE_INSTALL_LIBDIR` for the same thing [1]. Would probably make sense for FluidSynth to use that, would be less confusing from a user point of view.

### Other information
[1] https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

derselbst:
Why? `LIB_SUFFIX` is pretty common and it will only be applied if the user doesnt override `LIB_INSTALL_DIR`. I dont see a benefit moving to GNUInstallDirs in this case.

PR #969, opened 2021-09-01
 CMake Modernizations

derselbst:
This is a WIP for modernizing our cmake build system. The ideas emerged from #847. Open points:

- [x] [exported targets](https://cmake.org/cmake/help/git-stage/guide/importing-exporting/)
- [x] raise cmake_minimum_required ( VERSION 3.13 )
- [x] fix the over-linking issue (i.e. fix the unit tests)
- [x] better OpenMP support
- [x] Avoiding global functions (`link_directories`, `include_directories`, etc)
- [x] Using the more common `GnuInstallDirs` (requested already in #411)
- [x] Remove `enable-debug`
- [x] Remove `enable-pkgconfig`
- [x] fix `PC_REQUIRES_PRIV` related overlinking in `fluidsynth.pc`
- [x] document `CMAKE_MSVC_RUNTIME_LIBRARY`
- [x] ~~better clang-tidy support~~ (don't remember what was meant by that, or who came up with it)
- [x] ~~see if/how  #912 could be incorporated~~ (this will be handled separately, as there are too many dependencies / conflicts on / with the CMake related changes here)
- [x] ~~Rethink `enable-fpe-check` and `enable-trap-on-fpe`~~ (other than merging both options, I have no idea here... keep it for now)


pedrolcl:
In /CMakeLists.txt#L851 I don't understand why 3.13.3 is requested, when the enclosing if() is checking for version 3.12.0

This message was introduced in PR #933 merged last month.

pedrolcl:
I would expect 3.13.5 instead of 3.13.3:
https://cmake.org/cmake/help/v3.13/release/3.13.html#id4

derselbst:
Oh, you're right. Seems like my brain-memory is corrupted, sorry.

pedrolcl:
> 86 successful and 14 pending checks 

The OBS builds in *unresolvable* state are the 14 pending forever checks. Wouldn't be better to exclude them for this branch?

derselbst:
> Wouldn't be better to exclude them for this branch?

Done. I hope the pending ones will be gone after the next push.

derselbst:
Looks like the recent FindOpenMP modernization broke compilation on Mac. This wasn't a problem before, because I silently added openMP support to our Mac Builds when I merged #975.

And you can ignore the obsolete circleCI build for now.

pedrolcl:
> Looks like the recent FindOpenMP modernization broke compilation on Mac. This wasn't a problem before, because I silently added openMP support to our Mac Builds when I merged #975.

Succeeded in macOS 10.14 but failed in 10.15 and 11.0

Not sure where is the problem. Looks like the OpenMP include directory is not being searched in the CLI program, but it should be inherited from the object library. Something smells funny.

derselbst:
@pedrolcl There are two open points, which AFAIR were from you:
* better clang-tidy support
* CMAKE_MSVC_RUNTIME_LIBRARY

Could you let me know what exactly you had in mind?

pedrolcl:
> @pedrolcl There are two open points, which AFAIR were from you:
> 
>     * better clang-tidy support

I don't think so. I've never mentioned clang-tidy, and I don't have any proposals or plans about it. 
 
>     * CMAKE_MSVC_RUNTIME_LIBRARY
>
> Could you let me know what exactly you had in mind?

It is understandable that you link that with me, because I've opened #840 and #842 proposing a new CMake option to enable or disable linking the static MSVC runtime, but you didn't agree and implemented #848 in another way.  It was you who mentioned during the discussions `CMAKE_MSVC_RUNTIME_LIBRARY`as a better alternative, but your patch was only setting the policy CMP0091 to NEW, and indeed this option requires CMake >= 3.15, so in my opinion what is missing is a little bit of documentation including some advice for users of older CMake versions.

derselbst:
> so in my opinion what is missing is a little bit of documentation including some advice for users of older CMake versions.

Ok, I've added a hint to the [wiki](https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake#common-tips-for-compiling-from-source).

The [advice](https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake#what-is-cmake) for users of older CMake is to use CMake 3.15  - sry, but in former CMake versions the `/MD` flag is hardcoded, AFAIK. So manually supplying `/MT` via CFLAGS wouldn't help. I've linked the regex hack that we used before, just in case any WindowsXP user out there wants to do that.

derselbst:
Ready for review

derselbst (commented):
LGTM, I'm merging this now. Thanks @pedrolcl !

PR #970, opened 2021-09-02
cmake modernization, tasks 1 & 2

pedrolcl:
* raised cmake_minimum_required ( VERSION 3.13 )
* fixed overlinking of library, cmdline client, and unit tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@
 
 # CMake based build system. Pedro Lopez-Cabanillas <plcl@users.sf.net>
 
-cmake_minimum_required ( VERSION 3.13.5 )
+cmake_minimum_required ( VERSION 3.13 )
 # 3.1 because of CMAKE_C_STANDARD
 # 3.11 because COMPATIBILITY SameMinorVersion in CMakePackageConfigHelpers
 # 3.13.5 because it is the latest supported in Windows XP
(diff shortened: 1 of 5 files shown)


derselbst (commented):
Ok, looks good in general, thanks. Two thoughts:

1. When bumping the cmake version we should also remove all the `if (CMAKE_VERSION)` checks that we don't need anymore. To my surprise, LGTM.com now fails because it only uses 3.13.4. I don't want to loose LGTM, so we could either ignore this for now and hope that Semmle upgrades their CMake someday. Or we go with 3.13.4 right away. In any case, pls. make sure to reflect the CMake version requirement in the contrib/.dsc and contrib/.spec file.

2. Do you see a chance to change `libfluidsynth-OBJ` to a static library to achieve a more implicit dependency handling? Rather than that explicit `INTERFACE_LINK_LIBRARIES` dance?

pedrolcl:
>     1. When bumping the cmake version we should also remove all the `if (CMAKE_VERSION)` checks that we don't need anymore. To my surprise, LGTM.com now fails because it only uses 3.13.4. I don't want to loose LGTM, so we could either ignore this for now and hope that Semmle upgrades their CMake someday. Or we go with 3.13.4 right away. In any case, pls. make sure to reflect the CMake version requirement in the contrib/.dsc and contrib/.spec file.

In my opinion, we should not explicitly require a PATCH version number. The CMake API we use depends on  v3.13 alone, and shouldn't matter the other numbers unless serious bugs are present. In this case, the distributions or 3rd party services should care of updating their cmake version. We know  distros like Debian who do not shine for its diligence in this area, you only need to look at https://packages.debian.org/sid/source/fluidsynth, but still...

You are right about the CMAKE_VERSION checks, and the version requirements at contrib/* documents.
 
>     2. Do you see a chance to change `libfluidsynth-OBJ` to a static library to achieve a more implicit dependency handling? Rather than that explicit `INTERFACE_LINK_LIBRARIES` dance?

CMake doesn't provide an equivalence to autotools/libtool's convenience libraries. The object libraries are the most similar feature, which provides transitive dependencies (since version 3.12), but it doesn't always work as it should. The last patch to fix the static target is an example of that. Replacing the object library by a static library would complicate exporting targets (for which the `INTERFACE_LINK_LIBRARIES` property is a requirement). Maybe we can store the list of dependency libraries in a single cache variable or global property, but other than that...

derselbst (approved):
> In my opinion, we should not explicitly require a PATCH version number. 

Ok, I'm fine with that.

> Replacing the object library by a static library would complicate exporting targets

Ok, let's stay with object lib.

Thanks!

PR #973, opened 2021-09-05
No global macros

pedrolcl:
* replaced link_directories() and include_directories() global macros by target scoped ones
* dependencies are imported targets when possible
* property dancing has been fixed


CMakeLists.txt
```diff
@@ -698,20 +704,10 @@ if ( enable-threads )
 endif ( enable-threads )
 
 unset ( HAVE_OPENMP CACHE )
-find_package ( OpenMP QUIET )
+find_package ( OpenMP 4.0 QUIET )
```
derselbst, line 707: Carefully with this one. I remember that often it doesn't specify `OpenMP_C_VERSION`. Instead it uses `OpenMP_C_SPEC_DATE`. We will have to review all the CI builds and see whether openMP is still found properly.

derselbst (commented):
Looks good, thanks! However, the [vcpkg pipeline](https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=5975&view=logs&jobId=8ea4b35f-0335-5740-ca4c-0b3d3ee680be&j=8ea4b35f-0335-5740-ca4c-0b3d3ee680be&t=2f3c6ca8-0250-595e-3ed2-772e12994eb8) has a problem with the imported pkgconfig targets.

BTW, the cmake option `enable-pkgconfig` is obsolete and could be removed. AFAIR, it was originally introduced to allow for simpler compilation with vcpkg, but it's not needed anymore.

pedrolcl:
Not only the vcpkg builds fail, but also LGTM analysis. They fail because similar errors:

From the vcpkg build logs:

~~~
CMake Error in src/CMakeLists.txt:
  Imported target "PkgConfig::LIBSNDFILE" includes non-existent path
    "C:/vcpkg/installed/arm-windows/debug/../include/opus"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not
  provide.
~~~

and 

~~~
CMake Error in src/CMakeLists.txt:
  Imported target "PkgConfig::LIBINSTPATCH" includes non-existent path
    "C:/vcpkg/installed/x86-windows/debug/../include/opus"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not
  provide.
~~~

This one is from [LGTM](https://lgtm.com/projects/g/FluidSynth/fluidsynth/logs/rev/pr-1ff67ac17e26486760addfb6b72bcc46c69c706b/lang:cpp/stage:Build%20merge_b3a7e68ff7a833e162a15182a05503229b164992):

~~~
-- Checking for module 'lash-1.0>=0.3'
--   Found lash-1.0, version 0.6.0


[2021-09-05 10:12:44] [build-stderr] CMake Error in src/CMakeLists.txt:
[2021-09-05 10:12:44] [build-stderr]   Imported target "PkgConfig::LASH" includes non-existent path
[2021-09-05 10:12:44] [build-stderr]     "/usr/include/lash-1.0"
[2021-09-05 10:12:44] [build-stderr]   in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
[2021-09-05 10:12:44] [build-stderr]   * The path was deleted, renamed, or moved to another location.
[2021-09-05 10:12:44] [build-stderr]   * An install or uninstall procedure did not complete successfully.
[2021-09-05 10:12:44] [build-stderr]   * The installation package was faulty and references files it does not
[2021-09-05 10:12:44] [build-stderr]   provide.
~~~

Some dependencies provide wrong include paths in their .pc files, this was not a problem when using global include_directories() but it breaks imported targets. There are two solutions:

* omit the faulty dependencies
* clear the imported targets directory properties. The former is easier.

CMakeLists.txt
pedrolcl: FindOpenMP.cmake has been always distributed with the CMake program, and was seriously buggy sometimes.

https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html

Shouldn't be a problem in cmake 3.13 anymore.

pedrolcl:
There is a problem originated by the usage of imported targets: the generated file `fluidsynth.pc` is broken now.

pedrolcl:
After the last commit, vcpkg x64 still fails but I can't do anything about it. The 14 OBS pending builds are still there with 'unresolvable' state,

derselbst:
Yes, I only removed the OBS builds from the cmake-modernization branch. They will stay on the master branch and on all new future branches for the time being.

Do we need a message printed by cmake whenever it strips away / sanitizes an include path? Perhaps at `VERBOSE` or `DEBUG` level?

pedrolcl:
> Do we need a message printed by cmake whenever it strips away / sanitizes an include path? Perhaps at `VERBOSE` or `DEBUG` level?

Although it is trivial to include such message, I find it superfluous. The new macro `sanitize_target_dirs()` should not be needed at all. It is only a hack to pass the CI checks. It is the job of the wrong upstream packages to debug their stuff. 

I've applied the macro only on a few suspicious modules, like *sndfile* and the (for a long time) deprecated *lash*. What are you going to do with this message? report bugs to the upstream projects?

pedrolcl:
After this PR is merged, I propose this update to the list of pending tasks on CMake modernization attached to #969 :

* [x] [exported targets](https://cmake.org/cmake/help/git-stage/guide/importing-exporting/)
* [x] raise cmake_minimum_required ( VERSION 3.13 )
* [x] fix the over-linking issue (i.e. fix the unit tests)
* [x] better OpenMP support
* [x] Avoiding global functions (`link_directories`, `include_directories`, etc)
* [ ] better clang-tidy support
* [ ] `CMAKE_MSVC_RUNTIME_LIBRARY`
* [ ] Using the more common `GnuInstallDirs` (requested already a few years ago, cannot find the issue anymore)
* [ ] Remove `enable-debug`
* [ ] Rethink `enable-fpe-check` and `enable-trap-on-fpe`

derselbst (approved):
> Although it is trivial to include such message, I find it superfluous.

Ok, it was just a thought.

Thanks!

PR #977, opened 2021-09-08
WIP: applied instrumentation to cmake scripts, for debugging

pedrolcl:
* OpenMP and sndfile debugging
*  imported targets directories debugging


pedrolcl:
Summary:

* macOS 10.14: Found OpenMP v:3.1 date:201107
* macOS 10.15: Found OpenMP v:4.5 date:201511
* macOS 11.0:   Found OpenMP v:5.0 date:201811

All of them build, and pass the unit tests.  The former condition for a valid OpenMP was:
` if ( ( NOT OpenMP_C_SPEC_DATE LESS "201307" ) OR NOT ( OpenMP_C_VERSION VERSION_LESS "4.0" ) )`
but, seems that v3.1 is enough? Maybe there is something missing or malfunctioning before v4.0 ?

Another question about macOS: all the builds have `enable-framework=on` but HomeBrew uses traditional Unix libraries instead of frameworks, so I would like builds with ` enable-framework=off` as well.

winXP detects: Found OpenMP v:2.0 date:200203 and builds OK too.

derselbst:
Great job, thanks!

> seems that v3.1 is enough? Maybe there is something missing or malfunctioning before v4.0 ?

We use `omp simd aligned()` which requires 4.0. Unknown pragmas will be silently ignored. That's also why WinXP builds fine: it doesn't use any openMP features although in the cmake summary it says 'yes'. To avoid this confusion I'd prefer to keep the 4.0 limit.

> all the builds have enable-framework=on but HomeBrew uses traditional Unix libraries instead of frameworks, so I would like builds with enable-framework=off as well.

Sure, I'll add it to master. Or do you want to add it right here?

When merging this, I would only cherry-pick 01138f91a94efa716759054affb21229c572a975. Or do you want to keep the cmake instrumentation?

pedrolcl:
> > seems that v3.1 is enough? Maybe there is something missing or malfunctioning before v4.0 ?
> 
> We use `omp simd aligned()` which requires 4.0. Unknown pragmas will be silently ignored. That's also why WinXP builds fine: it doesn't use any openMP features although in the cmake summary it says 'yes'. To avoid this confusion I'd prefer to keep the 4.0 limit.

OK, I will add the former condition for a valid OpenMP dependency.

Another thing I miss is some unit test using OpenMP.

> > all the builds have enable-framework=on but HomeBrew uses traditional Unix libraries instead of frameworks, so I would like builds with enable-framework=off as well.
> 
> Sure, I'll add it to master. Or do you want to add it right here?

I would prefer to have it on the cmake-modernization branch and later in master.

> When merging this, I would only cherry-pick [01138f9](https://github.com/FluidSynth/fluidsynth/commit/01138f91a94efa716759054affb21229c572a975). Or do you want to keep the cmake instrumentation?

That commit only fixes mistakes introduced on the former commit b965cf5 so both are needed.  I will remove some of the instrumentation, but keep the infrastructure on place for future reference and debugging.

Issue #476, opened 2018-11-28
Inappropriate use of lib64 on Mac


RJVB:
LIB_SUFFIX is set to 64 also on Mac. This is not appropriate (I actually don't know where it *is* appropriate...).

derselbst:
This exists in fluidsynth for a long time. Even the 1.x releases made
use of this. The last time I've touched this was in
f125e8a9c6b63b3ae85a46181ee1bf6d93b85ebe, without intending to change
any existing behaviour. I'm no mac user. If you want this to be
removed permanently, I suggest you ask at the mailing list.

P.S.: Some linux distributions make use of it.

2018-11-28 17:52 GMT+01:00, René Bertin <notifications@github.com>:

RJVB:
I think that I didn't trip over this because I used the configure script until now, which presumably finds the most appropriate path for the platform you're on.

On Debian and Ubuntu the correct paths would be $prefix/lib/$arch-linux-gnu, for instance (where arch is i386, x86_64 etc) and on Mac it would be $prefix/lib ; the same would be true for other systems that only support a single architecture, or that support multiple architectures in a single binary file (like on Mac).

I thought this was the issue tracker?!

RJVB:
Hold on while I verify that it is indeed a bug with more consequences on Ubuntu...

PR #978, opened 2021-09-11
Adopting GNUInstallDirs

pedrolcl:
This PR has the following effects:
* removed and/or replaced most definitions from DefaultDirs.cmake and getting the values from GNUInstallDirs.cmake 
* fixed DEFAULT_SOUNDFOUNT in Windows, now in "C:/ProgramData/soundfonts"
* removed LIB_SUFFIX
* replaced LIB_INSTALL_DIR by CMAKE_INSTALL_LIBDIR. Same for INCLUDE_INSTALL_DIR, etc.

This issue may look somewhat cosmetic, but has some important implications. The key is that, most of the time, the variables like CMAKE_INSTALL_LIBDIR will be correctly and automatically computed for each distribution and architecture, ruling out the need of assigning a custom value. Depending on the kind of downstream user...
* if he is a packager, he already knows how and when to populate the CMAKE_INSTALL_* variables, and FS is just another package using well known variable names.
* if he is a final user compiling FS by himself, he gets the correct install directories automatically, without having to worry about it anymore.  See for instance #411 and #476

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,20 +115,6 @@ if ( CMAKE_SYSTEM MATCHES "OS2" )
     set ( enable-ipv6 off )
 endif ( CMAKE_SYSTEM MATCHES "OS2" )
 
-# Initialize the library directory name suffix.
-if (NOT MINGW AND NOT MSVC AND NOT CMAKE_SYSTEM_NAME MATCHES "FreeBSD|DragonFly")
-if ( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-  set ( _init_lib_suffix "64" )
-else ( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-  set ( _init_lib_suffix "" )
-endif ( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-else ()
-  set ( _init_lib_suffix "" )
-endif()
-set ( LIB_SUFFIX ${_init_lib_suffix} CACHE STRING
-      "library directory name suffix (32/64/nothing)" )
-mark_as_advanced ( LIB_SUFFIX )
-
 # the default C standard to use for all targets
 set(CMAKE_C_STANDARD 90)
 
@@ -156,7 +142,7 @@ if ( POLICY CMP0063 ) # since version 3.3, CMake version 3.21.2 warns when the p
   cmake_policy ( SET CMP0063 NEW )
 endif ( POLICY CMP0063 )
 
-# Default install directory names
+# Default install directory names, some provided by GNUInstallDirs
 include ( DefaultDirs )
 
 # Basic C library checks
@@ -373,7 +359,7 @@ unset ( MACOSX_FRAMEWORK CACHE )
 if ( CMAKE_SYSTEM MATCHES "Darwin" )
   set ( DARWIN 1 )
   set ( CMAKE_INSTALL_NAME_DIR
-        ${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR} )
+        ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR} )
   if ( enable-coreaudio )
     check_include_file ( CoreAudio/AudioHardware.h COREAUDIO_FOUND )
     if ( COREAUDIO_FOUND )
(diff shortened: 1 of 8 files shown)


derselbst (commented):
LGTM. What's the motivation behind "fix" of the DEFAULT_SOUNDFOUNT on Windows? I understand that ProgramData may make more sense. On the other hand, this breaks with the existing convention. And ProgramData is hidden by default.

pedrolcl:
> LGTM. What's the motivation behind "fix" of the DEFAULT_SOUNDFOUNT on Windows? I understand that ProgramData may make more sense. On the other hand, this breaks with the existing convention. And ProgramData is hidden by default.

It would be a bad practice in Linux/Unix to place the default soundfonts subdirectory directly at the root directory. Why should be different in Windows? 

"C:\soundfonts\" can't be created by a normal user (on modern Windows versions and filesystems). It may be created by an administrator or an hypothetical installer, but it would be a bad practice because, like in the Linux [FHS](https://refspecs.linuxfoundation.org/fhs.shtml), there are standard locations on Windows for data files that don't change in day to day use:

* The directory pointed by the %PROGRAMDATA% environment variable (usually "C:\ProgramData") is system wide, accessible by all users and programs. See: https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-shell-setup-folderlocations-programdata
* The directory pointed by %LOCALAPPDATA%  (usually C:\Users\{username}\AppData\Local) is a private data directory for a single user.

Both directories are hidden by default in the windows explorer, and they are meant to be accessed using the environment variables (in the explorer, directly at the address bar). Programs are supposed to use the environment variables also, and that should be an even better fix for the cmake issue, especially if some day we provide an installer for windows.

derselbst:
I think if we use %PROGRAMDATA% we should place it in a `fluidsynth` subfolder. I'm not entirely convinced that ProgramData is the right place. But I also have no better idea ATM.

pedrolcl:
> I think if we use %PROGRAMDATA% we should place it in a `fluidsynth` subfolder. I'm not entirely convinced that ProgramData is the right place. But I also have no better idea ATM.

If a Windows fluidsynth installer were installing soundfonts for fluidsynth's own use, then yes: `%PROGRAMDATA%\fluidsynth` would be the correct place. The same as in Linux: the soundfonts should be placed on `/usr/share/fluidsynth` when installed by a fluidsynth's RPM or DEB package. But this is not the case. Fluidsynth is not installing any soundfont, it is **searching** for a shared soundfont, installed by a 3rd party. That is why the location is `/usr/share/soundfonts` in Linux.

Just for reference, Microsoft ships a DLS file in Windows on `C:\Windows\System32\drivers\gm.dls`. But of course they can do whatever pleases them.  For the rest of us, the placement of shared soundfonts at `%PROGRAMDATA%\soundfonts` seems the only sensible place. I am not the only one arriving to this conclusion. See for instance: https://www.mozart.co.uk/music-theory/midi-coolsoft.htm

I'm posting here the contents of the soundfonts folder in one of my test computers. The operating system is Windows 10 and the drive is formatted with NTFS, so there are synlinks like `default.sf2` pointing to my favorite soundfont ATM.

~~~
C:\>dir %PROGRAMDATA%\SoundFonts
 Directorio de C:\ProgramData\SoundFonts

15/09/2021  16:06    <DIR>          .
15/09/2021  16:06    <DIR>          ..
03/11/2006  17:46    <DIR>          ChoriumRevA
03/09/2021  23:45    <SYMLINK>      default.sf2 [GeneralUser GS 1.471\GeneralUser GS v1.471.sf2]
17/07/2020  11:29    <DIR>          GeneralUser GS 1.471
06/03/2021  01:41    <SYMLINK>      GeneralUser.sf2 [GeneralUser GS 1.471\GeneralUser GS v1.471.sf2]
15/09/2021  16:06    <SYMLINK>      gm.dls [C:\Windows\System32\drivers\gm.dls]
27/08/2021  14:46    <DIR>          TOH
~~~

derselbst (approved):
> If a Windows fluidsynth installer were installing soundfonts for fluidsynth's own use, then yes: %PROGRAMDATA%\fluidsynth would be the correct place. The same as in Linux: the soundfonts should be placed on /usr/share/fluidsynth when installed by a fluidsynth's RPM or DEB package. But this is not the case. Fluidsynth is not installing any soundfont, it is searching for a shared soundfont, installed by a 3rd party. That is why the location is /usr/share/soundfonts in Linux.

Ok, you convinced me. Thanks!

PR #980, opened 2021-09-15
Per-channel ALL_SOUND_OFF when seeking/stopping player

albedozero:
- Only send all sound off on channels which had notes playing
- Send it as ALL_SOUND_OFF CC to a MIDI router to route it to a different synth channel afterwards

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -1660,6 +1660,11 @@ new_fluid_player(fluid_synth_t *synth)
         player->track[i] = NULL;
     }
 
+    for(i = 0; i < MAX_NUMBER_OF_CHANNELS; i++)
+    {
+        player->notesoff_channels[i] = -1;
+    }
+
     player->synth = synth;
     player->system_timer = NULL;
     player->sample_timer = NULL;
@@ -2091,7 +2096,14 @@ fluid_player_callback(void *data, unsigned int msec)
     {
         if(fluid_atomic_int_get(&player->stopping))
         {
-            fluid_synth_all_notes_off(synth, -1);
+            for(i = 0; i < MAX_NUMBER_OF_TRACKS; i++)
+            {
+                fluid_synth_all_notes_off(synth, player->notesoff_channels[i]);
+                if(player->notesoff_channels[i] < 0)
+                {
+                    break;
+                }
+            }
             fluid_atomic_int_set(&player->stopping, 0);
         }
         return 1;
@@ -2121,7 +2133,14 @@ fluid_player_callback(void *data, unsigned int msec)
         seek_ticks = fluid_atomic_int_get(&player->seek_ticks);
         if(seek_ticks >= 0)
         {
-            fluid_synth_all_notes_off(synth, -1); /* avoid hanging notes */
+            for(i = 0; i < MAX_NUMBER_OF_TRACKS; i++)
+            {
+                fluid_synth_all_notes_off(synth, player->notesoff_channels[i]);
+                if(player->notesoff_channels[i] < 0)
+                {
+                    break;
+                }
+            }
         }
 
         for(i = 0; i < player->ntracks; i++)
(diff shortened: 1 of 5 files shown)


albedozero:
Discussion/explanation in #981

src/midi/fluid_midi.c
```diff
@@ -2080,18 +2089,30 @@ fluid_player_callback(void *data, unsigned int msec)
...
+    fluid_midi_event_set_type(&mute_event, CONTROL_CHANGE);
+    mute_event.param1 = ALL_SOUND_OFF;
+    mute_event.param2 = 1;
 
     if(fluid_player_get_status(player) != FLUID_PLAYER_PLAYING)
     {
         if(fluid_atomic_int_get(&player->stopping))
         {
-            fluid_synth_all_notes_off(synth, -1);
+			for(i = 0; i < synth->midi_channels; i++)
```
mawe42, line 2108: Some more tabs here to remove :-) (and further down)

src/midi/fluid_midi.h
```diff
@@ -39,6 +39,7 @@ fluid_midi_event_t *fluid_midi_parser_parse(fluid_midi_parser_t *parser, unsigne
 
 
 #define MAX_NUMBER_OF_TRACKS 128
+#define MAX_NUMBER_OF_CHANNELS 256
```
mawe42, line 42: If the player is only used for standard MIDI files, then 16 channels should be enough. If there is the possibility that people could use more channels via some means, then I would vote to make this a dynamic list (maybe with fluid_list)?

src/midi/fluid_midi.c
```diff
@@ -1604,6 +1604,10 @@ fluid_track_send_events(fluid_track_t *track,
             if(player->playback_callback)
             {
                 player->playback_callback(player->playback_userdata, event);
+                if(event->type == NOTE_ON && !player->channel_isplaying[event->channel])
```
mawe42, line 1607: I think we should also send ALL_CTRL_OFF in some cases, so in my opinion we don't need to filter on the event type here. So any MIDI message on a channel marks that channel as `isplaying`.

mawe42 (commented):
Looks really good, just some smaller ideas/comments.

I think it could be extended to also send ALL_CTRL_OFF when seeking. That way we get around the hanging notes problem with sustain pedals.

src/midi/fluid_midi.c
albedozero: They keep sneaking in! thanks

src/midi/fluid_midi.h
albedozero: I agree 16 is probably more than enough, but a user could have lots of instruments player, multiple players, etc. - plus I wanted to be in keeping with MIDI 2.0's 256-track limit. I did want to make it dynamic, though - I'll try a `fluid_list`

albedozero: Oof - I tried using a fluid_list, but it seems beyond my skill, or perhaps it's unworkable to use a fluid_list to hold `int`s. I'll leave this as an int array for now, it's a lot cleaner, at least, and 256 channels is the max we should need.

derselbst: `fluid_midi_event_t` is a structure that resembles MIDI 1.0 events. MIDI 2.0 is a different story, for which we have much more to do. This shouldn't be intermixed.

If one really has multiple players running, the only way of intercepting events I see is by setting a custom callback function for the player. This would intercept events before they're sent out to the synth. Internally, every player would still deal with 0-15 channels.

You cannot use `fluid_list` either, as appending elements would call malloc() from synth context.

Therefore, I'd vote to change max channels to 16.

src/midi/fluid_midi.c
derselbst: Like I said in the discussion, I would be fine if we refrain from sending ALL_CRTL_OFF for now. Checking for the event type is ok for me. However, you also have to check for `velocity != 0` to be sure it's a real noteon.

src/midi/fluid_midi.c
```diff
@@ -1660,6 +1664,11 @@ new_fluid_player(fluid_synth_t *synth)
         player->track[i] = NULL;
     }
 
+    for(i = 0; i < synth->midi_channels; i++)
+    {
+        player->channel_isplaying[i] = 0;
```
derselbst, line 1669: You probably want to do this array-initialization in fluid_player_reset(). Otherwise this array will never be updated between different files.

derselbst: And for readability you could use `FALSE` instead of `0`. Same for `1` above.

src/midi/fluid_midi.h
albedozero: 16 it is!

src/midi/fluid_midi.c
albedozero: I added checking for note velocity as well - I like limiting the channels that receive SOUND_OFF to the minimum number necessary

src/midi/fluid_midi.c
albedozero: done

derselbst (approved):
LGTM. Thanks!

PR #982, opened 2021-09-16
[RFC] PipeWire audio driver

sykhro:
An implementation for an audio driver using the (relatively) new PipeWire libraries.
So far it has only been tested with playback, I'm planning on setting up a benchmark for latency soon.

As of now the functionality is basic but it can easily be extended to support more channels and audio formats.

### DIFF
--- a/src/drivers/fluid_pipewire.c
+++ b/src/drivers/fluid_pipewire.c
@@ -1,0 +1,307 @@
+/* FluidSynth - A Software Synthesizer
+ *
+ * Copyright (C) 2003  Peter Hanappe and others.
+ * Copyright (C) 2021  E. "sykhro" Melucci
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA
+ */
+
+/* fluid_pipewire.c
+ *
+ * Audio driver for PipeWire.
+ *
+ */
+
+#include "fluid_synth.h"
+#include "fluid_adriver.h"
+#include "fluid_settings.h"
+
+#if PIPEWIRE_SUPPORT
+
+#include <pipewire/pipewire.h>
+#include <spa/param/audio/format-utils.h>
+
+/* At the moment, only stereo is supported */
+#define NUM_CHANNELS 2
+static const int stride = sizeof(float) * NUM_CHANNELS;
+
+typedef struct
+{
+    fluid_audio_driver_t driver;
+    fluid_audio_func_t user_callback;
+    void *data;
+
+    /* Used only with the user-provided callback */
+    float *lbuf, *rbuf;
+
+    int buffer_period;
+    struct spa_pod_builder *builder;
+
+    struct pw_thread_loop *pw_loop;
+    struct pw_stream *pw_stream;
+    struct pw_stream_events *events;
+
+} fluid_pipewire_audio_driver_t;
+
+
+/* Fast-path rendering routine with no user processing callbacks */
+static void fluid_pipewire_event_process(void *data)
+{
+    fluid_pipewire_audio_driver_t *drv = data;
+    struct pw_buffer *pwb;
+    struct spa_buffer *buf;
+    float *dest;
+
+    pwb = pw_stream_dequeue_buffer(drv->pw_stream);
+    if (!pwb)
+    {
+        FLUID_LOG(FLUID_WARN, "No buffers!");
+        return;
+    }
+
+    buf = pwb->buffer;
+    dest = buf->datas[0].data;
+    if (!dest)
+    {
+        return;
+    }
+
+    fluid_synth_write_float(drv->data, drv->buffer_period, dest, 0, 2, dest, 1, 2);
+    buf->datas[0].chunk->offset = 0;
+    buf->datas[0].chunk->stride = stride;
+    buf->datas[0].chunk->size = drv->buffer_period * stride;
+
+    pw_stream_queue_buffer(drv->pw_stream, pwb);
+}
+
+/* Rendering routine with support for user-defined audio manipulation */
+static void fluid_pipewire_event_process2(void *data)
+{
+    fluid_pipewire_audio_driver_t *drv = data;
+    struct pw_buffer *pwb;
+    struct spa_buffer *buf;
+    float *dest;
+    float *channels[NUM_CHANNELS] = { drv->lbuf, drv->rbuf };
+    int i;
+
+    pwb = pw_stream_dequeue_buffer(drv->pw_stream);
+    if (!pwb)
+    {
+        FLUID_LOG(FLUID_WARN, "No buffers!");
+        return;
+    }
+
+    buf = pwb->buffer;
+    dest = buf->datas[0].data;
+    if (!dest)
+    {
+        return;
+    }
+
+    FLUID_MEMSET(drv->lbuf, 0, drv->buffer_period * sizeof(float));
+    FLUID_MEMSET(drv->rbuf, 0, drv->buffer_period * sizeof(float));
+
+    (*drv->user_callback)(drv->data, drv->buffer_period, 0, NULL, NUM_CHANNELS, channels);
+
+    /* Interleave the floating point data */
+    for (i = 0; i < drv->buffer_period; i++)
+    {
+        dest[i * 2] = drv->lbuf[i];
+        dest[i * 2 + 1] = drv->rbuf[i];
+    }
+
+    buf->datas[0].chunk->offset = 0;
+    buf->datas[0].chunk->stride = stride;
+    buf->datas[0].chunk->size = drv->buffer_period * stride;
+
+    pw_stream_queue_buffer(drv->pw_stream, pwb);
+}
+
+fluid_audio_driver_t *new_fluid_pipewire_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
+{
+    return new_fluid_pipewire_audio_driver2(settings, NULL, synth);
+}
+
+fluid_audio_driver_t *
+new_fluid_pipewire_audio_driver2(fluid_settings_t *settings, fluid_audio_func_t func, void *data)
+{
+    fluid_pipewire_audio_driver_t *drv;
+    int period_size;
+    int buffer_length;
+    int res;
+    int pw_flags;
+    int realtime_prio = 0;
+    double sample_rate;
+    char *media_role = NULL;
+    char *media_type = NULL;
+    char *media_category = NULL;
+    float *buffer = NULL;
+    const struct spa_pod *params[1];
+
+    drv = FLUID_NEW(fluid_pipewire_audio_driver_t);
+    if (!drv)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
+        return NULL;
+    }
+    FLUID_MEMSET(drv, 0, sizeof(*drv));
+
+    fluid_settings_getint(settings, "audio.period-size", &period_size);
+    fluid_settings_getint(settings, "audio.realtime-prio", &realtime_prio);
+    fluid_settings_getnum(settings, "synth.sample-rate", &sample_rate);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-role", &media_role);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-type", &media_type);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-category", &media_category);
+
+    drv->data = data;
+    drv->user_callback = func;
+    drv->buffer_period = period_size;
+
+    drv->events = FLUID_NEW(struct pw_stream_events);
+    if (!drv->events)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
+        goto driver cleanup;
+    }
+    FLUID_MEMSET(drv->events, 0, sizeof(*drv->events));
+    drv->events->version = PW_VERSION_STREAM_EVENTS;
+    drv->events->process = func ? fluid_pipewire_event_process2 : fluid_pipewire_event_process;
+
+    drv->pw_loop = pw_thread_loop_new("fluid_pipewire", NULL);
+    if (!drv->pw_loop)
+    {
+        FLUID_LOG(FLUID_ERR, "Failed to allocate PipeWire loop");
+        goto driver_cleanup;
+    }
+
+    drv->pw_stream = pw_stream_new_simple(
+    pw_thread_loop_get_loop(drv->pw_loop),
+    "FluidSynth",
+    pw_properties_new(PW_KEY_MEDIA_TYPE, media_type, PW_KEY_MEDIA_CATEGORY, media_category, PW_KEY_MEDIA_ROLE, media_role, NULL),
+    drv->events,
+    drv);
+    if (!drv->pw_stream)
+    {
+        FLUID_LOG(FLUID_ERR, "Failed to allocate PipeWire stream");
+        goto driver_cleanup;
+    }
+
+    buffer = FLUID_ARRAY(float, NUM_CHANNELS *period_size);
+    if (!buffer)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
+        goto driver_cleanup;
+    }
+    buffer_length = period_size * sizeof(float) * NUM_CHANNELS;
+
+    drv->builder = FLUID_NEW(struct spa_pod_builder);
+    if (!drv->builder)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
+        goto driver_cleanup;
+    }
+    FLUID_MEMSET(drv->builder, 0, sizeof(*drv->builder));
+    drv->builder->data = buffer;
+    drv->builder->size = buffer_length;
+
+    if (func)
+    {
+        drv->lbuf = FLUID_ARRAY(float, period_size);
+        drv->rbuf = FLUID_ARRAY(float, period_size);
+
+        if (!drv->lbuf || !drv->rbuf)
+        {
+            FLUID_LOG(FLUID_ERR, "Out of memory");
+            goto driver_cleanup;
+        }
+    }
+
+    params[0] = spa_format_audio_raw_build(drv->builder,
+                                           SPA_PARAM_EnumFormat,
+                                           &SPA_AUDIO_INFO_RAW_INIT(.format = SPA_AUDIO_FORMAT_F32,
+                                                                    .channels = 2,
+                                                                    .rate = sample_rate));
+
+    pw_flags = PW_STREAM_FLAG_AUTOCONNECT | PW_STREAM_FLAG_MAP_BUFFERS;
+    pw_flags |= realtime_prio ? PW_STREAM_FLAG_RT_PROCESS : 0;
+    res = pw_stream_connect(drv->pw_stream, PW_DIRECTION_OUTPUT, PW_ID_ANY, pw_flags, params, 1);
+    if (res < 0)
+    {
+        FLUID_LOG(FLUID_ERR, "PipeWire stream connection failed");
+        goto driver_cleanup;
+    }
+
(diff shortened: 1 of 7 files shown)


derselbst:
Oh, very nice! @SubhadeepJasu You're welcome to test this and give feedback.

SubhadeepJasu:
Audio is a bit glitchy when using higher number of channels. What settings would you recommend for this driver for average of 8 channels max 16 channel playback?

SubhadeepJasu:
These are my current settings:

```C
fluid_settings_setstr(style_synth_settings, "audio.driver", "pipewire");
fluid_settings_setint(style_synth_settings, "audio.period-size", 256);
fluid_settings_setint(style_synth_settings, "audio.realtime-prio", 1);
fluid_settings_setnum(style_synth_settings, "synth.gain", 2);
fluid_settings_setnum(style_synth_settings, "synth.overflow.percussion", 5000.0);
fluid_settings_setstr(style_synth_settings, "synth.midi-bank-select", "gs");
fluid_settings_setstr(style_synth_settings, "audio.pipewire.media-role", "Production");
fluid_settings_setstr(style_synth_settings, "audio.pipewire.media-type", "Audio");
```

SubhadeepJasu:
Update: It looks like lowering the period size reduces the glitches. But there's still a lot of crackling audio. Not sure how to fix that.

sykhro:
> Update: It looks like lowering the period size reduces the glitches. But there's still a lot of crackling audio. Not sure how to fix that.

Sounds like xruns, do you get the same with pulseaudio? Does setting `audio.realtime-prio` to 1 improve the situation? Does setting "Pro Audio" as pipewire profile change anything? (it can be done via pavucontrol).
I'll try and do some profiling to see if there's something I'm missing here.

SubhadeepJasu:
Setting "Pro Audio" in pavucontrol definitely fixes the issue, or at least I don't hear any crackling at all. I need to test it more to confirm. The priority settings only seem to affect the latency (which improves playing the instruments, priority 1 causes the synth output to be instant with no noticable delay, but without priority = 1 there is some delay), but I found no effect other than that when the style engine is playing with like 7 or 8 channels active and 16 channels in total.

As for the previous question. Pulseaudio had the opposite effect. Increasing period-size caused the audio crackling to go away, but if I increase it too much, there is much noticable latency which makes it difficult to play the app (as in play instruments in the app with a midi controller).

sykhro:
> The priority settings only seem to affect the latency

yes, that is intended.
I will try to ask around as the PipeWire API is severely undocumented but using the "Pro Audio" profile is probably the intended way when dealing with low-latency applications like yours.

SubhadeepJasu:
I need to know a bit more about "Pro Audio" profile. Is it something that is available to every distro regardless of whether pipewire is installed or not? Is it okay to recommend end users to configure pavucontrol to use this profile or would it be best that find a solution that works out of box with little configuration from end user?

sykhro:
> I need to know a bit more about "Pro Audio" profile. Is it something that is available to every distro regardless of whether pipewire is installed or not? Is it okay to recommend end users to configure pavucontrol to use this profile or would it be best that find a solution that works out of box with little configuration from end user?

It was added with PipeWire 0.3.22 (released 6 months ago), it's used to "have raw device access with the
maximum number of channels and no mixer controls"; it is, I believe, the option for people doing any sort of realtime audio work as it achieves very low latencies without straining the cpu much. I will try to see if there's a way to do this programmatically so that users don't have to do it, although I believe it's not supported. It is left, most likely, to people that know what they need

SubhadeepJasu:
I see. If we don't find a way to use pro audio out of box, I'll somehow have to direct users to pavucontrol. I plan to provide a simple button to launch pavucontrol. (if you know some other ways to do the same pls let me know because pavucontrol doesn't seem to be installed out of box in some distros). Anyway you are doing great work. Let's focus on the driver and performance for now.

mawe42:
@sykhro Thanks for this PR! I haven't looked at the code yet, but something you wrote got my attention. You said:
> [...] as the PipeWire API is severely undocumented [...]

That made me have a quick look at PipeWire with regard to API stability and I found [this](https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/FAQ#is-pipewire-ready-yet):
> The API/ABI has been declared stable since version 0.3.

That does sound ok. However, the FAQ also contains the [following](https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/FAQ#what-audio-api-do-you-recommend-to-use):
> ### What Audio API Do You Recommend To Use?
> The situation is a bit like GUI toolkits. There are many, each with different use cases. Nobody uses the native display server protocols directly (X11, Wayland) but always through an abstraction layer (GTK, Qt, etc).
> **We recommend that you continue to use PulseAudio, JACK and ALSA API's for now.**

So I must admit I do wonder if it is a good idea to implement a PipeWire driver in FluidSynth after all, especially if the PipeWire devs themselves recommend against using the native API.

sykhro:
> So I must admit I do wonder if it is a good idea to implement a PipeWire driver in FluidSynth after all, especially if the PipeWire devs themselves recommend against using the native API.

I guess it could be useful on embedded systems where you don't want to mess around too much with getting JACK to work(?). Besides that, honestly, I'm not too sure either. This was more of an exercise to see what the API looks like and see if there's any benefits wrt to latency (compared to pulseaudio, at least)

mawe42:
Ah, I just noticed that @SubhadeepJasu [found the same info](https://github.com/FluidSynth/fluidsynth/discussions/976#discussioncomment-1299280) 8 days ago already :-) 

> I guess it could be useful on embedded systems where you don't want to mess around too much with getting JACK to work(?).

Well, considering that the PipeWire devs say that PipeWire will "possibly" be as good as JACK (especially with regard to latency), JACK (or straight ALSA) might be the better choice for embedded until PipeWire is more mature. I don't know...

> Besides that, honestly, I'm not too sure either. This was more of an exercise to see what the API looks like and see if there's any benefits wrt to latency (compared to pulseaudio, at least)

Ah, I see. Ok.... maybe keep this driver around, either in this PR, in a pipewire-devel branch or in a forked repo or yours? What do you think, @derselbst ?

SubhadeepJasu:
I see minor improvements when using pipewire core as opposed to pipewire pulse. Like the fluidplayer plays at the correct tempo now (previously it was a slightly slower) . With the pro audio profile it seems like the best I have heard. Also have low level access to the core api which is also very well managed it nice. And it's definitely waay easy than jack. And I can hear other applications while doing production work.

derselbst:
> maybe keep this driver around, either in this PR, in a pipewire-devel branch or in a forked repo or yours? What do you think, @derselbst ?

I would leave this decision to @sykhro and @SubhadeepJasu. They seem to have a use-case for it. And once they have completed their testing and find it useful, I'm willing to merge this into master. We will probably (have to) keep it around in this PR for a few more weeks - I also want to give it a test, but I'm not sure when I find time for this.

Additionally, we should ensure that this driver will be built by our Linux CI build. So if the API unexpectedly changes, we see it instantly.

If everything goes well and we do merge it, we could also think about deprecating the OSS driver in turn.

SubhadeepJasu:
Crackling is back even with Pro Audio profile selected. Not sure why tho. And the period size is set to the lowest value, anything higher causes more crackling. May be something to do with that.

SubhadeepJasu:
> Crackling is back even with Pro Audio profile selected. Not sure why tho. And the period size is set to the lowest value, anything higher causes more crackling. May be something to do with that. May be my computer is weird. Or I need to recompile fluidsynth again.

Just Tested on my new laptop. Now it sounds great. No sound issues. So that probably was my old laptop dying.

sykhro:
> > maybe keep this driver around, either in this PR, in a pipewire-devel branch or in a forked repo or yours? What do you think, @derselbst ?
> 
> I would leave this decision to @sykhro and @SubhadeepJasu. They seem to have a use-case for it. And once they have completed their testing and find it useful, I'm willing to merge this into master. We will probably (have to) keep it around in this PR for a few more weeks - I also want to give it a test, but I'm not sure when I find time for this.
> 
> Additionally, we should ensure that this driver will be built by our Linux CI build. So if the API unexpectedly changes, we see it instantly.
> 
> If everything goes well and we do merge it, we could also think about deprecating the OSS driver in turn.

I'll try to do some benchmarking asap. For me it's a nice-to-have but definitely not super high priority

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    drv->pw_stream = pw_stream_new_simple(
+    pw_thread_loop_get_loop(drv->pw_loop),
+    "FluidSynth",
+    pw_properties_new(PW_KEY_MEDIA_TYPE, media_type, PW_KEY_MEDIA_CATEGORY, media_category, PW_KEY_MEDIA_ROLE, media_role, NULL),
+    &events,
+    drv);
+
+    buffer = FLUID_ARRAY(float, NUM_CHANNELS * period_size);
```
derselbst, line 196: Missing NULL check

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+
+    pw_init(NULL, NULL);
+    drv->pw_loop = pw_thread_loop_new("fluid_pipewire", NULL);
+    if (!drv->pw_loop)
+    {
+        FLUID_LOG(FLUID_ERR, "Failed to allocate PipeWire loop");
+        goto driver_cleanup;
+    }
+
+    drv->pw_stream = pw_stream_new_simple(
```
derselbst, line 189: Missing NULL chk

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+    pw_thread_loop_get_loop(drv->pw_loop),
+    "FluidSynth",
+    pw_properties_new(PW_KEY_MEDIA_TYPE, media_type, PW_KEY_MEDIA_CATEGORY, media_category, PW_KEY_MEDIA_ROLE, media_role, NULL),
+    &events,
+    drv);
+
+    buffer = FLUID_ARRAY(float, NUM_CHANNELS * period_size);
+    buffer_length = period_size * sizeof(float) * NUM_CHANNELS;
+
+    drv->builder = FLUID_NEW(struct spa_pod_builder);
```
derselbst, line 199: NULL chk

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+
+    if (drv->pw_loop)
+    {
+        pw_thread_loop_destroy(drv->pw_loop);
+    }
+
+    FLUID_FREE(drv->lbuf);
+    FLUID_FREE(drv->rbuf);
+    FLUID_FREE(drv->builder->data);
+    FLUID_FREE(drv->builder);
```
derselbst, line 265: NULL ptr deref if `drv->builder == NULL`

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+    fluid_settings_dupstr(settings, "audio.pipewire.media-type", &media_type);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-category", &media_category);
+
+    drv->data = data;
+    drv->user_callback = func;
+    drv->buffer_period = period_size;
+
+    events.process = func ? fluid_pipewire_event_process2 : fluid_pipewire_event_process;
+
+    pw_init(NULL, NULL);
```
derselbst, line 181: This is dangerous to do in here. When creating the driver multiple times, it would init multiple times. And every time it would change some static variables in global memory. I suggest to completely eleminate this call and leave it to downstream apps to properly init() and deinit(), like we decided to do for SDL2 driver.

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+
+    pw_stream_queue_buffer(drv->pw_stream, pwb);
+}
+
+fluid_audio_driver_t *new_fluid_pipewire_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
+{
+    return new_fluid_pipewire_audio_driver2(settings, NULL, synth);
+}
+
+static struct pw_stream_events events = {
```
derselbst, line 139: It is not clear to me why this needs to be static. Could you pls. add a comment?

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,277 @@
...
+} fluid_pipewire_audio_driver_t;
+
+/* Fast-path rendering routine with no user processing callbacks */
+static void fluid_pipewire_event_process(void *data)
+{
+    fluid_pipewire_audio_driver_t *drv = data;
+    struct pw_buffer *pwb;
+    struct spa_buffer *buf;
+    float *dest;
+    int stride;
```
derselbst, line 63: As `stride` used twice, consider making it `static const` in a global context.

derselbst (commented):
I gave it a try. Unfortunately pipewire 0.3.24 fails to create the driver on my machine:

```
[E][pipewire.c:335 add_interface()] can't get Spa:Pointer:Interface:Log interface -2
[E][pipewire.c:368 load_journal_logger()] can't get log interface -2
[E][pipewire.c:335 add_interface()] can't get Spa:Pointer:Interface:CPU interface -2
[E][loop.c:85 pw_loop_new()] loop 0x60400000cc50: can't make support.system handle: No such file or directory
fluidsynth: error: Failed to allocate PipeWire loop
```

There's a similar but shorter error in 0.3.37:

```
[E][loop.c:84 pw_loop_new()] loop 0x60400000f750: can't make support.system handle: No such file or directory
fluidsynth: error: Failed to allocate PipeWire loop
```

Don't know, what it means. No time to investigate further. In any case, fluidsynth crashes afterwards, because of NULL ptr deref. Hence I flagged some issues in your code, see below. Sry for being so nit-picky.

Sadly, only by including the pipewire header we get  a bunch of `warning: cast discards ‘const’ qualifier from pointer target type`

src/drivers/fluid_pipewire.c
SubhadeepJasu: How will we access the init and deinit functions downstream?

derselbst: You'll have to include the header.

src/drivers/fluid_pipewire.c
sykhro: it doesn't, actually, let me fix this

src/drivers/fluid_pipewire.c
sykhro: Done, will need to find a way to detect that pw hasn't been init.

sykhro:
> I gave it a try. Unfortunately pipewire 0.3.24 fails to create the driver on my machine:
> 
> ```
> [E][pipewire.c:335 add_interface()] can't get Spa:Pointer:Interface:Log interface -2
> [E][pipewire.c:368 load_journal_logger()] can't get log interface -2
> [E][pipewire.c:335 add_interface()] can't get Spa:Pointer:Interface:CPU interface -2
> [E][loop.c:85 pw_loop_new()] loop 0x60400000cc50: can't make support.system handle: No such file or directory
> fluidsynth: error: Failed to allocate PipeWire loop
> ```
> 
> There's a similar but shorter error in 0.3.37:
> 
> ```
> [E][loop.c:84 pw_loop_new()] loop 0x60400000f750: can't make support.system handle: No such file or directory
> fluidsynth: error: Failed to allocate PipeWire loop
> ```
> 
> Don't know, what it means. No time to investigate further. In any case, fluidsynth crashes afterwards, because of NULL ptr deref. Hence I flagged some issues in your code, see below. Sry for being so nit-picky.
> 
> Sadly, only by including the pipewire header we get a bunch of `warning: cast discards ‘const’ qualifier from pointer target type`

Is PipeWire your audio server?

derselbst:
> Is PipeWire your audio server?

I don't think so. I haven't made any special setup. Only installed pipewire-alsa and hoped it would work out-of-the-box. Perhaps I'll find time to try pipewire-jack later on.

sykhro:
> > Is PipeWire your audio server?
> 
> I don't think so. I haven't made any special setup. Only installed pipewire-alsa and hoped it would work out-of-the-box. Perhaps I'll find time to try pipewire-jack later on.

I see. `pipewire` is the main library, which needs to be set up as the system audio server. The other packages act as some sort of translation layer that make things work oob: `pipewire-alsa` is for ALSA apps, `pipewire-pulse` is for pulseaudio apps, `pipewire-jack` is for JACK apps.

SubhadeepJasu:
Yeah you need the one that just says "pipewire". The others simply act as "adapters".

src/drivers/fluid_pipewire.c
SubhadeepJasu: oh

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,299 @@
...
+        return NULL;
+    }
+    FLUID_MEMSET(drv, 0, sizeof(*drv));
+
+    fluid_settings_getint(settings, "audio.period-size", &period_size);
+    fluid_settings_getint(settings, "audio.realtime-prio", &realtime_prio);
+    fluid_settings_getnum(settings, "synth.sample-rate", &sample_rate);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-role", &media_role);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-type", &media_type);
+    fluid_settings_dupstr(settings, "audio.pipewire.media-category", &media_category);
```
derselbst, line 167: `fluid_settings_dupstr` requires you to `FLUID_FREE` the strings later on, which you currently not do.

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,299 @@
...
+
+    drv->data = data;
+    drv->user_callback = func;
+    drv->buffer_period = period_size;
+
+    drv->events = FLUID_NEW(struct pw_stream_events);
+    if (!drv->events)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
+        return NULL;
```
derselbst, line 177: You're leaking the `drv` object here. Consider using `goto driver_cleanup`.

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,299 @@
...
+    drv->pw_stream = pw_stream_new_simple(
+    pw_thread_loop_get_loop(drv->pw_loop),
+    "FluidSynth",
+    pw_properties_new(PW_KEY_MEDIA_TYPE, media_type, PW_KEY_MEDIA_CATEGORY, media_category, PW_KEY_MEDIA_ROLE, media_role, NULL),
+    drv->events,
+    drv);
+    if (!drv->pw_stream)
+    {
+        FLUID_LOG(FLUID_ERR, "Failed to allocate PipeWire stream");
+        return NULL;
```
derselbst, line 199: Consider `goto driver_cleanup`

src/drivers/fluid_pipewire.c
```diff
@@ -0,0 +1,299 @@
...
+    FLUID_FREE(drv->events);
+
+    FLUID_FREE(drv);
+}
+
+void fluid_pipewire_audio_driver_settings(fluid_settings_t *settings)
+{
+    fluid_settings_register_str(settings, "audio.pipewire.media-role", "Music", 0);
+    fluid_settings_register_str(settings, "audio.pipewire.media-type", "Audio", 0);
+    fluid_settings_register_str(settings, "audio.pipewire.media-category", "Playback", 0);
```
derselbst, line 296: Those new `audio.pipewire.*`  have no options registered (what we usually do with `fluid_settings_add_option()`). Is that intended? If they only ever default to the standard option, do we actually need them?
In case we do need them, they should be documented in /doc/fluidsettings.xml. Just placing a link to the pipewire documentation should be sufficient.

derselbst (commented):
There hasn't been any activity for quite some time, so it seems to me that this is ready for merge. I gave it one more review and found a few more tiny issues, pls. see below. 

Also, I updated pipewire on my machine to 0.3.40 and now at least the NULL deref and crash is gone. I still haven't figured out how to get it running. Setting up a VM for this is a bit of an overkill, sry. But if it works well for you I would be happy to take this.

src/drivers/fluid_pipewire.c
sykhro: Yeah, I didn't set options because application developers can use different values here. e.g. when using fluidsynth as the audio engine in a game, one can set "audio.pipewire.media-role" to "Game" as per [the freedesktop specification](https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/Clients/ApplicationProperties/). This is for integration with the system's audio mixer. Not sure what to do here, besides copying all the values from the spec..

derselbst: NP, I'll take care of the documentation after merge.

derselbst (approved):
LGTM. I'll merge it now. This will be part of fluidsynth 2.3.0. Not sure when it'll be released, we have a few other open issues.
Thanks for contributing!

shmerl:
So should pipewire be an option for the fluidsynth now?

I see this with fluidsynth 2.3.1 (in Debian testing):

```
fluidsynth -a help
-a options (audio driver):
   'alsa','file','jack','oss','pulseaudio','sdl2'
```

Issue #985, opened 2021-09-27
Segfault on exit after OpenMP was added


realnc:
(This is on latest git master (926581851ed1a095ef5b8659f77b38272d57e624). Gentoo Linux, GCC 11.2.0.)

I'm seeing a segfault that happens about 20% of time at program exit. I tracked it down to this:

https://github.com/FluidSynth/fluidsynth/blob/926581851ed1a095ef5b8659f77b38272d57e624/src/sfloader/fluid_defsfont.c#L397-L400

I'm not very familiar with openmp. If I delete the two pragmas, the problem disappears. It also disappears if I only delete `#pragma omp parallel` and leave `#pragma omp single` in. However, if I leave `#pragma omp parallel` in and delete `#pragma omp single`, the segfault is back.

The segfault itself is inside `/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/libgomp.so`.

Edit:
Oh, I should also mention that this is a heisenbug. It does not segfault when trying to debug. It only crashes when not running in gdb.

derselbst:
If you only remove `#pragma omp single` you create a race condition. A crash is not surprising in this case.

I assume due to the heisenbug nature, you cannot give steps to reproduce? In this case, pls recompile with cmake option `-Denable-ubsan=1`, hope for a crash and post the log output.

Do you only observe this when loading a particular soundfont?

Also, pls. try to use a different (older) compiler with a different (older) libomp runtime and see if it's still crashing.

realnc:
I tried GCC 9.4.0, but unfortunately when building with it, it still links with the 11.2 runtime. Then I tried clang 12.0.1 instead, which comes with its own omp runtime. It doesn't crash there.

I enabled `-Denable-ubsan=1` and it doesn't report anything other than minor memory leaks at exit. I then edited `CMakeLists.txt` and used these build flags to enable the thread sanitizer:

```cmake
set ( CMAKE_C_FLAGS "-fsanitize=thread,undefined ${CMAKE_C_FLAGS}" )
set ( CMAKE_CXX_FLAGS "-fsanitize=thread,undefined ${CMAKE_CXX_FLAGS}" )
set ( CMAKE_EXE_LINKER_FLAGS "-fsanitize=thread,undefined ${CMAKE_EXE_LINKER_FLAGS}" )
[...]

PR #986, opened 2021-09-28
Add cmake build option for disabling OpenMP

realnc:
In my case this is because of issue #985, but there's also the use case of being able to test without OpenMP. For example, to disable it on my system in order to test if the crash had anything to do with OpenMP, I had to manually edit `CMakeLists.txt` each time I wanted to turn it off and on again.


derselbst (approved):
Ok

realnc:
What happened to this commit? The option is still there, but it has no effect since the `if (enable-openmp` check is now gone. I can't find the commit that removed it. It's like it magically disappeared...

derselbst:
You're right. It's gone on the master branch but still exists on the 2.2.x branch. Seems like it was accidentally removed when we refactored the build system. I'll re-add it later on.

PR #989, opened 2021-10-17
Make building executables, documentation and tests optional

realnc:
This is very useful for cross-compiling. For example with mingw on Linux. It's also useful when we're only interested in building libfluidsynth as a dependency of another program.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,11 @@ option ( enable-readline "compile readline lib line editing (if it is available)
 option ( enable-threads "enable multi-threading support (such as parallel voice synthesis)" on )
 option ( enable-openmp "enable OpenMP support (parallelization of soundfont decoding, vectorization of voice mixing, etc.)" on )
 
+# Options for build components
+option ( build-bins "build binary programs" on )
+option ( build-docs "build docs" on )
+option ( build-tests "build tests" on )
+
 # Platform specific options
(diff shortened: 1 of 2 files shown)


derselbst:
What exactly makes this "useful"? I understand that you probably don't want to compile the tests, as you cannot execute them anyway. But the tests are only built if `make check` is explicitly issued.

The documentation is only built if doxygen is found and if `make doxygen` is explicitly issued.

The example programs are only built, if `make demo` is issued.

If you only want to compile the library (i.e. without the executable), just type `make libfluidsynth`.

Sry, but I really don't see what makes this change useful.

realnc:
> What exactly makes this "useful"? I understand that you probably don't want to compile the tests, as you cannot execute them anyway. But the tests are only built if `make check` is explicitly issued.
> 
> The documentation is only built if doxygen is found and if `make doxygen` is explicitly issued.
> 
> The example programs are only built, if `make demo` is issued.
> 
> If you only want to compile the library (i.e. without the executable), just type `make libfluidsynth`.
> 
> Sry, but I really don't see what makes this change useful.

It's more in line with the kind of options other projects provide. But if you don't like that, that's fine. However, `make install` will build everything, even if I initially only did a `make libfluidsynth`.

derselbst:
Ok, then let's close this for now. If someone feels too annoyed about it, we can discuss it again.

PR #994, opened 2021-10-25
Rewind playlist if calling fluid_player_play after all loops complete

albedozero:
This pull adds a check to `fluid_player_play` to determine whether the playlist has reached the end by checking whether the player's `currentfile` is null, and if so sets `loop=1` so that the player callback can restart the playlist when it calls `player_playlist_load`.


derselbst (approved):
Great, thanks!

PR #1010, opened 2021-12-21
Deprecate OSS and MidiShare

derselbst:
Suggesting to deprecate OSS and MidiShare drivers. They're old, unmaintained and believed to be unused downstream. Adding a deprecation notice in upcoming 2.2.5 release.

### DIFF
--- a/src/drivers/fluid_midishare.c
+++ b/src/drivers/fluid_midishare.c
@@ -87,6 +87,17 @@ new_fluid_midishare_midi_driver(fluid_settings_t *settings,
     fluid_midishare_midi_driver_t *dev;
     int i;
 
+    FLUID_LOG(FLUID_WARN,
+              "\n\n"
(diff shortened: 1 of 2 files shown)


derselbst:
No objections. Merging now.

alarixnia:
We use the OSS driver on BSD systems. I'm willing to maintain it but I'm kinda stretched maintaining most of the BSD sound backends in many different libraries :thinking:

derselbst:
@alarixnia  Oh, that surprises me. I thought BSD is shipping this custom SNDIO driver? (#470)

alarixnia:
sndio is a userspace server shipped by default on OpenBSD and optionally on FreeBSD. The native mechanism for accessing MIDI devices is still fundamentally an OSS compatible /dev/(r)midiX, and for PCM output, /dev/(audio|dsp)X

derselbst:
Ok, thanks for explaining!

PR #1011, opened 2021-12-21
Port the Sequencer to C++11

derselbst:
This is an MVP-like example of how a port of fluidsynth to C++11 may look like. See related discussion #847.

TODOs

- [x] restructure `extern C` crap to fix MacOS build
- [ ] MinGW fails to compile (could be related to https://stackoverflow.com/a/17986785)
- [ ] better integration of `fluid_seq_queue`
- [ ] remove `std::find_if` code duplication

Update: In the discussion I mentioned that porting to C++ will keep the API stable. This may not completely hold true:

* At least one function in the sequencer currently returns `char*` – in C++ this function must return a `const char*` (must have)
* Many functions in the sequencer currently return `void` – since in C++ an exception may be thrown pretty much everywhere, it would be helpful when those functions would return FLUID_OK / FLUID_FAILED (nice-to-have though)

However, I'd consider those API changes to be pretty cosmetic though.

### DIFF
--- a/src/midi/fluid_seq.cpp
+++ b/src/midi/fluid_seq.cpp
@@ -116,9 +116,9 @@ struct _fluid_sequencer_t
     ~_fluid_sequencer_t()
     {
         /* cleanup clients */
-        for(auto& client : this->clients)
+        while(!this->clients.empty())
         {
-            this->unregister_client(client->id);
+            this->unregister_client(this->clients.back()->id);
         }
         delete_fluid_seq_queue(this->queue);
     }
--- a/src/midi/fluid_seq.cpp
+++ b/src/midi/fluid_seq.cpp
@@ -40,6 +40,8 @@ extern "C" {
 }
 #endif
 
+#include "fluid_cxx_wrapper.hpp"
+
 #include <mutex>
 #include <vector>
 #include <atomic>
@@ -137,7 +139,7 @@ struct _fluid_sequencer_t
         return (id);
     }
     
-    void unregister_client(fluid_seq_id_t id)
+    int unregister_client(fluid_seq_id_t id)
     {
         fluid_event_t evt;
         unsigned int now = this->get_tick();
@@ -156,7 +158,7 @@ struct _fluid_sequencer_t
         if(result == this->clients.end())
         {
             // provided client id is not in the list, probably already deleted
-            return;
+            return FLUID_FAILED;
         }
 
         // client found, remove it from the list to avoid recursive call when calling callback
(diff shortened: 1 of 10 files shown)


src/midi/fluid_seq.cpp
```diff
@@ -39,44 +57,251 @@
...
+typedef struct _fluid_sequencer_client_t
+{
+    fluid_seq_id_t id;
+    std::string name;
+    fluid_event_callback_t callback;
+    void *data;
+} fluid_sequencer_client_t;
+
 /* Private data for SEQUENCER */
 struct _fluid_sequencer_t
```
paulsapps, line 70: prob want to add "final" to these `struct _fluid_sequencer_t final`

src/midi/fluid_seq.cpp
```diff
@@ -39,44 +57,251 @@
...
+        while(!this->clients.empty())
+        {
+            this->unregister_client(this->clients.back()->id);
+        }
+        delete_fluid_seq_queue(this->queue);
+    }
+    
+    fluid_seq_id_t register_client(const char *name, fluid_event_callback_t callback, void *data)
+    {
+        std::unique_ptr<fluid_sequencer_client_t> client(new fluid_sequencer_client_t);
```
paulsapps, line 128: you can do auto client = std::make_unique<fluid_sequencer_client_t>() which is the more idiomatic way

src/utils/fluid_cxx_wrapper.hpp
```diff
@@ -0,0 +1,58 @@
...
+    catch( const std::bad_alloc& e )
+    {
+        FLUID_LOG(FLUID_PANIC, "Out of memory\n");
+    }
+    catch( const std::exception& e )
+    {
+        ss << e.what();
+        FLUID_LOG(FLUID_ERR, "%s\n", ss.str().c_str());
+    }
+    catch(...)
```
paulsapps, line 52: catch(...) is usually bad practise - I can't see how anything other than what is being caught here would get thrown?

src/midi/fluid_seq.cpp
derselbst: Maybe, I'll think about it.

src/midi/fluid_seq.cpp
derselbst: std::make_unique is C++14, which is not applicable here.

src/utils/fluid_cxx_wrapper.hpp
derselbst: Throwing exceptions beyond extern C functions results in undefined behavior. catch(...) makes sure that this will be avoided under all circumstances. Even if this is not obvious when simply looking at our portion of code.

Issue #1017, opened 2022-01-04
Soundfonts that use loop offsets rather than sample loop points do not loop.


DaforLynx:
### FluidSynth version
FluidSynth runtime version 2.2.4
Copyright (C) 2000-2021 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

### Describe the bug
When playing samples with loop points set in the instrument's "loop start offset" and "loop end offset" (as described in Polyphone), with loop points set from 0 - 0 in the sample itself, the sample will not loop.

### Expected behavior
The sample should loop properly from the loop start offset to the loop end offset, if there is no loop points set in the sample itself. This is the case for other soundfont players including Musescore's, SynthFont's, and in Polyphone itself.

### Steps to reproduce

1. Download the sample MIDI file and sample SF2 file below.
2. Open a command window in the fluidsynth parent folder.
3. Type `./fluidsynth "DPPt.sf2" "Lake.mid"` with the appropriate file locations.

### Additional context
As stated before, this seems to be a common feature in most other soundfont players. Worth noting is that I used VGMTrans to create the soundfont.

When discovering this, I tested it with the loop points set in the sample itself rather than using offset values, and it worked perfectly. Try replacing the old soundfont with the "fixed" soundfont in which I changed the string sound to the working method. And for the record it'd be an extremely arduous process for me to go through every sample of every soundfont coming out of VGMTrans and change these values.

[Sample MIDI file](https://drive.google.com/file/d/13jvug2e3r4PhigXUTspFliryIXfmjwYk/view?usp=sharing)
[Sample SF2](https://drive.google.com/file/d/1z86EzZbKev9JUNDPLunY57H2PFU-XzVG/view?usp=sharing)
["Fixed" Sample SF2](https://drive.google.com/file/d/1w0Rxe9th68GHZYn1yBQNYuhjaVf7j7oq/view?usp=sharing)

![image](https://user-images.githubusercontent.com/13303368/148135816-e9db8394-e08f-4812-83a2-7dc134a115cd.png)

The method on the left does not work in fluidsynth, the method on the right does.

derselbst:
Sample loops - we're now entering the gray zone of the SoundFont spec.

[...]

PR #1018, opened 2022-01-05
Prevent samples accidentally having their loops disabled

derselbst:
If a SoundFont sets `loopstart == loopend` and then uses loop-offset-modulators to fix up those loops assigning them with a valid position, the sample was previously switched to unlooped mode erroneously.

For the long story, see #1017.


derselbst:
@mawe42 Your comment made me think: I'm now accepting `loopstart == loopend` without modifications. I print a debug message in this case. And in case we fixed up any other loops due to reasons, there will now be a warning displayed after processing all the samples. I think this should be simple enough to go into 2.2.5

PR #1023, opened 2022-01-11
Extend MIDI auto connect

pedrolcl:
* Add midi auto connect support for winmidi
* Make minor adjustments for alsa_seq driver
* Populate the setting `midi.alsa.device` with available devices

For full context, please see discussion in #414.

Before this patch, all external devices were connected to the last MIDI IN port, applying the same mapping to all events.
After the patch, the external devices are connected in order to each MIDI IN port. When there are more devices than MIDI IN ports, the connections are distributed among all MIDI IN ports, sequentially.

For instance, when `synth.midi-channels` is 32, then 2 MIDI IN ports are created. If the user has 3 controllers: device1, device2 and device3 and nothing more, then autoconnect will do:

MIDI IN port 1 connected to device1 and device3.
MIDI IN port 2 connected to device2.

### DIFF
--- a/src/drivers/fluid_alsa.c
+++ b/src/drivers/fluid_alsa.c
@@ -35,6 +35,7 @@
 #define ALSA_PCM_NEW_HW_PARAMS_API
 #include <alsa/asoundlib.h>
 #include <sys/poll.h>
+#include <math.h>
 
 #include "fluid_lash.h"
 
@@ -919,6 +920,13 @@ static void fluid_alsa_seq_autoconnect_port_info(fluid_alsa_seq_driver_t *dev, s
     }
 
     FLUID_LOG(FLUID_INFO, "Connection of %s succeeded", pname);
+
+    /* Calculate the next port to be auto-connected. When all ports have been connected
+     * in sequence, start again from port 0 for the next auto-connection. */
+    dev->autoconn_dest.port++;
+    if (dev->autoconn_dest.port >= dev->port_count) {
+        dev->autoconn_dest.port  = 0;
+    }
 }
 
 // Autoconnect a single client port (by id) to autoconnect_dest if it has right type/capabilities
@@ -939,16 +947,19 @@ static void fluid_alsa_seq_autoconnect_port(fluid_alsa_seq_driver_t *dev, int cl
     fluid_alsa_seq_autoconnect_port_info(dev, pinfo);
 }
 
-// Connect available ALSA MIDI inputs to the provided port_info
-static void fluid_alsa_seq_autoconnect(fluid_alsa_seq_driver_t *dev, const snd_seq_port_info_t *dest_pinfo)
+// Connect available ALSA MIDI inputs
+static void fluid_alsa_seq_autoconnect(fluid_alsa_seq_driver_t *dev)
 {
     int err;
     snd_seq_t *seq = dev->seq_handle;
     snd_seq_client_info_t *cinfo;
     snd_seq_port_info_t *pinfo;
 
-    // subscribe to future new clients/ports showing up
-    if((err = snd_seq_connect_from(seq, snd_seq_port_info_get_port(dest_pinfo),
+    /* Subscribe to system:announce for future new clients/ports showing up.
+     * This subscription is made to the first port, and does not rotate
+     * dev->autoconn_dest, because this subscription never provides
+     * MIDI channel messages, only system (ALSA) messages. */
+    if((err = snd_seq_connect_from(seq, dev->autoconn_dest.port,
                                    SND_SEQ_CLIENT_SYSTEM, SND_SEQ_PORT_SYSTEM_ANNOUNCE)) < 0)
     {
         FLUID_LOG(FLUID_ERR, "snd_seq_connect_from() failed: %s", snd_strerror(err));
@@ -956,8 +967,6 @@ static void fluid_alsa_seq_autoconnect(fluid_alsa_seq_driver_t *dev, const snd_s
 
     snd_seq_client_info_alloca(&cinfo);
     snd_seq_port_info_alloca(&pinfo);
-
-    dev->autoconn_dest = *snd_seq_port_info_get_addr(dest_pinfo);
 
     snd_seq_client_info_set_client(cinfo, -1);
 
@@ -1100,7 +1109,7 @@ new_fluid_alsa_seq_driver(fluid_settings_t *settings,
     FLUID_MEMSET(port_info, 0, snd_seq_port_info_sizeof());
 
     fluid_settings_getint(settings, "synth.midi-channels", &midi_channels);
-    dev->port_count = midi_channels / 16;
+    dev->port_count = ceil(midi_channels / 16);
 
     snd_seq_port_info_set_capability(port_info,
                                      SND_SEQ_PORT_CAP_WRITE |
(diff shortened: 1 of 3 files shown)


jjceresa:
I'm not familiar with alsa_seq or alsa_raw driver, so would you please teach how the auto connect future is intended to behave from a musician user point of view. Assuming the following setup:

1)The synth system having 2 physical MIDI inputs (i.e device 1, device 2) feeding a single fluid synth instance with 16 MIDI channels(0 to 15).

2)The musician connect 2 MIDI keyboards , both transmitting on MIDI channel 0.

With "auto connect" set to on, does the alsa_seq (or alsa_raw) driver redirect the 2 MIDI keyboards to the fluid synth's MIDI channel 0 input ?.

What happens with auto connect set to off ?

Is there really 2 distinct  alsa MIDI driver in fluidsynth library ?. Sorry for my ignorance.

pedrolcl:
> I'm not familiar with alsa_seq or alsa_raw driver, so would you please teach how the auto connect future is intended to behave from a musician user point of view. Assuming the following setup:
> 
> 1)The synth system having 2 physical MIDI inputs (i.e device 1, device 2) feeding a single fluid synth instance with 16 MIDI channels(0 to 15).

ALSA sequencer provides virtual MIDI ports, each one supports a maximum of 16 channels. When the user sets "synth.midi-channels" to a value greater than 16, then additional virtual ports are created by the alsa_seq driver.

If the value of synth.midi-channels" is the default (16), then only a single MIDI IN port is created by the alsa_seq driver. The user may connect both device1 and device2 to the single port, and ALSA merges both MIDI streams into one. No mapping will be performed (the synth receives the events exactly as they have been generated).

When the value of synth.midi-channels" is for instance 32, the alsa_seq driver creates two MIDI ports. The user may connect device1 to the first port and device2 to the second port, by hand (using the console command `aconnect`, or one of the graphic connection tools available). The events sent by device1 are received by the synth on channels 1-16. The events sent by device2 are received by the synth on channels 17-32.  The mappings are associated to the virtual MIDI ports.

**This was already implemented in the alsa_seq driver, and nothing changed on this area with my patch.**
 
> 2)The musician connect 2 MIDI keyboards , both transmitting on MIDI channel 0.
> 
> With "auto connect" set to on, does the alsa_seq (or alsa_raw) driver redirect the 2 MIDI keyboards to the fluid synth's MIDI channel 0 input ?.

Before my patch, all external devices were connected to the last MIDI IN port, applying the same mapping to all events.
After the patch, the external devices are connected in order to each MIDI IN port. When there are more devices than MIDI IN ports, the connections are distributed among all MIDI IN ports, sequentially.

For instance, when synth.midi-channels is 32, then 2 MIDI IN ports are created. If the user has 3 controllers: device1, device2 and device3 and nothing more, then autoconnect will do:

* MIDI IN port 1 connected to device1 and device3.
* MIDI IN port 2 connected to device2.
 
> What happens with auto connect set to off ?

The user needs to use an external connection tool. He may connect every controller to the first MID IN port, or not. The mappings are associated to the MIDI IN ports anyway. It doesn't matter if he uses autoconnect or does the connections by hand in relation to the MIDI channel mappings.
 
> Is there really 2 distinct alsa MIDI driver in fluidsynth library ?. Sorry for my ignorance.

Yes. The alsa raw MIDI driver in Linux is like the winmidi driver. It has no virtual MIDI ports, and the driver has to open the MIDI devices, read events from them, and close the devices when finished.  The alsa sequencer MIDI driver is totally different, more like the CoreMIDI driver in macOS providing virtual MIDI ports, and virtual cables (MIDI routing).

jjceresa:
Thanks Pedro, very nice explanations about alsa_seq (virtual ports) and alsa_raw behaviour !.

>if the value of synth.midi-channels" is the default (16), then only a single MIDI IN port is created by the alsa_seq driver. The user may connect both device1 and device2 to the single port, and ALSA merges both MIDI streams into one.

Ok, that means that device1 MIDI streams on Channel x are merged with device2 MIDI streams on Channel x (with x in the range: 1..16). If 2 musicians play at the same time on both devices, this could lead in MIDI streams conflicts. Does I understand well ?

pedrolcl:
> Ok, that means that device1 MIDI streams on Channel x are merged with device2 MIDI streams on Channel x (with x in the range: 1..16). If 2 musicians play at the same time on both devices, this could lead in MIDI streams conflicts. Does I understand well ?

Of course, that is something expected. And that is why each musician should use different MIDI channels in that scenario. There are 16 MIDI channels, and each MIDI controller should allow to select the base channel. And to provide more flexibility, the Fluidsynth synthesizer has a maximum of 256 MIDI channels, that you may request using the setting "synth.midi-channels" that you choose to ignore.

What are you telling me? That every [MIDI merge box](http://midisolutions.com/prodmrg.htm) in the market is wrong because doesn't solve this problem that comes from the very deep guts of the MIDI 1.0 specification?

jjceresa:
As already said I am not familiar at all with ALSA stuff. I have found your explanations very useful. I just needed your confirmation about virtual ports. Thanks.

derselbst (commented):
I really hoped that some more people would comment on those changes. Anyway, fluidsynth 2.3.0 is about to come and  I would really like to have those change in. I need to make a decision which way to go forward. In the past days I had a look at them again and I have to admit that the `mergebox` setting suggested by @jjceresa did not convince me. Therefore, I will apply Pedro's proposal with this PR. If need be, we can think about any `mergebox`-like settings later. But for the moment, Pedro's suggestion seems to be more straightforward.

There are a few issues in your PR though. E.g. calling `alloca` in a loop. I will take care of them. Anotherone below.

src/drivers/fluid_winmidi.c
```diff
@@ -458,6 +467,22 @@ fluid_winmidi_parse_device_name(fluid_winmidi_driver_t *dev, char *dev_name)
     return dev_count;
 }
 
+static void fluid_winmidi_autoconnect_build_name(char *name)
+{
+    int i, j, n = 0;
+    int num = midiInGetNumDevs();
+    if (num < 11) {
```
derselbst, line 474: Why are you checking for `num < 11` here?

src/drivers/fluid_alsa.c
```diff
@@ -1100,7 +1164,7 @@ new_fluid_alsa_seq_driver(fluid_settings_t *settings,
     FLUID_MEMSET(port_info, 0, snd_seq_port_info_sizeof());
 
     fluid_settings_getint(settings, "synth.midi-channels", &midi_channels);
-    dev->port_count = midi_channels / 16;
+    dev->port_count = ceil(midi_channels / 16);
```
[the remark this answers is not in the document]
pedrolcl: Yes, I think so, I forgot to change one of the operands to float because the original calculation was doing also an integer division, with the same result as floor(). The intention was to create an additional port for the remaining MIDI channels when the number of channels is not a multiple of 16.

src/drivers/fluid_winmidi.c
pedrolcl: If I recall correctly, this is because MAXPNAMELEN is defined as 32 in the classic windows multimedia interface, so this was to prevent exceeding the available room in the device name. But when using Unicode, the limit may be 8 instead?

pedrolcl:
I'm afraid I don't have time right now to work on this, sorry.

src/drivers/fluid_winmidi.c
derselbst: Ok, no problem. I'll think about a more explicit check here.

derselbst (approved):
Should be good now. Thanks!

PR #1025, opened 2022-01-14
Initialize Pipewire with pw_init()

insilications:
Before you can use any PipeWire functions, you need to call pw_init() (e.g. https://docs.pipewire.org/page_tutorial1.html)

> fluidsynth -a pipewire /usr/share/soundfonts/YamahaGrand.sf2 Queen_-_Bohemian_Rhapsody.mid
> FluidSynth runtime version 2.2.4
> Copyright (C) 2000-2021 Peter Hanappe and others.
> Distributed under the LGPL license.
> SoundFont(R) is a registered trademark of Creative Technology Ltd.
> 
> [E] pw.context [pipewire.c:256 load_spa_handle()] load lib: plugin directory undefined, set SPA_PLUGIN_DIR
> [E] pw.loop [loop.c:86 pw_loop_new()] 0x5555558478a0: can't make support.system handle: No such file or directory
> fluidsynth: error: Failed to allocate PipeWire loop
> Failed to create the audio driver. Giving up.

After the fix, it works perfectly.

### DIFF
--- a/src/drivers/fluid_pipewire.c
+++ b/src/drivers/fluid_pipewire.c
@@ -155,6 +155,8 @@ new_fluid_pipewire_audio_driver2(fluid_settings_t *settings, fluid_audio_func_t 
     float *buffer = NULL;
     const struct spa_pod *params[1];
 
+    pw_init(NULL, NULL);
+
     drv = FLUID_NEW(fluid_pipewire_audio_driver_t);
 
     if(!drv)


derselbst (changes requested):
Thanks for the hint. However, it is not that simple. Whenever there's an init() function, there typically also is an uninit() one. `pw_deinit()` in this case. Those functions manipulate global variables. Calling them from a shared library like fluidsynth is hazardous to the rest of the entire application. We have to go with a similar solution as we did in case of SDL2: Calling `pw_init()` and `pw_deinit()` must be the responsibility of downstream applications. Unlike SDL2 however, pipewire does not seems to have a way to detect whether it already has been initialized. So unfortunately, fluidsynth cannot even issue a warning in this case.

In case of the fluidsynth executable, we have to call pw_init and deinit similar as we did for SDL2: https://github.com/FluidSynth/fluidsynth/blob/015c6af52d8233eaae556242b5200290558ba415/src/fluidsynth.c#L404-L411

derselbst:
Superseded by #1032

PR #1027, opened 2022-01-14
cmake macos frameworks update

### LINKED ISSUE #1024: make install does not completely install components on MacOS
### FluidSynth version
FluidSynth runtime version 2.2.4

FluidSynth executable version 2.2.4
Sample type=double

### Describe the bug
Installation from sources on MacOS is incomplete. The framework is installed with no headers, so it is impossible to use the API.

### Expected behavior
`make install `  installs all required components. Ideally, headers should be placed inside the framework, but at least they should be installed in a public location (e.g. ${CMAKE_INSTALL_PREFIX}/include). 

### Steps to reproduce
- `make install`
- try to compile a program using the Fluidsynth API.

### Additional context
Flluidsynth is used as a dependency for the Csound plugin fluidOpcodes. Installation worked correctly for 1.x, but not so for 2.2.4. I had to copy the headers by hand, which is not ideal. Note that pointing to use `brew` is not a solution for cases where we want to provide binary packages for users.

pedrolcl:
The framework target needs to know about the headers, so cmake tries to copy the framework files into the bundle.
The macro set_source_files_properties() is a modern (cmake >= 3.0) replacement of the general set_property().

This hopefully...
Closes #1024

### DIFF
--- a/cmake_admin/DefaultDirs.cmake
+++ b/cmake_admin/DefaultDirs.cmake
@@ -15,12 +15,12 @@ endif ( WIN32 )
 mark_as_advanced (DEFAULT_SOUNDFONT)
 
 # BUNDLE_INSTALL_DIR - Mac only: the directory for application bundles 
-set (BUNDLE_INSTALL_DIR "/Applications" CACHE STRING 
+set (BUNDLE_INSTALL_DIR "Applications" CACHE STRING
      "The install dir for application bundles")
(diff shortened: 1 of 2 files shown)


derselbst (commented):
Thanks for the fix again. I'd close this PR in favor of #1029. I've cherry-picked your commit there.

PR #1029, opened 2022-01-14
Fix framework installation on MacOS

### LINKED ISSUE #1024: make install does not completely install components on MacOS

derselbst:
This includes Pedro's fix from #1027 along with adjustments to the CI pipeline and rebased against the `2.2.x` branch.

Resolves #1024


derselbst:
@vlazzarini CI builds are passing now. Feel free to test.

vlazzarini:
Thx. Here's what I did:

```
git clone https://github.com/FluidSynth/fluidsynth.git
cd fluidsynth
mkdir build 
cd build 
make -j 6 
sudo make install 
```

However, it doesn't look like the headers are there.  

```
victor@MU-LAP-7ZM build % ls /Library/Frameworks/FluidSynth.framework           
FluidSynth	Resources	Versions
```
I also looked at /Versions/3 to see if it was just them symlink that wasn't created.

am I in the wrong branch?

derselbst:
You need the `mac-make-install` as written in the PR.

vlazzarini:
oh, sorry. OK, now I get it built correctly, but installation in `/usr/local/Library/Frameworks` is non-standard. So I try then 

```
cmake .. -DCMAKE_INSTALL_PREFIX=/  
```

but that fails with

```
sudo make install
```
...

```
Install the project...
-- Install configuration: "RelWithDebInfo"
CMake Error at cmake_install.cmake:41 (file):
  file cannot create directory: /lib64/pkgconfig.  Maybe need administrative
  privileges.
```

The default install configuration does not seem to be what you would expect on MacOS (sounds like it's for a Debian installation).

derselbst:
Have you cleaned your build folder before trying? Also, you're not supposed to manually set `CMAKE_INSTALL_PREFIX`, as this implementation relies on the fact that it's default initialized.

vlazzarini:
yes, sorry, cleaned my build and now it installs  correctly, thanks.

derselbst (commented):
Great, thanks for testing!

Issue #999, opened 2021-10-27
The daemon could work with multiple users


es20490446e:
### Version
2.2.3

### How to reproduce
1. Log in as user1
2. systemctl --user start fluidsynth
3. Switch user to user2
4. systemctl --user start fluidsynth
5. systemctl --user status fluidsynth

### Output
fluidsynth.service: Failed with result 'protocol'

### Result
Only the first logged in user is able to play midi

es20490446e:
@derselbst It seems a bug, since I have tested that the fluidsynth daemon cannot also work by itself with multiple users running it.

It doesn't even need to be simultaneously. Only the first user that will daemonize fluidsynth will be able to play midi, even if the daemon of the second terminates.

es20490446e:
And changing audio o midi driver doesn't seem to affect this fact.

derselbst:
I consider this to be a systemd related issue, not a bug in fluidsynth. To reproduce, it would probably helpful if you share the fluidsynth.service file and the distribution you're on.
I have little knowledge about systemd, and no idea, how to investigate or solve this. Contributions are welcome.

es20490446e:
Observe that the behavior is reproducible without systemd. It's just reporting that Fluid is trying to claim the same resource for both users, and that's not possible.

Without systemd:
1. Log in user1, run fluidsynth on the terminal (it doesn't matter the arguments).
2. aplaymidi -p 128:0 song.mid (get sound)
3. Change to user2, without log off user1, and start fluidsynth on the terminal.
4. Log off user1.
5. On user1 session: aplaymidi -p 128:0 song.mid (no sound).

The systemd service I have on my Manjaro system is the official one in this repository:
```
[Unit]
Description=FluidSynth Daemon
Documentation=man:fluidsynth(1)
After=sound.target

[Service]
Type=notify
NotifyAccess=main
EnvironmentFile=/etc/conf.d/fluidsynth
EnvironmentFile=-%h/.config/fluidsynth
ExecStart=/usr/bin/fluidsynth -is $OTHER_OPTS $SOUND_FONT

[Install]
WantedBy=default.target
```

Any idea what that conflict may be?

derselbst:
The first fluidsynth instance will most likely have client ID 128. The second instance will have a different client ID. It doesn't matter in which user you start it. Even if you start it in the same user, the ID will be different. By executing
[...]

PR #1031, opened 2022-01-15
Add autoconnect to winmidi driver

jjceresa:
This PR is an alternative to commit d031d4b of PR https://github.com/FluidSynth/fluidsynth/pull/1023

When midi.autoconnect is set to true it works as commit d031d4b of PR https://github.com/FluidSynth/fluidsynth/pull/1023.
When midi.autoconnect is false it works as actual.

### DIFF
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -33,7 +33,7 @@
  * user trough the settings midi.winmidi.device.
  *
  * If midi.winmidi.mergebox is set to 0 the driver makes an aggregation of the
- * devices on separate MIDI channels to ensure no MIDI channels conflicts.
+ * devices on separate MIDI channels to ensure no MIDI channel conflict.
  * For example, let the following device names:
  * 0:Port MIDI SB Live! [CE00], 1:SB PCI External MIDI, default, x[;y;z;..]
  * Then the driver is able receive MIDI messages comming from distinct devices
@@ -43,13 +43,13 @@
  * We get a fictive device composed of real devices (0,1). This fictive device
  * behaves like a device with 32 MIDI channels whose messages are forwarded to
  * driver output as this:
- * - MIDI messages from real device 0 are output to MIDI channels set 0 to 15.
- * - MIDI messages from real device 1 are output to MIDI channels set 16 to 31.
+ * - MIDI messages from real device 0 are mapped to MIDI channels set 0 to 15.
+ * - MIDI messages from real device 1 are mapped to MIDI channels set 16 to 31.
  *
  * 1.2)Now another example with the name "1;0". The driver will forward
  * MIDI messages as this:
- * - MIDI messages from real device 1 are output to MIDI channels set 0 to 15.
- * - MIDI messages from real device 0 are output to MIDI channels set 16 to 31.
+ * - MIDI messages from real device 1 are mapped to MIDI channels set 0 to 15.
+ * - MIDI messages from real device 0 are mapped to MIDI channels set 16 to 31.
  * So, the device order specified in the setting allows the user to choose the
  * MIDI channel set associated with this real device at the driver output
  * according this formula: output_channel = input_channel + device_order * 16.
@@ -65,14 +65,14 @@
  * any synths that could be a fluid synth and hardware synth(s) as well.
  *
  * If midi.winmidi.mergebox is set to 1 (default value). The driver map the device
- * according the limit of the synth.midi-channels setting.
+ * MIDI channels according the limit of the synth.midi-channels setting.
  * For example, if the user chooses 2 devices at index 0 and 1 and a default value
  * of 16 for synth.midi-channels setting:
  * - MIDI messages from real device 0 are output to MIDI channels set 0 to 15.
  * - MIDI messages from real device 1 are output to MIDI channels set 0 to 15.
  * Device 1 MIDI channels are merged with device 0 MIDI channel. This could lead
  * to possible MIDI channels conflicts if both MIDI devices transmit at the same
- * on the same MIDI channels.
+ * time on the same MIDI channel.
  *
  * 2)Note also that the driver handles single device by putting the device name
  * in midi.winmidi.device setting.
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -69,6 +69,19 @@
  * or use the multi device naming "0" (specifying only device index 0).
  * Both naming choice allows the driver to handle the same single device.
  *
+ 3)If the setting "midi.autoconnect" is enabled and the device name is "default",
+ * then all the available devices are opened, applying the appropriate channel
+ * mappings to each device (the first device is mapped to the 16 first channels
+ * (0-15), the second one to the next 16 channels (16-31), and so on with the limit
+ * of the synth.midi-channels setting. After arriving to the channels limit, the
+ * mapping restarts with the channels 0-15.
+ * This makes all MIDI devices audible on the synth instance. However with more than
+ * 1 MIDI devices and the default value of synth.midi-channels of 16 then MIDI
+ * channels conflicts are possible if MIDI devices transmit on the same MIDI channel.
+
+ * If the device name other than "default" is specified, then midi.autoconnect
+ * and synth.midi-channels settings are ignored (see 1).
+ *
  */
 
 #include "fluidsynth_priv.h"
@@ -481,6 +494,8 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
     MMRESULT res;
     int i, j;
     int max_devices;  /* maximum number of devices to handle */
+    int autoconnect = 0;
+    int synth_midi_channels = 16;
     char strError[MAXERRORLENGTH];
     char dev_name[MAXPNAMELEN];
 
@@ -498,8 +513,18 @@ new_fluid_winmidi_driver(fluid_settings_t *settings,
         FLUID_STRCPY(dev_name, "default");
     }
 
-    /* parse device name, get the maximum number of devices to handle */
-    max_devices = fluid_winmidi_parse_device_name(NULL, dev_name);
+    fluid_settings_getint(settings, "midi.autoconnect", &autoconnect);
+    if ((autoconnect) && (strcmp(dev_name, "default") == 0))
+    {
+        fluid_settings_getint(settings, "synth.midi-channels", &synth_midi_channels);
+        max_devices = midiInGetNumDevs(); /* get number of real devices installed */
+    }
+    else
+    {
+        autoconnect = 0; /* disable autoconnect */
+        /* parse device name to get the number of devices to handle */
+        max_devices = fluid_winmidi_parse_device_name(NULL, dev_name);
+    }
 
     /* check if any device has be found	*/
     if(!max_devices)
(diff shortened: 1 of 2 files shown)


pedrolcl:
> I have to admit it's hard for me to comprehend your different views / use-cases and understand the differences. Which makes it hard to make a decission here. AFAIU, your change keeps backward compatibility with your original use-case?
> 
> It would make things easier if @pedrolcl could comment on this proposal as well.

Well, I've explained in https://github.com/FluidSynth/fluidsynth/issues/414#issuecomment-1011215646 why I think that  *"The synth.midi-channels winmidi independence of #677 was an important design choice."* creates a nasty bug. For instance: https://lists.nongnu.org/archive/html/fluid-dev/2021-11/msg00000.html . I am not sure if @es20490446e is using Windows or something else, but his problem is exactly the bug in winmidi that I tried to fix in #1023, no matter if autoconnect is enabled or not.

In this comment https://github.com/FluidSynth/fluidsynth/issues/414#issuecomment-1011523369 the argumentation is that MIDI channels collision is also a bug, but in my view it is a problem of the MIDI protocol, which has been affecting every MIDI installation for the last 30 years if the users don't understand MIDI 1.0 and its limitations.  The solution proposed for the winmidi bug, that users should use the midi router of the command line fluidsynth program to map the incoming event channels by hand, is preposterous, when there is a much more easy, universal and sensible solution for the user: to increase the value for the setting 'synth.midi-channels', which every MIDI driver should respect, whether  autoconnect is enabled or disabled.

es20490446e:
I'm on Linux.

src/drivers/fluid_winmidi.c
```diff
@@ -28,35 +28,60 @@
...
  * - MIDI messages from real device 1 are output to MIDI channels set 0 to 15.
- * - MIDI messages from real device 0 are output to MIDI channels set 15 to 31.
+ * - MIDI messages from real device 0 are output to MIDI channels set 16 to 31.
  * So, the device order specified in the setting allows the user to choose the
  * MIDI channel set associated with this real device at the driver output
  * according this formula: output_channel = input_channel + device_order * 16.
  *
+ * Note the driver ignores the setting synth.midi-channels which default value
+ * for this setting is 16, in which case the second MIDI device will not be hear
+ * on the fluid synth instance.
```
pedrolcl, line 58: You leave a comment in the source code, but this won't help the users.

IMO this "feature" is a plain and simple bug.

src/drivers/fluid_winmidi.c
```diff
@@ -28,35 +28,60 @@
...
+ * a) You have to rise the setting synth.midi-channels to accept the number
+ * of MIDI channels provided by the MIDI driver.
```
pedrolcl, line 60: You are saying that your driver policy rules, and the user should simply obey it and set the value of "synth.midi-channels" accordingly. 

IMO, the user setting of "synth.midi-channels" should be prevalent, and all the drivers should limit their channels mapping to this value, like does the ALSA sequencer driver.

pedrolcl:
> I'm on Linux.

Then, please open a new issue at https://github.com/FluidSynth/fluidsynth/issues with details. Thanks!

derselbst:
> Well, I've explained in #414 (comment) why I think that "The synth.midi-channels winmidi independence of #677 was an important design choice." creates a nasty bug

You have explained, that some channels will be ignored by the synth, yes. Might be not nice, but I wouldn't call it nasty bug either, as it can be fixed by simply increasing `synth.midi-channels`.

I also understand JJC's point of view, that a hardcoded modulo operation of the channel_map is a bad idea.

From what I see, the ALSA driver limits it's destination ports, but it doesn't do a modulo operation either.

> For instance: https://lists.nongnu.org/archive/html/fluid-dev/2021-11/msg00000.html .

There already was an issue #999. It's related to Systemd, not MIDI.

My proposal to handle the "insufficient number of midi channels" situation would be to either print a warning, or fail with an error during driver creation.

src/drivers/fluid_winmidi.c
```diff
@@ -28,35 +28,60 @@
...
  *
+ * 3)If the setting "midi.autoconnect" is enabled and the device name is "default",
+ * then all the available devices are opened, applying the appropriate channel
+ * mappings to each device (the first device is mapped to the 16 first channels
+ * (0-15), the second one to the next 16 channels (16-31), and so on with the limit
+ * of the synth.midi-channels setting. After arriving to the channels limit, the
+ * mapping restarts with the channels 0-15.
+ * This makes all MIDI devices audible on the synth instance. However with more than
+ * 1 MIDI devices and the default value of synth.midi-channels of 16 then MIDI
+ * channels conflicts are possible if MIDI devices transmit on the same MIDI channel.
```
pedrolcl, line 80: Think of this scenario: The user connects two MIDI controllers to the MIDI IN ports of a [MIDI merge box](https://www.thomann.de/intl/midi_solutions_merger_v2.htm), and the MIDI OUT of the merge box to the input of Fluidsynth (or any other synthesizer). Then the user plays both controllers in the same MIDI channel. This may produce event collisions, of course, because that is how MIDI works. Each MIDI controller should use a different MIDI channel.

pedrolcl:
> You have explained, that some channels will be ignored by the synth, yes. Might be not nice, but I wouldn't call it nasty bug either, as it can be fixed by simply increasing `synth.midi-channels`.

The user will only observe that fluidsynth is ignoring the events received from some MIDI controllers, and he won't know what to do about it. Your proposal is to return a warning message, or fail to start the driver. Well, that's better. Somewhat rude, though.

Let's summarize the problem with an example, including a simple (but common) use case.

The user runs Windows + FS with the default settings. He connects his USB MIDI keyboard, which uses the output MIDI channel 1 to the computer, and changes the setting midi.winmidi.device to match his device id. He plays the piano keys and FS produces sound. Now, his friend brings another USB MIDI keyboard, adjusting the output MIDI channel to 2, and connects it to the same computer. The user changes the setting midi.winmidi.device again to be "0;1" (assuming that this is the correct setting value) restarting the synth.  Now, playing both MIDI controllers at once only the first one produces sound. And there is nothing explaining why. The channel mappings applied by winmidi are not exposed at all, neither are evident or expected. Why should the user know that increasing the number of MIDI channels of the synth to 32 solves his problem?

Because the typical use case is that each MIDI controller sends events on one MIDI channel only. Using the default value of "synth.midi-channels" is enough for 16 MIDI controllers playing together with the same FS instance, using each one a different MIDI program. There is no need for channel mappings in this use case. Only basic knowledge of the MIDI standards.

derselbst:
> Why should the user know that increasing the number of MIDI channels of the synth to 32 solves his problem?

Ok, I get your point. This makes me think that we should do the channel modulo wrapping as Pedro suggested. And I would also tend to say that a possible risk of "MIDI channels collision" is an inherent problem of the MIDI standard, or better of the two friends playing with two instruments on the same MIDI channel.

jjceresa:
>AFAIU, your change keeps backward compatibility with your original use-case?

Right. The original use case is to have a driver behaviour that makes a straightforward  aggregation of the MIDI in devices on predictible separate MIDI channels to ensure no MIDI channels conflicts. I will explain why in next comments.

Of course having an "autoconnect" behaviour for the reason explained by Pedro is nice.
Obviously, "autoconnect" needs to take account of  synth.midi-channels and magically map any devices MIDI channel confined to synth.midi-channels.

>I don't like this to be initialized to magic 16.

Oh yes. That will be fixed

jjceresa:
>You leave a comment in the source code, but this won't help the users.

The comments are for developers not for end-users.

>Then the user plays both controllers in the same MIDI channel. This may produce event collisions, of course, because that is how MIDI works. Each MIDI controller should use a different MIDI channel.

In a perfect world each MIDI controller should use a different MIDI channel but there case were this is not true. Please see my comment below.

jjceresa:
One of my friend user is using multiples MIDI controllers (2 keyboards), and several custom MIDI cc controllers).
One constraint is that the custom MIDI cc transmit all on the same MIDI channel (and this cannot be changed).
At the output he is often using  2 synths (1 fluid synth +  hardware synth).

To map any MIDI input controllers (Keyboard layers+ CCs) to any synths (fluid synth or/and hardware synths)  this user prefers to use a central matrix (MIDI channel mapping/key split/ and routing) situated between the MIDI driver and the synths.
With the help of the fluidsynth library architecture it is possible to have a MIDI driver connected to a custom application and the output of this application connected to fluid synth + other synths.
The application allows a direct and fast access to the central matrix which is opaque to the human user
through a user interface.
Also the user never need to enter in MIDI controllers menu to change the MIDI channels setup.

In this case that means that the MIDI driver behaviour must be constant and predicitible. The MIDI driver should not do a magic device channel mapping merging dependent of  synth.midi-channels. This is the reason of the actual winmidi driver behaviour.

jjceresa:
I have added a "midi.winmidi.mergebox" setting toggle setting.
With this setting set to 1 (the default) the driver behaves like d031d4b of PR #1023.
With this setting set to 0 the driver behaves like actual.

pedrolcl:
> One of my friend user is using multiples MIDI controllers (2 keyboards), and several custom MIDI cc controllers). One constraint is that the custom MIDI cc transmit all on the same MIDI channel (and this cannot be changed). At the output he is often using 2 synths (1 fluid synth + hardware synth).
> 
> To map any MIDI input controllers (Keyboard layers+ CCs) to any synths (fluid synth or/and hardware synths) this user prefers to use a central matrix (MIDI channel mapping/key split/ and routing) situated between the MIDI driver and the synths. With the help of the fluidsynth library architecture it is possible to have a MIDI driver connected to a custom application and the output of this application connected to fluid synth + other synths. The application allows a direct and fast access to the central matrix which is opaque to the human user through a user interface. Also the user never need to enter in MIDI controllers menu to change the MIDI channels setup.
> 
> In this case that means that the MIDI driver behaviour must be constant and predicitible. The MIDI driver should not do a magic device channel mapping merging dependent of synth.midi-channels. This is the reason of the actual winmidi driver behaviour.

For the particular case of a custom application used only by your friend, you can provide him with a custom FS library, fine tuned as you see fit. Instead, you are forcing everyone to follow your particular policy. That is unfair. I am talking here as the packager of Qsynth for Windows, not only as a FS user.

pedrolcl:
> I have added a "midi.winmidi.mergebox" setting toggle setting. With this setting set to 1 (the default) the driver behaves like [d031d4b](https://github.com/FluidSynth/fluidsynth/commit/d031d4bab75c5f09f5dee1d97f6440847defd91d) of PR #1023. With this setting set to 0 the driver behaves like actual.

So, you think that it is fine to add and obey a new user setting in the winmidi driver, but you insist on ignoring in some cases the setting "synth.midi-channels"?

Why not change "synth.midi-channels" to the maximum value of 256 channels in your friend's custom application instead, and always fulfill this setting?

es20490446e:
I would do the least surprising thing for the most general case. While not enforcing any particular policy, but making the most common the default.

jjceresa:
>you are forcing everyone to follow your particular policy....That is unfair. I am talking here as the packager of Qsynth for Windows.
>So, you think that it is fine to add and obey a new user setting in the winmidi driver...

With the help of midi.winmidi.mergebox setting everyone aren't forced to follow my particular policy. The default value of this setting is 1 and in this case the driver behave as a `merging box` that is the policy of  PR #1023  who is making a silent magic 
MIDI channel mapping (for ALSA seq also if understood well).

Note that the MIDI driver `merging box` policy nature makes the use of fluid router or custom router impossible because the MIDI channel mapping at the router input isn't easily predictable. 

With the midi.winmidi.mergebox setting  presence it is now more clear how the driver behaves.

>Why not change "synth.midi-channels" to the maximum value of 256 channels in your friend's custom application instead, and always fulfill this setting?

This excessive and sensible solution is a workaround, because if  "synth.midi-channels" is accidentally changed for any reason than MIDI channel mapping at the MIDI driver output becomes unpredictable again. The MIDI driver output is connected to MIDI input of this custom application. So the important point is to keep all the MIDI in configuration routing independent of the  synths MIDI channels capabilities. This user use a fluid synth + others synth. 

Please, as the packager of Qsynth for Windows why midi.winmidi.mergebox setting presence is a problem ?

derselbst:
Superseded by #1023

PR #1032, opened 2022-01-16
Initialize Pipewire with pw_init()

derselbst:
Before using any PipeWire functions, one must call pw_init() (e.g. https://docs.pipewire.org/page_tutorial1.html)

Thanks to @insilications for noticing.


insilications:
@derselbst sorry for taking too long to respond. I agree with all your points in https://github.com/FluidSynth/fluidsynth/pull/1025. I should have looked into the problem more deeply, especially the Fluidsynth source for which I have zero experience with. I plan to add a PR to https://github.com/dosbox-staging/dosbox-staging, in order to add pipewire audiodriver setting and usage. They use the shared library. With these dev docs and your explanation, it should be easy to do it.

Issue #1021, opened 2022-01-10
Further cmake modernizations


pedrolcl:
After #969 already merged, I've found a few rusty pieces. The first one is:

https://github.com/FluidSynth/fluidsynth/blob/015c6af52d8233eaae556242b5200290558ba415/CMakeLists.txt#L35

For CMake 3.13 and later, there is another `project()` command signature:  https://cmake.org/cmake/help/v3.13/command/project.html

This also would simplify the following lines (and related):

https://github.com/FluidSynth/fluidsynth/blob/015c6af52d8233eaae556242b5200290558ba415/CMakeLists.txt#L41-L46

Another proposal is to remove the following code which is utterly deprecated:

https://github.com/FluidSynth/fluidsynth/blob/015c6af52d8233eaae556242b5200290558ba415/CMakeLists.txt#L886-L903

It may be replaced by a `git export` command, but anyway the `cpack` targets (package, package_source) already provide a quick and easy way of packaging as tarballs both the sources and the binaries. It only needs modernization this line:

https://github.com/FluidSynth/fluidsynth/blob/015c6af52d8233eaae556242b5200290558ba415/CMakeLists.txt#L921

If nobody needs/uses it, the alternative is to remove the cpack support entirely:

https://github.com/FluidSynth/fluidsynth/blob/015c6af52d8233eaae556242b5200290558ba415/CMakeLists.txt#L909-L931

derselbst:
Thanks for bringing this up.

1. I don't see too much benefit in moving to the new project() syntax.
2. true, can be removed
3. I never used cpack before. Github takes care of creating the source tarballs. I just tried creating the binary tarball and this looks promising. Maybe I'll use it for tarball creation in our CI build in the future. So I'd prefer to keep it.

pedrolcl:
> Thanks for bringing this up.
> 
>     1. I don't see too much benefit in moving to the new project() syntax.

It is only a cosmetic thing for now, but the advantage to adopt it early (well, it has been around since cmake 3.0) may be that the moment cmake deprecates the old syntax, there will be no hurry to migrate.

>     3. I never used cpack before. Github takes care of creating the source tarballs. I just tried creating the binary tarball and this looks promising. Maybe I'll use it for tarball creation in our CI build in the future. So I'd prefer to keep it.

[...]

PR #1040, opened 2022-01-28
Use project-specific SOURCE_DIR and BINARY_DIR

devingryu:
This change is for compatibility of adding fluidsynth as subdirectory in other cmakefiles.
If fluidsynth is added as subdirectory of other projects, it would produce error since CMAKE_SOURCE_DIR is different from the value when built solely.
I have nothing to complain about even if you don't accept it, but it would be grateful if you consider it.

Thanks in advance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,6 @@ cmake_minimum_required ( VERSION 3.13 )
 # 3.11 because COMPATIBILITY SameMinorVersion in CMakePackageConfigHelpers
 # 3.13.5 because it is the latest supported in Windows XP
 
-set ( FLUIDSYNTH_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR} )
-set ( FLUIDSYNTH_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR} )
-
 if(POLICY CMP0075) # CMake version 3.13.5 warns when the policy is not set or value is OLD
   cmake_policy(SET CMP0075 NEW)
 endif()
@@ -36,7 +33,7 @@ if(POLICY CMP0091) # new in CMake 3.15, defaults to OLD
 endif()
 
 project ( FluidSynth C CXX )
-list( APPEND CMAKE_MODULE_PATH ${FLUIDSYNTH_SOURCE_DIR}/cmake_admin )
+list( APPEND CMAKE_MODULE_PATH ${FluidSynth_SOURCE_DIR}/cmake_admin )
 
 # FluidSynth package name
 set ( PACKAGE "fluidsynth" )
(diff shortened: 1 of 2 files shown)


derselbst:
@pedrolcl Opinions?

pedrolcl:
> Opinions?

It is a common use case to embed cmake projects inside other projects, for instance as subdirectories or git submodules. That is why the global-process variables CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR should be avoided. But...

The [project()](https://cmake.org/cmake/help/v3.13/command/project.html) command already defines suitable replacements:
* `PROJECT_SOURCE_DIR`, `${PROJECT_NAME}_SOURCE_DIR`
* `PROJECT_BINARY_DIR`, `${PROJECT-NAME}_BINARY_DIR`

In our case, https://github.com/FluidSynth/fluidsynth/blob/91515b09c45c403bbd5dad84875988b95253f3b9/CMakeLists.txt#L35 the two variables already created are:
* `FluidSynth_SOURCE_DIR`
* `FluidSynth_BINARY_DIR`

Because in cmake, [variable names are case sensitive](https://cmake.org/cmake/help/v3.13/manual/cmake-language.7.html#variables).

By the way, there is still an open issue regarding the `project()` command: #1021

CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ cmake_minimum_required ( VERSION 3.13 )
 # 3.11 because COMPATIBILITY SameMinorVersion in CMakePackageConfigHelpers
 # 3.13.5 because it is the latest supported in Windows XP
 
+set ( FLUIDSYNTH_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR} )
+set ( FLUIDSYNTH_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR} )
+
```
pedrolcl, line 29: In my opinion, there is no need to create those variables, when there are already two created:
* `FluidSynth_SOURCE_DIR`
* `FluidSynth_BINARY_DIR`
These variables are created for every cmake project, when the command [project()](https://cmake.org/cmake/help/v3.13/command/project.html) is called.

devingryu:
@pedrolcl Thanks for commenting.
I didn't realized that there are predefined variables in `project()`... Then are you willing to accept the alternative method of using FluidSynth_*_dir variables?

pedrolcl:
> are you willing to accept the alternative method of using FluidSynth_*_dir variables?

I have no objections.

pedrolcl (changes requested):
Please reference the already existing variables instead of the new ones that you have created, Thanks!

derselbst:
Using FluidSynth_SOURCE_DIR and  FluidSynth_BINARY_DIR sounds good to me as well.

> By the way, there is still an open issue regarding the project() command: #1021

I see no dependency between those two tasks. So, I would prefer to address #1021 separately.

devingryu:
> Please reference the already existing variables instead of the new ones that you have created, Thanks!

It seems that the build has gone well(except for broken prerequisite links for win builds)! Thanks for letting me know about project variables too. Helped me a lot for my project!

derselbst (approved):
LGTM, Thanks!

PR #1057, opened 2022-02-26
Use int64_t as fluid_long_long_t in C99

devingryu:
Currently, fixed type of long long is used for fluid_long_long_t, as it is guranteed to have 64-bit wide in C99.
However, stdint.h defines int64_t as long, so it has posibillity to break codes which uses 64-bit type with stdint.
For example, jni.h defines jlong with int64_t, so some functions which uses fluid_long_long_t as parameter type breaks when binding them into jlong.

So, even though using long long as 64-bit type is reasonable, I suggest you to use int64_t since it has potential to break in some build targets.

Thanks in advance.

### DIFF
--- a/include/fluidsynth/types.h
+++ b/include/fluidsynth/types.h
@@ -27,6 +27,9 @@
 extern "C" {
 #endif
 
+#if !(defined(_MSC_VER) && (_MSC_VER < 1800))
+#include <stdint.h>
+#endif
 
 /**
  * @defgroup Types Types
@@ -73,7 +76,6 @@ typedef __int64 fluid_long_long_t; // even on 32bit windows
  * A typedef for C99's type long long, which is at least 64-bit wide, as guaranteed by the C99.
  * @p __int64 will be used as replacement for VisualStudio 2010 and older.
  */
-#include <stdint.h>
 typedef int64_t fluid_long_long_t;
 #endif
 
--- a/include/fluidsynth/types.h
+++ b/include/fluidsynth/types.h
@@ -73,7 +73,8 @@ typedef __int64 fluid_long_long_t; // even on 32bit windows
  * A typedef for C99's type long long, which is at least 64-bit wide, as guaranteed by the C99.
  * @p __int64 will be used as replacement for VisualStudio 2010 and older.
  */
-typedef long long fluid_long_long_t;
+#include <stdint.h>
+typedef int64_t fluid_long_long_t;
 #endif
 
 /* @} */


derselbst:
> For example, jni.h defines jlong with int64_t, so some functions which uses fluid_long_long_t as parameter type breaks when binding them into jlong.

Sry, I don't get this point. Why would you want to bind `fluid_long_long_t` to `jlong`?

An why would this break something? You're essentially defining `fluid_long_long_t`  == `jlong` == `int64_t` == `long`. The only chance I see this would go wrong is when fluidsynth was compiled on a machine where `long long` is actually an 128 bit integer, whereas `long` is only 64 bit.

You're basically suggesting an ABI breakage for everyone. So I really need to understand your motivation.

derselbst:
> When calling functions (or setting callback functions) which includes fluid_long_long_t, the value of java long(which came from java) should be explicitly type converted in order to be passed to fluidsynth C functions.

It's pretty uncommon to mess around in 3rd party library headers. Have you considered to typecast `fluid_long_long_t` to `int_least64_t`? If so, what does your jni.h say about this type?


> But come to think of it, other programs using fluidsynth could produce build error if the type of fluid_long_long_t is changed (though the build process of the library seems to go well)

I do breaking change releases of fluidsynth from time to time. The next breaking change will probably be when porting to C++ (in case that ever happens, see #847). I will consider your request if or once we do so.

devingryu:
> It's pretty uncommon to mess around in 3rd party library headers. Have you considered to typecast fluid_long_long_t to int_least64_t? If so, what does your jni.h say about this type?

It differs per target platform, as it refers to stdint.h. in armv7-eabi and x86 it is not auto casted, but in aarch64 and amd64, it works well(since both `int64_t` and `jlong` is `long`)

The point of my opinion is not in jni.h, but in stdint.h. The best choice of 64-bit type varies per platform, so my point is it would be better to choose type of 64-bit type with stdint.h. So for portability and clarity, I suggest you to change `fluid_long_long_t` to `int64_t` (or `int_fast64_t`) when next breaking change comes. Since stdint.h is standard both in C99 and C++, it would increase portability for other compilers and stability would be guranteed when other architectures with 128-bit based comes up.

However, this is just a future talk, and I realized that the cost of the change is larger than maintaining the current situation. So I also think it is not a think that should be considered now. But when a breaking change happens, I hope that you could consider this.

> I do breaking change releases of fluidsynth from time to time. The next breaking change will probably be when porting to C++ (in case that ever happens, see https://github.com/FluidSynth/fluidsynth/discussions/847). I will consider your request if or once we do so.

Glad to hear from you! I'll be looking forward to it.

derselbst:
Ok thanks, I've added it to the discussion so we don't forget about it.

devingryu:
OK, Thanks.

PR #1063, opened 2022-03-05
Fix oboe driver build issue in latest glib

devingryu:
I recently managed to build latest glib(2.71.2) on Android with meson, and found that oboe driver is not building well on it.

It is because glib is included (indirectly in `fluidsynth_priv.h`) in the scope of `extern "C"`. In glib header `glib-2.0/glib/glib-typeof.h`, `<cstddef>` is included, which has cpp template code. So it generates following error:

> error: templates must have C++ linkage
> `template <class _Tp> struct __libcpp_is_integral                     { enum { value = 0 }; };`

I think this error is not generated in usual since oboe driver is not used in desktop environments, and it is the only driver that is written in cpp. But to consider updating glib in Android, I think the code should be changed to prevent further issues.

I hope this change does not make a big issue to other platform builds...

Thanks in advance.

### DIFF
--- a/src/drivers/fluid_adriver.h
+++ b/src/drivers/fluid_adriver.h
@@ -23,6 +23,10 @@
 
 #include "fluidsynth_priv.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /*
  * fluid_audio_driver_t
  */
@@ -175,6 +179,8 @@ fluid_audio_driver_t *new_fluid_file_audio_driver(fluid_settings_t *settings,
 void delete_fluid_file_audio_driver(fluid_audio_driver_t *p);
 #endif
 
-
+#ifdef __cplusplus
+}
+#endif
 
 #endif  /* _FLUID_AUDRIVER_H */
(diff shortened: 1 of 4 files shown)


derselbst (approved):
LGTM, thanks. I'll cherry-pick it to the `2.2.x` branch later on.

Pls. note that we do have an [Android CI pipeline](https://github.com/FluidSynth/fluidsynth/blob/master/.azure/azure-pipelines-android.yml), which also compiles and links Oboe. But it uses glib 2.58.3, the last  version which uses autotools, because I failed to get it working with meson. So if you did manage to get it built, I'd welcome a PR on that as well.

devingryu:
> Pls. note that we do have an [Android CI pipeline](https://github.com/FluidSynth/fluidsynth/blob/master/.azure/azure-pipelines-android.yml), which also compiles and links Oboe. But it uses glib 2.58.3, the last version which uses autotools, because I failed to get it working with meson. So if you did manage to get it built, I'd welcome a PR on that as well.

Of course I'll make PR if you want, but that meson cross-file is just made with my immature googling and brute-force testing. So please make sure it doesn't make any problems before accepting. Also, it needs extra libpcre build process. (Building goes well anyway, since it has fallback subproject feature).

Thanks for accepting!

derselbst:
Cherry-picked and merged back to master. Pls. update your local master.

PR #1081, opened 2022-04-15
Multi-channel output for the CoreAudio driver.

mattrtaylor:
Hi, this pull request does two things:

* Enable the _synth.audio-channels_ and _synth.audio-groups_ settings to work with CoreAudio as they do on other platforms.

    * This required modifying the `AudioStreamBasicDescription`, which was hardcoded as interleaved stereo. A nice efficiency improvement we can make here is adding the `kAudioFormatFlagIsNonInterleaved` flag, which allows us to skip the allocation and interleave + copy step and pass the buffers filled by the fluidsynth render callback directly.

* Adds a _audio.coreaudio.channel-map_ setting to make use of CoreAudio's standard [channel mapping](https://developer.apple.com/library/archive/documentation/MusicAudio/Conceptual/CoreAudioOverview/ARoadmaptoCommonTasks/ARoadmaptoCommonTasks.html#//apple_ref/doc/uid/TP40003577-CH6-SW16) facility to arbitrarily route output channels.

As an example of why these changes are helpful, I use them for live performances to route MIDI instruments, samples, and cues to different channels on a digital rack mixer where they can be adjusted by a front-of-house sound engineer as if they were separate instruments.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -453,6 +453,27 @@ Developers:
             </desc>
         </setting>
         <setting>
+            <name>coreaudio.channel-map</name>
+            <type>str</type>
+            <def>0,1</def>
+            <desc>
+                This setting is a comma-separated integer list that maps CoreAudio channels
+                to output device channels. The value of each position in the list, up to the number of
+                output channels available on your device, is the zero-based index of the fluidsynth
+                output channel to route there. Additionally, the special value of -1 will turn off
+                an output.
+
+                For example, the default map for a single stereo output is "0,1". A value of "0,0" will
+                copy the left channel to the right, a value of "1,0" will flip left and right, and a
+                value of "-1,1" will play only the right channel.
+
+                With a six channel output device, and the synth.audio-channels and synth.audio-groups
+                settings both set to "2", a channel map of "-1,-1,0,1,2,3" will result in notes from odd
+                MIDI channels being sent to outputs 3 and 4, and even MIDI channels being send to
+                outputs 5 and 6.
+            </desc>
+        </setting>
+        <setting>
             <name>dart.device</name>
             <type>str</type>
             <def>default</def>
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -467,10 +467,14 @@ Developers:
                 copy the left channel to the right, a value of "1,0" will flip left and right, and a
                 value of "-1,1" will play only the right channel.
 
-                With a six channel output device, and the synth.audio-channels and synth.audio-groups
+                With a six-channel output device, and the synth.audio-channels and synth.audio-groups
                 settings both set to "2", a channel map of "-1,-1,0,1,2,3" will result in notes from odd
-                MIDI channels being sent to outputs 3 and 4, and even MIDI channels being send to
-                outputs 5 and 6.
+                MIDI channels (audible on the first stereo channel, i.e. mono-indices 0,1) being sent to
+                outputs 3 and 4, and even MIDI channels (audible on the second stereo channel, i.e. mono-indices 2,3)
+                being sent to outputs 5 and 6.
+
+                If the list specifies fewer than the number of available outputs channels, outputs
+                beyond those specified will maintain the default channel mapping.
             </desc>
         </setting>
         <setting>
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -455,6 +455,7 @@ Developers:
         <setting>
             <name>coreaudio.channel-map</name>
             <type>str</type>
+            <def>(empty string)</def>
             <desc>
                 This setting is a comma-separated integer list that maps CoreAudio channels
                 to output device channels. The value of each position in the list, up to the number of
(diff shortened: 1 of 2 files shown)


doc/fluidsettings.xml
```diff
@@ -452,6 +452,27 @@ Developers:
...
+                an output.
+
+                For example, the default map for a single stereo output is "0,1". A value of "0,0" will
+                copy the left channel to the right, a value of "1,0" will flip left and right, and a
+                value of "-1,1" will play only the right channel.
+
+                With a six channel output device, and the synth.audio-channels and synth.audio-groups
+                settings both set to "2", a channel map of "-1,-1,0,1,2,3" will result in notes from odd
+                MIDI channels being sent to outputs 3 and 4, and even MIDI channels being send to
+                outputs 5 and 6.
```
derselbst, line 473: It took quite some time for me to understand this last part. In case I understood it correctly, I'd suggest to phrase it a bit differently:

```
                This setting is a comma-separated integer list that maps fluidsynth mono-channels
                to CoreAudio device output channels. Each position in the list represents the output channel 
                of the CoreAudio device.
                The value of each position indicates the zero-based index of the fluidsynth
                output mono-channel to route there (i.e. the buffer index used for fluid_synth_process()).
                Additionally, the special value of -1 will turn off an output.

[...your single stereo output example...]

                With a six-channel output device, and the synth.audio-channels and synth.audio-groups
                settings both set to "2", a channel map of "-1,-1,0,1,2,3" will result in notes from odd
                MIDI channels (audible on the first stereo channel, i.e. mono-indices 0,1) being sent to
                outputs 3 and 4, and even MIDI channels (audible on the second stereo channel, i.e. mono-indices 2,3)
                being sent to outputs 5 and 6.

                The list should specify as many positions as there are output channels for the corresponding
                CoreAudio device. Otherwise, the behavior is undefined.
```

src/drivers/fluid_coreaudio.c
```diff
@@ -114,6 +115,75 @@ get_num_outputs(AudioDeviceID deviceID)
...
+    }
+
+    channel_map = FLUID_ARRAY(int, number_of_channels);
+    if(channel_map == NULL)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory.\n");
+        return;
+    }
+
+    FLUID_MEMSET(channel_map, 0, property_size);
```
derselbst, line 151: You are zero-initing the map here. If the CoreAudio device provides 6 output channels, but the user has only supplied a list with two entries, outputs 2,3,4,5 would all be mapped to the first left mono channel. Is this intended? Shouldn't you better init with `0xFF`?

derselbst (commented):
That looks pretty excellent. Thank you!  I have some minor comments below. And I'm pinning @pedrolcl  as he might be interested in this change as well.

pedrolcl:
> I'm pinning @pedrolcl as he might be interested in this change as well.

I don't use macOS anymore. My MBP is useless after the update to Catalina, so I'm planning to install Linux on it. Bye, bye, Apple!

src/drivers/fluid_coreaudio.c
```diff
@@ -125,6 +195,7 @@ fluid_core_audio_driver_settings(fluid_settings_t *settings)
     pa.mElement = kAudioObjectPropertyElementMain;
 
     fluid_settings_register_str(settings, "audio.coreaudio.device", "default", 0);
+    fluid_settings_register_str(settings, "audio.coreaudio.channel-map", "", 0);
```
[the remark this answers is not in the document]
mattrtaylor: Thank you for your feedback, all good points.

In retrospect, I agree it's better to give the real default value of the empty string.

doc/fluidsettings.xml
mattrtaylor: The updated version has a variation on this. Rather than being undefined, if the list of outputs is too short, it doesn't change the mapping away from the default for the higher channels.

The reasoning for this is that you might have a lot of outputs, and generally don't care about those above however many fluidsynth is producing on due to _synth.audio-channels_.

It makes nice logical sense with the empty string too, in that it's just leaving all channels as they are by default.

src/drivers/fluid_coreaudio.c
mattrtaylor: Good call.

Memsetting this array is primarily being defensive, since the following call to _AudioUnitGetProperty_ fills it out with the default mapping from the OS. I've only ever seen that be 0,1,2,3...n-1 prior to being set, but I don't know if that is guaranteed.

Silence is definitely the better option in the remote chance CoreAudio fails to fill the buffer without returning an error code though. In testing this out, I learned the hard way that it will happily allow you to set an invalid channel number, and the resulting sound is... unpleasant.

doc/fluidsettings.xml
derselbst: You haven't applied the first paragraph I suggested. In case that was not intentional, I would amend it after merging, if you don't mind.

derselbst (approved):
LGTM, Thank You!

doc/fluidsettings.xml
mattrtaylor: Thanks, sorry I missed it. Looks great!

PR #1082, opened 2022-04-17
bump micro version to 2.2.6

albedozero:
Seems the latest micro version didn't get increased with the new release. Assume this is an oversight, but perhaps has something to do with the cmake updates proposed in #1021 ?


derselbst (changes requested):
Nope, sry. It exists on the `2.2.x` branch, but this one hasn't been merged back to master yet. Will be done in the coming days or so.

PR #1093, opened 2022-05-05
Update to latest fluidsynth

timothyschoen:


derselbst (changes requested):
Breaking all builds?! Breaking all use-cases of fluidsynth?! Removing all dependencies??! Applying patch that we declined long ago without talking to the original author?! You must be kidding.

PR #1113, opened 2022-06-10
Modify select drivers to pass system realtime messages

albedozero:
FluidSynth's various MIDI drivers don't appear to pass along system realtime messages (i.e. MIDI clock, start, stop, etc.) to the driver's handler/callback. These messages could be useful if, for example, the user writes a custom callback that can use these messages to synchronize a fluid_sequencer to an external clock.

This pull modifies the _alsa_seq_ and _winmidi_ drivers, as well as any that use `fluid_midi_parser_parse`, to pass along system realtime messages (status byte F8-FF) to the callback. I don't think it will cause problems further down the chain, since `fluid_midi_router_handle_midi_event` will just [return them as NULL](https://github.com/FluidSynth/fluidsynth/blob/eda2fb21b0c41114add966dbaae009dd443dfc94/src/midi/fluid_midi_router.c#L661). This behavior could be added to other drivers in future if desired.

### DIFF
--- a/.azure/azure-pipelines-android.yml
+++ b/.azure/azure-pipelines-android.yml
@@ -53,6 +53,7 @@ variables:
   # This is a symlink pointing to the real Android NDK
   # Must be the same as $ANDROID_NDK_HOME see:
   # https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
+  # We cannot use $ANDROID_NDK_HOME because this is an environment variable, but here, we need a compile-time constant.
   NDK: '/usr/local/lib/android/sdk/ndk-bundle'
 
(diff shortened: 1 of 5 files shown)


derselbst (commented):
Looks good to me, thanks! If you want to have those changes in fluidsynth 2.2.8 I would recommend that you rebase your changes against the `2.2.x` branch. Currently it's based against `master`, meaning 2.3.0, and I don't know when this version will be ready.

albedozero:
Sounds good, although I think I'm not clear on how to rebase properly. It seems like I'm going to bring in all the 70+ commits from master if I do. I should learn how to do this - but for the moment I think it will be easier for me if I close this PR and recreate my fork from 2.2.x.

PR #1115, opened 2022-06-10
Modify select drivers to pass system realtime messages

albedozero:
FluidSynth's various MIDI drivers don't appear to pass along system realtime messages (i.e. MIDI clock, start, stop, etc.) to the driver's handler/callback. These messages could be useful if, for example, the user writes a custom callback that can use these messages to synchronize a fluid_sequencer to an external clock.

This pull modifies the _alsa_seq_ and _winmidi_ drivers, as well as any that use `fluid_midi_parser_parse`, to pass along system realtime messages (status byte F8-FF) to the callback. I don't think it will cause problems further down the chain, since `fluid_midi_router_handle_midi_event` will just [return them as NULL](https://github.com/FluidSynth/fluidsynth/blob/eda2fb21b0c41114add966dbaae009dd443dfc94/src/midi/fluid_midi_router.c#L661). This behavior could be added to other drivers in future if desired.

### DIFF
--- a/src/drivers/fluid_alsa.c
+++ b/src/drivers/fluid_alsa.c
@@ -1339,13 +1339,33 @@ fluid_alsa_seq_run(void *d)
                     break;
 
                 case SND_SEQ_EVENT_PORT_START:
-                {
-                    if(dev->autoconn_inputs)
                     {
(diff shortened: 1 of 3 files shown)


albedozero:
re-creating this PR with 2.2.x as the base per comments in #1113

derselbst (approved):
Thanks! I've just formatted it a bit.

Issue #1120, opened 2022-06-22
Unwanted timing quantization when using large buffers


pedrolcl:
### FluidSynth version
2.2.x, probably older versions too.

### Describe the bug
When using large audio buffers, the start time of many notes is quantized. This affects severely to the musical rhythm, and it can be perceived as degraded quality. This has been observed with several audio drivers, but PulseAudio is specially impacted because it often requires larger buffers.

### Expected behavior
When using large buffers big latency is expected, but quantization should not happen.

### Steps to reproduce
Start fluidsynth on a Linux terminal, using large buffers (>4K):
~~~
fluidsynth -p fluid -m alsa_seq -a pulseaudio -z 4800 ~/.local/share/soundfonts/GeneralUser.sf2
~~~
From another terminal instance, use aplaymidi to play MIDI files via the ALSA Sequencer:
~~~
aplaymidi -p fluid schubert_avemaria.mid TwistAndShout.mid
~~~

### Additional context

This is somewhat related to the ticket #1 but instead of duration, the starting time is impacted when using large audio buffers.

Sometimes large buffers are wanted in despite of the latency, probably because latency doesn't matter on every situation. Or perhaps can't be avoided, like when using PulseAudio in many scenarios. And in this case, the PulseAudio driver may be easily worked around to avoid this issue. See:  pedrolcl/fluidsynth@adffd1c

The number 128 is arbitrary. The minimum buffer for the synthesizer is 64 frames, and I am not advocating for using it as a fixed constant.

The PulseAudio driver is a particular case where the problem can be easily solved because PulseAudio supports it out of the box, but other drivers may be harder to fix (and I've not tried any other).

Here are two MIDI files where the rhythm is negatively impacted by this issue:

https://members.tripod.com/~jpgr_beatles/midis/files/TwistAndShout.mid
https://github.com/pedrolcl/dmidiplayer/raw/master/examples/schubert_avemaria.mid

pedrolcl:
> Pls set -o audio.pulseaudio.adjust-latency=0 and try again. If that doesn't fixes it, pls try a different audio driver.

Perhaps I've not been clear on this point: this has nothing to do with latency, and all drivers are affected when using large audio buffers.

derselbst:
Ok, after giving it a listen I hear what you mean by "quantized".

[...]

PR #1126, opened 2022-06-25
Add warnings and extend documentation for issue #1120

derselbst:
Addresses #1120

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -418,7 +418,7 @@ and commit the results. Refresh with the following command:
             <min>64</min>
             <max>8192</max>
             <desc>
-                The size of the audio buffers (in frames).
+                This is the number of audio samples most audio drivers will request from the synth at one time. In other words, it's the amount of samples the synth is allowed to render in one go when no state changes (events) are about to happen. Because of that, specifying too big numbers here may cause MIDI events to be poorly quantized (=untimed) when a MIDI driver or the synth's API directly is used, as fluidsynth cannot determine when those events are to arrive. This issue does not matter, when using the MIDI player or the MIDI sequencer, because in this case, fluidsynth does know when events will be received.
             </desc>
         </setting>
         <setting>
--- a/src/drivers/fluid_adriver.c
+++ b/src/drivers/fluid_adriver.c
@@ -324,16 +324,16 @@ new_fluid_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
     if(def)
     {
         fluid_audio_driver_t *driver;
-        double srate, latency;
+        double srate, midi_event_latency;
         int period_size;
         
         fluid_settings_getint(settings, "audio.period-size", &period_size);
         fluid_settings_getnum(settings, "synth.sample-rate", &srate);
         
-        latency = period_size / srate;
-        if(latency >= 0.05)
+        midi_event_latency = period_size / srate;
+        if(midi_event_latency >= 0.05)
         {
-            FLUID_LOG(FLUID_WARN, "You have chosen 'audio.period-size' to be %d samples. Given a sample rate of %.1f this results in a latency of %.1f ms, which will cause MIDI events to be poorly quantized (=untimed) in the synthesized audio (also known as the 'drunken-drummer' syndrome). To avoid that, you're strongly advised to increase 'audio.periods' instead, while keeping 'audio.period-size' small enough to make this warning disappear.", period_size, srate, latency*1000.0);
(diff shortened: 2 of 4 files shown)


src/drivers/fluid_adriver.c
```diff
@@ -323,7 +323,20 @@ new_fluid_audio_driver(fluid_settings_t *settings, fluid_synth_t *synth)
...
+        double srate, latency;
+        int period_size;
+        
+        fluid_settings_getint(settings, "audio.period-size", &period_size);
+        fluid_settings_getnum(settings, "synth.sample-rate", &srate);
+        
+        latency = period_size / srate;
+        if(latency >= 0.05)
+        {
+            FLUID_LOG(FLUID_WARN, "You have chosen 'audio.period-size' to be %d samples. Given a sample rate of %.1f this results in a latency of %.1f ms, which will cause MIDI events to be poorly quantized (=untimed) in the synthesized audio (also known as the 'drunken-drummer' syndrome). To avoid that, you're strongly advised to increase 'audio.periods' instead, while keeping 'audio.period-size' small enough to make this warning disappear.", period_size, srate, latency*1000.0);
```
pedrolcl, line 336: 'audio.periods' is ignored by half of the audio drivers: pulseaudio and pipewire are the most relevant among Linux users.

doc/fluidsettings.xml
```diff
@@ -418,7 +418,7 @@ and commit the results. Refresh with the following command:
             <min>64</min>
             <max>8192</max>
             <desc>
-                The size of the audio buffers (in frames).
+                This is the number of audio samples most audio drivers will request from the synth at one time. In other words, it's the amount of samples the synth is allowed to render in one go when no state changes (events) are about to happen. Because of that, specifying too big numbers here may cause MIDI events to be poorly quantized (=untimed) when a MIDI driver or the synth's API directly is used, as fluidsynth cannot determine when those events are to arrive. This issue does not matter, when using the MIDI player or the MIDI sequencer, because in this case, fluidsynth does know when events will be received.
```
pedrolcl, line 421: You may add as well that the issue doesn't matter because Fluidsynth is not anymore a real time MIDI synthesizer. Instead, Fluidsynth is now optimized for its internal MIDI file player and sequencer, and all other use cases are irrelevant.

jjceresa:
> latency = period_size / srate;

Pherhaps, 'midi_event_latency' would be a better name ?. A new name emphasis the fact that the issue leads to an "unwanted quantisation" effect rather to a latency issue.

jjceresa:
>if(latency >= 0.05)

As already said in issue 1120, a threshold of 50 ms or any other value is subjective and depend highly of the music content.

>'audio.periods' is ignored by half of the audio drivers: pulseaudio and pipewire are the most relevant among Linux users.

Because issue 1120 is only relevant to rendering functions API, I suggest not to insert this warning in the midi driver nor in audio driver code. Simply perhaps will could add a chapter in the Wiki closer to the latency chapter ?.

doc/fluidsettings.xml
derselbst: "Now"? This is the way fluidsynth works since at least version 1.1.3. It took 12 years until somebody complained about it. Therefore, I don't believe that using a real-time synth with an insanely high latency is an issue out there and I'm esp. not willing to change this behavior by potentially breaking other people's use-case as you suggested earlier. Sorry.

src/drivers/fluid_adriver.c
derselbst: The pipewire hasn't even been released yet in fluidsynth. I consider the ALSA driver to be most relevant on Linux, esp. because it's the default. Yet, I welcome your PR to add support to other drivers.

derselbst:
> Pherhaps, 'midi_event_latency' would be a better name ?.

Changed.

> Because issue 1120 is only relevant to rendering functions API, I suggest not to insert this warning in the midi driver nor in audio driver code. Simply perhaps will could add a chapter in the Wiki closer to the latency chapter ?.

For now, I prefer to have a warning issued by commandline, rather than "hiding" it somewhere in the wiki. Explaining it in the wiki might be a good mid- to long-term solution, so you're welcome to write up something.

doc/fluidsettings.xml
pedrolcl: >  potentially breaking other people's use-case as you suggested earlier. Sorry.

I've not suggested that.

src/drivers/fluid_adriver.c
pedrolcl: PulseAudio PR: #1127

PR #1127, opened 2022-06-26
Adding audio.periods setting to the fluid_pulse audio driver

pedrolcl:
This common setting was ignored by the PulseAudio driver.
It doesn't change the synthesis buffer size, which is still
the number of frames specified by audio.period-size, but
the number of periods may be raised to request a higher latency
amount, unless audio.pulseaudio.adjust-latency is 1 (the
default). In turn, PulseAudio will grant or not this latency
request. See: [PulseAudio Latency
Control](https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/Clients/LatencyControl/).

### DIFF
--- a/src/drivers/fluid_pulse.c
+++ b/src/drivers/fluid_pulse.c
@@ -83,12 +83,12 @@ new_fluid_pulse_audio_driver2(fluid_settings_t *settings,
     pa_sample_spec samplespec;
     pa_buffer_attr bufattr;
     double sample_rate;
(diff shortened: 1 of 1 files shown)


pedrolcl:
This is related to #1120 and is on to of PR #1126

derselbst:
Oh, Github automatically closes stacked PRs. Sry. Will cherry-pick it manually. Thanks.

PR #1131, opened 2022-07-27
Fix regression in WinMIDI driver

albedozero:
In #1115 I created a bug in the winmidi driver that causes all incoming voice messages to have channel, param1, and param2 = 0. In fluid_winmidi_callback() the check for `msg_param < 0xf0` fails for all MIDI messages (I had the byte order of dwParam1 backwards, status byte is the lowest byte, not the highest), so the channel and params are never set.

Fixed using the `msg_type` macro to check for voice category message. Also have to mask the status byte with `0xff` for system messages to get the type properly.

### DIFF
--- a/src/drivers/fluid_winmidi.c
+++ b/src/drivers/fluid_winmidi.c
@@ -145,7 +145,7 @@ fluid_winmidi_callback(HMIDIIN hmi, UINT wMsg, DWORD_PTR dwInstance,
         break;
 
     case MIM_DATA:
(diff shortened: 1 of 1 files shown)


derselbst (approved):
Ok, thanks

albedozero:
Thanks, sorry for breaking stuff

derselbst:
No problem. It also escaped my attention when reviewing your original PR. Don't worry.

PR #1137, opened 2022-08-04
Fix minor bugs in fluid_player_callback()

albedozero:
This modifies `fluid_player_callback()` to call `fluid_track_send_events()` _before_ checking if `!fluid_track_eot()`, so that if the user is seeking backward `fluid_track_send_events()` can reset the track. After returning, `!fluid_track_eot()` is checked to see if there are events left in the track to play, and if so `status` is set to `FLUID_PLAYER_PLAYING` as normal.

Similarly, this also modifies `fluid_player_callback()` to set the respective `channel_isplaying` flags back to FALSE after sending an ALL_SOUNDS_OFF control change, to avoid sending unnecessary messages if the player is stopped again before notes are played on a channel, or after beginning a new song in the playlist (which might use completely different channels).

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -2116,6 +2116,7 @@ fluid_player_callback(void *data, unsigned int msec)
                 {
                     fluid_midi_event_set_channel(&mute_event, i);
                     player->playback_callback(player->playback_userdata, &mute_event);
+                    player->channel_isplaying[i] = FALSE;
                 }
             }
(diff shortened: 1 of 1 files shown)


albedozero:
Fixes #1138. I realize the second point above could be split into a second PR, but since both were minor tweaks to the same function I bundled them together. Happy to split them up if desired.

derselbst (approved):
Sorry, it took a while until I had time looking into this. I was able to reproduce the bug and your change seems to fix it correctly. Thanks!

PR #1144, opened 2022-08-20
Cross-platform build using GitHub workflows

kmturley:
I had difficulty locating variables, urls and commands needed for cross-platform builds in a GitHub pipeline.

You have pipeline files I could refer to, but each file is long, complex, has many custom commands and they use different pipelines azure vs GitHub.
https://github.com/FluidSynth/fluidsynth/blob/master/.azure/azure-pipelines-win.yml
https://github.com/FluidSynth/fluidsynth/blob/master/.azure/azure-pipelines-mac.yml
https://github.com/FluidSynth/fluidsynth/blob/master/.github/workflows/linux.yml

Full discussion of my difficulties is here:
https://github.com/FluidSynth/fluidsynth/discussions/1140

Therefore I've created a GitHub workflow template with the following features:

1. bash as the default terminal
2. Fewer, shared commands
3. List of dependency packages for each platform (Linux, Mac, Win)
4. Upload artifacts to a url for download/testing

If you wanted, you can take this a step further an automatically generate a release based on a `git tag v1.0.0` command to automate your workflows.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -1,0 +1,90 @@
+name: FluidSynth All
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+
+env:
+  dir_build: build
+  dir_source: .
+  type: Release
+
+jobs:
+  build:
+    name: All platforms
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            name: linux
+            generator: Unix Makefiles
+            flags: -Denable-portaudio=1 -Denable-ladspa=1
+            path: ./build/src
+          - os: macos-latest
+            name: mac
+            generator: Unix Makefiles
+            flags: -Denable-portaudio=1 -Denable-ladspa=1
+            path: ./build/src
+          - os: windows-latest
+            name: win
+            generator: Visual Studio 17 2022
+            flags: -Denable-jack=0 -Denable-pulseaudio=0 -Denable-ladspa=0 -Denable-dbus=0 -Denable-readline=0 -Denable-sdl2=0 -Denable-libinstpatch=0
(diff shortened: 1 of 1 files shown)


derselbst (changes requested):
While this does look cleaner, I see a couple of issues:

* The Windows and Mac builds are now lacking all the different configuration variations that the existing Azure pipelines are building. You've just taken the least effort to get a green cross-platform build, but for me as a maintainer, it doesn't tell me anything whether fluidsynth can really be built correctly in all different circumstances.
* There are no more x86 Windows builds.
* The Windows binaries cannot be signed anymore.
* I want to have three nice status badges in the Readme (one for each OS), but here I can only get a single badge given that single pipeline.

Even though triggering a release workflow upon a `git tag` command is nice I'm currently doing it with two additional mouse clicks. Even if we did it with Github Actions, we would still need to migrate the Android pipeline as well, in order to have the Android binaries published correctly. Not to forget that the signing logic is also built in Azure pipeline.

So sorry, but I don't see a benefit in accepting this pipeline.

kmturley:
Totally understand. I spent quite a bit of time getting it this far. Probably not going to spend time adding the missing features. But wanted to note how if would be possible if you did want to continue down this path:

**Multiple architectures**
```
matrix:
  os: [ubuntu-latest, macos-latest, windows-latest]
  arch: [arm, arm64, x86, x86_64]
```
https://github.com/xyzz/openmw-android/actions/runs/65368513/workflow

**Signing code**
There are already actions for that, or you could add your own:
https://github.com/marketplace/actions/sign-windows-binary-with-timestamp-server-rotation

**Multiple badges**
I know it's possible to have a badge per branch, and per event. But don't think per matrix is supported.
However you could workaround by either:
1. Copy/pasta same commands into three separate files (not ideal)
2. Create a reusable build workflow, and link to it from three separate platform files.
- linux.yml -> build.yml
- mac.yml -> build.yml
- win.yml -> build.yml
https://docs.github.com/en/actions/using-workflows/reusing-workflows

Thanks for you time and assistance with helping me.

PR #1146, opened 2022-08-30
Fix source comment

luzpaz:
Found via downstream https://github.com/Ardour/ardour/pull/721#pullrequestreview-1090606322


derselbst (approved):
Oh, what a nice code duplication we've got here. Thanks.

PR #1156, opened 2022-09-19
Add linear interpolation to convex and concave transfer functions

derselbst:
While looking through forks, I found a development branch from @md1872b. I don't understand most of the changes there, but this one here makes complete sense to me. Originally only applied to `fluid_convex` I also applied it to `fluid_concave`.

@mawe42 FYI


md1872b:
Great, there are some small changes I applied, maybe some are application-specific. Free to discuss them.

Issue #24, opened 2010-05-26
MIDI file player ends before all notes have been played.


derselbst:
*Originally created by:* David Henningsson

Two problems must be addressed:  

- [x] The MIDI player should wait for all voices to terminate after the last MIDI note, plus some extra time for reverb.  
- [ ] Client programs (like [FluidSynth](/apps/trac/fluidsynth/wiki/FluidSynth) CLI) should wait for all audio buffers to be played before terminating, it's not enough that the player has finished.  

Reference:  
<http://lists.gnu.org/archive/html/fluid-dev/2009-03/msg00040.html>  

Reported by: *anonymous

Original Ticket: [fluidsynth/tickets/24](https://sourceforge.net/p/fluidsynth/tickets/24)

tstarling:
I found this task because I was listening to some rendered audio files on Wikipedia and I was annoyed about the abrupt way that they terminate. See e.g. https://en.wikipedia.org/wiki/Mozart%27s_starling . Our Score extension uses LilyPond to generate MIDI files which are then converted to Ogg using FluidSynth with the -F option. I suppose I could add 500ms of silence as a second MIDI file to the command line, although it's a bit of a hack since we don't know how long the sound will take to decay. 

I appreciate that it's nice to have a way to generate files that end exactly on the beat, for looping. For our purposes, it would be fine to have a command line option or config setting.

Foxy6670:
Hate to revive an old thread, but this is STILL HAPPENING in FluidSynth 2.1.7.

derselbst:
> Hate to revive an old thread, but this is STILL HAPPENING in FluidSynth 2.1.7.

That's why this ticket is still open. I welcome your PR.

topaz:
Other than situations where FluidSynth reverb settings are high, I've alleviated big chunks of this issue in #1159 and #1167. They cause FluidSynth to wait for active voices to finish plus an additional two seconds (for effects like reverb) before stopping playback. Previously, FluidSynth would stop playback immediately once the last MIDI event was processed. The "wait two seconds" part would be better handled by a mechanism for detecting when audio actually drops below some minimum threshold, but I'm not equipped to implement such a feature at this time; ideas for that logic in particular are being captured in #1165.

36000000:
I have problem only with Jack2 or the pulse driver. With the alsa driver fluidsynth 2.1.7 is playing fine.
good:
[...]

PR #1159, opened 2022-09-20
Don't stop playing until zero voices are active and reverb has a chance to fade out

topaz:
In an attempt to alleviate 12-year-old #24 without a deep understanding of FluidSynth internals, this PR adds a `player.extend-playback` setting (and corresponding `-x` / `--extend-playback` CLI option) which takes a number of milliseconds to wait after the last MIDI event before stopping playback.  This PR makes no attempt to solve any of the more complex problems like automatic silence detection; `-x 2000` is more than enough for my use cases, and so I hope this change is useful to others as well.

Reviewers beware: I'm not 100% fluent in C, knew almost nothing about FluidSynth internals a week ago, and have very likely caused a mess of things, so please be careful when reviewing this PR. In particular, I'm not sure whether claiming `-x` was appropriate, whether I've filled in documentation in all the right places/ways, whether the added logic is even in the right place, whether I've caused weird behavior in edge cases involving stopping/restarting/reloading a player or playing multiple files, or how to write tests for any of those situations.

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -1687,6 +1687,7 @@ new_fluid_player(fluid_synth_t *synth)
     player->deltatime = 4.0;
     player->cur_msec = 0;
     player->cur_ticks = 0;
+    player->end_msec = -1;
     player->last_callback_ticks = -1;
     fluid_atomic_int_set(&player->seek_ticks, -1);
     fluid_player_set_playback_callback(player, fluid_synth_handle_midi_event, synth);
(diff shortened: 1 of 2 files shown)


topaz:
I think I've made the changes you requested. For my use cases, waiting for zero active voices plus two seconds for reverb leaves almost 2 seconds of silence; when reverb is turned way up, 2 seconds isn't nearly enough and the audio still cuts out. I'm not sure how to handle it better beyond putting back a setting that lets the user choose the amount of extra time.

derselbst (commented):
Ok, that looks better already, thanks!

Regarding the reverb / grace time issue, you could try whether a simple weighting of the hardcoded value would give better results, like so:

`fluid_synth_get_reverb_roomsize(player->synth) * fluid_synth_get_reverb_level(player->synth) * FLUID_PLAYER_STOP_GRACE_MS`

`roomsize` and `level` will in range of about [0;1], so the grace time would represent the maximum value in this case (probably you need to increase it then, 2 sec was just a wild guess).

I think having absolutely no silence in between is not possible that way, but I don't consider it to be a use-case either. Also, I would question whether we should respect the case of "turning reverb all the way up". Just having a reasonably well working default behavior would be nice. Perhaps you could have a try.

In case it turns out to be not that simple, I would like to have a try myself. In this case pls. give me some more information of how you test this, e.g. which reverb settings, which soundfont and possibly also with MIDI files you use.

topaz:
I'm happy to do the modeling, at least initially.

In the interest of maybe doing the more complete solution after all, is there somewhere we can currently (or add a feature to) query the volume of the audio currently coming out of the synth?  It seems like we could just wait until it drops below some threshold. I don't know enough about audio synthesis to even know whether that's a sane question to ask.

mmontag:
I think this would be a nice addition to fluidsynth.

Just for fun...I have my own hacky solution to the general problem:
https://github.com/mmontag/chip-player-js/blob/master/src/players/MIDIFilePlayer.js#L236
It just waits for samples to stay below value X for a duration Y, but it helps a lot with the whole "gapless playback" experience.

derselbst (approved):
Thanks for your investigations! Since you're happy with this initial approach, I'm fine as well. To take care of this "silence detection" part, I've opened  #1165. Your input is welcome!

PR #1162, opened 2022-09-21
Remove last references to obsolete CMake option LIB_SUFFIX

h3xx:
This option has been ignored since 563592aa3d5e585282877cd454b50369e3b19cdd (2021-09-03) when the project was changed to use GNUInstallDirs instead.


derselbst (approved):
Ok, thanks!

PR #1164, opened 2022-09-24
two more small examples for libfluidsynth

pedrolcl:
fluidsynth_instruments.c: enumeration of banks and programs of a soundfont to the standard output.

fluidsynth_enumsettings.cxx: enumeration of settings and default values in C++11 to the standard output.


derselbst (approved):
Thanks!

PR #1167, opened 2022-09-27
Disable sustain/sostenuto pedals after last MIDI event

topaz:
While working with a larger set of MIDI files with the changes in #1159, I discovered that for some of the files, `fluid_synth_get_active_voice_count(player->synth)` never goes to zero (at least, not after as long as I was willing to wait for it, maybe 15-30 seconds). This causes the player to appear to terminate automatically on some songs bug hang indefinitely on others.

In an attempt to prevent that situation, this PR changes the post-MIDI-events player logic in #1159 from "wait until `fluid_synth_get_active_voice_count(player->synth)` is zero, then wait until `FLUID_PLAYER_STOP_GRACE_MS` have elapsed" to instead "wait until _either_ `fluid_synth_get_active_voice_count(player->synth)` _or_ `FLUID_PLAYER_STOP_GRACE_MS` have elapsed".

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -1688,6 +1688,7 @@ new_fluid_player(fluid_synth_t *synth)
     player->cur_msec = 0;
     player->cur_ticks = 0;
     player->end_msec = -1;
+    player->end_pedals_disabled = 0;
     player->last_callback_ticks = -1;
     fluid_atomic_int_set(&player->seek_ticks, -1);
     fluid_player_set_playback_callback(player, fluid_synth_handle_midi_event, synth);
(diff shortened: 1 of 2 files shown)


derselbst:
> I discovered that for some of the files, fluid_synth_get_active_voice_count(player->synth) never goes to zero

Oh really? Sounds like a bug. Could you provide an example MIDI where you've observed that?

topaz:
Sure! I'm using a recent commit on master (`89a2c4b4`) that includes my change in #1159.

Just for fun, add a debug line at `src/midi/fluid_midi.c` line 2184:

```
        /* Once we've run out of MIDI events, keep playing until no voices are active */
        if(status == FLUID_PLAYER_DONE && fluid_synth_get_active_voice_count(player->synth) > 0)
        {
            printf("waiting for zero voices; voices active: %d\n", fluid_synth_get_active_voice_count(player->synth));
            status = FLUID_PLAYER_PLAYING;
        }
```

Then, I tried to fast-render [jtbach.mid](https://modland.com/incoming/vault/MIDI/Everquest/jtbach.mid) using the FluidSynth-provided SF2 file `sf2/VintageDreamsWaves-v2.sf2` (not sure if this is important; another SF2 file I tried also had this behavior):

```
path/to/fluidsynth -F tmp path/to/sf2/VintageDreamsWaves-v2.sf2 path/to/jtbach.mid
```

(Realtime-render does the same thing if you wait for the song to end, and other SF2 files also seem to do the same thing.)

When I do this, it eventually just repeatedly spams:
```
waiting for zero voices; voices active: 34
```

derselbst:
Ok thanks! The problem is that MIDI channel 0 activates the sustain pedal pretty much at the end of the track, but doesn't disable it, causing some notes to keep playing in sustain forever. Therefore I would suggest the following more straightforward fix:

```
        /* Once we've run out of MIDI events, keep playing until no voices are active */
        if(status == FLUID_PLAYER_DONE && fluid_synth_get_active_voice_count(player->synth) > 0)
        {
            int i;
            /* disable hold pedals on all channels to release any playing voices */
            for(i = 0; i < synth->midi_channels; i++)
            {
                fluid_synth_cc(player->synth, i, SUSTAIN_SWITCH, 0);
                fluid_synth_cc(player->synth, i, SOSTENUTO_SWITCH, 0);
            }
            status = FLUID_PLAYER_PLAYING;
        }
```

It worked fine with the MIDI you provided. You might want to test a few more.

derselbst (approved):
Great, thanks!

Issue #1157, opened 2022-09-19
Support different reverb engines


derselbst:
### Related discussion
https://lists.nongnu.org/archive/html/fluid-dev/2022-05/msg00014.html

### Is your feature request related to a problem?
I want to implement a reverbator that produces similar results as the Ultra 64 SDK.

### Describe the solution you'd like
It would be nice to select different reverb engines when creating the synth. This would allow to re-enable the old freeverb engine, and for a potentially higher quality FDN reverb (by using 12 delay lines rather than only 8), and it would allow for a "Lexicon 224"-like reverbator, that I have been playing around with.

### Describe alternatives you've considered
Alternative 1)
Feed the dry audio to some external program. This seems to have been suggested [here](https://lists.nongnu.org/archive/html/fluid-dev/2022-05/msg00016.html) (although I didn't fully understand how). In any case, this approach has the drawback that it would be almost impossible to get the amplitudes for the individual fx buffers right and therefore respect CC91 correctly.

Alternative 2)
Feed the dry audio to LADSPA plugins and process it there. Not sure if it would be possible to respect CC91 correctly. Additionally, Reinhold made the following note:
> Using the LADSPA architecture on Windows is possible
according to the documentation, but dealing with an additional plugin (dll)
would have a wider impact (building, intregrating into installers). I guess
that the overall file size would be bigger than option (1). Setting the
reverb level per effect-groups (effect-group == midi channel == instrument)
is key when using the MIDI CC event 91 as reverb indicator for each
instrument.  

### Additional context
To implement this, one should rewrite the reverbator into an abstract C++ class, so that derived classes can implement a particular reverbator. Class template would allow to change previously hard-coded values like `NBR_DELAYS`.

The existing setting `synth.reverb.active` could be used to indicate which engine should be used at synth creation. Alternatively, yet another setting could be introduced.

@mawe42 FYI

mmontag:
Strong vote for option 1 (Extend synth.reverb.active from boolean to an enum integer, to allow 
selecting different reverb engines when the synth is created).

Different reverb options would be great.  BTW, do you know any open source implementations with a Lexicon 224-like architecture?

derselbst:
> BTW, do you know any open source implementations with a Lexicon 224-like architecture?

Unfortunately not. But I've seen a simple one in the N64 dev kit, which we may use as "source of inspiration".

mmontag:
[...]

PR #1170, opened 2022-10-08
Support stereo fx send to LADSPA

derselbst:
For details, see #1161

Open questions:

* Should this become fluidsynth's default way of handling fx?
* Or should we make this feature configurable?
  * at runtime via a setting?
  * or at compile-time via a `#define` / CMake option?

### DIFF
--- a/src/rvoice/fluid_rvoice.h
+++ b/src/rvoice/fluid_rvoice.h
@@ -131,9 +131,23 @@ struct _fluid_rvoice_dsp_t
     fluid_real_t phase_incr;	/* the phase increment for the next FLUID_BUFSIZE samples */
 };
 
+#define STEREO_FX
+
+typedef enum  {
+DRY_L = 0,
+DRY_R,
+REV_L,
+CHO_L,
+#ifdef STEREO_FX
+REV_R,
+CHO_R,
+#endif
+RBUF_COUNT
+}  fluid_rvoice_bufidx_t;
+
 /* Currently left, right, reverb, chorus. To be changed if we
    ever add surround positioning, or stereo reverb/chorus */
-#define FLUID_RVOICE_MAX_BUFS (4)
+#define FLUID_RVOICE_MAX_BUFS (RBUF_COUNT)
 
 /*
  * rvoice mixer-related parameters
--- a/src/rvoice/fluid_rvoice_mixer.c
+++ b/src/rvoice/fluid_rvoice_mixer.c
@@ -306,6 +306,24 @@ fluid_mixer_buffers_prepare(fluid_mixer_buffers_t *buffers, fluid_real_t **outbu
             ? &base_ptr[(fx_idx + SYNTH_CHORUS_CHANNEL) * FLUID_BUFSIZE * FLUID_MIXER_MAX_BUFFERS_DEFAULT]
             : NULL;
     }
+    #ifdef STEREO_FX
+    base_ptr = fluid_align_ptr(buffers->fx_right_buf, FLUID_DEFAULT_ALIGNMENT);
+
+    for(i = 0; i < buffers->mixer->fx_units; i++)
+    {
+        int fx_idx = i * fx_channels_per_unit;
+
+        outbufs[offset + fx_idx + SYNTH_REVERB_CHANNEL + 2] =
+            (with_reverb)
+            ? &base_ptr[(fx_idx + SYNTH_REVERB_CHANNEL) * FLUID_BUFSIZE * FLUID_MIXER_MAX_BUFFERS_DEFAULT]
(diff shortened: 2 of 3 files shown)


derselbst:
@mawe42 In case you have some time these days, could you perhaps have a look at these changes and the related discussion? I'm struggling a bit whether this is useful in general as this is open for over a year now and no comments or feedback have arrived.

mawe42:
So far I've only looked at the motivation behind this change and the diff, so I haven't tried it out yet. But from what I can see, this looks like a valuable change. And I'm leaning towards having the stereo fx send always enabled, without config option or compile-time switch.

But that would either require three fx sends (mono, left, right), or we would need to change the multi-channel audio drivers as well. For example, the jack driver only supports one reverb and chorus channel at the moment. If we would take the change as-is and always define `#STEREO_FX`, then the jack sends would not be mono anymore but only send the left channel.

Not sure what it would mean in terms of performance, but it might be easiest to go for three sends (mono, left + right). Expose all via the LADSPA interface (which would also remove any backwards incompatibilities) and later extend the multichannel audio drivers to also expose the left + right sends.

derselbst:
Thanks Marcus! I think the jack driver is the only one to handle effects channels, i.e. dsound, wasapi and waveout only handle dry channels currently.

So changing jack driver to expose the left and right fx channels should be straight forward.

The only benefit for an additional mono channel I see would be to retain backward compatibility in LADSPA, for e.g. `Reverb:Send`, correct?

mawe42:
I think the mono channels would also be required for the internal effects. Currently the sends for the internal reverb and chorus effects only consider `voice->reverb_send` for the amp calculation. With this change, the first effect buffer (which is used by the internal effects) becomes the left effect send and also considers `voice->pan` and `voice->balance`. Voices that pan hard right would not send anything to the internal effects anymore.

Instead of a mono send, we could also sum the left and right sends together before handing them to the internal effects. But do we might run into phase problems here. Hm... more thinking required :-)

mawe42:
Coming back to this... maybe a way forward would be to upgrade the internal effects to stereo? It would change the "sound" of FluidSynth, but probably for the better. And I wouldn't be too concerned about performance, as computing power even on embedded devices has increased quite a lot since FluidSynth was first written. If we are concerned about performance, we could leave the complile-time switch in but make it default to stereo fx. What do yo think?

derselbst:
Those who are really concerned about performance would disable reverb and chorus anyway. I'm all in favor to upgrade internal fx to stereo. If you've got some time to spare, that would be great :)

mawe42:
But maybe some input from @mrbumby409 and @jjceresa would be good before I start moving the internal effects to stereo. From what I've read so far, using a stereo reverb instead of mono would quite drastically alter the sound of FluidSynth. Stereo reverbs seem to give more of a "physical room" effect while mono reverbs contribute more of a "sound depth" instead. Having mono reverb (and chrous?) might actually be much better for the default effects of FluidSynth.

Maybe the better way forward would be to expose stereo fx sends via LADSPA, but sum the stereo sends before feeding them into our internal effects?

mawe42:
After more reading on stereo vs. mono reverbs, it seems to me that summing the two fx sends would be the better solution here. But I'll wait for other opinions before going any further.

mawe42:
#1157 seems to be closely related to this PR.

derselbst:
> After more reading on stereo vs. mono reverbs, it seems to me that summing the two fx sends would be the better solution here.

The Lexicon like reverb that I was also planning to implement as part of the linked ticket is just a mono reverb as well (but it reads from stereo input). So the summing option you've suggested sounds fine to me.

jjceresa:
>Marcus, After more reading on stereo vs. mono reverbs, it seems to me that summing the two fx sends would be the better solution here....

Yes, as actually, internal effects (reverb and chorus) are able to produce true stereo output from mono input . This is due to multiple delays lines presence in those effects from which it is quite easy to pick internal multiple signal to produce multiple output  (i.e stereo) out of phase. So, I think it is sufficient to keep mono input for this kind of internal effects.

For LADSPA unit input, it is better to expose stereo input.  Indeed, as we cannot predict the internal capabilities of the LADSPA loaded effect, it is good for this effect not to be constrained only by mono input.

RoqueDeicide:
Here are my 2 cents on this. I've been working on a soundfont, and in the process of testing it with various tracks, I've noticed, that FluidSynth's current chorus implementation nearly completely neuters most panning effects used in the tracks. It makes tracks sound less dynamic and requires some programs to be made louder, to make them stand out more without the benefit of being panned away from programs on other channels. So, I think chorus should be fully stereo, to allow tracks to take full advantage of panning effects.

PR #1174, opened 2022-10-15
drivers/fluid_alsa: don't try to close invalid ctls

Bob131:
snd_ctl_close is called unconditionally after snd_ctl_open in fluid_alsa_rawmidi_driver_settings, resulting in a segfault if snd_ctl_open failed to open the card. This commit has the function call snd_ctl_close only when the ctl was successfully opened.


derselbst (approved):
Ok, thanks!

PR #1181, opened 2022-11-03
systemd: restore sandboxing options and add PrivateUsers=yes to make them work as user units

bluca:
The alternative is to also remove ProtectSystem, which also requires PrivateUsers. At some point we'll make it implicit, and only a handful of user units use it, so we want to make sure it is intentional to have this and not accidental.


derselbst (approved):
Thank you very much! I don't really know about systemd, glad you took care of this :)

PR #1199, opened 2023-01-07
Implement MSGS-style Drum Note Cut

derselbst:
This PR implements the drum note cut used by Microsoft's software synth as discussed in #1196. Whenever a NoteOn is played on a drum channel, any previously playing voice for the same note is now being canceled / cut.

I am not sure if this make musically sense, whether this is permitted by the SF2 spec and whether it breaks any use-case. Any feedback is therefore highly welcome!


vladkosi:
I have tested this PR on Doom soundtrack and the drums sound closer to a real Roland SC-55 now. Good job!

Comparison between version 2.3.1, this PR and a recording from an SC-55:

fdelapena:
+1 to this. It improves playback when using Windows' gm.dls _inspired_ soundfonts, drum note length matches the Microsoft Synthesizer behaviour with this patch.

derselbst:
Thanks for the feedback. Will become part of fluidsynth 2.4.0.

PR #1209, opened 2023-02-01
Enabled processing all sysex messages with synth.device-id=127

trolley813:
As discussed in #1206. 
When `synth.device-id` is set to the "broadcast" value of 127, fluidsynth will handle all SYSEX messages, regardless of the actual device ID specified.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -108,10 +108,10 @@ Developers:
             <type>int</type>
             <def>0</def>
             <min>0</min>
-            <max>126</max>
+            <max>127</max>
             <realtime/>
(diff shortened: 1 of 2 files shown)


derselbst (commented):
Yeah, almost. You forgot to adjust the new max limit here:
https://github.com/FluidSynth/fluidsynth/blob/9ddf46d0905af377b7609b1a95caf53f6542ff32/src/synth/fluid_synth.c#L232
Then it should work.

trolley813:
Thank you! Finally got it to fix.

derselbst (approved):
Looks good, thanks!

Issue #1212, opened 2023-02-19
Support also Windows drivers on CYGWIN


carlo-bramini:
At the time of writing, Fluidsynth can be compiled for running on CYGWIN and it uses the traditional drivers available on an UNIX platform. However, CYGWIN is also able to use native audio drivers for Windows because it supports a version of WIN32API that it is able to exist together with POSIX includes. Hopefully, it is just needed to do few small changes for supporting the drivers for Windows also on CYGWIN, together with the ones that run on POSIX.

EDIT:
Supporting the drivers for Windows on CYGWIN requires these steps:

1. Replacing all references to `WIN32` with `_WIN32`.
This is required because `WIN32` is defined after including `Windows.h` on CYGWIN, but `_WIN32` isn't.
2. Fixing `CMakelists.txt` for including also CYGWIN when dsound, winmidi, waveout and wasapi are checked.
3. Include `Windows.h` into `src/utils/fluid_sys.h` when `__CYGWIN__` is defined.

After that, CYGWIN will be able to run also the drivers for Windows.

PR #1213, opened 2023-02-19
Support Windows Driver on CYGWIN.

carlo-bramini:
`_WIN32` is an intrinsic compiler macro that is always provided by the compiler. Instead, `WIN32` is defined by the project template or after you included Windows.h (for example). I suggest to use `_WIN32` instead of `WIN32` macro.
Nothing will change at compile time, but this is the first step to take for supporting Windows drivers also on CYGWIN, see issue #1212. Next, a second commit adds support for Windows drivers under CYGWIN, as explained into issue #1212.
Tested without problems with MSVC 2019/2022 for x86/x64 and i686/x86_64 under MinGW-w64.
Tested second commit on the latest CYGWIN 64-bits.


derselbst (approved):
LGTM, thanks Carlo!

PR #1220, opened 2023-03-11
Added fluid_player_get_division

Fruchtzwerg94:
Added getter for division which is helpful to calculate the MIDI track length.

See also [#1112](https://github.com/FluidSynth/fluidsynth/discussions/1112)


derselbst (approved):
Ok, thanks!

PR #1223, opened 2023-03-16
CYGWIN: set correct prefix for library name.

carlo-bramini:
Under CYGWIN, the name of a library must start with the "cyg" prefix.


derselbst (approved):
Ok, thanks.

PR #1225, opened 2023-03-21
CMake: Use upstream name for mpg123

FtZPetruska:
The upstream project uses `mpg123` for package name, despite the uppercase namespace (`MPG123::`).
We should match that naming for better interoperability with the upstream config.

I also change the logic for the `*_FOUND` variables, since all we're interested in is `MPG123::libmpg123` for SndFile.
Now:
- `mpg123_libmpg123/libout123/libsyn123_FOUND` is set to true when the respective library is found
- `mpg123_FOUND` is set to true when libmpg123 and its include directories are found


derselbst (approved):
Thanks!

Issue #1224, opened 2023-03-18
Incomplete pkg-config module when using OpenMP.


FtZPetruska:
### FluidSynth version
_Execute `fluidsynth --version` and provide the output._
```
FluidSynth runtime version 2.3.2
Copyright (C) 2000-2023 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.3.2
Sample type=double
```

### Describe the bug
_Provide a clear and concise description of the current situation, e.g. how the bug manifests._

When building a programs against fluidsynth using the flags provided by pkg-config, the link step fails with undefined references to OpenMP.

### Expected behavior
_Provide a clear and concise description of what you expected to happen._

The program builds successfully.

### Steps to reproduce
_Please explain the steps required to duplicate the issue, esp. if you are able to provide a sample application. E.g. how to start fluidsynth, what shell commands to enter, what midi events to send, etc._

- Build fluidsynth statically with OpenMP enabled, OpenMP needs to be an external library (such as LLVM's libomp)
- Try to link a program using `pkg-config --libs --static fluidsynth`
- Linking fails due to undefined references to OpenMP functions

### Additional context
_If you are able to illustrate the bug with an example, please provide simple
source code below or as attached file. List any other information that is relevant to your issue, e.g. stack traces, related issues, build logs, suggestions on how to fix, links to related discussions at fluid-dev, etc._

I noticed the following snippet in `cmake_admin/PkgConfigHelpers.cmake`:
```cmake
get_target_property (_libs ${target} INTERFACE_LINK_LIBRARIES)
set(_cleanlibs)
foreach(_lib IN LISTS _libs)
    if (TARGET ${_lib})
        # All the imported PkgConfig target are explicitly added to PC_REQUIRES_PRIV.
        # Do not duplicate them into the Libs.private section, as they will be already part of Requires.private
    else()
        list(APPEND _cleanlibs ${_lib})
    endif()
endforeach()
```

This excludes all targets, which includes `OpenMP_C`.

Similarly, this snippet will add PulseAudio's libraries since they aren't linked through a target, despite PulseAudio providing a pkg-config module.

A possible solution could be to build a list similar to `PC_REQUIRES_PRIV` that contains libraries which do not have pkg-config module files, instead of relying on the `INTERFACE_LINK_LIBRARIES` of the libfluidsynth target. For example:

```cmake
[...]

PR #1226, opened 2023-03-26
Add an integration test for consuming fluidsynth via pkg-config

derselbst:
This adds an additional build step to MacOS and Linux CI to make sure a static libfluidsynth can be consumed via `pkg-config`.

It's intended as a regression test for #1224.

### DIFF
--- a/.azure/azure-pipelines-mac.yml
+++ b/.azure/azure-pipelines-mac.yml
@@ -26,10 +26,10 @@ jobs:
     matrix:
       UnixLibs11:
         imageName: 'macos-11'
-        CMakeFlags: '-Denable-framework=0'
+        CMakeFlags: '-Denable-framework=0 -Denable-readline=0'
       UnixLibs11_static:
         imageName: 'macos-11'
-        CMakeFlags: '-Denable-framework=0 -DBUILD_SHARED_LIBS=0'
+        CMakeFlags: '-Denable-framework=0 -DBUILD_SHARED_LIBS=0 -Denable-readline=0'
       #UnixLibs12:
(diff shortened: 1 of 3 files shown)


.azure/azure-pipelines-mac.yml
```diff
@@ -42,14 +51,15 @@ jobs:
...
         brew install $PACKAGES
       displayName: 'Prerequisites'
     - script: |
         set -ex
         mkdir build && cd build
-        export PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig"
-        cmake -Werror=dev $(CMakeFlags) -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=1 -DNO_GUI=1 ..
+        export PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig:/usr/local/opt/readline/lib/pkgconfig"
+        # assist cmake in finding openmp installed via homebrew
+        cmake -Werror=dev $(CMakeFlags) -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=1 -DNO_GUI=1 -DCMAKE_PREFIX_PATH=/usr/local/opt/libomp ..
```
FtZPetruska, line 62: It may be preferable to use `$(brew --prefix libomp)` instead of hardcoding the path. It should make it a little more future-proof if either Homebrew or Azure changes their layout.

derselbst: Done, as well as for readline and libffi. Thanks!

PR #1228, opened 2023-04-01
Fix consuming static libfluidsynth via pkg-config

### LINKED ISSUE #1224: Incomplete pkg-config module when using OpenMP.

derselbst:
This fixes static linking problems that came up with issue #1224, in particular:

* OpenMP libraries are now included as private libs in fluidsynth.pc file
* Readline libraries are included as private libs in fluidsynth.pc, unless readline is provided via pkg-config

Additionally, the Mac CI now builds with and without a pkg-config version of deadline.

Resolves #1224


derselbst:
@FtZPetruska FYI

FtZPetruska (approved):
Thank you for the heads-up!

The changes look good. They work great on my machines. Thanks a lot for your help!

derselbst:
You're welcome!

PR #1230, opened 2023-04-05
Fix choppy sound when selecting pipewire output

nilninull:
Because forgot to set the period size in pipewire.


derselbst (approved):
Thanks!

Issue #1229, opened 2023-04-04
Build failure on Alpine Linux


PureTryOut:
### FluidSynth version
2.3.2

### Describe the bug
Build failure.

```
[...]

PR #1244, opened 2023-05-11
Fix implicit declaration of function strcasecmp with musl

listout:


derselbst:
Well yeah, that's the same fix as mine in 0b40da5 . I would be more interested in a way to reproduce this build failure in a CI pipeline, particularly in #1243. But the Alpine build succeeds there: https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=9127

I don't see how the question of C99 compliance is relevant here, esp. because fluidsynth is a C89 project currently.

listout:
> I don't see how the question of C99 compliance is relevant here, esp. because fluidsynth is a C89 project currently.

You're correct, by bad. Let me correct that.

derselbst:
Ok, thanks for the pointer. I'll double check any compiler flags with the apline build when I find the time.

According to #1229 this is a problem on Alpine as well, but this issue lacks precise instructions on reproducing it.

CMakeLists.txt
```diff
@@ -159,6 +159,7 @@ include ( CheckIncludeFile )
 include ( CheckSymbolExists )
 include ( CheckTypeSize )
 check_include_file ( string.h HAVE_STRING_H )
+check_include_file ( string.h HAVE_STRINGS_H )
```
sevz17, line 162: This should be `strings.h`

thesamesam: whoops, thanks

derselbst:
Superseded by #1245

PR #1259, opened 2023-07-01
Add fluid_thread

kyledecot:
There have been a lot of discussions about wanting to get rid of glib (my personal reasons are to make FS work w/ WASM). As others have pointed out most of the use of glib is for threading. As a (very small) first step I think that some indirection needs to be added so that `fluid_sys` is not riddled with `g_*` calls. 

If creating `fluid_thread.h` seems like a good idea to others the various other `fluid_thread_*` functions can be moved over as well.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -121,6 +121,7 @@ set ( libfluidsynth_SOURCES
     utils/fluidsynth_priv.h
     utils/fluid_sys.c
     utils/fluid_sys.h
+    utils/fluid_thread.h
     sfloader/fluid_defsfont.c
     sfloader/fluid_defsfont.h
     sfloader/fluid_sfont.h
--- a/src/utils/fluid_sys.c
+++ b/src/utils/fluid_sys.c
@@ -1002,7 +1002,7 @@ void fluid_profile_start_stop(unsigned int end_ticks, short clear_data)
 fluid_cond_t *
 new_fluid_cond(void)
 {
-    if(!g_thread_supported())
+    if(!fluid_thread_supported())
     {
         g_thread_init(NULL);
     }
@@ -1048,7 +1048,7 @@ new_fluid_thread(const char *name, fluid_thread_func_t func, void *data, int pri
     /* Make sure g_thread_init has been called.
      * Probably not a good idea in a shared library,
      * but what can we do *and* remain backwards compatible? */
-    if(!g_thread_supported())
+    if(!fluid_thread_supported())
     {
         g_thread_init(NULL);
     }
--- a/src/utils/fluid_sys.h
+++ b/src/utils/fluid_sys.h
@@ -40,6 +40,7 @@
 #define _FLUID_SYS_H
 
 #include "fluidsynth_priv.h"
+#include "fluid_thread.h"
 
 #if HAVE_MATH_H
 #include <math.h>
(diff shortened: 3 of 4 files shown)


derselbst:
What exactly is the intention behind this? You're abstracting away `g_thread_supported()`, but you keep `g_thread_init()` or `g_static_private_init()`. If people want to get rid of glib in their custom forks, they keep replacing entire functions or macros, like `fluid_mutex_init` or `new_fluid_cond`, in which case the content of those doesn't matter.

I'm sorry, but I don't really see a benefit in this change.

kyledecot:
@derselbst thanks for your quick reply. I only moved a single function / opened the PR to start the conversation before moving all other thread functions. 

While moving these functions out of the current file wouldn't have any **immediate** benefit  it does corral some of the logic  that is currently backed by `g_` into a single place where future work could make the threading backend swappable, overridable, etc.

derselbst:
> Since a good chunk of the glib code has to do with threading I think it makes sense to corral that into it's own file for a couple of reasons:

My problem with this approach is that whatever functions you put into fluid_thread, they will be custom: custom signatures, custom implementation, non-standard behaviour. That's not good.

If you really want to factor out that threading stuff, why not mocking a C11 compatible API? I.e. fluidsynth would call `thrd_create()` to create a thread. The implementation of that function lives in `fluid_thread` and contains the regular [glib based implementation](https://github.com/FluidSynth/fluidsynth/blob/a85aa9811ca511809964fd55f692120baa362640/src/utils/fluid_sys.c#L1038). Instead of typedefing `fluid_thread_t`, we'd typedef `thrd_t` to be glib's type. Similar approach for atomics and cond variables.

People then have the option to

1. Reimplement that C11 compatible API with C++11 and submit this back to upstream, as this would provide most portable implementation.
2. Do not reimplement that C11 API and instead use the C11 implementation of a compatible standard library (musl, glib?, MSVC??)
3. Reimplement that C11 compatible API with a customized OS specific implementation.

Even though option 1. would not solve, that we'd still need glib for DLS support (libinstpatch), I would still accept it, because:
* It would be WinXP compatible, and
* It would not necessitate us to move entirely to C++.

What do you think about this? Would you want to do this? I can't promise I can  assist much in coding though... busy schedule.

ReinholdH:
Let me add my 2 cents here, too.

We use the libfluidsynth dlls with glib for our Windows apps for almost 10 years. This setup is very stable and robust. No issue has been observed in the context of fluidsynth/glib over this period.

Once, I also voted to find a solution w/o glib in order to remove complexity and to reduce the size. But I changed my mind. Our Windows apps in msi format have a size of ~40MB. Because the Windows app installer highly compresses the files, without glib our msi file is only 1.5% smaller than with glib.  

That said, given the existing robustness and on the other side the complexity associated by removing glib, I strongly vote to leave the setup fluidsynth/glib as is.

My 2 cents from our point of view... 

Thx

luzpaz:
Maybe making a parallel branch without glib support and providing downloads so testers can sort out bugs that arise ?

derselbst:
> Maybe making a parallel branch without glib support and providing downloads so testers can sort out bugs that arise ?

I'm not willing to maintain parallel branches, sry. If all you want is a "branch", pls. go for one of the various glib-free forks.

I do welcome a clean and backward-compatible approach as I've outlined above. But it does seem to me that people are losing interest when they realize that I'll not accept incomplete, quick and dirty approaches. And since there is no update from the author either, I'm closing this.

If anyone has similar ideas or solutions, pls. bring them up in our related discussion  #847 before spending time implementing anything.

PR #1261, opened 2023-07-10
Fix incorrect way of turning CMAKE_INSTALL_LIBDIR absolute

OPNA2608:
`${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}` is not guaranteed to make sense because `CMAKE_INSTALL_LIBDIR` may be specified as an absolute path, in which case this concatenation is nonsense.

Use `CMAKE_INSTALL_FULL_LIBDIR` instead, which will use `CMAKE_INSTALL_LIBDIR` if it's already absolute, or concatenate it onto `CMAKE_INSTALL_PREFIX` if it isn't.


derselbst:
Thanks!

PR #1265, opened 2023-08-04
Fallback to IPv4 when creating socket if IPv6 is not available (#1208)

### LINKED ISSUE #1208: Implement IPv6 to IPv4 fallback
### Related discussion
#1205 

The `enable-ipv6` cmake option disables IPv4 support, causing fluidsynth's server to fail starting on system which have disabled IPv6 runtime support

Ideally, get rid of the `enable-ipv6` option altogether. This can be done by
* either always trying to open IPv4 and IPv6 sockets, or
* first trying IPv6 and if it fails with `EAFNOSUPPORT` use an IPv4 socket.

none

### Additional context
Low priority issue, contribution required.

ivan-zaera:
Fixes #1208

### DIFF
--- a/src/utils/fluid_sys.c
+++ b/src/utils/fluid_sys.c
@@ -1606,12 +1606,12 @@ fluid_server_socket_t *
 new_fluid_server_socket(int port, fluid_server_func_t func, void *data)
 {
     fluid_server_socket_t *server_socket;
+    struct sockaddr_in addr4;
 #ifdef IPV6_SUPPORT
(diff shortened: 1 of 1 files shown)


ivan-zaera:
I can only compile the project in Linux, so I'm going to need some guidance on how to make failing platforms compile...

derselbst (approved):
Looking good, thanks!

The failing builds were already failing before. I'll try to find some time looking into that.

PR #1266, opened 2023-08-11
Fix issue with CMake Xcode generator

bradhowes:
There is a known issue with CMake and its Xcode generator when the target has no sources of its own -- see

https://gitlab.kitware.com/cmake/cmake/-/issues/23688

A workaround is to define an empty source file that can be used to trigger the build path for the framework. Without this hack, the framework is missing the shared library component.

With this change, the following now runs successfully:

```
% cmake -G Xcode -Denable-sdl2=NO -Denable-readline=NO -Denable-dbus=NO -DCMAKE_INSTALL_PREFIX=$PWD .. 
% xcodebuild
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -205,7 +205,7 @@ set ( public_main_HEADER
     ${FluidSynth_BINARY_DIR}/include/fluidsynth.h
 )
 
-configure_file ( ${FluidSynth_SOURCE_DIR}/include/fluidsynth/version.h.in 
+configure_file ( ${FluidSynth_SOURCE_DIR}/include/fluidsynth/version.h.in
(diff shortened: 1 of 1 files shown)


derselbst (approved):
Thanks Brad! I just added this to our MacCI and it works well.

Without your change:
https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=9499&view=logs&j=7055962f-51ce-5353-72ce-09bdafd645f3&t=d31a795d-4576-5a97-343e-1c458f9914b8

With your change:
https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=9510&view=logs&j=576a9cda-93b0-5727-f1c8-22833b424115&t=dff66e23-3473-587f-f9c1-9e625e695534

Issue #1272, opened 2023-08-29
The maximum playback duration has been reached


denybear:
### FluidSynth version
FluidSynth executable version 2.3.3
Sample type=double

### Describe the bug
Playing a midi song, then stopping it, then playing it again cause the following message: **fluidsynth: error: The maximum playback duration has been reached. Terminating player!**

### Expected behavior
message should not display, and song should be played again

### Steps to reproduce
Please refer to code below

### Additional context
code snippet

```
/* includes */
#include <stdio.h>
#include <errno.h>
#include <stdlib.h>
#include <string.h>
#include <math.h>
#include <signal.h>
#include <dirent.h>
#include <time.h>
#ifndef WIN32
#include <unistd.h>
#endif

#include <fluidsynth.h>


int main ( int argc, char *argv[] )
{
// FLUIDSYNTH player
fluid_player_t* player;
fluid_settings_t* settings;
fluid_synth_t* synth;
fluid_audio_driver_t* adriver;
int sf2_id;		// id of sf2 file currently loaded

// init fluidsynth
settings = new_fluid_settings();
synth = new_fluid_synth(settings);
// jack as audio driver
// sample rate as the one defined in jack
fluid_settings_setstr(settings, "audio.driver", "jack");

// start the synthesizer thread
adriver = new_fluid_audio_driver(settings, synth);

// load sound font				
sf2_id = fluid_synth_sfload(synth, "./soundfonts/00_FluidR3_GM.sf2", TRUE);

// create new player
player = new_fluid_player(synth);

// load midi file
fluid_player_add(player, "./songs/5B_Rock1.mid");
// set endless looping of current file
fluid_player_set_loop (player, -1);

// play the midi file
fluid_player_seek (player, 0);
fluid_player_play (player);
sleep (10);

fluid_player_stop (player);
sleep (5);

// play the midi file
// error pops up here !
fluid_player_seek (player, 0);
fluid_player_play (player);

while (1);
}
```
[5B_Rock1.mid.zip](https://github.com/FluidSynth/fluidsynth/files/12467138/5B_Rock1.mid.zip)

Snoucher:
I tested fluidsynth-2.3.0-2-x86_64 dating 22-Sep-2022 10:34 on the arch linux archive, It works wonderfully. This is probably a regression.

denybear:
Hi, indeed this is likely a regression.
[...]

Issue #1270, opened 2023-08-19
Daemon is broken with MIDI & WINE


satheevrin:
I used to be able to play really ancient programs with music to them, but ever since this updated past 2.3.0 , its broken the way the Daemon functions. To repro, simply do the following...

On Arch, of course

sudo pacman -S fluidsynth lib32-fluidsynth soundfont-fluid

edit SOUND_FONT=/usr/share/soundfonts/FluidR3_GM.sf2   in /etc/conf.d/fluidsynth

create the symbolic link: sudo ln -s FluidR3_GM.sf2 /usr/share/soundfonts/default.sf2

systemctl --user enable fluidsynth.service

reboot...

Launch an old game through WINE (visual novel, space cadet). The Daemon reports that the service is running, but MIDI music is not playing. With the newest versions, every I had to do systemctl --user restart fluidsynth.service  in order to get MIDI music to work in WINE

the intended behavior is for the Daemon to work at boot in WINE and you should not have to restart the service.

Downgrading to v 2.3.0 has restored this Daemon to its intended behavior where I would go into WINE, launch an app and the MIDI music would play

derselbst:
Pls. share the logs of the fluidsynth daemon after boot and after first manual restart.

satheevrin:
Where do I get that information? I am still kind of nooby. Where is the log saved...?

satheevrin:
I figured it out. Here are the logs

The first file is fluidsynth at boot and running WINE to play a game with MIDI:

[fluidsynthboot.txt](https://github.com/FluidSynth/fluidsynth/files/12392797/fluidsynthboot.txt)

the second file is fluidsynth ater having to forcibly restart the daemon using systemctl --user restart fluidsynth.service and then receiving sound in WINE while playing the same game with MIDI music:

[fluidsynthreboot.txt](https://github.com/FluidSynth/fluidsynth/files/12392811/fluidsynthreboot.txt)

derselbst:
Ok, great. Given all the errors it surprises me that it works at all. Esp, because:

`error: Failed to find an audio format supported by alsa`

i.e. fluidsynth cannot output any sound to ALSA, and:

`warning: Failed to allocate a synthesis process`

i.e. it cannot allocate a voice. Ergo, you should get no sound at all. To investigate further, pls.

1. Post the exact commandline that the fludisynth daemon is started with (it should be in the `fluidsynth.service`, but there may be additional variables included, which are expanded from I don't know where).
[...]

PR #1273, opened 2023-09-03
Fix player termination after stopping and restarting

albedozero:
If using the sample timer, reset cur_msec=0 in fluid_player_play() since the sample timer is also reset. This way fluid_player_callback() will not misinterpret the player's cur_msec < msec as being due to a fluid_synth_get_ticks overflow.

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -2291,6 +2291,7 @@ fluid_player_play(fluid_player_t *player)
     if(!player->use_system_timer)
     {
         fluid_sample_timer_reset(player->synth, player->sample_timer);
+        player->cur_msec = 0;
     }
 
     /* If we're at the end of the playlist and there are no loops left, loop once */


albedozero:
Fixes #1272 and perhaps #1270?

albedozero:
#1236 added a check that _fluid_synth_get_ticks()_ has not overflowed by comparing the sample timer's `msec` provided to _fluid_player_callback()_ is not less than the `cur_msec` of the player. However, _fluid_player_play()_ calls _fluid_sample_timer_reset()_, which resets the `startticks` of the timer. Before, this was fine because the player callback updates `cur_msec` on its own, but now we need to reset the player's `cur_msec` as well so it's not misinterpreted as a ticks overflow.

derselbst (changes requested):
First of all, thanks for looking into this. However, I don't think this is correct. `cur_msec` is needed to process tempo change events correctly. We can't just reset it because the user has called e.g. `fluid_player_stop(); fluid_player_play();`. Yes, the `fluid_player_callback` restores `cur_msec`, but still I believe this could easily break something else that we currently don't see (we don't know because there is not a single test for the player).

I prefer a more conservative change here, which is to simply memorize the `msec` send by the synth in `fluid_player_callback` and compare against that accordingly.

#1270 doesn't use the player, so this is unrelated.

albedozero:
Sorry it's been a while for me to come back to this. I still think `cur_msec` _is_ supposed to be the thing that memorizes the last `msec` sent by the synth - that's why it's set in `fluid_player_callback()`. `cur_msec` is used by `fluid_player_update_tempo()` to set the synth-time of the tempo change, but that is based on the assumption that `fluid_player_callback()` has kept it updated. `cur_msec` isn't used anywhere else.

If we are as of #1236 also going to use `cur_msec` as a flag to determine whether the sample timer has overflowed, we have to reset it if we reset the sample timer. A more conservative solution would be to add an `unsigned int last_msec` to `fluid_player_t`, reset _that_ in `fluid_player_play()`, and use it to check for overflow in `fluid_player_callback()`. I can do that if you think it's safer.

derselbst:
> A more conservative solution would be to add an unsigned int last_msec to fluid_player_t, reset that in fluid_player_play(), and use it to check for overflow in fluid_player_callback(). I can do that if you think it's safer.

Well, if you're sure that directly manipulating `cur_msec` is fine and doesn't break any other existing use-case, tell me and you can keep it. Otherwise, use `last_msec`.

I'm thinking of use-cases like seeking in the file with and without stopping / starting the player; manipulating the tempo via an external timesource and switching back to internal, etc.

derselbst (approved):
I changed my opinion. I once again looked through the code, checked any side-effects but couldn't find one. It should work, just as you said. I'll take this now hoping to release 2.3.4 later today.

Excuse the confusion and thanks again!

albedozero:
Thanks! I was confident it would work, but I was trying to find time to write a player test to prove it. I will hopefully do this for an upcoming commit.

PR #1276, opened 2023-09-08
[RFC] Using Sustain Pedal like Soft Pedal

derselbst:
This attempt to implement the method described in  #1271.

Unfortunately, it doesn't work. In my tests I've found that manipulating the release time does not affect voices already playing in release phase.

I currently have no time for a deeper investigation.

### DIFF
--- a/src/synth/fluid_voice.c
+++ b/src/synth/fluid_voice.c
@@ -321,7 +321,7 @@ fluid_voice_init(fluid_voice_t *voice, fluid_sample_t *sample,
     voice->mod_count = 0;
     voice->start_time = start_time;
     voice->has_noteoff = 0;
-    voice->ignore_sustain = 0;
+    voice->note_off_sustain= 0;
     UPDATE_RVOICE0(fluid_rvoice_reset);
 
     /*
@@ -613,15 +613,24 @@ fluid_voice_calculate_runtime_synthesis_parameters(fluid_voice_t *voice)
         int dest_gen_index = mod->dest;
         fluid_gen_t *dest_gen = &voice->gen[dest_gen_index];
         dest_gen->mod += modval;
-        
-        // issue blablabla: If a modulator is in use, which responds to CC64 and intended to manipulate the release envelope, this voice should not be sustained,
-        // to allow the modulator to achieve its desired behavior while the sustain switch is being depressed.
-        if((mod->src1 == SUSTAIN_SWITCH && mod->flags1 & (FLUID_MOD_CC | FLUID_MOD_SWITCH))
-            || (mod->src2 == SUSTAIN_SWITCH && mod->flags2 & (FLUID_MOD_CC | FLUID_MOD_SWITCH))
-            && (mod->dest == GEN_VOLENVRELEASE))
-        {
-            voice->ignore_sustain = 1;
-        }
+      	
+    // Issue #1276: Using Sustain Pedal like Soft Pedal 
+    //
+    //    If a modulator is in use that responds
+    //    to the CC64 and intends to manipulate the release envelope,
+    //    It's voice should change the volume envelope during the sustain fase.
+    //     to allow the modulator to achieve the desired behavior 
+    //    it uses this modulator amount in new decay value and set attenuation,
+    //     creating smooth release.
+    // 
+    
+    if(mod->src1 == SUSTAIN_SWITCH &&      
+            (mod->flags1 & FLUID_MOD_CC) &&
+            mod->dest == GEN_VOLENVRELEASE)
+    {
+        voice->note_off_sustain = mod->amount;
+    }
+
         /*      fluid_dump_modulator(mod); */
     }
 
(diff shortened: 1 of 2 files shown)


derselbst:
@RobsonFBP FYI

RobsonFBP:
-----
Thanks for this PR.
I liked the name of the created variable ignore_sutain because it is more intuitive than cancel_default_hold.
-----
`        if((mod->src1 == SUSTAIN_SWITCH && mod->flags1 & (FLUID_MOD_CC | FLUID_MOD_SWITCH))
            || (mod->src2 == SUSTAIN_SWITCH && mod->flags2 & (FLUID_MOD_CC | FLUID_MOD_SWITCH))
            && (mod->dest == GEN_VOLENVRELEASE))
        {
            voice->ignore_sustain = 1;
        }`

I think if `(mod->src1 == SUSTAIN_SWITCH)` is true, `mod->flags1 & (FLUID_MOD_CC | FLUID_MOD_SWITCH)` will always be true and could be simplified to:

        if ((mod->src1 == SUSTAIN_SWITCH || mod->src2 == SUSTAIN_SWITCH) && mod->dest == GEN_VOLENVRELEASE)
        {
            voice->ignore_sustain = 1;
        }

I would be even more restrictive using both SUSTAIN_SWITCH sources


        if (mod->src1 == SUSTAIN_SWITCH && mod->src2 == SUSTAIN_SWITCH && mod->dest == GEN_VOLENVRELEASE)
        {
            voice->ignore_sustain = 1;
        }

But if you want to keep the code without simplifying, you could fix the brackets like this.

        if(((mod->src1 == SUSTAIN_SWITCH && mod->flags1 & (FLUID_MOD_CC | FLUID_MOD_SWITCH))
            || (mod->src2 == SUSTAIN_SWITCH && mod->flags2 & (FLUID_MOD_CC | FLUID_MOD_SWITCH)))
            && mod->dest == GEN_VOLENVRELEASE)
        {
            voice->ignore_sustain = 1;
        }

src/synth/fluid_voice.c
```diff
@@ -1364,7 +1374,7 @@ fluid_voice_noteoff(fluid_voice_t *voice)
         voice->status = FLUID_VOICE_HELD_BY_SOSTENUTO;
     }
     /* Or sustain a note under Sustain pedal */
-    else if(fluid_channel_sustained(channel))
+    else if(fluid_channel_sustained(channel) && !voice->ignore_sustain)
     {
         voice->status = FLUID_VOICE_SUSTAINED;
     }
```
RobsonFBP, line 1380:
```suggestion
    else if(fluid_channel_sustained(channel))
    {
        voice->status = FLUID_VOICE_SUSTAINED;        
        //  Here it changes the envelope to simulate ignore_sustain        
        if(voice->note_off_sustain)
        {
            fluid_voice_gen_set(voice, GEN_VOLENVDECAY, voice->note_off_sustain);
            fluid_voice_gen_set(voice, GEN_VOLENVSUSTAIN, 960); 
            fluid_voice_update_param(voice, GEN_VOLENVSUSTAIN); 
        }
    }
```

RobsonFBP (changes requested):
These proposed changes are to create a smooth fit when releasing the key with the sustain pedal depressed. Useful for smooth tuning on instruments that sound indefinitely while the sustain pedal is depressed.

derselbst:
Ok, I applied your suggestions, thanks. However, I'm currently unable to test this. It may take me a few weeks before coming back to this.

derselbst:
I found some time to look into this again. I can confirm that the current change results in an audible manipulation of the sustain and release phases for new voices only (i.e. not for already playing voices). The most noticeable effect is that notes which were played while the sustain pedal was depressed, keep playing even when releasing the pedal afterwards and playing further notes. I'm not quite sure if this is supposed to be the effect intended by the original discussion.

One problem I observed is that fluidsynth easily runs out of polyphone, when playing notes while the pedal is depressed:
`fluidsynth: debug: Polyphony exceeded, trying to kill a voice`

This doesn't happen when playing with the regular sustain pedal behavior and is probably due to decay and release phases being both expanded by the current implementation.

Honestly, I don't really find this behavior intuitive or comprehensive. But I would like to hear what others think about it?

RobsonFBP:
I would also like to be able to test each change, but I don't know how to do it. If you could send me a compiler with all the settings and all the necessary files and dependencies, so that I could just make the changes and compile, I could perhaps deliver the finished result.

Eu também gostaria de poder testar cada mudança, mas não sei como fazer. Se você pudesse me mandar um compilador com todas as configurações  e todos os arquivos e dependências necessários de modo que eu apenas fizesse as mudanças e compilasse, poderia entregar o resultado pronto.

derselbst:
The process of compiling is explained in our wiki:
https://github.com/FluidSynth/fluidsynth/wiki/BuildingWithCMake

RobsonFBP:
Obrigado pela atenção. Grande abraço.

Obter o Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

derselbst:
Sorry, but this PR stayed open for over a year without a working solution in mind, without activity over the past 8 months. I currently don't see this PR will reach a shape where it can be accepted.

PR #1287, opened 2023-11-28
Make settings more pleasing, natural and subtler

es20490446e:
Demos:
- [Records](https://drive.google.com/file/d/1wy9XaB9WWG4XYXSRmeV7LjC1eeCRree8/view?usp=sharing)
- [Sources](https://github.com/FluidSynth/fluidsynth/files/13485479/demos-sources.tar.gz)


derselbst (approved):
I finally had some time coming back to this. I did some listening tests myself and quite like result. Thank you!

I've also added the previous default settings to the documentation, just in case...

PR #1292, opened 2023-12-13
Update FreeBSD CI: Drop 12.4 and add 14.0

emaste:
FreeBSD 12.4 is EOL at the end of 2023, and 14.0 was released recently.


derselbst (approved):
Thanks!

PR #1309, opened 2024-03-26
Stop the pipewire loop before destroying the stream

### LINKED ISSUE #1305: The Pipewire audio output driver does not cleanly destroy when FluidSynth exits
### FluidSynth version
```
FluidSynth executable version 2.3.4
Sample type=double
```

### Describe the bug
Various PipeWire functions related to destroying the context produce numerous errors when FluidSynth exits or otherwise wants to destroy the audio context.

```
*** pw_stream_destroy called from wrong context, check thread and locking: Operation not permitted
*** impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted
*** impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted
```

Occasionally, this can result in a segfault or abort.

### Expected behavior
The PipeWire context should be cleanly removed, and not produce these error messages.

### Steps to reproduce
Launch FluidSynth in a interactable shell with the PipeWire audio driver `$ fluidsynth -a pipewire`
Type quit into the shell

### Additional context
The Mumble project ran into similiar pitfalls; more information here: https://github.com/mumble-voip/mumble/pull/6103
https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3150#note_1862593

mawe42:
Should fix #1305

### DIFF
--- a/src/drivers/fluid_pipewire.c
+++ b/src/drivers/fluid_pipewire.c
@@ -296,7 +296,7 @@ void delete_fluid_pipewire_audio_driver(fluid_audio_driver_t *p)
 
     if (drv->pw_loop)
     {
-        pw_thread_loop_stop(drv->pw_loop);
+        pw_thread_loop_lock(drv->pw_loop);
     }
 
     if(drv->pw_stream)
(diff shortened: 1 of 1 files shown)


derselbst (approved):
Thanks Marcus! I'll cherry-pick this to `release/2.3` and then merge it back to master.

PR #1314, opened 2024-04-05
Fix behavior of volume envelope delay phase

### LINKED ISSUE #1312: volume envelope delay is not to spec
FluidSynth's volume envelope delay does not behave according to the SoundFont specification. Currently, FluidSynth starts sample playback at the start of the attack phase rather than the start of the envelope generator. Section 9.1.7 of the SoundFont 2.04 specification defines the envelope generator as such:

> [...] An envelope generates a control signal in six phases. When key-on occurs, a delay period begins during which the envelope
value is zero. The envelope then rises in a convex curve to a value of one during the attack phase. [...]

Nowhere in the document is it suggested that sample playback should start at the attack phase, and as far as I have seen, only FluidSynth behaves in this manner. In every other SoundFont synth I have tried (including Sound Blaster hardware and also other sampler formats such as SFZ), the sample is actually silently playing during the volume envelope's delay phase.

Now, here is where I have to caveat that I actually *prefer* FluidSynth's method here, as I have wished for the envelope delay to actually delay the start of sample playback since I started building SoundFonts back in the late '90s. From a sound design perspective, it actually makes more sense to do it this way. How often do I want to delay a sound and lose the start of the sample at the same time? If it weren't for that darned compatibility factor, I'd argue in favor of keeping things as-is. In fact, I'm kind of hating myself right now for reporting this! :laughing:

[Here is an audio comparison](https://github.com/FluidSynth/fluidsynth/files/14805105/Church.Bells.comparison.zip) of a legacy preset that becomes problematic with FluidSynth's delay method, the Roland GS preset **008:014 Church Bells**, here demonstrated using [GeneralUser GS](https://schristiancollins.com/generaluser). The sound you hear is a single press of the note C4. [Sobanth](https://blog.rosseaux.net/page/e5ca75d98990e33b31dadc78a8df1333/Sobanth) plays this correctly, with each of the subsequent echoing bells quieter and more mellow due to the attack phase occurring later in the bell waveform. On Fluidsynth, you clearly hear each bell hit identically as each voice's attack phase is starting at the beginning of the sample.

The preset could be reprogrammed to sound more natural in FluidSynth, but that would break the playback in all other SoundFont synths. And this is the reason why—as much as it pains me to say it—we should consider changing the delay envelope behavior to match the spec.

Thoughts?

### FluidSynth version
Any, as far as I know.

derselbst:
Previously, fluidsynth used an optimization that allowed silent voices to be completely skipped rendering. While this behavior might make sense from a performance point of view, it does break the expected behavior of the delay phase demanded by the SF2, see #1312 for details. Essentially, the delay phase shall not postpone the sample, rather the sample should be played "silently". And that's what this PR does. However, as the rendered samples will always become zero, the nearest neighbor interpolation is used as a minor optimization in this case.

Fixes #1312.

### DIFF
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -355,10 +355,6 @@ fluid_rvoice_write(fluid_rvoice_t *voice, fluid_real_t *dsp_buf)
 
     count = fluid_rvoice_calc_amp(voice);
 
-    if(count <= 0)
-    {
-        return count; /* return -1 if voice is quiet, 0 if voice has finished */
-    }
 
     /******************* phase **********************/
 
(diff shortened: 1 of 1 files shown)


derselbst:
@mrbumpy409 @mawe42 FYI

PR #1316, opened 2024-04-20
add a new windows workflow based on msys2

pedrolcl:


derselbst (approved):
Looking good, thanks!

Issue #1322, opened 2024-04-27
Wasapi driver outputs device names encoded in Windows ANSI default codepage


pedrolcl:
### FluidSynth version
```
$ ./fluidsynth.exe --version
FluidSynth runtime version 2.3.5
Copyright (C) 2000-2024 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.3.5
Sample type=double
```
### Describe the bug
This comes from a Qsynth ticket:  **_Device initialization fails when audio driver is set to WASAPI and output destination is other than default_**. https://github.com/rncbc/qsynth/issues/82

The reporter includes a screenshot where the selection of audio devices appears corrupted, due to inconsistent character decoding.
When the users tries to select any audio device other than "default", the initialization of the audio driver fails with an error message:

`Qsynth1: Failed to create the audio driver (wasapi). Cannot continue without it.`

### Expected behavior

The list of available audio devices is correctly presented to the user, and any device can be selected and initialized, if all other parameters are supported.

### Steps to reproduce
The original reporter has a Japanese Windows 11 version, using codepage 932. I've been able to reproduce the issue renaming in my system (which uses codepage 1252) the audio devices to include some tricky characters.

### Additional context

All other Fluidsynth MIDI and audio drivers report the audio devices as settings options encoded as UTF-8 except the Wasapi driver, that uses the default windows codepage.

```
$ grep "WideCharToMultiByte" src/*/*
src/drivers/fluid_wasapi.c:        nsz = WideCharToMultiByte(CP_ACP, 0, var.pwszVal, -1, 0, 0, 0, 0);
src/drivers/fluid_wasapi.c:        WideCharToMultiByte(CP_ACP, 0, var.pwszVal, -1, name, nsz, 0, 0);
src/drivers/fluid_wasapi.c:    nsz = WideCharToMultiByte(CP_ACP, 0, var.pwszVal, -1, 0, 0, 0, 0);
src/drivers/fluid_wasapi.c:    WideCharToMultiByte(CP_ACP, 0, var.pwszVal, -1, name, nsz, 0, 0);
src/drivers/fluid_waveout.c:            WideCharToMultiByte(CP_UTF8, 0, caps.szPname, -1, dev_name, MAXPNAMELEN, 0, 0);
[...]

Issue #983, opened 2021-09-25
fluidsynth for windows failed to load soundfont path containing non-ASCII chars


tsingakbar:
### FluidSynth 2.2.3
os: win10
locale(ANSI encoding): CP936(default for widows for china since windows xp?)

### Describe the bug
```cmd
d:\opt\fluidsynth-2.2.3\bin>fluidsynfluidsynth.exe d:\冷备\音乐制作\软件\soundfont2\GeneralUser-GS-Live-Audigy-v1.43.sf2
FluidSynth runtime version 2.2.3
Copyright (C) 2000-2021 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

fluidsynth: error: fluid_is_soundfont(): fopen() failed: 'File does not exist.'
Parameter 'd:\冷备\音乐制作\软件\soundfont2\GeneralUser-GS-Live-Audigy-v1.43.sf2' not a SoundFont or MIDI file or error occurred identifying it.
```
But that file indeed exists.

### Expected behavior
Soundfont file should be loaded.

### Additional context
seems related to https://github.com/FluidSynth/fluidsynth/issues/880

tsingakbar:
## another encoding problem: wasapi enumeration can not display properly

### FluidSynth 2.2.3
os: win10
locale(ANSI encoding): CP936(default for widows for china since windows xp?)

### Describe the bug
```cmd
d:\opt\fluidsynth-2.2.3\bin>fluidsynth.exe -Q
FluidSynth runtime version 2.2.3
Copyright (C) 2000-2021 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

Available audio devices:
 default
 杩滅▼闊抽

Shared mode:
        default
          8000Hz, 16bits ................................ OK(I)
(omit....)
        杩滅▼闊抽
          8000Hz, 16bits ................................ OK(I)
(omit....)
```

```
> info audio.waveout.device
audio.waveout.device:
Type:          string
Value:         default
Default value: default
Options:       default, 远程音频
Real-time:     no
> info audio.dsound.device
audio.dsound.device:
Type:          string
Value:         default
Default value: default
Options:       default, 远程音频, 主声音驱动程序
Real-time:     no
> info audio.wasapi.device
audio.wasapi.device:
Type:          string
Value:         default
Default value: default
[...]

PR #1325, opened 2024-04-29
Fixed  audio and MIDI drivers; encode device names as CP_UTF8

pedrolcl:
This also sets unicode (utf8) for CLI program's console output. It should be the proper fix for #983 wasapi problems.
cc: @tsingakbar 

Solves #1322


derselbst:
Thanks Pedro! @tsingakbar If you have some time, could you pls. run `fluidsynth.exe -Q` and let us know if the reported WASAPI device names still look correct? For convenience, you may use the `fluidsynth-x64` prebuilt binaries: https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=10122&view=artifacts&pathAsName=false&type=publishedArtifacts

pedrolcl:
CC  @nanitaro , please test as well

nanitaro:
> CC @nanitaro , please test as well

I ran fluidsynth.exe -Q on a test version (fluidsynth-x64).
Japanese information was output without problems as in 2.3.5 (official release version).
All variants (x64,mingw-x64,x86) have been checked.
I will upload the output log just to be sure.
[fluidsynth.txt](https://github.com/FluidSynth/fluidsynth/files/15163562/fluidsynth.txt)

src/drivers/fluid_dsound.c
```diff
@@ -137,6 +155,7 @@ fluid_dsound_enum_callback2(LPGUID guid, LPCTSTR description, LPCTSTR module, LP
             return FALSE;
```
derselbst, line 155: `name` will be leaked when returning false here.

include/fluidsynth/settings.h
```diff
@@ -48,6 +48,9 @@ extern "C" {
  *       // ...
  *     }
  * @endcode
+ * All string settings are encoded in UTF-8. This includes the names
+ * of the audio and MIDI devices, exposed as setting options.
```
derselbst, line 52: Probably it's also worth mentioning in the `fluidsettings.xml`, that starting with 2.3.6 all device names are expected to be UTF8 encoded.

src/drivers/fluid_dsound.c
```diff
@@ -107,8 +107,18 @@ typedef struct
 BOOL CALLBACK
 fluid_dsound_enum_callback(LPGUID guid, LPCTSTR description, LPCTSTR module, LPVOID context)
 {
-    fluid_settings_t *settings = (fluid_settings_t *) context;
-    fluid_settings_add_option(settings, "audio.dsound.device", (const char *)description);
+    fluid_settings_t *settings = (fluid_settings_t *)context;
+    char *name;
+#if _UNICODE
+    int nsz = WideCharToMultiByte(CP_UTF8, 0, description, -1, 0, 0, 0, 0);
+    name = FLUID_ARRAY(char, nsz + 1);
```
derselbst, line 114: Is there a reason for allocating +1 byte? From my understanding, when `cchWideChar` is set to -1, the value returned by `WideCharToMultiByte` will always include the NULL byte. So technically, the very last byte of  `name` would be uninitialized. Not sure, if that's intended?

derselbst (commented):
Nice, thank you! A few minor comments.

src/drivers/fluid_dsound.c
pedrolcl: fixed

include/fluidsynth/settings.h
pedrolcl: done

src/drivers/fluid_dsound.c
pedrolcl: This was the result of copy/paste. But you are right: that is what the documentation states, so there should be no need for this precaution.

All drivers fixed.

pedrolcl:
With these changes, the users may enjoy utf8 output and input when using cmd.exe (or other shells) in Windows. The input is relevant for Fluidsynth's own shell.  And this would be enough for libfluidsynth.dll clients, like Qsynth.

But there is still a loose thread: the CLI program's command line arguments are only processed/converted into utf8 for file names, but not other arguments. For instance, this would not work:

C:> fluidsynth.exe -o audio.wasapi.device=GroßeStraßenlautsprecher

derselbst:
Sorry for the late response, busy schedule.

It looks good to me now, thanks! If you want and have some time feel free to also UTF8 convert the commandline args.

@nanitaro Would you mind running `fluidsynth.exe -Q` again with these binaries? @tsingakbar CC
https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=10135&view=artifacts&pathAsName=false&type=publishedArtifacts

If everything goes well, I would apply the commits onto the `release/2.3` and then merge it back to master. But probably not before next weekend, though.

nanitaro:
@derselbst
I ran the new test binary.
I think there is no problem about outputting Japanese information.
[fluidsynth_05112024.txt](https://github.com/FluidSynth/fluidsynth/files/15279719/fluidsynth_05112024.txt)

Compared to the previous binary, the output result of the default device in shared and exclusive mode seems to have changed.
This time the result seems to be more accurate.

For your information, the performance and settings of the sound devices in the test environment are as follows
Default device = SPDIF interface (FX-AUDIO-DAC-X6)
Windows sound properties. Default format=2ch,24bit,96000Hz

pedrolcl:
After the commits on May 1st, not only WASAPI but also the other Windows drivers are now updated to use UTF8. There is no equivalent of `-Q` for the other drivers, but it should be possible to use the shell, starting fluidsynth.exe without the argument and...

```
> settings
[...]
> info midi.driver
Options: winmidi
> info midi.winmidi.device
[...]
> info audio.driver
Options: dsound, waveout, wasapi
[...]
>  info audio.dsound.device
>  info audio.waveout.device
>  info audio.wasapi.device
```

The output of all these commands should be encoded now in UTF8.

derselbst (approved):
I have just applied the commit to the `release/2.3` branch and merged it to master. Thanks @pedrolcl  for the fix and @nanitaro for the tests!

Issue #1331, opened 2024-06-04
Chorus malfunctioning for single precision builds


RoqueDeicide:
### FluidSynth version
2.3.5 x64 and x86 Windows builds. Most likely affects previous versions as well.

### Describe the bug
An issue had been bugging me for about half a year: playing MIDI tracks using FluidSynth would always result in a very left-heavy audio. I initially thought that my soundfont wasn't very well balanced, until I encountered a track with a channel that had center pan and a program which I had just centered. Realizing something was wrong with FluidSynth, I first turned off chorus, and audio became centered! Thinking that "synth.chorus.nr" being set to an odd number resulted in an extra output on the left side, I set it to 4 from default 3. The audio became centered! Then I switched back to 3, and audio went back to the left, then I switched it to 2... and audio stayed on the left side! After this I tested all "nr" settings from 1 to 20. Here is how they affected the audio balancing:

 * 1-3 and 9-10: audio stayed on the left.
 * 6 and 14 : audio stayed on the right.
 * 4-5, 7-8, 11-12, 15-16 and 19: audio stayed centered.
 * 13, 20: audio started on the right, but moved to the center after a few seconds.
 * 17: audio started on the left, but moved to the center after a few seconds.
 * 18: audio started centered, but moved to the left after a few seconds.

I've looked at the code, to see what's wrong, but I couldn't find anything wrong with it. So I cloned the repository and compiled the code myself to try to debug it. Neither debug nor release builds, that I've made however, were affected by the above issue! The only potential relevant difference between my builds and official ones, is that mine definitely don't use OpenMP. I don't know anything about the latter, so I cannot investigate further.

### Expected behavior

There supposed to a stereophonic effect, where chorus output is moving from side to side. Prior to using my builds of FluidSynth I didn't even know what that sounded like.

### Steps to reproduce

Play any MIDI track with at least some center panned channels, using official 2.3.5 x86 or x64 Windows builds. The chorus settings are defaults. The soundfont should have all of its programs use at lease 25% chorus. This can be achieved by selecting all presets in Polyphone editor, selecting a chorus row and typing 25, and saving the soundfont afterward.

RoqueDeicide:
I forgot to mention, I've built my FluidSynth using MSVC in Visual Studio 2022.

derselbst:
Thanks for the report. I have reproduced this and can confirm that something isn't correct. It doesn't seem to be related to OpenMP though:

[...]

PR #1339, opened 2024-06-20
Fix chorus when compiling with single precision

derselbst:
The default chorus parameters yield an oscillator phase close to PI/2. When compiling with single precision, the phase becomes exactly PI/2, which causes the sinusoidal oscillation to get stuck, causing attenuated signals on the left or right channels, see #1331.

This only avoids a phase of PI/2 for the default values. I.e. it doesn't fix the chorus behavior when the phase really evaluates equal to PI/2.

### DIFF
--- a/src/rvoice/fluid_chorus.c
+++ b/src/rvoice/fluid_chorus.c
@@ -169,7 +169,7 @@
 /* modulator */
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


jjceresa:
@derselbst

Hello Tom, I'm afraid this change won't be enough. For detail, please read my comment in issue #1331.

Issue #1342, opened 2024-06-26
AWE32 NRPN Compatibility Layer


derselbst:
### Related discussion
As discussed in #1290, a compatibility layer for supporting AWE32-style NRPNs shall be attempted.

### Is your feature request related to a problem?
Mimic the behavior of AWE32 Midis in fluidsynth.

### Describe the solution you'd like
Fluidsynth being able to understand AWE32 NRPNs and produce similar output.

### Describe alternatives you've considered
There are projects that emulate the Emu8000 chip that should be used for more accurate reproduction, found e.g. [here](https://git.javispedro.com/cgit/vmusic.git/tree/emu8k.c?id=44a797ceaad46fcc9d9faf80efb4f8ad74f611b6).

### Additional context

Implementation details of AWE32 NRPNs:
* [SB AWE32 Developer's Information Pack](https://github.com/user-attachments/files/15757220/adip301.pdf)
* [Frequently Asked Questions for SB AWE32](http://archive.gamedev.net/archive/reference/articles/article445.html)

Example MIDIs: https://ftp.lysator.liu.se/pub/awe32/songs/

Example MIDI Renderings on the ~AWE32~ SB Live! using GeneralUser GS v1.471:

[NRPN MIDIs Github.7z.001.zip](https://github.com/user-attachments/files/15991421/NRPN.MIDIs.Github.7z.001.zip)
[NRPN MIDIs Github.7z.002.zip](https://github.com/user-attachments/files/15991422/NRPN.MIDIs.Github.7z.002.zip)
[NRPN MIDIs Github.7z.003.zip](https://github.com/user-attachments/files/15991424/NRPN.MIDIs.Github.7z.003.zip)

(The `.zip` extension is for Github to accept these files - remove it - it's a multi-part 7z archive)

klerg:
Nice to see this and I know it is confirmed as an issue that will be resolved at some point. One thing to note, the example renderings were on a SB Live! and not an AWE32.

derselbst:
Actually, this was quite interesting. Here is what I have so far

[awe32-nrpn_.zip](https://github.com/user-attachments/files/16040838/awe32-nrpn_.zip)

There are a few implementation details that I'm not sure about:

#### 1. Where and when to apply the AWE32 NRPNs?

[...]

PR #1345, opened 2024-06-29
Remove interpolation of IIR filter coefficients

derselbst:
During implementation of #1342, I found the IIR filter producing strange cracks and clicks. I haven't done a deep investigation but it seems that this was caused by big and frequent changes of the cutoff frequency by `fres_mod`.

I found that the clicks were produced by the code that interpolates the filter coefficients. I.e. setting `if(dsp_filter_coeff_incr_count > 0)` to `if(FALSE)` didn't cause any more clicks.

Without a deeper investigation of the numeric logic I don't know how to fix this. Yet, I don't think this justifies the more complex logic, and therefore I'm suggesting to remove that interpolation logic.

Attached is a ZIP with two audio samples. DF2.ogg (Direct Form 2) is the filter as implemented before this PR. DF2-nointerp.ogg is the same filter without any interpolation.

[filter-coeff-interp.zip](https://github.com/user-attachments/files/16039681/filter-coeff-interp.zip)

### DIFF
--- a/src/rvoice/fluid_iir_filter.c
+++ b/src/rvoice/fluid_iir_filter.c
@@ -65,7 +65,6 @@ fluid_iir_filter_apply(fluid_iir_filter_t *iir_filter,
         fluid_real_t dsp_a2 = iir_filter->a2;
         fluid_real_t dsp_b02 = iir_filter->b02;
         fluid_real_t dsp_b1 = iir_filter->b1;
-        int dsp_filter_coeff_incr_count = iir_filter->filter_coeff_incr_count;
 
         fluid_real_t dsp_centernode;
         int dsp_i;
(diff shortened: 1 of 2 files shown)


klerg:
Very good to see you are already trying to put this into Fluidsynth. I'm not sure what IIR means or what that does to the filter. And I know even less on what fres_mod changes to the cutoff frequency but I'm sure you got this down to a tee.

As you can tell, at this point I'm pretty much lost on what all this means but glad you were able to find the issue and stop anymore clicks and pops to the sound as well. 

Wow this seems like it will not be easy to find what is behind the problem at all then.  Sure, on stuff like this it will be your call to do what you assume is the best way forward. So you had to remove it. 

I just listened to both samples and find it hard to hear that much of difference, in other words, to me each sample has clicks and pops, but the second has barely any less, why is that ?

derselbst:
You need to listen till the end. The first clicks appear in DF2.ogg after 22 sec.

klerg:
Yes I'm able to hear the clicks in DF2.ogg but can also get clicks in DF2-nointerp.ogg as well

PR #1358, opened 2024-07-24
Fix LSB being ignored in pitch bend range

### LINKED ISSUE #1357: FluidSynth ignores LSB in Pitch Bend Range
### FluidSynth version

```
FluidSynth runtime version 2.3.5

FluidSynth executable version 2.3.5
Sample type=double
```

### Describe the bug

Fluidsynth only allows to specify the pitch bend range in semitones, ignoring the cent tuning via data entry.
[Test MIDI file](https://github.com/user-attachments/files/16366815/test-rpn-00-00-pitch-bend-range.zip) The second note in this file doesn't change its pitch as the bend range is set to 50 cents.

### Expected behavior

The second note should have a noticeable pitch change when the range is set to 50 cents.

### Steps to reproduce

1. Get the file
2. `fluidsynth microsoft_gm.sf2 test-rpn-00-00-pitch-bend-range.mid `
3. No audible pitch change on the second note

### Additional context
Soundfont spec doesn't say anything about the way pitch bend range should be stored so simply doing
```c
float pitchBendRange = semitones + (cents / 100);
```
should easily fix the problem without further modifications.

spessasus:
fixes #1357

also my IDE added more tabs... sorry!

### DIFF
--- a/src/synth/fluid_chan.h
+++ b/src/synth/fluid_chan.h
@@ -105,7 +105,7 @@ struct _fluid_channel_t
     enum fluid_interp interp_method;                    /**< Interpolation method (enum fluid_interp) */
 
     unsigned char channel_pressure;                 /**< MIDI channel pressure from [0;127] */
-    unsigned char pitch_wheel_sensitivity;          /**< Current pitch wheel sensitivity */
+    float pitch_wheel_sensitivity;          /**< Current pitch wheel sensitivity */
     short pitch_bend;                      /**< Current pitch bend value */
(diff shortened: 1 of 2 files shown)


spessasus:
Also I had to merge the logic for data entry lsb and msb since in the test file I provided MSB came first, so the previous logic wouldn't work. Now the data entry should work regardless of the order.

derselbst (approved):
Looking good, thanks!

>  I had to merge the logic for data entry lsb and msb

Just checked that, it seems to be in line with what the SF spec says. Not sure why it was implemented differently before.

PR #1377, opened 2024-09-08
Fix modLfoToVol behavior

### LINKED ISSUE #1374: ModLfoToVolume incorrect - the volume doesn't rise
### FluidSynth version
```

Sample type=double
```

### Describe the bug
A bit of backstory:

I've been working on making my own synth work perfectly with spec, and I've found [this blogpost](https://schristiancollins.wordpress.com/2016/03/02/using-soundfonts-in-2016/) by Christian Collins. He also linked to his [test files](https://www.dropbox.com/s/qmf0ej91h2ka5xw/SoundFont_2.01_Implementation_Test_v2.01.zip?dl=1) for testing the synths. And he also provided a rendered audio using a hardware Audigy 2 ZS.

I've ran the test in fluidsynth and it worked great, except for one part: modLfoToVolume:
image
As you can see, fluidsynth's audio is strange in this section, compared to Audigy.

From the soundfont spec:
> modLfoToVolume
> This is the degree, in centibels, to which a full scale excursion of the Modulation
LFO will influence volume. A positive number indicates a positive LFO excursion
increases volume; a negative number indicates a positive excursion decreases
volume. Volume is always modified logarithmically, that is the deviation is in
decibels rather than in linear amplitude. For example, a value of 100 indicates that
the volume will first rise ten dB, then fall ten dB.

So it should rise 10 dB, then fall 10 dB. Fluid only does the latter (possibly the initialAttenuation clamping to 0?)

I've managed to replicate fluidsynth's behavior in my code and it seems like:
image
- Red is the triangle wave that should be used
- Blue is the wave fluidsynth uses for some reason

As you can see, it ignores the negative values which should result in volume increase, like Audigy does.

### Expected behavior
Fluidsynth should rause the volume when the modLfoToVol generator is used.

### Steps to reproduce
1. get the [test files](https://www.dropbox.com/s/qmf0ej91h2ka5xw/SoundFont_2.01_Implementation_Test_v2.01.zip?dl=1)
2. `fluidsynth testv2_01.sf2 testv2_01.mid -F fluid_test.wav -o synth.gain=1`
3. compare the two

### Additional context
I think this might have to do something with initialAttenuation being capped at 0 cB, thus ignoring negative values.

Also, modLfoToPitch is described the same (rises, then falls) and it works fine.

derselbst:
Previously, `modLfoToVol` never caused the volume to increase, because amplitude calculation is performed in `fluid_cb2amp()` which previously forbids negative values. This PR changes this behavior and permits negative values. I believe this is a rare case, hence I'm calculating the values on the fly rather than using lookup tables as we usually do in such cases

Fixes #1374.

### DIFF
--- a/src/utils/fluid_conv.c
+++ b/src/utils/fluid_conv.c
@@ -111,9 +111,20 @@ fluid_cb2amp(fluid_real_t cb)
      */
 
     /* minimum attenuation: 0 dB */
-    if(cb < 0)
+    if(FLUID_UNLIKELY(cb < 0))
     {
-        return 1.0;
+        /* Issue #1374: it seems that by using modLfoToVolEnv, the attenuation can become negative and
+         * therefore the signal needs to be amplified.
+         * In such a rare case, calculate the attenuation on the fly.
+         *
+         * This behavior is backed by the spec saying:
+         * modLfoToVolume: "A positive number indicates a positive LFO excursion increases volume;
+         * a negative number indicates a positive excursion decreases volume.
+         * [...] For example, a value of 100 indicates that the volume will first rise ten dB, then fall ten dB."
+         *
+         * And in order to rise, a negative attenuation must be permitted.
+         */
+        return FLUID_POW(10.0f, cb / -200.0f);
     }
 
     if(cb >= FLUID_CB_AMP_SIZE)


derselbst:
@mrbumpy409 FYI. Seems like [test case column nr. 5](https://docs.google.com/spreadsheets/d/1QzqA9IetT7MVDhG-oSFAWrciyg6O-oaQ_hGsl8fKeSo/edit?gid=2011028303#gid=2011028303) should have been red previously ;)

mrbumpy409:
> @mrbumpy409 FYI. Seems like [test case column nr. 5](https://docs.google.com/spreadsheets/d/1QzqA9IetT7MVDhG-oSFAWrciyg6O-oaQ_hGsl8fKeSo/edit?gid=2011028303#gid=2011028303) should have been red previously ;)

Good point, haha! At the time, I saw the same volume clamping with Audigy 2 when the track volume was set to 127, so I figured FluidSynth was emulating that behavior.

With CC7 = 127 and velocity = 127, the modLfoToVol test on Audigy 2 ZS looks like this (note, my newer, yet-to-be-published version of this test uses a sine instead of saw wave):
image

Anyway, I actually understand the SoundFont 0dB voice volume ceiling now, and future versions of my test will properly account for that.

PR #1381, opened 2024-09-17
Fix a regression introduced by #1314

derselbst:
This fixes a regression introduced in #1314, most notable in test case 1 of [Christian's new test suite](https://github.com/mrbumpy409/SoundFont-Spec-Test), but also present in others.

Broken behavior introduced by #1314: The short saw wave voice finished processing, but this was not properly attributed and therefore remained in the dsp loop, causing audible sound until the noteoff was received.

Screenshot_20240917_121558

Correct behavior before #1314 and with this fix: The saw wave voice is now removed from the dsp loop once it has finished. Voices which are in delay phase (#1312) are still processed to have their dsp phase updated correctly.

Screenshot_20240917_121606

### DIFF
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -354,6 +354,12 @@ fluid_rvoice_write(fluid_rvoice_t *voice, fluid_real_t *dsp_buf)
     /******************* amplitude **********************/
 
     count = fluid_rvoice_calc_amp(voice);
+    if(count == 0)
+    {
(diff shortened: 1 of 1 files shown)


derselbst:
There is a short saw wave audible for noteon 58. It decays quickly and reaches zero gain. Before #1314, the behavior was to turn off that voice. After #1314, it remains active, as it was assumed to be silently active. Need to figure out why its volume increases a bit shortly after it reached zero gain.

PR #1399, opened 2024-10-08
Fix Default Pan Modulator Amount

### LINKED ISSUE #1395: Fluidsynth's pan modulator amount is not correct
### FluidSynth version
2.3.6

### Describe the bug
`Sfspec24.pdf` section 8.4.6:
> 8.4.6 MIDI Continuous Controller 10 to Pan Position
Source Enumeration = 0x028A (type=0, P=1, D=0, CC=1, index = 10)
Destination Enumeration = Initial Attenuation
Amount = 1000 tenths of a percent
Amount Source Enumeration = 0x0 (No controller)
Transform Enumeration = 0 (Linear)

Yet
https://github.com/FluidSynth/fluidsynth/blob/66adce6ad9772dd7285370d3f78b6824b907c0e8/src/synth/fluid_synth.c#L396-L409

Spec says 1000, so why is it set to 500? 

Amount of 1000 does not mean that the full range of the modulator is 1000, with 0 being -500, 64=0 and 127=500. 
The range is twice that because of bipolar transform: 0=-1000 64=0 127=1000, like with the pitch modulator.

With the current behavior, you can clearly notice this:
- [stereo_square.zip](https://github.com/user-attachments/files/17267418/stereo_square.zip)
- In polyphone, panning both stereo and mono squares hard left results in right channel being silent.
- In fluidsynth, that's not the case. `Stereo Square` sounds like it's a bit on the left but not fully. `Mono Square` fully goes to the left.

This is because, amount of 1000 always forces the sample to be on the desired channel. That is:
- hard left for the right sample is: `500 + (-1000) = -500` which is left.
- hard right for the left sample is `-500 + 1000 = 500` which is right.

Where with amount being 500 like fluid sets:
- hard left for the right sample is: `500 + (-500) = 0` which is the middle. That is not what we want.
- hard right for the left sample is `-500 + 500 = 0` which is the middle. That is not what we want.

I think this is why creative chose 1000 and not 500.

### Expected behavior
It should be 1000 and both of the attached soundfont's presets should pan in the same way.

### Steps to reproduce
1. boot up fluid
2. run it with the attached soundfont
3. select stereo square
4. set CC10 to 0 or 127
5. play a few notes
6. repeat 3-5 with mono square
7. hear the difference

### Additional context
None

spessasus:
This PR fixes the incorrect amount in the default modulator for pan in fluidsynth.

Resolves #1395

### DIFF
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -422,6 +422,12 @@ fluid_rvoice_write(fluid_rvoice_t *voice, fluid_real_t *dsp_buf)
         voice->dsp.phase_incr = 1;
     }
 
+    /* loop mode release? if not in release, the voice is silent */
+    if( voice->dsp.samplemode == FLUID_START_ON_RELEASE && fluid_adsr_env_get_section(&voice->envlfo.volenv) < FLUID_VOICE_ENVRELEASE)
+    {
+        return -1;
+    }
+
     /* voice is currently looping? */
     is_looping = voice->dsp.samplemode == FLUID_LOOP_DURING_RELEASE
                  || (voice->dsp.samplemode == FLUID_LOOP_UNTIL_RELEASE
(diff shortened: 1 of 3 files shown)


spessasus:
Oh sorry, I pushed another fix to the wrong branch. Closing this one, I'll make a new one. Sorry!

PR #1400, opened 2024-10-08
Add support for loop mode release

### LINKED ISSUE #1398: Add support for sample mode release
_A clear and concise description of what the problem is._
Polyphone 2.4 added a new sample mode:

image

_A clear and concise description of what you want to happen._
Sample mode 2: release

_A clear and concise description of any alternative solutions or features you've considered._
None

### Additional context
_Add any other context or screenshots about the feature request here._
In the discussion.

spessasus:
Adds support for the new "sample mode release" introduced by Polyphone 2.4

Resolves #1398

### DIFF
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -423,7 +423,7 @@ fluid_rvoice_write(fluid_rvoice_t *voice, fluid_real_t *dsp_buf)
     }
 
     /* loop mode release? if not in release, the voice is silent */
-    if( voice->dsp.samplemode == FLUID_START_ON_RELEASE && fluid_adsr_env_get_section(&voice->envlfo.volenv) < FLUID_VOICE_ENVRELEASE)
+    if(voice->dsp.samplemode == FLUID_START_ON_RELEASE && fluid_adsr_env_get_section(&voice->envlfo.volenv) < FLUID_VOICE_ENVRELEASE)
     {
         return -1;
     }
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -422,7 +422,9 @@ fluid_rvoice_write(fluid_rvoice_t *voice, fluid_real_t *dsp_buf)
         voice->dsp.phase_incr = 1;
     }
 
-    /* loop mode release? if not in release, the voice is silent */
+    /* loop mode release? if not in release, the voice is silent
+     * note: this intentionally processes the volenv before returning silence,
+     * since that's what polyphone does (PR #1400) */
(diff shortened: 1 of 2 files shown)


derselbst (changes requested):
Disclaimer: This is only my personal static code analysis - I haven't tested this yet.

IMO, the change is inaccurate because for two reasons:

* You're not implementing the 0.02s attack phase.
* Polyphone claims that the "playback begins with the release phase". But you implemented it in such a way that the voice plays back silently until the release phase begins (because of the `dsp `updates happening just before that). This behavior would be pretty equivalent to the delay phase, but I don't think this is the intended behavior by Polyphone.

spessasus:
> Polyphone claims that the "playback begins with the release phase". But you implemented it in such a way that the voice plays back silently until the release phase begins (because of the `dsp `updates happening just before that). This behavior would be pretty equivalent to the delay phase, but I don't think this is the intended behavior by Polyphone.

And that's exactly what this patch does. It only delays the wavetable oscillator. The volenv works normally. Feel free to test in it. Put an attack time of 1 and hold the key for 1 second then release, then try instantly releasing it after the keypress. You can hear that the attack phase advances while the key is pressed.

derselbst (commented):
Ok, so I've created a soundfont and it does indeed sound just like in Polyphone because I overlooked that the actual dsp update will happen in the interpolation functions below. I'm providing the test soundfont I've used: 
[polyphone-release-test.zip](https://github.com/user-attachments/files/17449574/polyphone-release-test.zip)

```
noteon 0 23 127
noteoff 0 23
```

Will start playing the at its beginning sample with no attack and no hold.

```
noteon 0 24 127
noteoff 0 24
```

Will use 1s attack and 1s hold. When executed quickly consecutively, one can hear the attack phase. I find that behavior a bit strange. I mean we are already in  release phase, so why do we hear attack and hold?

Yet, I would buy it. Pls. remove the extra space in `if( voice....)`

spessasus:
> Pls. remove the extra space in `if( voice....)`

Done!

> Will use 1s attack and 1s hold. When executed quickly consecutively, one can hear the attack phase. I find that behavior a bit strange. I mean we are already in release phase, so why do we hear attack and hold?

It makes sense IMO. Loop modes only affect wavetable, never envelope. And for example if we want an instrument with long attack time and then some sort of release noise at the end, this approach will make the noise play at the same volume as the main sample if the voice is released during the attack phase.

derselbst (approved):
Ok, LGTM. Thanks!

davy7125:
Thank you so much for implementing this into FluidSynth!
I'm using it regularly for playing organ (with jOrgan) and having now the possibility to use custom releases will make the instrument more realistic.

PR #1409, opened 2024-10-18
Add support for absolute transform

### LINKED ISSUE #1391: Support SF2.04 absolute value modulators
### Related discussion
--

Fluidsynth doesn't support absolute value modulator transforms. In fact, it actively disables them during loading:
https://github.com/FluidSynth/fluidsynth/blob/ce97c6d43715d550d6a3f1cc8d952f68b57e3e68/src/sfloader/fluid_defsfont.c#L1651-L1658

SF2.04 8.3:
> The output value of the multiplier is to be the absolute value of the input value, as
defined by the relationship:

--

### Additional context
--

Please upvote this issue if you're lacking this functionality.

spessasus:
Fixes #1391

This PR adds support for absolute value modulator transform.

Also fixes duplicated "to ensure" in a comment.

### DIFF
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -1652,11 +1652,11 @@ fluid_zone_mod_import_sfont(char *zone_name, fluid_mod_t **mod, SFZone *sfzone)
          * *** Transform Type ***
          * Only 2 types of transform are defined in the sf2 specification.
          */
-        if(mod_src->trans != FLUID_MOD_LINEAR_TRANSFORM && mod_src->trans != FLUID_MOD_ABS_VALUE)
+        if(mod_src->trans != FLUID_MOD_TRANSFORM_LINEAR && mod_src->trans != FLUID_MOD_TRANSFORM_ABS)
         {
             /* disable the modulator as the transform is invalid */
             mod_dest->amount = 0;
-            mod_dest->trans = FLUID_MOD_LINEAR_TRANSFORM;
+            mod_dest->trans = FLUID_MOD_TRANSFORM_LINEAR;
         }
         else
         {
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -1648,8 +1648,20 @@ fluid_zone_mod_import_sfont(char *zone_name, fluid_mod_t **mod, SFZone *sfzone)
             mod_dest->flags2 &= ~FLUID_MOD_BIPOLAR;
         }
 
-        /* *** Transform Type *** */
-        mod_dest->trans = mod_src->trans;
+        /**
+         * *** Transform Type ***
+         * Only 2 types of transform are defined in the sf2 specification.
+         */
+        if(mod_src->trans != FLUID_MOD_LINEAR_TRANSFORM && mod_src->trans != FLUID_MOD_ABS_VALUE)
+        {
+            /* disable the modulator as the transform is invalid */
+            mod_dest->amount = 0;
+            mod_dest->trans = FLUID_MOD_LINEAR_TRANSFORM;
+        }
+        else
+        {
+            mod_dest->trans = mod_src->trans;
+        }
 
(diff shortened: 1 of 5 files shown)


src/synth/fluid_mod.h
```diff
@@ -34,6 +34,7 @@ struct _fluid_mod_t
     unsigned char flags1;         /**< Source controller 1 flags */
     unsigned char src2;           /**< Source controller 2 */
     unsigned char flags2;         /**< Source controller 2 flags */
+    unsigned char trans;          /**< Output transform flag */
```
derselbst, line 37: For the sake of backward compatibility, this field must be initialized to zero in `new_fluid_mod()`.

src/synth/fluid_mod.c
```diff
@@ -470,8 +472,14 @@ fluid_mod_get_value(fluid_mod_t *mod, fluid_voice_t *voice)
...
     }
 
-    /* it's as simple as that: */
-    return (fluid_real_t) mod->amount * v1 * v2;
+    /* it indeed is as simple as that: */
+    final_value = (fluid_real_t) mod->amount * v1 * v2;
+    /* check for absolute value transform*/
+    if(mod->trans == 2)
+    {
+        final_value = fabs(final_value);
```
derselbst, line 480:
```suggestion
        final_value = FLUID_FABS(final_value);
```

src/synth/fluid_mod.c
```diff
@@ -470,8 +472,14 @@ fluid_mod_get_value(fluid_mod_t *mod, fluid_voice_t *voice)
         v2 = 1.0f;
     }
 
-    /* it's as simple as that: */
-    return (fluid_real_t) mod->amount * v1 * v2;
+    /* it indeed is as simple as that: */
+    final_value = (fluid_real_t) mod->amount * v1 * v2;
+    /* check for absolute value transform*/
+    if(mod->trans == 2)
```
derselbst, line 478: Once an enum exists, please use that instead of directly comparing with 2.

derselbst (commented):
Looks good, thanks. A few comments below. 

Also, a public interface for setting and getting the new transform field needs to be added as well. I.e.

```
fluid_mod_set_transform(fluid_mod_t *mod, int flags);
int fluid_mod_get_transform(const fluid_mod_t *mod);
```

Where `flags` is an enum representing the two possible values. And with API documentation.

src/sfloader/fluid_defsfont.c
```diff
@@ -1648,14 +1648,8 @@ fluid_zone_mod_import_sfont(char *zone_name, fluid_mod_t **mod, SFZone *sfzone)
...
-        /* *** Transform *** */
-        /* SF2.01 only uses the 'linear' transform (0).
-         * Deactivate the modulator by setting the amount to 0 in any other case.
-         */
-        if(mod_src->trans != 0)
-        {
-            mod_dest->amount = 0;
-        }
+        /* *** Transform Type *** */
+        mod_dest->trans = mod_src->trans;
```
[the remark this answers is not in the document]
spessasus: Sure, added.

src/synth/fluid_mod.h
spessasus: I did add `mod->trans = 0` in the function. Not sure if that's what you meant since other properties do not get the same treatment, but I didn't find any constructor for fluid_mod.

spessasus:
I've added enums and the checks you wanted. The functions are added too (with the comments), but if there's a separate .md or .xml file with the docs, I didn't find it. And lastly, check my comment regarding new_fluid_mod in your review.

src/synth/fluid_mod.h
derselbst: You change is correct. The point here is that applications calling `new_fluid_mod` need to initialize the structure manually. This is common sense. Not initializing this new field would have required applications to change their code, i.e. would have been a breaking change.

derselbst (approved):
I tweaked the docs a bit and renamed the enums to be more consistent / easier to memorize. LGTM now, thanks!

Issue #1388, opened 2024-09-27
MIDI files with non-ASCII characters in the filename fail to play on Windows.


ghost:
### FluidSynth version
fluidsynth.exe --version
FluidSynth runtime version 2.3.6
Copyright (C) 2000-2024 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.3.6
Sample type=float

### Describe the bug
On Windows, playback failures due to non-ASCII characters in filenames occur with MIDI files, but not with SoundFont files.
MIDI files that use only ASCII characters in the filename will play correctly.

When _WIN32 macro is enabled.
fluid_is_midifile() is called twice in main() of fluidsynth.c.
The first time, u8_path is set as the argument, which is argv[i] converted by win32_ansi_to_utf8().
The second time, argv[i] is set as the argument without being converted.
This difference is a bit concerning.

Thank you.

derselbst:
@stardusteyes Thanks for the report. Can you test the [fluidsynth-x64 binaries](https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=10592&view=artifacts&pathAsName=false&type=publishedArtifacts) and confirm that this fixes the problem?

stardusteyes:
It might be polite to add the following at the end of the loop.

#if defined(_WIN32)
        free(u8_buf);
#endif

- - -
I only tested the current issue.
The test binary worked as expected.
I think the problem is fixed.

Thank you.

derselbst:
Ok, thanks for confirming. I indeed missed the free(), thanks!

pedrolcl:
@derselbst , reopen request:

```
$ ./fluidsynth.exe -F test.wav ~/fluidsynth-git/test/ⓉⒺⓈⓉ.mid
fluidsynth: debug: adding audio.dsound.device=Controlador primario de sonido
fluidsynth: debug: adding audio.dsound.device=GroßeStraßenlautsprecher (HDMI) (2- High Definition Au
dio Device)
fluidsynth: debug: Testing audio device: GroßeStraßenlautsprecher (HDMI)
FluidSynth runtime version 2.4.0
Copyright (C) 2000-2024 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

[...]

PR #1411, opened 2024-10-21
Replaces the command line interface from ANSI to UNICODE on Windows.

### LINKED ISSUE #1410: Unicode command line string support on Windows.
### Related discussion
https://github.com/FluidSynth/fluidsynth/issues/880#issuecomment-835744776

When using the Fluidsynth on windows :
Now, the command line supports ANSI strings, but not non-ANSI strings.
This can sometimes cause problems in multilingual environments.

To be precise, this is not a bug, is a limitation of the ANSI command line interface.
Windows users generally only use ANSI strings, so they will have less trouble using Fluidsynth.
The maintainers seem to be very busy with more important issues.
So this update may not be a high priority.

If the command line supported UNICODE strings, the Fluidsynth users would be able to use UNICODE strings in filenames or devicenames, and then more freeing from character encoding issues.

Updating the command line interface from ANSI to UNICODE.
The libfluidsynth functions already support utf-8 strings, so there aren't many updates.

### Additional context
The code below need some changes:
* include/fluidsynth/misc.h
* src/bindings/fluid_filerenderer.c
* src/fluidsynth.c

[new_codes.zip](https://github.com/user-attachments/files/17467138/new_codes.zip)

stardusteyes:
This pull request replaces the command line interface from ANSI to UNICODE on Windows.

Closes #1410

All additional changes are between #ifdef _WIN32 and #endif.
So this only affects Windows-specific code, especially code that relies on command line strings.

***include/fluidsynth/misc.h :***
Added export of fluid_alloc().
This fix is for use the macro FLUID_ARRAY.
Without this fix, when linking import library for the libfluidsynth, an error would occur that the symbol "fluid_alloc" is not found.

***src/bindings/fluid_filerenderer.c :***
* Converts the utf-8 filename to utf-16, and calls sf_wchar_open() which is the utf-16 version of sf_open().
sf_open() cannot be used, because it converts the filename to an ANSI string.
* Exceptionally calls sf_open() if the filename is "-".
This is because sf_open() does support pipes, but not  sf_wchar_open().

***src/fluidsynth.c :***

<details>
<summary>This procedure is inspired by FFmpeg code.</summary>

```
#include <shellapi.h>
/* Will be leaked on exit */
static char** win32_argv_utf8 = NULL;
static int win32_argc = 0;

/**
 * Prepare command line arguments for executable.
 * For Windows - perform wide-char to UTF-8 conversion.
 * Input arguments should be main() function arguments.
 * @param argc_ptr Arguments number (including executable)
 * @param argv_ptr Arguments list.
 */
static void prepare_app_arguments(int *argc_ptr, char ***argv_ptr)
{
    char *argstr_flat;
    wchar_t **argv_w;
    int i, buffsize = 0, offset = 0;

    if (win32_argv_utf8) {
        *argc_ptr = win32_argc;
        *argv_ptr = win32_argv_utf8;
        return;
    }

    win32_argc = 0;
    argv_w = CommandLineToArgvW(GetCommandLineW(), &win32_argc);
    if (win32_argc <= 0 || !argv_w)
        return;

    /* determine the UTF-8 buffer size (including NULL-termination symbols) */
    for (i = 0; i < win32_argc; i++)
        buffsize += WideCharToMultiByte(CP_UTF8, 0, argv_w[i], -1,
                                        NULL, 0, NULL, NULL);

    win32_argv_utf8 = av_mallocz(sizeof(char *) * (win32_argc + 1) + buffsize);
    argstr_flat     = (char *)win32_argv_utf8 + sizeof(char *) * (win32_argc + 1);
    if (!win32_argv_utf8) {
        LocalFree(argv_w);
        return;
    }

    for (i = 0; i < win32_argc; i++) {
        win32_argv_utf8[i] = &argstr_flat[offset];
        offset += WideCharToMultiByte(CP_UTF8, 0, argv_w[i], -1,
                                      &argstr_flat[offset],
                                      buffsize - offset, NULL, NULL);
    }
    win32_argv_utf8[i] = NULL;
    LocalFree(argv_w);

    *argc_ptr = win32_argc;
    *argv_ptr = win32_argv_utf8;
}
```

https://github.com/FFmpeg/FFmpeg/blob/8d940a07d19023a98689f353e4425a14688547e9/fftools/cmdutils.c#L171C1-L223C2
</details>

* Regenerates the utf-8 command arguments from the utf-16 command line string, and overwrites argc and argv at main() startup.
The new utf-8 arguments are created on the heap, and released at the end of main().
* Code that relies on ANSI are deleted or changed.
* Initialize the local variable "settings" in main() to NULL.
This is because, there is a possibility that delete_fluid_settings() will be called without calling new_fluid_settings().

- - - 
The operation of these codes was tested using binaries built with Microsoft Visual studio and MSYS2.
Calls from Win32API, and Windows command prompt were confirmed to work as expected.
It has not been confirmed whether the codes will operate normally in other build and execution environments.

Tested with non-ANSI characters.
for example:
* ♬.sf2
* ♬.mid
* ♬.wav

In this case, an error will occur when running with Fluidsynth 2.3.7.
And the file names are not displayed correctly.

>Parameter '?.sf2' not a SoundFont or MIDI file or error occurred identifying it.
Parameter '?.mid' not a SoundFont or MIDI file or error occurred identifying it.
Rendering audio to file '?.wav'..
fluidsynth: error: Failed to open audio file '?.wav' for writing

The code in the pull request fixes these issues.


stardusteyes:
There are some failures in the above workflow.
And even in the success, there are dangerous warnings.
It is that sf_wchar_open() is not found, when compiling fluid_filerenderer.c.

When the macro _WIN32 is not defined, sf_wchar_open() in sndfile.h is not declared.
It seems that the build in workflows are compiled without the macro _WIN32 being defined.

I'm not familiar with Azure DevOps / Pipelines, but this error/warning is a bit worrying.
If it's not a problem, please ignore my worries.

include/fluidsynth/misc.h
```diff
@@ -67,6 +67,7 @@ extern "C" {
 
 FLUIDSYNTH_API int fluid_is_soundfont(const char *filename);
 FLUIDSYNTH_API int fluid_is_midifile(const char *filename);
+FLUIDSYNTH_API void* fluid_alloc(size_t len);
```
derselbst, line 70: I'm hesitant to make fluid_alloc public for everyone. It is only used inside libfluidsynth. I'd prefer that you use plain `malloc()` and `free()` in `fluidsynth.c` and do not make it public.

derselbst (commented):
Thanks for the PR, one comment below. @pedrolcl You also committed encoding related changes in the past, would you like to have a look at this as well?

pedrolcl:
> Thanks for the PR, one comment below. @pedrolcl You also committed encoding related changes in the past, would you like to have a look at this as well?

This is just another method of fixing the same problems that were already reported in #128 #983 #1322 #1388 and then supposedly fixed by #718 #1325 and #1390. If the reported problems were already fixed, I don't see the need to re-fix it again. But this is more a comment for the issue report instead of the pull request.

About the PR itself: it is quite intrusive: +99 −87! and I agree with your comment about `fluid_alloc()`: this change requires bumping the library minor version number. We now that `fluid_free()` needs to be exposed, because client applications sometimes need to use it to ensure the use of the same runtime library, but I don't see the need of exposing `fluid_alloc()`. Please replace that.

The core of this method is ` CommandLineToArgvW(GetCommandLineW(),&argc)` which I dislike a lot, so I would never have used it.

include/fluidsynth/misc.h
stardusteyes: I agree, and thank you for checking.
I'll replace the code with malloc and free, test it, and then make a pull request again.
Please wait a while.

stardusteyes: Completed.
https://github.com/stardusteyes/fluidsynth/pull/1/files

Please confirm.

src/bindings/fluid_filerenderer.c
```diff
@@ -289,7 +289,32 @@ new_fluid_file_renderer(fluid_synth_t *synth)
...
+            if (u16_count != MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, filename, -1, filename_wchar, u16_count))
+                FLUID_LOG(FLUID_ERR, "Failed to convert UTF8 string to wide char string");
+            else
+                dev->sndfile = sf_wchar_open(filename_wchar, SFM_WRITE, &info); // file open
+            FLUID_FREE(filename_wchar);
+        }
+    }
+#else
     dev->sndfile = sf_open(filename, SFM_WRITE, &info);
+#endif
```
pedrolcl, line 317: in http://libsndfile.github.io/libsndfile/api.html#open : 
> The sf_open() function opens the sound file at the specified path. The filename is byte encoded, but may be utf-8 on Linux, while on Mac OS X it will use the filesystem character set. On Windows, there is also a Windows specific sf_wchar_open() that takes a UTF16_BE encoded filename.

It is quite absurd that `libsndfile` does not allow UTF-8 file names in Windows like it does on Linux. It may be acceptable this workaround in Fluidsynth, but this should be fixed upstream (`libsndfile`) as well.

Even more funny: in [1.1.0](https://github.com/libsndfile/libsndfile/releases/tag/1.1.0) you may find this change:

>Use UTF-8 as internal path encoding on Windows platform.

>This is an internal change to unify and simplify the handling of file paths.

>On the Windows platform, the file path is always converted to UTF-8 and converted to UTF-16 only for calls to WinAPI functions.

> The behavior of the functions for opening files on other platforms does notchange.

stardusteyes:
I'm closing this pull request due to the comment that the code is not good.
Thanks for checking, maintainer.
I wish Fluidsynth success.

PR #1414, opened 2024-10-30
Enable ALSA support by default.

0-wiz-0:
pkgsrc, a portable packaging system, supports ALSA on quite a number of platforms (including e.g. NetBSD, Solaris, Illumos). I think it's easier to just allow the option by default than to try collecting the allowed platforms.

Add two compatibility defines for error codes that do not exist on all those platforms.


derselbst (approved):
Ok, let's try it. Thanks!

PR #1421, opened 2024-11-05
Convert fluidsynth CLI arguments

### LINKED ISSUE #1388: MIDI files with non-ASCII characters in the filename fail to play on Windows.

pedrolcl:
Closes #1388

### DIFF
--- a/src/fluidsynth.c
+++ b/src/fluidsynth.c
@@ -46,9 +46,6 @@ void print_help(fluid_settings_t *settings);
 void print_welcome(void);
 void print_configure(void);
 void fluid_wasapi_device_enumerate(void);
-#ifdef _WIN32
-static char* win32_ansi_to_utf8(const char* ansi_null_terminated_string);
-#endif
 
 /*
  * the globals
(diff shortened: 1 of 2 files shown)


derselbst (approved):
I like that solution, thanks!

src/fluidsynth.c
```diff
@@ -403,11 +356,40 @@ int main(int argc, char **argv)
...
+    {
+        fprintf(stderr, "Out of memory\n");
+        goto cleanup;
+    }
+    else
+    {
+        for (i = 0; i < argc; ++i)
+        {
+            int u8_count = 0;
+            if (1 > (u8_count = WideCharToMultiByte(CP_UTF8, WC_ERR_INVALID_CHARS, wargv[i], -1, NULL, 0, NULL, NULL)))
```
derselbst, line 376: @pedrolcl We have a minor oversight: `WC_ERR_INVALID_CHARS` is not available for WindowsXP, causing the CI to fail:
https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=10875&view=logs&j=ae601b3d-8756-52a4-4bcc-59c4d24c3f8c&t=1e42d489-b2c6-5e77-903e-6d797975357c&l=653

JJC needs WindowsXP compatibility. I haven't seen this earlier, because the build failed due to an issue in libinstpatch related to your bump of libsndfile min version: https://github.com/swami/libinstpatch/pull/78

pedrolcl: Then, the only option is to replace the WC_ERR_INVALID_CHARS flag by 0.

derselbst: Would something like

```c
#ifndef WC_ERR_INVALID_CHARS
#define WC_ERR_INVALID_CHARS 0
#endif
```

make sense, so that at least on modern Windows it reports a proper error?

Issue #1427, opened 2024-11-11
voice with high filter Q killed too early, causing audible click


mrbumpy409:
### FluidSynth version
Affects 2.3 (and presumably older) and 2.4 versions of FluidSynth including current master and iir-tests branches.

### Describe the bug
By default FluidSynth cuts off a voice after the volume envelope release has reached -96 dB (as I recall... I'm going off of memory here). Normally the volume envelope being at -96 dB would be an inaudible cutoff, but when high filter Q is used, the filter hasn't stopped "ringing" at this point, and so the still audible sound is abruptly cut off, resulting in a click.

### Steps to reproduce
1. Download and extract the following zip file: 
[high_Q_note_cutoff_test.zip](https://github.com/user-attachments/files/17704226/high_Q_note_cutoff_test.zip)
2. Load the .sf2 file into FluidSynth and play the .mid file. I have also included an audio file `fluidsynth_iir-tests.flac` showing the result of this test using FluidSynth's current iir-tests branch.

### Current behavior
The voice cuts off while the filter is still resonating, causing an audible click. You can see this quite clearly by viewing the waveform in Audacity: 
![image](https://github.com/user-attachments/assets/72d7b9a5-81c1-4a02-a922-b37e0a1286c6)

### Expected behavior
The voice shouldn't be cut off until the filter output has also reached -96 dB.

### Additional context
Related to bug #1415.

spessasus:
This is strange. It seems that the filter is not being affected by the final gain amplifier and instead is being sent directly to the final output.

According to Section 9.1.8 of the sf2 specification:
![image](https://github.com/user-attachments/assets/49ef80c2-612c-42e9-8a41-70fa2548127b)

The gain should come after the filtered output, so the ringing should ramp to -96dB too. This happens when playing the `Whistlewind` in Polyphone.

derselbst:
Thanks again for the report. I didn't found time to look into this, but here are two thoughts:

> By default FluidSynth cuts off a voice after the volume envelope release has reached -96 dB (as I recall... I'm going off of memory here)

We wanted to in #319, but it didn't happen. On the other hand, there is that logic around "noise floor" that might explain the effect observed here:

https://github.com/FluidSynth/fluidsynth/blob/168183c3b63efdf43ec07449a1a921c2c7b63868/src/rvoice/fluid_rvoice.c#L52

> It seems that the filter is not being affected by the final gain amplifier and instead is being sent directly to the final output.

[...]

Issue #1424, opened 2024-11-06
Loud clicks on rapid ModEnv FilterFc changes


spessasus:
### FluidSynth version
_Execute `fluidsynth --version` and provide the output._
2.4.0

### Describe the bug
The attached preset `Echo Drops` will audibly click when the notes are played rapidly. See the video below:

https://github.com/user-attachments/assets/8511a80a-4803-4e33-bea6-76f89833c68f

### Expected behavior
No clicks, like in polyphone.

### Steps to reproduce
1. Load the attached soundfont
2. Play a short note
3. Hear clicks

[echo_drops.sf2.zip](https://github.com/user-attachments/files/17653522/echo_drops.sf2.zip)
IMPORTANT: Remove `.zip` extension. This is just `.sf2` file, not an archive!!

### Additional context
Definitely related to #1415 and #1417 , but this one is:
- not a distortion like Christian's report
- not related to AWE NRPNs like derselbst's report
- this is the filter clicking with modEnv release set to the minimum value.

So this is a new bug report. If it is similar enough to the two bugs mentioned above, feel free to close it.

spessasus:
The fix I implemented for this was to copy the volEnv code:
https://github.com/FluidSynth/fluidsynth/blob/d2c56ca8190d3bdda8bb9733fa714c3fd6718852/src/synth/fluid_voice.c#L37-L38

Fluid should to that for modEnv too IMO. Not noticeable and should fix the issue.

Issue #1417, opened 2024-11-04
strange audio distortion when using filter envelope after merging #1345


mrbumpy409:
### FluidSynth version
2.4 (and master)

### Describe the bug
Since #1345 was merged, there is now a strange audio distortion when the filter is modulated by the modulation envelope. I created a demo MIDI file and audio renderings that should hopefully make this quite evident: [filter envelope noise.zip](https://github.com/user-attachments/files/17612822/filter.envelope.noise.zip)

Inside the zip file, you will find the following files:
* `filter envelope noise.mid` – A MIDI file that uses the GS preset **008:038 Acid Bass** featured in [GeneralUser GS 2.0.1](https://www.schristiancollins.com/generaluser). This preset uses high filter emphasis and has both the modulation envelope and modulation LFO modulating the filter.
* `filter envelope noise-FluidSynth 2.4.flac` – The above MIDI file rendered using the latest FluidSynth master (one commit ahead of 2.4) via FluidSynth Plugin in REAPER. You can hear a strange, grainy distortion throughout correlating to the filter being modulated by the modulation envelope.
* `filter envelope noise-1345 reverted.flac` – Same as above but with #1345 reverted. You can hear no more distortion. Everything sounds as it should.
* `filter envelope noise-Roland SC8820.flac` – The above MIDI file rendered on Roland Sound Canvas VA (SC8820 model, reverb disabled). Filter popping with high Q and quick modulation is a known issue with IIR filters. I included this recording to show that FluidSynth handles filter popping much better than the Roland Sound Canvas, even without #1345.

It is unfortunate that such a substantial change was made to FluidSynth's behavior so soon before a major new release. From my perspective, FluidSynth's filter behavior was far better than any other SoundFont synth I had tested, and now it's broken. I'm really upset that I didn't have the time to test this until now, otherwise I would have been able to say something sooner.

ReinholdH:
I can confirm the distortion in the sample and agree to all your points. More to #1415.

derselbst:
> It is unfortunate that such a substantial change was made to FluidSynth's behavior so soon before a major new release

Let me clarify a few things:
* 2.4.0 is a minor release, technically, since it's a minor version bump.
* #1345 stayed open for a month, which I usually consider long enough, and it was already merged 4 months ago, which I don't consider as "soon".
* I was convinced - given the linked example rendering - that  #1345 would make things better.

Yet, I appreciate your input Chris and will look into that when I find the time.

[...]

Issue #1415, opened 2024-10-31
IIR filter: potential for producing clicks and cracks


derselbst:
### FluidSynth version
2.4.0 (recent master), 2.3.x is not affected, because it doesn't implement #1342 

### Describe the bug
Even though I have attempted to fix several clicks produced by the IIR filter on rapid fc changes in  #1345 , there still seems to be a potential for clicks and cracks.

### Expected behavior
Clean audio.

### Steps to reproduce

1. Checkout and compile recent master (or 2.4.0 when it's released)
2. Grab the ZIP below
3. `fluidsynth -F render.wav -R 0 -C 0   "The Nervous Filter trimmed.mid"   "GeneralUser GS v1.471.sf2"`

The first 5 seconds of the MIDI are sufficient. The original untrimmed MIDI is provided as well. The click may be more audible when reverb is enabled.

### Additional context
To my understanding, fluidsynth's implementation of the IIR filter is the result of an analog transformation of a 2nd order Butterworth filter into the z-domain. The exact cause of the clicks is currently unknown.

[The Nervous Filter.zip](https://github.com/user-attachments/files/17587614/The.Nervous.Filter.zip)

![Screenshot_20241031_132723](https://github.com/user-attachments/assets/b519dad5-e37e-4575-9362-1fae957cfaf1)

klerg:
Well, the clicks are quite obvious in this short The Nervous Filter trimmed.ogg

mrbumpy409:
IIR filters causing pops when modulated very quickly is a known issue. See [here](https://forum.juce.com/t/solved-what-causes-the-iirfilter-class-to-pop-when-filter-frequency-is-changed-quickly-and-how-can-i-fix-it/23660) as an example. FluidSynth's filter seems to handle this better than most. Unfortunately, #1345 introduces audible distortions (as I reported in bug #1417).

In all of my SoundFont programming, I have only ever created very slight occasional pops on my most extreme presets (such as **008:038 Acid Bass** in [GeneralUser GS 2.0](https://www.schristiancollins.com/generaluser)), and nothing that would really be perceivable in a mix. If NRPN control of filter cutoff is causing pops, you might want to incorporate "smoothing" logic to limit the speed at which a NRPN or modulator can increase/decrease the cutoff, though I would advise against also applying this smoothing logic to modulation from the modulation envelope, as this may mess with the sound of presets.

[...]

PR #1429, opened 2024-11-16
Add regression test cases for IIR filter

derselbst:
This adds a custom target to the build system `check_manual` that invokes rendering several test-MIDI files. The test files, along with potentially pre-rendered demonstration audio clips are checked into git lfs. The audio files rendered during build-time are stored in `build/test/manual/` and can be used as reference, whenever a part of fluidsynth's  logic is about to change. Esp. useful when changing legacy code, as in e.g. #1345.

For the beginning, test files for the IIR filter have been added, taken from issues #1415, #1417, #1424, and #1427.

Additionally, an interactive Bode-Plot for the IIR filter has been added written in Matlab.


spessasus:
#1427 is not related to the filter changes as it affects 2.3 and older, it is related to the order of operations in rvoice_write. It still should be included as a test case, but it is not related in any way to #1345, so maybe in a different folder?

derselbst:
> #1427 is not related to the filter changes

I don't think I have claimed that. However, #1427 seems to be caused by the filter's behavior, or at least the filter is involved here. That's why I currently believe it makes sense to keep it in the `iir_filter` subfolder.

PR #1430, opened 2024-11-16
Reenable AWE32 NRPN 23 and 24

derselbst:
After dealing with filters extensively and reading through the AWE32 docs yet again, I'm pretty sure that what they claim as "phase [...] filter modulation" should actually mean "filter cutoff modulation." Therefore, this PR enables AWE32 NRPN 23 and 24, causing them to behave just like the generators found in SF2.


klerg:
Interesting, here we find another case of the misleading and tricky AWE32 docs, will this result in a more accurate interpretation of the AWE32 NRPN's or what effect is this going to have on the overall filter ?

derselbst:
> what effect is this going to have on the filter overall ?

I'm currently unable to provide you binaries, because the Windows build is broken.

klerg:
I see when will the Windows build get fixed so you can provide the binaries ?

spessasus:
> I see when will the Windows build get fixed so you can provide the binaries ?

If you have windows, you can use WSL2 and build for Linux, then render with the `-F` option to a wav file. Works fine for me

derselbst:
@klerg Binaries are available again. Feel free to test, I can barely hear a difference:
https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=11032&view=artifacts&pathAsName=false&type=publishedArtifacts

klerg:
I'm unable to run any of three binaries, it gives the following error:

image

Any idea how to fix this ?

derselbst:
Sry, I fixed it: https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=11073&view=artifacts&pathAsName=false&type=publishedArtifacts

klerg:
That's fine, it loads but now the first 5 seconds or of The Nervous Filter.mid is malformed and mangled, is #1345 reverted in these binaries, or are they using #1432 smoothing ?

klerg:
That's fine, it loads but now the first 6 seconds or so of The Nervous Filter.mid is malformed and mangled, is https://github.com/FluidSynth/fluidsynth/pull/1345 reverted in these binaries, or are they using https://github.com/FluidSynth/fluidsynth/pull/1432 smoothing ?

derselbst:
The smoothing logic is not included here. It will be included in the final release.

klerg:
I see so is the issue with The Nervous Filter.mid caused by #1345 being reverted and can you send binaries with smoothing before final release ?

derselbst:
Sorry, I was very busy the past days. I cannot postpone the release any further, so I'll merge this now. Binaries will be available in the release.

klerg:
Sure that is fine and not a big deal here. I do not see a need to release a new version so soon. I just tested 2.4.1 and the 1st 7 seconds of The Nervous Filter.mid is still malformed and mangled so I will post it as a new issue

klerg:
Sure that is fine and not a big deal here. I do not see a need to release a new version so soon. I just tested 2.4.1 and the 1st 7 seconds or so of The Nervous Filter.mid is still malformed and mangled so I will post it as a new issue

PR #1432, opened 2024-11-16
Smooth linear filter parameter change

derselbst:
This PR implements a smooth change of the IIR filter parameters `fres` (=filter cutoff frequency) and `Q`. It is meant as a proper replacement for the previous linear smoothing of the individual filter coefficients, see #1345.

This PR is supposed to fix the clicks and pops heard in those related issues:

Fixes #1415 
Fixes #1417 
Fixes #1424

PR Depends on:
#1429
#1430

Open TODOs:

- [x] Investigate if the maximum smoothing range of `8*FLUID_BUFSIZE` can be shortened
- [x] Watch out for performance optimizations (e.g. writing and re-reading the filter coeffs should not go via global memory)
- [x] Publish rendered audio samples for comparison (native 2.4.0 release + 2.4.0 with #1345 reverted + changes as of this  PR) -  see my comment here for the test tunes: https://github.com/FluidSynth/fluidsynth/issues/1417#issuecomment-2495454180

PR #1434, opened 2024-11-17
Add test files for AWE32 NRPN

derselbst:
This adds three test files for the AWE32 NRPN support. `ALTITUDE` and `UPLIFT` are using their original soundbanks, converted to SF2 Emu's "SoundFont Librarian" `sflib10c.zip` utility.


derselbst:
@mrbumpy409 Could you do me a favor and pls. render `Uplift.mid` on your Audigy2 and upload it as OGG?
https://github.com/FluidSynth/fluidsynth/tree/awe32-tests/test/manual/awe32_nrpn/Uplift

@klerg Could you pls. render `Altitude.mid` *and* `Uplift.mid` on your SB Live and upload the rendered audio?
https://github.com/FluidSynth/fluidsynth/tree/awe32-tests/test/manual/awe32_nrpn/Altitude
https://github.com/FluidSynth/fluidsynth/tree/awe32-tests/test/manual/awe32_nrpn/Uplift

If possible pls. prefer using the SF2 files in the respective directories, so that it's as close as possible to fluidsynth output. I would like to store the rendered audio for future reference.

klerg:
Sure that will be fine, and I may render all the midis with the original banks, as frenetic.mid, Marathon.mid, MYSTRAVE.mid, trancpar.mid, and TRUTH.mid sound totally different, will MP3 be fine ?

derselbst:
Uplift and Altitude would be sufficient. MP3 is fine as well.

klerg:
I see that is fine but it says "Uploads are disabled" what to do here ?

derselbst:
You need to ZIP them. If it still doesn't work, feel free to send them via e-mail.

mrbumpy409:
> @mrbumpy409 Could you do me a favor and pls. render `Uplift.mid` on your Audigy2 and upload it as OGG? https://github.com/FluidSynth/fluidsynth/tree/awe32-tests/test/manual/awe32_nrpn/Uplift

I will try to get this recorded over the next couple of days.

klerg:
Same here I will email them very soon and you can put that in the above links

mrbumpy409:
SBK files converted to SF2 all need further edits in order for the SoundFont to render correctly on a SoundFont 2.01 or 2.04-compliant device. This is also true for version 2.00 SoundFonts such as those created on the Sound Blaster Live!. I have been working on proper conversions of all the songs contained in [this archive](https://ftp.lysator.liu.se/pub/awe32/songs/) along with some others that I had collected back in the day. Here is a list of the changes I have had to make across all of the SoundFonts to approximate SoundFont 1.0 / 2.0 behavior:

1. Disable default *velocity→filter cutoff* modulator for each instrument (both 2.01 and 2.04 spec versions).
2. Enable *velocity→filter cutoff* modulator (2.04 spec version) for any samples with volume envelope attack at 0.008 seconds or greater.
3. Re-route mod wheel (CC1) and channel aftertouch to modulate modLFO (LFO1) instead of vibLFO (LFO2). (Was CC1 switched over to using vibLFO before SoundFont version 2.01? If someone with a SB Live! could run [this test](https://github.com/mrbumpy409/SoundFont-Spec-Test), I would be much obliged.)
4. Increase reverb (CC91) and chorus (CC93) controller sensitivity to better match AWE32 fx levels when using Audigy2 or FluidSynth.
5. Replace ROM samples with RAM copies.
6. In the original SoundFont implementation, if note was played on a drum kit (MIDI channel 10) on bank 1 or higher that didn't have an assigned sample in the preset, the corresponding note from bank 0 would be played instead. This is not true for SoundFont 2.01 or 2.04 devices, where the notes missing from the bank 1 preset will just play silence. I'm not sure whether this was changed in SoundFont revision 2.0 or 2.01. Anyway, some AWE32 songs rely on this behavior, so the missing percussion sounds must be copied into the song's SoundFont bank.
7. Fix various SoundFont Librarian conversion errors as I encounter them. This part is very hard to do without reference to any AWE32 recordings. For example, I was able to find and fix a very bad pitch error in *Tekkniko* by Niko Boese thanks to [this YouTube video](https://www.youtube.com/watch?v=FZKDBpK9nbA). Without that, I might never have noticed one instrument playing inaudibly 5 octaves below its correct pitch.

[Altitude_and_Uplift-Audigy2.zip](https://github.com/user-attachments/files/17856885/Altitude_and_Uplift-Audigy2.zip) contains SoundFont 2.01/2.04-compliant versions of both Altitude and Uplift as well as renderings from my Audigy2 ZS.

### Uplift

Uplift was a bit of a mess. There were a number of samples in the SF2 that had out-of-spec loop points (e.g., loop end was set beyond the end of the sample) leading to horribly out-of-tune playback on Audigy2. This affected presets **000:004 HARSH SQUARE** and **000:006 SAWTOOTH VIBRATO**. I'm not sure how the AWE32 handled the out-of-spec loops, but whatever the case was, Ian Chisholm appears to have tuned the samples around this specific rendering. Once the loop points were fixed, Ian's tuning values were no longer correct and the samples had to be retuned. The instruments containing these samples now play in tune, but if there was any deliberate detuning in the original preset versions, this has been lost. Only access to an AWE32 might reveal the truth.

Uplift.sf2 was also missing referenced percussion samples (see point 6 above), so I copied them over from by AWE32 `synthgs.sbk` SF2 conversion, which is also included in the above ZIP archive. The Audigy2 ZS recording of Uplift uses `synthgs-sf2_04-compat.sf2` as the GM bank, since a couple of tracks require bank 0 GM voices. For the recording, I also used my fixed copy of the MIDI file (`Uplift-fixed.mid`), which removes the quantization errors that caused frequent "hiccups" in the tempo. It's possible that a few errors might remain, but not enough to be noticeable.

derselbst:
Wow, many thanks Christian! I admit, the difference is clearly audible! Also thanks for the explanations, I thought it would be worth to store that in the wiki, for future reference:
https://github.com/FluidSynth/fluidsynth/wiki/SoundFont#differentiation-to-sbk--sf1

I would merge this now so I can proceed with #1432 and #1430, though I haven't received @klerg's renderings yet. Looking forward to your SB Live! renderings Klerg, no need to hurry though. Pls use Christians updated SF2 for rendering Uplift and Altitude, which you can find here:

https://github.com/FluidSynth/fluidsynth/raw/51242f7603c4f48141dadedfefed1c9a6e286a34/test/manual/awe32_nrpn/Altitude/Altitude.sf2?download=
https://github.com/FluidSynth/fluidsynth/raw/51242f7603c4f48141dadedfefed1c9a6e286a34/test/manual/awe32_nrpn/Uplift/Uplift.sf2?download=

Additionally, pls do Christian a favor and render his midi file with the Soundfont found here:
https://github.com/mrbumpy409/SoundFont-Spec-Test

If uploading to Github doesn't work, just mail me and I'll forward it.

mrbumpy409:
> I'm pretty sure that the Sound Blaster Live! is a SoundFont 2.01 and not 2.00 sound card as well.

I'm pretty sure the first SoundFont 2.01-compliant Sound Blaster was the Audigy. The Live! supports SoundFont 2.00 only, as I recall. I believe this will be confirmed when you run my SoundFont spec test on your Live!

>     2. So I assume you keep this modulator on but only for samples that have a volume envelope attack of 0.008 or higher and my guess is makes it sound better overall then ?

In SoundFont 2.0 and earlier, velocity-based filtering was enabled by setting the volume envelope attack greater than 0.001, though the threshold seemed to differ between AWE32 and Live! for some reason. On AWE32, it was 8ms or higher attack to enable the filtering, and on Live! it was 7ms? I can't remember for sure, but I'll post back here when I am back home and can check my notes. Anyway, the modulators I add are to emulate this behavior on SoundFont 2.01 and 2.04 devices.

>     3. No clue why SF 2.01 had to re-route Mod wheel (CC1) and channel aftertouch to modLFO (LFO1) to vibLFO (LFO2)

The bigger mystery is why CC1 was ever being routed to the multi-purpose LFO in the first place. Routing to the dedicated vibrato LFO just makes more sense. The SoundFont 2.01 standard set this straight. Your test results will reveal if they changed this in SoundFont 2.00 or 2.01.

> Sure I got a SB Live! and can run this test, and will contact you soon on this. Just do not know what to expect from it so will get a hold of you outside of github on this.

Thanks, I appreciate it!

>     6. Obviously, the original soundfont specification had this the right way, and not sure why in SoundFont 2.01 and above it was changed for the worse. It makes sense to fall back on bank 0 when bank 1 does not have the required percussion sounds. I cannot see any good reason to take away a feature like this. I bet every AWE32 song here with a Ch.10 track expects this as well.

No, only some. I have already copied the missing percussion sounds where necessary for all of the AWE32 songs in the archive linked previously. I just need to figure out the best way to share my converted collection and then create proper documentation, etc. If only I had more free time, I would have done so already.

> I'm not sure how you pulled off making Altitude and Uplift to be soundfont 2.01 and 2.04 but always nice to know that it can be done and hope it was not too hard at all

Well, it helps that I spent my teenage years with an AWE32 and Live! sound card, and then migrated to Audigy and then Audigy 2, paying very close attention to the behavior of each. I even created some of my own AWE32 MIDI files (one of which received 6th placing in the Creative Labs international SoundFont composition contest in... I think it was 1997?).

> Sure but it looks like with my SB Live! it is better to record the version of Uplift linked in here. It makes no sense to do this if it is going to end with all the issues you mentioned. But my hope is on the SB Live! it will sound closer to the AWE32 and I will be able to confirm this in a better way.

My fixed SoundFont files should be backwards-compatible with SoundFont 2.0 devices like the SB Live!, which should just ignore my custom modulators. So, it will be best to use my fixed SoundFonts on SB Live! as well.

klerg:
I hope you are right on this it looks like the SB Live! came out in August 1998, but the SoundFont 2.01 spec was released in July 1998. Yet, the EMU10K1 is dated from 1997, sure I will contact you later to run that test.

Yes, I heard about this in what was most likely one of your detailed and thorough write-ups on SoundFont 2.0 and much more than that. Sure, Creative or E-mu wanted the AWE32 (which is SoundFont 1.0, right?) to sound different in when and where the velocity-filtering starts. No clue on the AWE32, but pretty sure the Live! had that at around 0.007 or so. That's fine, take some time to find out this out. I see that makes sense and the modulators must be needed at this point. 

Sure it is good to ask why CC1 is routed to the mod LFO at all and what the goal of that even is. Yes as CC1 is made to adjust the pitch, and vibrato is also related to that as well. Glad that SoundFont 2.01 sorted all this out too. Sure we can get to the spec test fairly soon then.

No problem, I'm happy to try this

I guess some songs have no percussion then ? Nice that you were able to add the necessary percussion sounds back into all of the AWE32 songs then. It will not be hard to share the collection but the documetation may take some time, and I'm sure you are going to find some free time soon and it will be interesting to hear and read up on this

Cool, I was still a kid when the AWE32 came out but was a teenager at the time the Live! was released, I have an Audigy and Audigy 2 ZS just no empty slots on my systems to install them. Nice to know that you made some AWE32 MIDI files and that they made it to 6th place in the Creative Labs competition for SoundFonts and would be awesome to hear them


I see but not sure if that seems to be the case as the files all have the two modulators you mentioned before, yes I guess the SB Live! ignores the modulators then. Ok, here is Altitude and Uplift using your soundfont files then

[Altitude and Uplift 2.zip](https://github.com/user-attachments/files/17893869/Altitude.and.Uplift.2.zip)

klerg:
I hope you are right on this it looks like the SB Live! came out in August 1998, but the SoundFont 2.01 spec was released in July 1998. Yet, the EMU10K1 is dated from 1997, sure I will contact you later to run that test.

Yes, I heard about this in what was most likely one of your detailed and thorough write-ups on SoundFont 2.0 and much more than that. Sure, Creative or E-mu wanted the AWE32 (which is SoundFont 1.0, right?) to sound different in when and where the velocity-filtering starts. No clue on the AWE32, but pretty sure the Live! had that at around 0.007 or so. That's fine, take some time to find out this out. I see that makes sense and the modulators must be needed at this point. 

Sure it is good to ask why CC1 is routed to the mod LFO at all and what the goal of that even is. Yes as CC1 is made to adjust the pitch, and vibrato is also related to that as well. Glad that SoundFont 2.01 sorted all this out too. Sure we can get to the spec test fairly soon then.

No problem, I'm happy to try this

I guess some songs have no percussion then ? Nice that you were able to add the necessary percussion sounds back into all of the AWE32 songs then. It will not be hard to share the collection but the documetation may take some time, and I'm sure you are going to find some free time soon and it will be interesting to hear and read up on this

Cool, I was still a kid when the AWE32 came out but was a teenager at the time the Live! was released, I have an Audigy and Audigy 2 ZS just no empty slots on my systems to install them. Nice to know that you made some AWE32 MIDI files and that they made it to 6th place in the Creative Labs competition for SoundFonts and would be awesome to hear them


I see but not sure if that seems to be the case as the files all have the two modulators you mentioned before, yes I guess the SB Live! ignores the modulators then. Ok, here is Altitude and Uplift using your soundfont files recorded on my SB Live!

[Altitude and Uplift 2.zip](https://github.com/user-attachments/files/17893875/Altitude.and.Uplift.2.zip)

mrbumpy409:
> I guess some songs have no percussion then ?

I think you're misunderstanding the problem. Most of the percussion sounds used by the AWE32 MIDI files are included in the SoundFont. It's just that sometimes they also try to access sounds that are not a part of the bespoke kit, and it is _those_ percussion sounds that must be manually added back in.

> Ok, here is Altitude and Uplift using your soundfont files recorded on my SB Live!

I can hear that you have reverb and chorus disabled on your Live! It would be a better comparison if you recorded these MIDI files with those effects enabled on your card. Also, the previous version of the *Uplift* recording you shared has lots of audio dropouts. Notice how much shorter it is than your second recording.

Now that I have your comparison files, it appears that the Audigy is playing the wrong percussion sounds in *Uplift*. Once I get back home to my desktop PC, I will investigate why this is the case. Once I figure it out, I'll probably have to share another fix and Audigy2 recording. Besides the missing reverb and chorus, it would seem the other differences I perceive are due to you using a different GM bank (can you confirm?).

klerg:
Well, I do not have anything called "What U Hear" as a mixer recording source, so looks like will not be able to record chorus and reverb then ?

mrbumpy409:
Perhaps it is called "Stereo Mix"? It's been a while since I owned a Live, so I don't remember.

klerg:
I do not get "Stereo Mix" either here are all my options: SPDIF-In, CD Digital, MIDI, CD Audio, Line-In, Microphone, Auxiliary, TAD-In, and Wave. MIDI is selected, but effects are not recorded, anything else to try ?

mrbumpy409:
Are you using the Windows mixer or the Creative Labs mixer? I think you might need to use the latter to see the option as a recording source. Or it might be hidden. Here's a discussion on the matter, if it's of any help: https://www.sevenforums.com/sound-audio/20437-record-what-u-hear.html

What OS are you using?

klerg:
I'm using the Windows mixer and Creative Labs mixer too, and the only extra option I get with it is the PC Speaker and nothing else even close to "What U Hear" or "Stereo Mix" That is a good forum topic on this matter, but does not help.

Windows XP SP3

I did nor even know the Audigy2 works on Windows 7, let alone Windows 10, I guess Creative brought back "What u Hear" as a recording source or something like that

mrbumpy409:
[Here](https://github.com/user-attachments/files/17967453/Uplift-fix2.zip) is another fix to the *Uplift* MIDI file and Audigy2 recording. The MIDI file was attempting to access preset 1 on the percussion channel while the actual percussion sounds are on preset 0. On AW32/Live!, this falls back to bank 1, preset 0, but on SoundFont 2.01 devices, this falls back to bank 0, leading to the wrong percussion sounds being heard.

Until I can get somebody to submit a recording of my [SoundFont Spec Test](https://github.com/mrbumpy409/SoundFont-Spec-Test) using an AWE32/64 or Live! with reverb and chorus enabled ([link to my request here](https://github.com/mrbumpy409/SoundFont-Spec-Test/issues/2)), I will not be able to accurately set reverb and chorus send levels in my AWE32 song conversions, including the two songs I have uploaded here.

spessasus:
> Until I can get somebody to submit a recording of my [SoundFont Spec Test](https://github.com/mrbumpy409/SoundFont-Spec-Test) using an AWE32/64 or Live! with reverb and chorus enabled ([link to my request here](https://github.com/mrbumpy409/SoundFont-Spec-Test/issues/2)), I will not be able to accurately set reverb and chorus send levels in my AWE32 song conversions, including the two songs I have uploaded here.

You should try asking over at [VOGONS forum](https://www.vogons.org/viewforum.php?f=62). This is a forum about retro games and music, plenty of people discuss the old soundcards (especially in the category i linked above). You will probably find someone with AWE32 willing to record these :)

mrbumpy409:
> You should try asking over at [VOGONS forum](https://www.vogons.org/viewforum.php?f=62). This is a forum about retro games and music, plenty of people discuss the old soundcards (especially in the category i linked above). You will probably find someone with AWE32 willing to record these :)

That's a great idea, Spessasus! I [just posted over there](https://www.vogons.org/viewtopic.php?t=103797) thanks to your suggestion.

klerg:
Nice, Uplift.mid basically sounds almost perfect now. Yes, I could tell
the MIDI tries to use the wrong preset for Ch.10 percussion for sure when
everything is on preset 0. That's right the AWE32/Live! falls to bank 1 and
preset 0, but for Audigy and above it goes to bank 0, and that results in the
wrong sounds and we get no sound on them.

I still am not able to record the effects on my Live! card even when they are 
enabled, My only option may be to use Total Recorder or something like it that
has a driver for "What U Hear" or 'Stereo Mix' I'd rather not mess with the Live! 
drivers until I've went through every other recourse then.

Yes, that is a great tip to ask on VOGONS forum. I'm going to post here for some help on my issues with this Live! Value Dell OEM card and a way to get "What U Hear"as a recording source on Windows XP on here as well

mrbumpy409:
Thanks to user @Kuokka77, I was able to measure the AWE64 effects response for internal vs. CC91/93 values (which should also be the same for Sound Blaster Live!, but I will need a recording with FX on that hardware to confirm). From this test, I can ascertain that the CC91/93 modulation of reverb/chorus amount is 400 (40%), compared to the SoundFont 2.01/2.04 default of 200 (20%).

Furthermore, this test allowed me to attempt to balance FluidSynth's effects levels and settings to better match the AWE32/64's "Hall 1" and "Chorus 1" presets. With the SoundFont configured for 40% CC91/93 FX send, the following FluidSynth effects settings should result in a decent effects level correspondence to the AWE32/64:
```
set synth.reverb.damp 0.3
set synth.reverb.level 0.6
set synth.reverb.room-size 0.65
set synth.reverb.width 0.8
set synth.chorus.depth 4.7
set synth.chorus.level 0.43
set synth.chorus.nr 3
set synth.chorus.speed 0.4
```
Of course, it will never be a perfect match due to differences between the synthesizers' effects engines. I have also updated my Audigy2 recordings with similar, new FX settings. [Here is a zip file](https://drive.google.com/file/d/1GL5VUMY33w6msr4wJHLBxLqaU1m0UAkK/view?usp=sharing) containing the following:
* `Altitude-FluidSynth-1427.ogg` – `Altitude.mid` played using the latest FluidSynth build from the 1427 branch with the updated SoundFont and effects settings above.
* `Altitude-SB Audigy2 ZS.ogg` – `Altitude.mid` played on Sound Blaster Audigy2 ZS with the updated SoundFont and FX settings.
* `Altitude.sf2` – SoundFont updated for 40% FX modulation via CC91/93.
* `synthgs-sf2_04-compat.sf2` – SoundFont updated for 40% FX modulation via CC91/93.
* `Uplift-fixed-FluidSynth-1427.ogg` – `Uplift-fixed.mid` played using the latest FluidSynth build from the 1427 branch with the updated SoundFont and effects settings above. FluidSynth's internal MIDI player would not play this MIDI file correctly, resulting in the wrong instruments on some tracks. I had to use a separate MIDI player (Sekaiju) for this recording.
* `Uplift-fixed-SB Audigy2 ZS.ogg` – `Uplift-fixed.mid` played on Sound Blaster Audigy2 ZS with the updated SoundFont and FX settings.
* `Uplift.sf2` – SoundFont updated for 40% FX modulation via CC91/93. Note that this SoundFont features some very high internal reverb and chorus values, so the piece is quite effects heavy (especially the "Mega Strings" preset, which is set to 100% internally for both reverb and chorus).

Comparing Audigy2 and FluidSynth playback, I can hear that the NRPN filter control is bringing the cutoff much lower on Audigy2 than FluidSynth (and possibly higher as well?). I may make some tests for this when I have some more time.

derselbst:
Thanks for the update Chris! Currently, I'm subtracting 1000Hz from the cutoff, because I only had a single hardware rendered comparison tune. I'm open to change this.

mrbumpy409:
I will make a proper test for this that I can render using my Audigy2 for comparison. That should make it much easier to match the behavior in FluidSynth. I just need to get through a couple projects and student recitals first.

klerg:
I'm glad someone with an AWE64 Gold helped on this as I, and many others
only have an AWE64 Value with 512k RAM. Yes, I tried Total Recorder and Virtual
Audio Cable, but none of them gave me a "What U Hear" recording source. I also
"force" installed the LiveWare 3.0 and LiveDrvUni-Pack drivers, and none of them 
gave the elusive "What U Hear" The only other thing to try to get this CT4780 working
is in Windows 98SE as that may enable "What U Hear"

Well, it is nice to know of a way that FluidSynth can be I guess sort of molded to fit
a certain sound card, like the AWE32/64, although I would not use the settings here for
every MIDI but only ones that were made for the AWE32/64 like these. And 40% for reverb
and chorus does not seem too extreme but it will be noticeable for sure and nice to hear


That's fine, we have no need for anything to be perfect so long as it sounds close enough
to the original sound. I see it will be good to hear how the Audigy2 differs from Fluidsynth
and also how similar both of them sound as well. 

I do not know where to obtain the 1427 branch of Fluidsynth, but it will be great if you could
point me to the right place to find this. My Live! Value sounds pretty close to this as well

I really like how Altitude.mid sounds on the Audigy 2 ZS, I have one too but may need to stick it 
in a VIA chipset

Yes, I see the modulator for 40% reverb and chorus on all the instruments too.

Did you add the modulator of 40% reverb and chorus to every instrument or is there a faster way ?

I see this looks a bug in FluidSynth, where it plays the wrong instruments in Uplift.mid, and I did
mention this in a comment from another issue. I may report this as a formal bugI never heard of Sekaiju, 
but it looks really cool and does come in handy in a lot of cases.

Again, I really like how Uplift.mid sounds on the Audigy 2 ZS, and many points sound better than Fluidsynth
which I will bring up later.

Yea, it is rare to find any soundfont with such high reverb and chorus values of 100%. I personally think that
is way too high and is going to drown out the sound at some point. But I guess a preset like "Mega Strings" can
lend itself to extreme effects levels. 


Yes especially in Altitude.mid at around the 1:10 mark up to 2:20 and again from 2:25 up to very close to the end
of the piece. The filter cutoff definitely is lower on the Audigy 2 ZS, hopefully is it not by too much, and look
forward to your tests to get a better idea on the amount of the filter cutoff.

mrbumpy409:
I made a test file for testing the AWE32 filter cutoff NRPN behavior, but there is a FluidSynth issue that is preventing this test from providing the information I need. When setting filter cutoff via AWE32 NRPN, FluidSynth appears to be setting filter emphasis to 0, overriding the value I have specified in the preset. I need this value to not be overridden in order to accurately identify the filter cutoff point in a spectrogram. Let me know if I should file a separate bug report about this. I am experiencing this using the 1427 branch (compiled from before it was merged).

derselbst:
> When setting filter cutoff via AWE32 NRPN, FluidSynth appears to be setting filter emphasis to 0, overriding the value I have specified in the preset

Yes. From my understanding, the NRPN sets an absolute value for its respective generator, i.e. if an NRPN is present, it overrides the value of that SF2 generator during noteon, here:

https://github.com/FluidSynth/fluidsynth/blob/0f308a953d737a99cce49fd35781ab2cb65a80b1/src/sfloader/fluid_defsfont.c#L1028-L1032

Currently, it's not possible to undo this override, unless you send a system reset event. If you prefer a differnet way to handle this, feel free to file a new issue.

mrbumpy409:
I have created an AWE32 filter cutoff (FC) NRPN test: [AWE32_FC_NRPN_test.zip](https://github.com/user-attachments/files/18201667/AWE32_FC_NRPN_test.zip)

Here are the test results in a spectrogram, Audigy2 on top, FluidSynth on bottom:
image

Here are details on each test and how FluidSynth currently fares. Each test uses a white noise waveform with 24 dB filter emphasis applied within the preset. This emphasis is to make the filter cutoff point more visible when viewed in a spectrogram:

1. You will hear four filter sweeps, each one using a different filter cutoff in the preset. If the AWE32 FC NRPN is properly implemented, there should be no difference between the four sweeps:
   1. FC = 20000 Hz
   2. FC = 8000 Hz
   3. FC = 2000 Hz
   4. FC = 200 Hz

   **FluidSynth result:** The wrong filter curve is used, and the filter cutoff NRPN is causing the preset-level filter emphasis to be canceled, so it is hard to measure the actual cutoff point.

2. You will hear a fifth filter sweep that emulates the AWE32 FC NRPN filter curve via CC1 modulator. The result should be identical to the test 1 sweeps above.

   **FluidSynth result:** As expected, FluidSynth renders the CC1 modulator correctly. Since this modulator perfectly emulates the AWE32 FC NRPN behavior on Audigy2, perhaps it could be used as a model for correcting the FluidSynth's AWE32 FC NRPN behavior. The configuration is essentially this:
   * Preset filter cutoff = 100 Hz
   * Modulator curve = unipolar linear positive, amount = 7646 (timecents), destination = filter cutoff

3. You should hear a sustained white noise wave while the filter switches every two seconds through the following values:
   * 0 (100 Hz)
   * 13 (157 Hz)
   * 25 (237 Hz)
   * 38 (371 Hz)
   * 51 (581 Hz)
   * 64 (910 Hz)
   * 76 (1377 Hz)
   * 89 (2156 Hz)
   * 102 (3376 Hz)
   * 114 (5108 Hz)
   * 127 (8000 Hz)
   
   Each jump in the filter should sound roughly like the interval of a fifth.

   **FluidSynth result:** Same issue as test 1.

4. The same as the above test, but with the AWE32 FC NRPN filter curve emulated via CC1 modulator. The result should be identical to test 3 above.

   **FluidSynth result:** Same conclusion as test 2.

mrbumpy409:
I have improved my filter NRPN test and updated it to include tests for the filter Q NRPN. It made sense to put my findings in their own bug report, so you can read the details in issue #1473.

PR #1444, opened 2024-12-07
Fix filter processing order in DSP chain

### LINKED ISSUE #1427: voice with high filter Q killed too early, causing audible click

derselbst:
Previously, the IIR filter was applied after the gain of the volume envelope was applied to the signal. This could have caused audible clicks when a voice is turned off while its filter was using a high Q value, see #1427.

This PR fixes that by applying the filter on the interpolated signal but before the volEnv gain is applied. This also matches the processing order of the DSP chain closer to the diagram of section 9.1.8 in the sf2 specification.

`fluid_iir_filter_apply()` which is actually meant to process an array of samples is now used to process only a single sample. One might consider to define this function in header file to allow the compiler to inline it...

Fixes #1427

### DIFF
--- a/src/rvoice/fluid_rvoice.c
+++ b/src/rvoice/fluid_rvoice.c
@@ -435,6 +435,16 @@ fluid_rvoice_write(fluid_rvoice_t *voice, fluid_real_t *dsp_buf)
                  || (voice->dsp.samplemode == FLUID_LOOP_UNTIL_RELEASE
                      && fluid_adsr_env_get_section(&voice->envlfo.volenv) < FLUID_VOICE_ENVRELEASE);
 
+    /*************** resonant filter ******************/
+
+    fluid_iir_filter_calc(&voice->resonant_filter, voice->dsp.output_rate,
+                          fluid_lfo_get_val(&voice->envlfo.modlfo) * voice->envlfo.modlfo_to_fc +
+                          modenv_val * voice->envlfo.modenv_to_fc);
+
+
+    /* additional custom filter - only uses the fixed modulator, no lfos... */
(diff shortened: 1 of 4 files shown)


spessasus:
> fluid_iir_filter_apply() which is actually meant to process an array of samples is now used to process only a single sample. 

Won't that hurt the performance? If the function was designed to handle blocks of data, won't calling it repeatedly with only one sample make it slower?

derselbst:
I don't think so. The function is inherently serial, i.e. it cannot be vectorized. Hence, the most expensive part is reading the samples from memory, which is (hopefully) avoided with the current approach as the sample might be ideally located in a register.

Ofc, you're free to benchmark this.

PR #1459, opened 2024-12-30
Remove orphaned ramsfont typedefs

yuan-xy:
remove typedef of fluid_ramsfont_t and fluid_rampreset_t because the ramsfont/rampreset has been removed.


derselbst (approved):
Oh, a leftover. Thanks!

PR #1460, opened 2025-01-01
Update documentation about reverb width

### LINKED ISSUE #1455: reverb width range discrepency between effect behavior and documentation
### FluidSynth version
2.4.1

### Describe the bug
All of FluidSynth's reverb parameters have a range of 0–1, with the exception of width, which has a range of 0–100. However, the effective range for this parameter is actually 0–1, with 1 being 100% effect width. Values above 1 begin introducing an odd feedback into the reverb, and should be avoided.

At the very least, I believe this should be mentioned in the documentation, as the developers of software using FluidSynth frequently set the controls to the full 0–100 range. This often leads to control knobs where setting values between 0–1 is difficult or impossible due to the granularity of the control, and most of the usable knob range is in the weird feedback zone. QSynth has since [limited its reverb width control](https://github.com/rncbc/qsynth/issues/32) to 0–1 based on my advice, and Drumstick MIDI Player will now also be [doing the same](https://github.com/pedrolcl/drumstick/issues/26).

Here is a [test MIDI file and FluidSynth recordings](https://github.com/user-attachments/files/18222555/reverb.width.test.zip) rendered at the following reverb width settings:

* **0** – 0% stereo width in the reverb effect (mono)
* **0.5** – 50% stereo width in the reverb effect
* **1** – 100% stereo width in the reverb effect
* **5** – strange feedback/delay introduced into the effect
* **10** – even more feedback
* **50** – ...and more
* **100** – ...a crap ton of feedback

### Expected behavior
One of the following should happen, IMO:

1. Clearly document that a reverb width setting of 0–1 corresponds to 0%–100% reverb stereo width and above that goes into undesired jank territory.
2. Change the documentation to report the reverb width range as 0–1, leaving support for values up to 100 in the synthesizer for compatibility purposes.

derselbst:
This PR adds two test files for reverb and updates the docs about reverb width as suggested in #1455.

Fixes #1455

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -348,7 +348,7 @@ Developers:
             <max>100.0</max>
             <realtime/>
             <desc>
-                Sets the stereo spread of the reverb signal. A value of 0 indicates no stereo-separation causing the reverb to sound like a monophonic signal. A value of 1 indicates maximum separation between the the left and right channels (i.e. reverb is still a monophonic effect). This subrange [0;1] is recommended for general usage. Values bigger 1 increase (or exaggerate) the perception of the uncorrelated left and right signals. Otherwise, this setting should be considered as dimensionless quantity, with its maximum value existing due to historical reasons. Please note that under some circumstances, values bigger than 1 may induce a feedback into the signal which can be perceived as unpleasant.</desc>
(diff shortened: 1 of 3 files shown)


mrbumpy409:
Shouldn't the doc instead say "A value of 1 indicates maximum separation between left and right channels"? It seems the separation is at full stereo width at 1, and then only gets louder with a sort of feedback/delay attribute above 1.

mrbumpy409:
One more typo I found: "Values bigger 1 increase" should be "Values bigger **than** 1 increase". Otherwise, the updated description looks great.

derselbst:
Great, thanks for the feedback!

PR #1474, opened 2025-01-23
Silence "Referenced but unset environment variable" warning

andrew-sayers:
Systemd v254+ warns about unset environment variables, so we need to provide an explicit default value.

Based on https://github.com/analogdevicesinc/libiio/pull/1144


derselbst (approved):
Ok, thanks!

PR #1478, opened 2025-01-26
SDL3 Support

andyvand:
Add SDL3 support


src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,368 @@
...
+    fluid_audio_callback_t write_ptr;
+
+    SDL_AudioDeviceID devid;
+    SDL_AudioStream *stream;
+
+    int frame_size;
+} fluid_sdl3_audio_driver_t;
+
+AudioDeviceList AudioSDL3PlaybackDevices;
+AudioDeviceList AudioSDL3RecordingDevices;
```
derselbst, line 56: I don't think it's a good idea to have these two declared globally. They are potentially modified concurrently by `GetNumAudioDevices` when called from `new_fluid_sdl3_audio_driver`. I'd suggest to move them into the `fluid_sdl3_audio_driver_t` struct.

derselbst (commented):
Wow, that was quick. Thanks a lot! I've only got one comment below. Before merging it, I'll adapt the Window CI pipeline in the next days, to make it compile with SDL3 support.

src/drivers/fluid_sdl3.c
andyvand: The sources have been updated to use the structure and some code was rewritten.

.DS_Store
CoelacanthusHex, line 1: Remove it and add it to your global gitignore file. macOS always generates this annoying file automatically and it's totally useless for any software outside your macOS.

src/.DS_Store
CoelacanthusHex, line 1: Same as above.

.DS_Store
andyvand: Done

src/.DS_Store
andyvand: Done

src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,431 @@
...
+    fluid_sdl3_audio_driver_t *dev = NULL;
+    fluid_audio_callback_t write_ptr;
+    double sample_rate;
+    int period_size, sample_size;
+    SDL_AudioSpec aspec;
+    char *device;
+    const char *dev_name;
+
+    /* Check if SDL library has been started */
+    if(!SDL_WasInit(SDL_INIT_AUDIO))
```
FtZPetruska, line 280: You probably need to change the logic around this.

The Audio subsystem no longer gets automatically init'd when you open a device (see the [migration guide](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md#sdl_audioh)).

So you most likely want to have a call to `SDL_InitSubSystem(SDL_INIT_AUDIO)` when creating a new driver, and `SDL_QuitSubSystem(SDL_INIT_AUDIO)` when deleting the driver. They are refcount'd so it's safe to call them multiple times.

src/CMakeLists.txt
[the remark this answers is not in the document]
andyvand: Added

CMakeLists.txt
```diff
@@ -683,6 +684,26 @@ if ( enable-sdl2 )
     endif ( SDL2_FOUND )
 endif ( enable-sdl2 )
 
+unset ( SDL3_SUPPORT CACHE )
+if ( enable-sdl3 )
+    find_package ( SDL3 QUIET )
```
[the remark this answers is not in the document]
andyvand: Change done

src/drivers/fluid_sdl3.c
andyvand: SDL init and quit subsystem added

CMakeLists.txt
```diff
@@ -683,6 +684,26 @@ if ( enable-sdl2 )
...
+    if ( SDL3_FOUND )
+        if ( NOT DEFINED SDL3_VERSION )
```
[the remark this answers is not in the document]
andyvand: Removed if (NOT DEFINED SDL3_VERSION) check

cmake_admin/GetSDL3VersionFromHeaders.cmake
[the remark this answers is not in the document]
andyvand: Removed

src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,449 @@
...
+        {
+            /* Stop audio and close */
+            SDL_PauseAudioDevice(dev->devid);
+            SDL_CloseAudioDevice(dev->devid);
+        }
+
+        FLUID_FREE(dev);
+    }
+
+    SDL_QuitSubSystem(SDL_INIT_AUDIO);
```
derselbst, line 446:
```suggestion
        SDL_QuitSubSystem(SDL_INIT_AUDIO);
    }
```

A malicious user or app could have triggered calls to QuitSubSystem by just passing NULL pointers to this function, forcing the ref count to zero and therefore eventually crashing SDL while audio is still in use. Guard it by NULL check to avoid that.

src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,449 @@
...
+        SDL_memcpy(list, &newlist, sizeof (AudioDeviceList));
+    }
+
+    return num_devices;
+}
+
+static void
+SDLAudioCallback(void *data, SDL_AudioStream *stream, int add_len, int len)
+{
+    void *buffer = malloc(len);
```
derselbst, line 157: Memory allocation during audio playback is not real-time friendly. Is it possible to calculate the maximum `len` beforehand and use it to preallocate the biggest possible buffer?

src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,449 @@
...
+    SDL_AudioDeviceID *devices = NULL;
+
+    fluid_settings_register_str(settings, "audio.sdl3.device", "default", 0);
+    fluid_settings_add_option(settings, "audio.sdl3.device", "default");
+
+    if(!SDL_WasInit(SDL_INIT_AUDIO))
+    {
+#if FLUID_VERSION_CHECK(FLUIDSYNTH_VERSION_MAJOR, FLUIDSYNTH_VERSION_MINOR, FLUIDSYNTH_VERSION_MICRO) < FLUID_VERSION_CHECK(2,2,0)
+        FLUID_LOG(FLUID_WARN, "SDL3 not initialized, SDL3 audio driver won't be usable");
+#endif
```
derselbst, line 191:
```suggestion
        FLUID_LOG(FLUID_WARN, "SDL3 not initialized, SDL3 audio driver won't be usable");
```

This version check is used to prevent log messages when using an old fluidsynth executable and a recent libfluidsynth. I don't think it's needed here, so we can safely remove it.

src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,449 @@
...
+        return;
+    }
+
+    if(!SDL_InitSubSystem(SDL_INIT_AUDIO))
+    {
+#if FLUID_VERSION_CHECK(FLUIDSYNTH_VERSION_MAJOR, FLUIDSYNTH_VERSION_MINOR, FLUIDSYNTH_VERSION_MICRO) < FLUID_VERSION_CHECK(2,2,0)
+        FLUID_LOG(FLUID_WARN, "SDL3 subsystem not initialized, SDL3 audio driver won't be usable");
+#endif
```
derselbst, line 199:
```suggestion
        FLUID_LOG(FLUID_WARN, "SDL3 subsystem not initialized, SDL3 audio driver won't be usable");
```
Same

src/drivers/fluid_sdl3.c
```diff
@@ -0,0 +1,449 @@
...
+    {
+        FLUID_LOG(FLUID_ERR, "Failed to create SDL3 audio driver, because the audio subsystem of SDL3 is not initialized.");
+        return NULL;
+    }
+
+    if(!SDL_InitSubSystem(SDL_INIT_AUDIO))
+    {
+#if FLUID_VERSION_CHECK(FLUIDSYNTH_VERSION_MAJOR, FLUIDSYNTH_VERSION_MINOR, FLUIDSYNTH_VERSION_MICRO) < FLUID_VERSION_CHECK(2,2,0)
+        FLUID_LOG(FLUID_WARN, "SDL3 subsystem not initialized, SDL3 audio driver won't be usable");
+#endif
```
derselbst, line 298:
```suggestion
        FLUID_LOG(FLUID_WARN, "SDL3 subsystem not initialized, SDL3 audio driver won't be usable");
```

derselbst (commented):
Thanks for addressing the review findings. It's getting into shape. I found a few more minor issues.

CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ option ( enable-wasapi "compile Windows WASAPI support (if it is available)" on
 option ( enable-waveout "compile Windows WaveOut support (if it is available)" on )
 option ( enable-winmidi "compile Windows MIDI support (if it is available)" on )
 option ( enable-sdl2 "compile SDL2 audio support (if it is available)" off )
+option ( enable-sdl3 "compile SDL3 audio support (if it is available)" off )
```
derselbst, line 98:
```suggestion
option ( enable-sdl3 "compile SDL3 audio support (if it is available)" on )
```

CMakeLists.txt
```diff
@@ -857,6 +868,10 @@ else ()
   set ( includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}" )
 endif ()
 
+if ( SDL3_SUPPORT )
+    target_link_libraries ( libfluidsynth-OBJ PUBLIC SDL3::SDL3 )
```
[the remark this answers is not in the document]
derselbst: @FtZPetruska  is correct.  In addition, you need the same logic for the `fluidsynth` target into `src/CMakeLists.txt` as well, to inherit SDL3 linkage and include dirs, just like we did it for SDL2, i.e. something like this:

```cmake
if ( SDL2_SUPPORT ) # because SDL_Init() etc.
    target_include_directories ( fluidsynth PRIVATE ${SDL2_INCLUDE_DIRS} )
    target_link_libraries ( fluidsynth PRIVATE ${SDL2_LIBRARIES} )
endif()

if ( SDL3_SUPPORT ) # because SDL_Init() etc.
    target_link_libraries ( fluidsynth PRIVATE SDL3::SDL3 )
endif()
```

derselbst:
@FtZPetruska I quickly smashed together a Windows CI build that works with SDL3. Even when resolving the cmake issues mentioned above, all the unit tests will be failing because of missing DLL - presumably SDL3.dll, e.g. [here](https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=11560&view=logs&j=ae601b3d-8756-52a4-4bcc-59c4d24c3f8c&t=0c4024e7-76d6-5840-9660-e36e440de66f&l=2013)

The unit tests inherit their link dependencies from the object lib `libfluidsynth-OBJ` by linking against it here:
https://github.com/FluidSynth/fluidsynth/blob/de1daa4449ae48c5eeaab82aff6613f21417ae1f/cmake_admin/FluidUnitTest.cmake#L11

This would carry over all `INTERFACE_LINK_LIBRARIES` from the object lib to the unit tests. However, when looking at the definition of the sdl3 cmake target, `INTERFACE_LINK_LIBRARIES` only contains the headers' include dir. The DLL is instead declared by the `IMPORTED_LOCATION` property, which from my understanding is not used to specify transitive dependencies and therefore isn't carried over from fluidsynth's object lib to the unit test:

```cmake
add_library(SDL3::SDL3-shared SHARED IMPORTED)
        set_target_properties(SDL3::SDL3-shared
            PROPERTIES
                INTERFACE_LINK_LIBRARIES "SDL3::Headers"
                IMPORTED_IMPLIB "${_sdl3_implib}"
                IMPORTED_LOCATION "${_sdl3_dll}"
                COMPATIBLE_INTERFACE_BOOL "SDL3_SHARED"
                INTERFACE_SDL3_SHARED "ON"
                COMPATIBLE_INTERFACE_STRING "SDL_VERSION"
                INTERFACE_SDL_VERSION "SDL3"
        )
```

Is my understanding correct? And why isn't `_sdl3_dll` passed to `INTERFACE_LINK_LIBRARIES`?

FtZPetruska:
> @FtZPetruska I quickly smashed together a Windows CI build that works with SDL3. Even when resolving the cmake issues mentioned above, all the unit tests will be failing because of missing DLL - presumably SDL3.dll, e.g. [here](https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=11560&view=logs&j=ae601b3d-8756-52a4-4bcc-59c4d24c3f8c&t=0c4024e7-76d6-5840-9660-e36e440de66f&l=2013)
> 
> The unit tests inherit their link dependencies from the object lib `libfluidsynth-OBJ` by linking against it here:
> 
> https://github.com/FluidSynth/fluidsynth/blob/de1daa4449ae48c5eeaab82aff6613f21417ae1f/cmake_admin/FluidUnitTest.cmake#L11
> 
> This would carry over all `INTERFACE_LINK_LIBRARIES` from the object lib to the unit tests. However, when looking at the definition of the sdl3 cmake target, `INTERFACE_LINK_LIBRARIES` only contains the headers' include dir. The DLL is instead declared by the `IMPORTED_LOCATION` property, which from my understanding is not used to specify transitive dependencies and therefore isn't carried over from fluidsynth's object lib to the unit test:
> 
> ```cmake
> add_library(SDL3::SDL3-shared SHARED IMPORTED)
>         set_target_properties(SDL3::SDL3-shared
>             PROPERTIES
>                 INTERFACE_LINK_LIBRARIES "SDL3::Headers"
>                 IMPORTED_IMPLIB "${_sdl3_implib}"
>                 IMPORTED_LOCATION "${_sdl3_dll}"
>                 COMPATIBLE_INTERFACE_BOOL "SDL3_SHARED"
>                 INTERFACE_SDL3_SHARED "ON"
>                 COMPATIBLE_INTERFACE_STRING "SDL_VERSION"
>                 INTERFACE_SDL_VERSION "SDL3"
>         )
> ```
> 
> Is my understanding correct? And why isn't `_sdl3_dll` passed to `INTERFACE_LINK_LIBRARIES`?

This SDL3 target looks properly set up:
- [`IMPORTED_IMPLIB`](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_IMPLIB.html) is the .lib that is actually linked with.
- [`IMPORTED_LOCATION`](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_LOCATION.html) is the .dll that is needed at runtime.
- [`INTERFACE_LINK_LIBRARIES`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html) lists all the dependencies for linking, in this case, only the `SDL3::Headers` target. Since it is a target, "linking" to it also adds the public include directories.

What is most likely happening is that all the DLLs need to be copied to the same directory as the test executable. In CMake >=3.21 it can be done with:
```cmake
add_custom_command(TARGET ${_test} POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:${_test}> $<TARGET_FILE_DIR:${_test}>
  COMMAND_EXPAND_LISTS
 )
```

Otherwise, if you know about the lists of DLLs needed, you can just use:
```cmake
add_custom_command(TARGET ${_test} POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy
    $<TARGET_FILE:SDL3::SDL3>
    $<TARGET_FILE:other::dependency>
    ...
    $<TARGET_FILE_DIR:${_test}>
)
```

Also on the topic of CI, you may be interested in this Github Action: https://github.com/libsdl-org/setup-sdl

src/drivers/fluid_sdl3.c
andyvand: Removed version checks

src/drivers/fluid_sdl3.c
andyvand: Added it to additional len as buffer.

CMakeLists.txt
andyvand: Set now standard on

src/drivers/fluid_sdl3.c
derselbst: Sry, I think you misunderstood me. I was asking to get rid of the `malloc` from the callback and instead allocate the buffer in `new_fluid_sdl3_audio_driver`. Also note that `dev->write_ptr` takes the number of samples (audio frames) to synthesize, not the buffer length in bytes. See here for an example:
https://github.com/libsdl-org/SDL/blob/be991239d9bc6df06b0ca7a9ae9dbb7251e93c12/examples/audio/02-simple-playback-callback/simple-playback-callback.c#L25-L34

andyvand: I have replaced the malloc with a static buffer

CMakeLists.txt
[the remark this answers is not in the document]
andyvand: I have moved this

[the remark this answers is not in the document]
andyvand: I have fixed this also.
fluidsynth binary itself also needs SDL3_SUPPORT and target_link_libraries.
I have also added SDL3::Headers includes.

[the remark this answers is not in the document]
FtZPetruska: Linking to `SDL3::Headers` is not necessary if you're already linking to `SDL3::SDL3`.

In other words, `SDL3::Headers` is automatically added to the target when you link to `SDL3::SDL3` since it is part of the latter's `INTERFACE_LINK_LIBRARIES`.

src/drivers/fluid_sdl3.c
FtZPetruska: Since it's a static buffer, you may remove the check for NULL in the same function since it generates a warning:
```c
if (buffer == NULL)
{
  return;
}
```
The actual warning:
```c
fluidsynth/src/drivers/fluid_sdl3.c:165:13: warning: comparison of array 'buffer' equal to a null pointer is always false [-Wtautological-pointer-compare]
if (buffer == NULL)
    ^~~~~~    ~~~~
```

On the topic of warnings, the following call also triggers a `-Wsign-compare` warning on clang:
```c
fluidsynth/src/drivers/fluid_sdl3.c:163:19: warning: comparison of integers of different signs: 'int' and 'unsigned long' [-Wsign-compare]
buf_len = SDL_min(add_len, SDL_arraysize(buffer));
```

andyvand: Fixed

CMakeLists.txt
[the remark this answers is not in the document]
andyvand: Removed SDL3::Headers

andyvand:
It is working!
I have done a small test on the built fluidsynth binary.
It plays using the SDL3 audio driver everything beautifully.

`fluidsynth /Users/andyvand/Downloads/fluidsynth/sf2/VintageDreamsWaves-v2.sf2 /Users/andyvand/Downloads/opentyrian2000/extras/music40.mid -a sdl3
FluidSynth runtime version 2.4.3

fluidsynth: warning: Sample 'SineWave': ROM sample ignored
Type 'help' for help topics.
`

src/drivers/fluid_sdl3.c
derselbst: Sry for nit-picking again. But you cannot use a buffer with a constant size. The size of the buffer must depend on `audio.period-size` that you've read in `new_fluid_sdl3_audio_driver`, otherwise there's no point in reading that setting. I suppose allocating a buffer size of `period-size * frame_size` would be reasonable.

andyvand: Buffer got updated to this.
I've added period_size to the dev structure.

andyvand: This gives an issue with MinGW and older VS compilers.

andyvand: Reverted to malloc since the buffer can't be preallocated.

derselbst:
Ok, thanks so far. I'll merge this to a feature branch and will take care of the correct buffer allocation and fixing the CI build.

PR #1490, opened 2025-02-16
Haiku port

mmuman:
It seems we had patches getting dust. I rebased them and dropped stuff that didn't seem needed anymore.
All tests pass though I haven't tried the audio or MIDI outputs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ option ( enable-dsound "compile DirectSound support (if it is available)" on )
 option ( enable-wasapi "compile Windows WASAPI support (if it is available)" on )
 option ( enable-waveout "compile Windows WaveOut support (if it is available)" on )
 option ( enable-winmidi "compile Windows MIDI support (if it is available)" on )
+option ( enable-haiku "compile Haiku MediaKit audio support (if it is available)" on )
 option ( enable-sdl2 "compile SDL2 audio support (if it is available)" off )
 option ( enable-pulseaudio "compile PulseAudio support (if it is available)" on )
 option ( enable-pipewire "compile PipeWire support (if it is available)" on )
@@ -186,8 +187,10 @@ endif ()
 include ( CMakePrintHelpers ) # for cmake_print_properties() and cmake_print_variables()
 include ( TestInline )
 include ( TestVLA )
+if ( NOT HAIKU )
 include ( TestBigEndian )
 test_big_endian ( WORDS_BIGENDIAN )
+endif ( NOT HAIKU )
 
 unset ( LIBFLUID_CPPFLAGS CACHE )
 unset ( LIBFLUID_LIBS CACHE )
@@ -248,6 +251,11 @@ if ( CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_C
     endif ()
   endif (CMAKE_C_COMPILER_ID STREQUAL "Intel" )
 endif ( CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID STREQUAL "Intel" )
+
+# Haiku
+if ( HAIKU )
+set ( CMAKE_CXX_FLAGS "-fpermissive")
+endif ( HAIKU )
 
 # Windows
 unset ( WINDOWS_LIBS CACHE )
@@ -394,6 +402,12 @@ if ( CMAKE_SYSTEM MATCHES "SunOS" )
   set ( FLUID_LIBS "${FLUID_LIBS};nsl;socket" )
   set ( LIBFLUID_LIBS "${LIBFLUID_LIBS};nsl;socket" )
 endif ( CMAKE_SYSTEM MATCHES "SunOS" )
+
+# Haiku
+if ( CMAKE_SYSTEM MATCHES "Haiku" )
+  set ( FLUID_LIBS "${FLUID_LIBS};network;bsd;be;media" )
+  set ( LIBFLUID_LIBS "${LIBFLUID_LIBS};network;bsd;be;media" )
+endif ( CMAKE_SYSTEM MATCHES "Haiku" )
 
 # Apple Mac OSX
 unset ( COREAUDIO_SUPPORT CACHE )
@@ -707,6 +721,11 @@ if ( enable-readline )
     endif ()
   endif ( Readline_FOUND )
 endif ( enable-readline )
+
+unset ( HAIKU_SUPPORT CACHE )
+if ( enable-haiku )
+  set ( HAIKU_SUPPORT 1 )
+endif ( enable-haiku )
 
 unset ( AUFILE_SUPPORT CACHE )
 if ( enable-aufile )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,7 @@ option ( enable-wasapi "compile Windows WASAPI support (if it is available)" on 
 option ( enable-waveout "compile Windows WaveOut support (if it is available)" on )
 option ( enable-winmidi "compile Windows MIDI support (if it is available)" on )
 option ( enable-haiku "compile Haiku MediaKit audio support (if it is available)" on )
+option ( enable-midikit2 "compile Haiku MidiKit2 MIDI support (if it is available)" on )
 option ( enable-sdl2 "compile SDL2 audio support (if it is available)" off )
 option ( enable-pulseaudio "compile PulseAudio support (if it is available)" on )
 option ( enable-pipewire "compile PipeWire support (if it is available)" on )
@@ -407,6 +408,9 @@ endif ( CMAKE_SYSTEM MATCHES "SunOS" )
 if ( CMAKE_SYSTEM MATCHES "Haiku" )
   set ( FLUID_LIBS "${FLUID_LIBS};network;bsd;be;media" )
   set ( LIBFLUID_LIBS "${LIBFLUID_LIBS};network;bsd;be;media" )
+  if ( enable-midikit2 )
+    set( LIBFLUID_LIBS "midi2" ${LIBFLUID_LIBS} )
+  endif ( enable-midikit2 )
 endif ( CMAKE_SYSTEM MATCHES "Haiku" )
 
 # Apple Mac OSX
@@ -723,9 +727,14 @@ if ( enable-readline )
 endif ( enable-readline )
 
 unset ( HAIKU_SUPPORT CACHE )
+unset ( MIDIKIT2_SUPPORT CACHE )
 if ( enable-haiku )
   set ( HAIKU_SUPPORT 1 )
 endif ( enable-haiku )
+
+if ( enable-midikit2 )
+  set ( MIDIKIT2_SUPPORT 1 )
+endif ( enable-midikit2 )
 
 unset ( AUFILE_SUPPORT CACHE )
 if ( enable-aufile )
--- a/cmake_admin/report.cmake
+++ b/cmake_admin/report.cmake
@@ -154,6 +154,12 @@ else ( DBUS_SUPPORT )
   set ( MISC_REPORT "${MISC_REPORT}  D-Bus:                 no\n" )
 endif ( DBUS_SUPPORT )
 
+if ( HAIKU_SUPPORT )
+  message ( "Haiku MediaKit support: yes" )
+else ( HAIKU_SUPPORT )
+  message ( "Haiku MediaKit support: no" )
+endif ( HAIKU_SUPPORT )
+
 if ( LADSPA_SUPPORT )
   set ( MISC_REPORT "${MISC_REPORT}  LADSPA support:        yes\n" )
 else ( LADSPA_SUPPORT )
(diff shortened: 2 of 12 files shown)


mmuman:
Ugh, seems they missed the proper Haiku check :shrug:

mmuman:
That should fix it.

mmuman:
For some reason tests 25 & 27 fail now… !?

mmuman:
Oh, yeah… nothing `LANC=C LC_ALL=C` can't fix.

CMakeLists.txt
```diff
@@ -186,8 +188,10 @@ endif ()
 include ( CMakePrintHelpers ) # for cmake_print_properties() and cmake_print_variables()
 include ( TestInline )
 include ( TestVLA )
+if ( NOT HAIKU )
 include ( TestBigEndian )
 test_big_endian ( WORDS_BIGENDIAN )
+endif ( NOT HAIKU )
```
derselbst, line 194: Why is that if-ed out?

CMakeLists.txt
```diff
@@ -249,6 +253,11 @@ if ( CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_C
   endif (CMAKE_C_COMPILER_ID STREQUAL "Intel" )
 endif ( CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID STREQUAL "Intel" )
 
+# Haiku
+if ( HAIKU )
+set ( CMAKE_CXX_FLAGS "-fpermissive")
+endif ( HAIKU )
```
derselbst, line 259: Why is that needed?

cmake_admin/report.cmake
```diff
@@ -154,6 +166,12 @@ else ( DBUS_SUPPORT )
   set ( MISC_REPORT "${MISC_REPORT}  D-Bus:                 no\n" )
 endif ( DBUS_SUPPORT )
 
+if ( HAIKU_SUPPORT )
+  message ( "Haiku MediaKit support: yes" )
+else ( HAIKU_SUPPORT )
+  message ( "Haiku MediaKit support: no" )
+endif ( HAIKU_SUPPORT )
```
derselbst, line 173: This is redundant to above and should be removed.

CMakeLists.txt
```diff
@@ -94,6 +94,8 @@ option ( enable-dsound "compile DirectSound support (if it is available)" on )
 option ( enable-wasapi "compile Windows WASAPI support (if it is available)" on )
 option ( enable-waveout "compile Windows WaveOut support (if it is available)" on )
 option ( enable-winmidi "compile Windows MIDI support (if it is available)" on )
+option ( enable-haiku "compile Haiku MediaKit audio support (if it is available)" on )
```
derselbst, line 97:
```suggestion
option ( enable-mediakit "compile Haiku MediaKit audio support (if it is available)" on )
```

IMO, the naming is unfortunate. `enable-haiku` would enable the audio driver and not support for Haiku in general. I'd suggest to rename it `mediakit`.

CMakeLists.txt
```diff
@@ -94,6 +94,8 @@ option ( enable-dsound "compile DirectSound support (if it is available)" on )
...
+option ( enable-midikit2 "compile Haiku MidiKit2 MIDI support (if it is available)" on )
```
derselbst, line 98: Both of these cmake options should be wraped inside a `if ( CMAKE_SYSTEM MATCHES "Haiku" )`, as we do it below, since they are platform-specific options.

cmake_admin/report.cmake
```diff
@@ -26,6 +26,18 @@ else ( DSOUND_SUPPORT )
     set ( AUDIO_MIDI_REPORT "${AUDIO_MIDI_REPORT}  DSound:                no\n" )
 endif ( DSOUND_SUPPORT )
 
+if ( HAIKU_SUPPORT )
+    set ( AUDIO_MIDI_REPORT "${AUDIO_MIDI_REPORT}  Haiku:                 yes\n" )
+else ( HAIKU_SUPPORT )
+    set ( AUDIO_MIDI_REPORT "${AUDIO_MIDI_REPORT}  Haiku:                 no\n" )
```
derselbst, line 32:
```suggestion
    set ( AUDIO_MIDI_REPORT "${AUDIO_MIDI_REPORT}  Haiku Mediakit:        yes\n" )
else ( HAIKU_SUPPORT )
    set ( AUDIO_MIDI_REPORT "${AUDIO_MIDI_REPORT}  Haiku Mediakit:        no\n" )
```

src/drivers/fluid_haiku.cpp
```diff
@@ -0,0 +1,137 @@
...
+static void playerProc(void *cookie, void *buffer, size_t len, const media_raw_audio_format &format)
+{
+    fluid_haiku_audio_driver_t *dev = (fluid_haiku_audio_driver_t *)cookie;
+    len /= dev->frame_size;
+    dev->write_ptr(dev->synth, len, buffer, 0, 2, buffer, 1, 2);
+}
+
+void fluid_haiku_audio_driver_settings(fluid_settings_t *settings)
+{
+    fluid_settings_register_str(settings, "audio.haiku.device", "default", 0);
```
derselbst, line 54: `audio.haiku.device` should be documented in `doc/fluidsettings.xml`

src/drivers/fluid_haiku.cpp
```diff
@@ -0,0 +1,137 @@
...
+	dev->m_player->Start();
+
+	return (fluid_audio_driver_t *) dev;
+}
+
+
+void delete_fluid_haiku_audio_driver(fluid_audio_driver_t *d)
+{
+    fluid_haiku_audio_driver_t *dev = (fluid_haiku_audio_driver_t *) d;
+	if (dev->m_player != NULL) {
```
derselbst, line 130: Potential nullptr deref if `d` is NULL.

src/drivers/fluid_haiku.cpp
```diff
@@ -0,0 +1,137 @@
...
+    fluid_haiku_audio_driver_t *dev = NULL;
+    fluid_audio_callback_t write_ptr;
+    
+    double sample_rate;
+    int period_size;
+    int periods;
+    int sample_size;
+    
+    char *device;
+    const char *dev_name;
```
derselbst, line 70: Looks like these two variables were intended to be used with `audio.haiku.device`, but they are unused. If so, pls. remove them, and then also remove the `audio.haiku.device` setting.

src/drivers/fluid_midikit2.cpp
```diff
@@ -0,0 +1,161 @@
...
+            }
+        }
+    }
+public:
+    fluid_midikit2_driver_t* callback;
+};
+
+void fluid_midikit2_driver_settings(fluid_settings_t *settings)
+{
+    fluid_settings_register_str(settings, "midi.midikit2.id", "pid", 0);
```
derselbst, line 68: `midi.midikit2.id` should be documented in `doc/fluidsettings.xml`

src/drivers/fluid_midikit2.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+static void fluid_coremidi_autoconnect(fluid_midikit2_driver_t *dev)
```
derselbst, line 71: Wrong function name.

```suggestion
static void fluid_midikit2_autoconnect(fluid_midikit2_driver_t *dev)
```

src/drivers/fluid_midikit2.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+fluid_midi_driver_t*
+new_fluid_midikit2_driver(fluid_settings_t *settings, handle_midi_event_func_t handler, void *data)
+{
+    fluid_midikit2_driver_t* driver = NULL;
+    char* id = NULL;
+    char clientid[128];
+
+    memset(clientid, 0, sizeof(clientid));
```
derselbst, line 89:
```suggestion
    FLUID_MEMSET(clientid, 0, sizeof(clientid));
```

src/drivers/fluid_midikit2.cpp
```diff
@@ -0,0 +1,161 @@
...
+    else strncpy(clientid, "FluidSynth virtual port", sizeof(clientid));
+
+    driver->m_consumer = new fluid_midikit2_input;
+    if(!driver->m_consumer->IsValid())
+    {
+        FLUID_LOG(FLUID_ERR, "Failed to create the MIDI consumer");
+        driver->m_consumer->Release();
+        delete_fluid_midi_parser(driver->parser);
+        FLUID_FREE(driver);
+        return NULL;
```
derselbst, line 133: You're leaking `driver->m_consumer` here. Pls. avoid these manual partial cleanup logic and instead, properly do the cleanup in `delete_fluid_midikit2_driver`

src/drivers/fluid_midikit2.cpp
```diff
@@ -0,0 +1,161 @@
...
+delete_fluid_midikit2_driver(fluid_midi_driver_t *p)
+{
+    fluid_return_if_fail(p != NULL);
+    if (p)
+    {
+        fluid_midikit2_driver_t* driver = (fluid_midikit2_driver_t*)p;
+        driver->m_consumer->Unregister();
+        driver->m_consumer->Release();
+        delete_fluid_midi_parser(driver->parser);
+        FLUID_FREE(driver);
```
derselbst, line 159: You're leaking `driver->m_consumer`.

src/synth/fluid_synth.c
```diff
@@ -218,8 +222,25 @@ void fluid_synth_settings(fluid_settings_t *settings)
...
+			FILE *inFile = fopen(midiSettings, "rt");
+			if (inFile) {
+				char sf2file[PATH_MAX];
+				if (fscanf(inFile, "# Midi\n\tsoundfont \"%[^\"]\"", sf2file)) {
+					fluid_settings_register_str(settings, "synth.default-soundfont", sf2file, 0);
+				}
+				fclose(inFile);
+			}
+		}
+	}
```
derselbst, line 239: What does that do??

derselbst (changes requested):
Thanks for submitting. However, this is not yet in a shape that is ready to accept. It needs some work. Comments below. Also, it would be very helpful to have a Haiku CI build in place, to avoid breaking stuff in the future.

src/drivers/fluid_midikit2.cpp
mmuman: Release() will delete it by itself as [it's ref-counted](https://www.haiku-os.org/docs/api/group__midi2.html#book_midi2refcount).

src/synth/fluid_synth.c
mmuman: Seemingly fetches the first path to the sound fonts [from the Media preferences](https://www.haiku-os.org/docs/userguide/en/preferences/media.html#midi).

mmuman:
> Thanks for submitting. However, this is not yet in a shape that is ready to accept. It needs some work. Comments below. 

Well I didn't write most of this code, just trying to upstream before it bit-rots, so I'll have to figure it out as well.

> Also, it would be very helpful to have a Haiku CI build in place, to avoid breaking stuff in the future.
> Edit: Here's an example of how SDL coped with Haiku CI: [libsdl-org/SDL#6739](https://github.com/libsdl-org/SDL/pull/6739)

Yes, though our cross-compiler is only built as part of our build system, but there are a few solutions now.

src/drivers/fluid_midikit2.cpp
mmuman: Fixed.

src/drivers/fluid_midikit2.cpp
mmuman: Fixed.

src/drivers/fluid_haiku.cpp
mmuman: Seems like some unimplemented stuff yeah.

mmuman:
poke @Cacodemon345 @threedeyes :-)

CMakeLists.txt
Cacodemon345: IIRC it was needed because the compiler would start spitting out errors for what was seemingly valid C++ code for non-permissive builds.

CMakeLists.txt
Cacodemon345: The macros were broken on Haiku I believe.

src/drivers/fluid_midikit2.cpp
derselbst: Meanwhile I discovered the implementation of BMidiEndpoint: https://github.com/haiku/haiku/blob/c8252422b8b3a2a293fb3be039eec6f87b30b467/src/kits/midi2/MidiEndpoint.cpp#L118-L135

So it turns out that they do make a hard call to `delete this`. If you used `new[]` to allocate the endpoint this will cause undefined behavior. And if you stack allocated this you might get a log message: https://github.com/haiku/haiku/blob/c8252422b8b3a2a293fb3be039eec6f87b30b467/src/kits/midi2/MidiEndpoint.cpp#L258-L260

Poor implementation, IMO. But still: You must declare an empty destructor for `fluid_midikit2_input` and mark it `virtual` to make this cleanup work as intended.

derselbst:
Understood. Yet, I cannot keep this open forever. As said, feel free to re-file a PR.

PR #1491, opened 2025-02-25
Fix errors when switching user

andrew-sayers:
Fluidsynth includes a systemd user service, which listens on TCP port 9800 by default.  So if a computer starts a second user session, that session's service immediately fails because port 9800 is already in use.  Lightdm implements its login prompt as a user session, so this can happen just by misclicking the "switch user" button.  I assume other display managers are the same, but haven't checked.

This PR is a set of loosely-connected suggestions that would have helped as I went down this particular rabbit hole.  I figure they're easier to review together, but happy to edit/split out/remove anything that doesn't fit.

### DIFF
--- a/fluidsynth.service.in
+++ b/fluidsynth.service.in
@@ -3,6 +3,8 @@ Description=FluidSynth Daemon
 Documentation=man:fluidsynth(1)
 After=sound.target
 After=pipewire.service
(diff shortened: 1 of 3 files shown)


derselbst (approved):
LGTM, thanks!

PR #1494, opened 2025-03-01
Fix interruptions of real-time playback

### LINKED ISSUE #1481: High CPU usage may impair real-time playback
### FluidSynth version
2.4.3

### Describe the bug
When playing Altitude.mid in Fluidsynth 2.4.3 with the default audio driver on some Windows 10 systems it will crackle at various points throughout playback of the MIDI.

### Expected behavior
When playing Altitude.mid in Fluidsynth 2.4.3 with the default audio driver on some Windows 10 systems Fluidsynth should not crackle throughout playback of the MIDI.

### Steps to reproduce
Download Altitude.mid below, open Fluidsynth 2.4.3, load GeneralUser-GS v1.47x.sf2 or GeneralUser-GS.sf2, play Altitude.mid and you may hear it crackle on some Windows 10 systems. 

### Additional context
It will also crackle with waveout driver, although it is a bit less, also this issue first appeared in FluidSynth 2.4.1 and continued in 2.4.2. AFAIK, cannot hear any crackling on FluidSynth 2.4.0. And turning on/off chorus or reverb has no effect on the crackling. One way to lower most crackling is to set audio buffers at 64, but no need for this in 2.4.0

[ALTITUDE.zip](https://github.com/user-attachments/files/18627757/ALTITUDE.zip)

derselbst:
PR #1432 introduced an on-the-fly recalculation of the filter coefficients, whenever the filter parameters change. It turns out this consumes too much CPU for voices that heavily depend on e.g. modulating the IIR filter.

This PR addresses this problem by avoiding direct calls to `sin()` and `cos()` and instead using a lookup table for calculating the filter coefficients.

TODO:
- [x] Eliminate the dependency on the sample rate in `fluid_iir_filter_init_table()`

Fixes #1481


src/rvoice/fluid_iir_filter.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    LOG_FILTER("%f + %f = %f cents = %f Hz | Q: %f", iir_filter->fres, fres_mod, iir_filter->fres + fres_mod, fres, iir_filter->last_q);
+    
+    fres = fluid_hz2ct(fres);
+    
+    /* if filter enabled and there is a significant frequency change.. */
+    fres_diff = fres - iir_filter->last_fres;
+    if(iir_filter->filter_startup)
+    {
+        // The filer was just starting up, make sure to calculate initial coefficients for the initial Q value, even though the fres may not have changed
```
spessasus, line 407: typo: it should be "filter" instead of "filer

derselbst: Thanks!

PR #1501, opened 2025-03-14
Apply *_FOR_BUILD environment variables to gentables

fabiangreffrath:
This is the last attempt (I promise!) to address this unfortunate topic. This patch has been applied to the fluidsynth_2.4.3+dfsg-3 Debian package since February 25th and has proved doing what it's supposed to do:

https://salsa.debian.org/multimedia-team/fluidsynth/-/jobs/7156765

It has also proved to work well in a cross building environment:

http://crossqa.debian.net/src/fluidsynth

### DIFF
--- a/src/gentables/CMakeLists.txt
+++ b/src/gentables/CMakeLists.txt
@@ -2,13 +2,13 @@ cmake_minimum_required(VERSION 3.5)
 
 # remove $CC from the current environment and by that force cmake to look for a (working) C compiler,
 # which hopefully will be the host compiler
-unset(ENV{CC})
+set(ENV{CC} "$ENV{CC_FOR_BUILD}")
 
 # also unset $CFLAGS to avoid passing any cross compilation flags to the host compiler
-unset(ENV{CFLAGS})
+set(ENV{CFLAGS} "$ENV{CFLAGS_FOR_BUILD} $ENV{CPPFLAGS_FOR_BUILD}")
 
 # linker flags as well
-unset(ENV{LDFLAGS})
+set(ENV{LDFLAGS} "$ENV{LDFLAGS_FOR_BUILD}")
 
 project (gentables C)
 
--- a/src/gentables/CMakeLists.txt
+++ b/src/gentables/CMakeLists.txt
@@ -1,13 +1,13 @@
 cmake_minimum_required(VERSION 3.5)
 
-# remove $CC from the current environment and by that force cmake to look for a (working) C compiler,
-# which hopefully will be the host compiler
+# Set the CC, CFLAGS and LDFLAGS variables to the corresponding *_FOR_BUILD environment variables.
+# If the CC_FOR_BUILD environment variable is unset or empty, this will unset the CC CMake variable
+# and by that force CMake to look for a (working) C compiler, which hopefully will be the host compiler.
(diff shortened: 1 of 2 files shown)


derselbst (commented):
Since you're very persistent about this one, I would be willing to buy this proposal. Yet, I don't like it, because introduces a second way to specify the host compiler, next to already existing `FLUID_HOST_COMPILER`. This is an inconsistent Mischmasch. Pls. decide for one approach and show some love to the comments, too.

(Superseding `FLUID_HOST_COMPILER` with this approach is a valid option.)

src/gentables/CMakeLists.txt
```diff
@@ -2,13 +2,13 @@ cmake_minimum_required(VERSION 3.5)
 
 # remove $CC from the current environment and by that force cmake to look for a (working) C compiler,
 # which hopefully will be the host compiler
-unset(ENV{CC})
+set(ENV{CC} "$ENV{CC_FOR_BUILD}")
 
 # also unset $CFLAGS to avoid passing any cross compilation flags to the host compiler
-unset(ENV{CFLAGS})
+set(ENV{CFLAGS} "$ENV{CFLAGS_FOR_BUILD} $ENV{CPPFLAGS_FOR_BUILD}")
```
derselbst, line 8: How about combining them to one env var?

src/gentables/CMakeLists.txt
```diff
@@ -2,13 +2,13 @@ cmake_minimum_required(VERSION 3.5)
 
 # remove $CC from the current environment and by that force cmake to look for a (working) C compiler,
 # which hopefully will be the host compiler
-unset(ENV{CC})
+set(ENV{CC} "$ENV{CC_FOR_BUILD}")
```
derselbst, line 5: Naming it `HOST_CC` or `FLUID_HOST_CC` makes it clearer. Likewise for the other vars below.

src/gentables/CMakeLists.txt
fabiangreffrath: CMake is infamous for ignoring the CPPFLAGS variable. However, some cross building environments assign separate parameters to CFLAGS and CPPFLAGS, because that is natively supported by Autotools. I'd like to assign both to CFLAGS at once and be explicit about this.

src/gentables/CMakeLists.txt
fabiangreffrath: Sorry, but the names are already taken. I haven't chosen them myself and I would have chosen them differently. But the *_FOR_BUILD suffices are established since literally decades, at least in the Autotools ecosystem.

https://www.sourceware.org/autobook/autobook/autobook_143.html#Supporting-Cross-Compiler-in-Make
https://www.gnu.org/software/autoconf-archive/ax_prog_cc_for_build.html
https://manpages.debian.org/bookworm/autotools-dev/config.guess.1.en.html
https://stackoverflow.com/a/40954603

derselbst: I see. I'm no autotools guy, so thanks for the pointer.

derselbst (approved):
LTGM. For the record, here's the motivation for this change: https://github.com/FluidSynth/fluidsynth/pull/1479#issuecomment-2628917797

fabiangreffrath:
Thank you, your patience is appreciated. 😽

Issue #1473, opened 2025-01-21
filter cutoff and Q NRPNs don't behave correctly


mrbumpy409:
### FluidSynth version
2.4.2

### Details
Control of filter cutoff and resonance (Q) via NRPNs was introduced with the AWE32 and continued to be supported in all later Sound Blaster cards containing a hardware SoundFont synth, as far as I am aware. Recent versions of FluidSynth now support these NRPNs, but its behavior is not correct when compared to my Sound Blaster Audigy2 ZS, and likely not accurate to the original AWE32 either.

As for defining the "correct" behavior of these NRPNs, one must decide which of Creative's sound cards to model. I believe FluidSynth should model the behavior of the later, SoundFont 2.01–2.04 devices such as the Audigy for the following reasons:
1. FluidSynth's SoundFont implementation is already based on the 2.01 and 2.04 spec, and the Audigy's NRPN implementation is aligned with the behavioral characteristics of this spec.
2. I have a Sound Blaster Audigy2 ZS that can be used for testing and comparing the behavior with FluidSynth, but no access to an AWE32/64 or Live!.
3. The Audigy's NRPN behavior can be recreated easily using modulators, so no need for lookup tables, etc. of the AWE32.

### Test Results
I have created a [filter NRPN test](https://github.com/mrbumpy409/SoundFont-Spec-Test/tree/main/NRPN%20test%20-%20filter). It contains a `recordings` folder in which you can compare recordings of the test run on FluidSynth 2.4.2 and Audigy2 ZS. Through this, we can see several differences between the two synthesizers.

I will list each test from the accompanying README, along with my notes on the results of both FluidSynth and Audigy2 ZS. The tests are arranged into pairs, with the first of each pair testing NRPN behavior and the second emulating that behavior using a modulator.

#### Tests #&NoBreak;1 & #&NoBreak;2
**Test #&NoBreak;1:** If the AWE32 FC NRPN is implemented, you should hear a filter sweep from 100 to 8000 Hz and then back down to 100 Hz.
**Test #&NoBreak;2:** You will hear a filter sweep that emulates the AWE32 FC NRPN filter curve via CC1 modulator.

Here is a spectrogram view of tests #&NoBreak;1 (left) and #&NoBreak;2 (right) between Audigy2 (top) and FluidSynth (bottom):
![Image](https://github.com/user-attachments/assets/9cf42da4-1ef4-4a52-8fcd-11930eb0cef0)

[...]

Issue #1452, opened 2024-12-19
Use of AWE32 filter cutoff NRPN overrides both cutoff and emphasis


mrbumpy409:
### FluidSynth version
2.4.1 (1427 branch)

### Describe the bug
FluidSynth now supports AWE32 NRPN control of both filter cutoff (FC) and emphasis (Q). The use of the FC NRPN should override the preset's FC, and the use of the Q NRPN should override the preset's Q. However, using the FC NRPN overrides both FC and Q (which ends up being set to 0, as far as I can tell).

### Expected behavior
Using the FC NRPN should only override the preset's FC and not the Q.

### Steps to reproduce
This bug can be reproduced pretty easily by running the test linked in [this comment](https://github.com/FluidSynth/fluidsynth/pull/1434#issuecomment-2555507683). A spectrogram view of tests 1 and 3 both show the presets' Q values being overridden, as these two tests make use of the FC NRPN.

derselbst:
Right, the reason I was implementing it this way, is because of how the "NRPN LSB 22 (Initial Filter Resonance Coefficient)" is documented:

* [SB AWE32 Developer's Information Pack](https://github.com/user-attachments/files/15757220/adip301.pdf)
* [Frequently Asked Questions for SB AWE32](http://archive.gamedev.net/archive/reference/articles/article445.html)

There, they specify a coefficient table with a high and low Q, and a high and low filter fc respectively. Therefore, I assumed that the NRPN implementation of the filter uses a varying filter Q, i.e. when setting the filter fc to its lower bound, you'll get the low Q. If you set fc to its upper bound, you'll get a high filter Q. And any fc in between just linearly interpolates the Q within its given range.

I didn't had time so far to look into your provided test. But I'll probably find time to do so over Christmas. In addition, you may try to add another test case to it which sets the NRPN LSB 22 (i.e. initializes the Filter Resonance Coefficient), and then vary the filter fc via NRPN LSB 21. I would assume that this also has an effect on the filter's emphasis.

mrbumpy409:
At least on the Audigy, the filter when modulated by NRPN LSB 21 acts the same as if it were modulated by any other source (via modulator, etc.), with no affect on the filter Q. You can see this in the test results on the other thread. With FluidSynth aiming to be a SoundFont 2.01/2.04-compliant synthesizer, the Audigy1/2 behavior would be the ideal model to emulate, IMO. When I get some time, I'll update the test to include filter Q tests.

[...]

PR #1516, opened 2025-03-22
More accurate AWE32 NRPN filter implementation

derselbst:
This PR adresses the issues found in #1452 and #1473


klerg:
Can an option, setting, or switch be put in Fluidsynth to set the filter Q ?

derselbst:
You mean the maximum Q? This would complicate things from a user perspective. I prefer to keep things simple, but functioning well in most cases. And with 22dB we have reached a quite good compromise, IMO.

klerg:
Not for a maximum Q.  My hope is this will keep things simple. I see, yes it  is the same way for me as well.  I only ask for a [synth.filter.Q](https://www.fluidsynth.org/api/fluidsettings.xml) setting, so the Q can be set to 20, 22, 24 or whatever is best for the music, does that make sense ?

derselbst:
You _are_ asking for a maximum Q setting. Q is not hardcoded to 22dB, the value sent via AWE32 NRPN decides how high the Q will be. However, Q's maximum value is currently set to 22dB, since we previously chose this value as compromise.

klerg:
I’m asking to set default filter Q to any value. I see that is good to know, yes it makes sense for AWE32 NRPN to put the value of Q then. Sure, can a setting such as [synth.filter.Q](https://www.fluidsynth.org/api/fluidsettings.xml) be added to change the Q from 22 dB to something else ?

derselbst:
No, sry.

klerg:
Ok, why not ?

derselbst:
Pls. see my initial comment. I'm not convinced that this would add any benefit. It would rather unnecessarily complicate things. We have agreed on  22 dB as maximum. +-2dB difference is very subtle anyway.

klerg:
Yes, I did see that comment. Sure, what will it take to convince you it adds a benefit ? Well, no one needs to use this setting. True, it is very hard to hear a difference from 24 or 22 dB, but why did not go > 24 or < 20 dB for filter Q ?

PR #1519, opened 2025-03-23
Implement mapping of GS NRPNs to CC

### LINKED ISSUE #1420: Map GS NPRN Params to CC numbers
### Discussed in https://github.com/FluidSynth/fluidsynth/discussions/1406

<div type='discussions-op-text'>

<sup>Originally posted by **spessasus** October 13, 2024</sup>
## The problem
Currently, there's no way to make fluidsynth react to GS NPRNs.

For example, take a listen to [this MIDI file being played on a GS device](https://youtu.be/0jsOL5Og-0Q). At around 18 seconds in, the trumpets can be heard with clear vibrato. Playing [it](https://github.com/AyHa1810/touhou-midi-collection/blob/main/1%20-%20Shooter%20Games/7%20-%20Perfect%20Cherry%20Blossom/14.%20Border%20of%20Life%20(ZUN).mid) on fluidsynth even with stgiga's HiDef (the SC-88Pro soundfont) does not result in any vibrato.

And that's fine, fluidsynth is not said to be GS compatible.

But, even as a soundfont creator, I *can't* make it GS compatible with modulators. You see, modulators only support CCs as sources, not NPRNs. This is why I propose the following:

## The solution
In this [SC8850 owner's manual](https://cdn.roland.com/assets/media/pdf/SC-8850_OM.pdf#page=227&zoom=133,-64,835) you can see the CC numbers on page 227 and the correspoding NPRNs on page 228.

It would be great if fluidsynth *mapped* those NPRNs to their respective controllers. That is:
- NPRN 01H 08H to CC 76 (vibrato rate)
- NPRN 01H 09H to CC 77 (vibrato depth)
- NPRN 01H 0AH to CC 78 (vibrato delay)

`note: numbers with "H" at the end indicate hex`

And by mapping i mean directly copying the data entry value to the CC:
- received an NRPN LSB: 01H
- received an NRPN MSB: 08H (this is vibrato rate)
- received a data entry MSB: 10H: this is effectively interpreted as CC change: CC 76 is set to 10H

**Note: this would only apply if the synth is in the `gs` bank select mode (I.E. after receiving a GS sysex)

## Rationale
- does not mess up other systems (the change only applies after a GS sysEx)
- does not add any default modulators or anything like that
- allows the soundfont creators to implement support for the said GS NRPNs
- does not break any existing soundfonts

What do you think?</div>

derselbst agreed to give it a try, hence I'm opening an issue here.

derselbst:
This implements the mapping of vibrato-related Roland GS NRPNs to their corresponding CC parameters.

Resolves #1420

- [X] TODO: before merging, document this behavior in wiki (done [here](https://github.com/FluidSynth/fluidsynth/wiki/FluidFeatures#nrpn-control-change-implementation-chart))

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -1870,6 +1870,44 @@ fluid_synth_cc_LOCAL(fluid_synth_t *synth, int channum, int num)
                     }
                 }
             }
+            else if(fluid_channel_get_cc(chan, NRPN_MSB) == 1 && synth->bank_select == FLUID_BANK_STYLE_GS)
+            {
+                int nrpn2cc = -1;
+                int nrpn_lsb = fluid_channel_get_cc(chan, NRPN_LSB);
+                // cf. SC8850 owner's manual pages 227 + 228
+                switch(nrpn_lsb)
+                {
+                case 8:
+                    // vibrato rate
+                    nrpn2cc = SOUND_CTRL7;
+                    break;
+                case 9:
+                    // vibrato depth
+                    nrpn2cc = SOUND_CTRL8;
+                    break;
+                case 10:
+                    // vibrato rate
+                    nrpn2cc = SOUND_CTRL9;
+                    break;
+                default:
+                    break;
+                }
+                if(nrpn2cc != -1)
+                {
+                    if(synth->verbose)
+                    {
+                        FLUID_LOG(FLUID_INFO, "Translating Roland GS NRPN %d to CC %d", nrpn_lsb, nrpn2cc);
+                    }
+                    fluid_synth_cc(synth, channum, nrpn2cc, msb_value);
+                }
+                else
+                {
+                    if(synth->verbose)
+                    {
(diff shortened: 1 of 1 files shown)


src/synth/fluid_synth.c
```diff
@@ -1870,6 +1870,44 @@ fluid_synth_cc_LOCAL(fluid_synth_t *synth, int channum, int num)
...
                 }
             }
+            else if(fluid_channel_get_cc(chan, NRPN_MSB) == 1 && synth->bank_select == FLUID_BANK_STYLE_GS)
+            {
+                int nrpn2cc = -1;
+                int nrpn_lsb = fluid_channel_get_cc(chan, NRPN_LSB);
+                // cf. SC8850 owner's manual pages 227 + 228
+                switch(nrpn_lsb)
+                {
+                case 8:
```
spessasus, line 1880: Maybe using names rather than hardcoding numbers would be better if there will be requests for more NPRNs with this functionality?

derselbst: _Iff_, there will be more such requests, I would surely consider it. But right now, I'm fine with hardcoding the numbers :)

PR #1528, opened 2025-04-04
Put fluidsynth.lock in its own directory

### LINKED ISSUE #1527: /run/lock/fluidsynth.lock  issue.
My distro upgraded to Fluidsynth 2.4.4  two weeks  ago, and since
fluidsynth no longer starts in daemon mode due to permission issues
for  /run/lock/fluidsynth.lock.
Error:
Can't  touch /run/lock/fluidsynth.lock.

Downgrade to 2.4.3 solves the issue.

andrew-sayers:
In some distributions (e.g. Fedora), /run/lock is only writable by root.

Create /run/lock/fluidsynth with appropriate permissions.

Fixes #1527.

Some issues with rolling this out:

I can't see where the `cmake` installer installs the `.service` files, so haven't added an equivalent to install `fluidsynth.tmpfiles`.  If the installer doesn't do that, this is fine.

I've added the relevant line to `contrib/fluidsynth.spec`, and [the documentation claims `%{_tmpfilesdir}` is the correct incantation](https://docs.fedoraproject.org/en-US/packaging-guidelines/Tmpfiles.d/), but I don't have a Fedora system to test on.  Should we give the Fedora maintainer a heads-up somehow?

I haven't added the relevant line to `contrib/debian.fluidsynth.install`, because Debian's version of that file has drifted out of sync.  To see their version, go to [Debian's package page](https://packages.debian.org/sid/fluidsynth), download [fluidsynth_2.4.4+dfsg-1.debian.tar.xz](http://deb.debian.org/debian/pool/main/f/fluidsynth/fluidsynth_2.4.4+dfsg-1.debian.tar.xz) and take a look at `debian/fluidsynth.install`.  The relevant line is probably `*/fluidsynth.tmpfiles usr/lib/tmpfiles.d/fluidsynth.conf` - do you want to resync these, or shall I just send them a bug report explaining they'll need to fix their version?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -917,6 +917,9 @@ if ( UNIX )
         configure_file ( fluidsynth.service.in
         ${FluidSynth_BINARY_DIR}/fluidsynth.service @ONLY )
 
+        configure_file ( fluidsynth.tmpfiles.in
+        ${FluidSynth_BINARY_DIR}/fluidsynth.tmpfiles @ONLY )
+
         configure_file ( fluidsynth.conf.in
         ${FluidSynth_BINARY_DIR}/fluidsynth.conf @ONLY )
 
--- a/contrib/fluidsynth.spec
+++ b/contrib/fluidsynth.spec
@@ -95,9 +95,10 @@ This package contains the shared library for Fluidsynth.
 
 %if 0%{?suse_version}
 
-# manually install systemd service files
+# manually install systemd files
 install -Dm 644 build/fluidsynth.conf %{buildroot}%{_fillupdir}/sysconfig.%{name}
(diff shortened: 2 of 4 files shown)


derselbst (approved):
Thanks for taking care of this!

I can confirm that the system service file isn't installed by CMake.

For RPM-based distros, this looks fine. I'll include a hint about that in the release notes of the next version, so no need to ping the maintainers, IMO.

andrew-sayers:
For the record, I've used this PR as the basis of the fix for [an equivalent issue in bluez](https://lore.kernel.org/linux-bluetooth/20250416152039.971257-1-kernel.org@pileofstuff.org/T/).  The feedback here has helped significantly over there - thanks!

Issue #1544, opened 2025-04-14
`fluidsynth --version` fails on MSVC builds on Windows X64


cclauss:
### FluidSynth version
_Execute `fluidsynth --version` and provide the output._

v2.4.3:
```
FluidSynth runtime version 2.4.3
Copyright (C) 2000-2025 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.4.3
Sample type=float
```
v2.4.4:
```
Error: Process completed with exit code 1.
```
v2.4.5: 
```
Usage: fluidsynth [options] [soundfonts]
Try -h for help.
Unknown switch '-'
```
https://github.com/cclauss/fluidsynth-tests/actions

Related to:
* #1510 
* #1543

derselbst:
The MSVC binaries are compiled without getopt support. See the notice for `fluidsynth -h`

```
FluidSynth runtime version 2.4.5
Copyright (C) 2000-2025 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

Usage:
  fluidsynth [options] [soundfonts] [midifiles]

Note:
  This version of fluidsynth was compiled without getopt support.
  Thus, the long options are not supported.
```

Therefore, this is not a bug, it's a feature request for getting getopt support for MSVC builds.

pedrolcl:
There is a [getopt-win32](https://vcpkg.io/en/package/getopt-win32.html) package in https://vcpkg.io/ but it is not a blind drop-in. It will need [some tweaking](https://github.com/microsoft/vcpkg/blob/41e267049e8edaf37193794d46d33a3582985d5e/ports/getopt-win32/usage) on the build system.

pedrolcl:
Another alternative would be [getopt_port](https://github.com/kimgr/getopt_port) (BSD licensed).

PR #1545, opened 2025-04-14
Update bug_report.md to be compliant with non-getopt builds

cclauss:
* https://github.com/FluidSynth/fluidsynth/issues/1544#issuecomment-2802115630


derselbst (approved):
Totally valid, thanks!

PR #1546, opened 2025-04-15
Fix inconsistent library naming on Windows

### LINKED ISSUE #1543: Inconsistent library naming for MSVC vs. MinGW builds
### FluidSynth version
FluidSynth runtime version 2.4.5

### Describe the bug
Dynamic DLL loading on Windows fails because the file name changed from 'libfluidsynth-3.dll' to 'fluidsynth-3.dll'.

### Expected behavior
The DLL filename should not have changed.

### Steps to reproduce
Rename 'fluidsynth-3.dll' to 'libluidsynth-3.dll'.

### Additional context
None

derselbst:
This PR reverts an unintended change introduced by #1469. It also adds an extra check to the CI build to verify the filename of the DLL.

Fixes #1543.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -323,17 +323,10 @@ elseif ( WIN32 )
   set_target_properties ( libfluidsynth
     PROPERTIES
       PUBLIC_HEADER "${public_HEADERS}"
-      ARCHIVE_OUTPUT_NAME "libfluidsynth-${LIB_VERSION_CURRENT}"
-      OUTPUT_NAME "libfluidsynth-${LIB_VERSION_CURRENT}"
-	  # Do NOT use prefix!
-	  # PREFIX applies to
-	  # import lib + static lib of MinGW, and
-	  # the DLL itself obviously.
-	  # But it does not apply to the import lib MSVC!
-	  # In order to name everything "libfluidsynth" consistently, specify an empty prefix.
-	  # See #1546.
-      PREFIX ""
-      IMPORT_PREFIX ""
+      ARCHIVE_OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
+      OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
+      PREFIX "lib"
+      IMPORT_PREFIX "lib"
       VERSION ${LIB_VERSION_INFO}
       SOVERSION ${LIB_VERSION_CURRENT}
     )
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -323,10 +323,10 @@ elseif ( WIN32 )
   set_target_properties ( libfluidsynth
     PROPERTIES
       PUBLIC_HEADER "${public_HEADERS}"
-      ARCHIVE_OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
-      OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
       PREFIX "lib"
+      OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
       IMPORT_PREFIX "lib"
+      ARCHIVE_OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
       VERSION ${LIB_VERSION_INFO}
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -323,6 +323,8 @@ elseif ( WIN32 )
   set_target_properties ( libfluidsynth
     PROPERTIES
       PUBLIC_HEADER "${public_HEADERS}"
+      ARCHIVE_OUTPUT_NAME "libfluidsynth-${LIB_VERSION_CURRENT}"
+      PREFIX "lib"
```
pedrolcl, line 327: This has the effect of naming the import library in MinGW as:

`-- Installing: D:/a/fluidsynth/fluidsynth/fluidsynth_install/lib/liblibfluidsynth-3.dll.a`

https://github.com/FluidSynth/fluidsynth/actions/runs/14477269150/job/40605816818

pedrolcl (changes requested):
The static library name in MSVC is also weird:

`libfluidsynth.vcxproj -> D:\a\1\s\build\src\Release\liblibfluidsynth-3.lib`

https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=12360&view=logs&j=9975b921-e497-50f7-062d-b73be56f9742&t=b986791a-45f4-5bb1-8a12-94a22ca1df04&l=531

src/CMakeLists.txt
derselbst: So the `PREFIX` applies to
* import lib + static lib of MinGW, and
* the DLL itself obviously.

But it does not apply to the import lib MSVC, which is [named correctly](https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=12360&view=logs&j=e765dd6c-0c81-5220-d856-9f489fdad6be&t=61587a34-96e9-564a-c9f3-8ebf5db1c48a&l=132). Very strange. I'll add more checks. The documentation for [PREFIX](https://cmake.org/cmake/help/v3.31/prop_tgt/PREFIX.html) is also pretty much non-existent. Let's see if we can avoid using it, at least on Windows.

derselbst:
Empty PREFIX and libfluidsynth for both, OUTPUT and ARCHIVE name set: MinGW still calls the import library [liblibfluidsynth-3.dll.a](https://dev.azure.com/tommbrt/tommbrt/_build/results?buildId=12372&view=logs&j=feaa8cae-2cce-5249-44f7-7ee14c3a4611&t=424b4271-db2b-5c4e-0d26-273a4dfa6e41&l=51), while everything else is correctly named, even the static lib of MinGW!

pedrolcl (commented):
For import libraries, the target property is [IMPORT_PREFIX](https://cmake.org/cmake/help/v3.13/prop_tgt/IMPORT_PREFIX.html)

There are also global variables like [CMAKE_IMPORT_LIBRARY_PREFIX](https://cmake.org/cmake/help/v3.13/variable/CMAKE_IMPORT_LIBRARY_PREFIX.html) but it is better to avoid globals.

pedrolcl (approved):

derselbst:
The library will now consistently be named `libfluidsynth-3` (with the proper file extension), no matter if it's a DLL, import lib, or static lib, compiled with MinGW or MSVC. The CI pipelines will also validate this now. Thanks for reporting, and Pedro for reviewing!

src/CMakeLists.txt
```diff
@@ -323,7 +323,10 @@ elseif ( WIN32 )
   set_target_properties ( libfluidsynth
     PROPERTIES
       PUBLIC_HEADER "${public_HEADERS}"
+      PREFIX "lib"
       OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
+      IMPORT_PREFIX "lib"
+      ARCHIVE_OUTPUT_NAME "fluidsynth-${LIB_VERSION_CURRENT}"
```
madebr, line 329: This change causes static or import libraries on Windows to always be versioned.
So it's currently impossible to link to fluidsynth as `-lfluidsynth`.

See [the contents of the msys2 mingw64 fluidsynth package](https://packages.msys2.org/packages/mingw-w64-x86_64-fluidsynth): no `libfluidsynth.a` or `libfuildsynth.dll.a` is present.

PR #1552, opened 2025-04-21
Align structures for 64-bit platforms reduce cost move, copy, create objects

GermanAizek:
@derselbst, hi. Thanks for huge contribution to open source development FluidSynth for Linux and others platforms.

## About PR changes:

Smaller size structure or class, higher chance putting into cpu cache.

Most processors are already 64 bit, so the change won't make it any worse.

More info about: https://stackoverflow.com/a/20882083

## Affected structs:

- _fluid_sample_t 128 to 120 bytes
- _fluid_samplecache_entry_t 80 to 72 bytes

### DIFF
--- a/src/sfloader/fluid_samplecache.c
+++ b/src/sfloader/fluid_samplecache.c
@@ -48,9 +48,9 @@ struct _fluid_samplecache_entry_t
     int sample_type;
     /*  End of cache key members */
 
+    int sample_count;
     short *sample_data;
     char *sample_data24;
-    int sample_count;
 
     int num_references;
     int mlocked;
--- a/src/sfloader/fluid_sfont.h
+++ b/src/sfloader/fluid_sfont.h
@@ -165,13 +165,13 @@ struct _fluid_sample_t
     unsigned int loopstart;       /**< Loop start index */
     unsigned int loopend;         /**< Loop end index, first point following the loop (superimposed on loopstart) */
 
+    short *data;                  /**< Pointer to the sample's 16 bit PCM data */
+    char *data24;                 /**< If not NULL, pointer to the least significant byte counterparts of each sample data point in order to create 24 bit audio samples */
     unsigned int samplerate;      /**< Sample rate */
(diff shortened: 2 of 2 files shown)


derselbst (changes requested):
Well honestly, I don't think this is worth the effort. The majority of the structs you have touched will only ever have a single instance... others maybe two or four. There are a few structs I can think of where this would actually matter, because they can have `N` instances:

* fluid_sample_t
* fluid_samplecache_entry_t
* fluid_voice_t
* fluid_rvoice_t
* fluid_event_t

And that's it. No need to change all the other structs. If you really want to get into it you should look through all the individual member vars and identify those which are being declared as `int` while being used as boolean flags, so that they could actually be demoted to `uchar`.

Also, please refrain from using compiler-specific attributes as this breaks compilation with other compilers (e.g. MSVC in your case). Whether or not to pack a struct should preferably be a decision of the compiler (and its optimization level).

GermanAizek:
> Also, please refrain from using compiler-specific attributes as this breaks compilation with other compilers (e.g. MSVC in your case). Whether or not to pack a struct should preferably be a decision of the compiler (and its optimization level).

@derselbst,
I just used keyword because visual style will break there, perhaps code readability is a higher priority for you.

GermanAizek:
@derselbst, check this commit https://github.com/FluidSynth/fluidsynth/pull/1552/commits/bf03d6d04b55d42766002addbbfb307e246c62a4

derselbst (commented):
That would be fine for me. Are you intending to look into the other mentioned structs as well?

GermanAizek:
> That would be fine for me. Are you intending to look into the other mentioned structs as well?

You suggest reducing `int` fields to range 0 - 255 (1 byte)? I can do this in other PR.

derselbst:
For example, yes. There might be a few more places were it might be reasonable to turn an enum member, which is declared as `int` into its true `enum` type.

Ok then, I'd merge this now, thanks.

PR #1556, opened 2025-04-22
Reduce structs sizes using scoped enums and strict types

GermanAizek:
@derselbst,
I started reducing the structures as you suggested, there is a lot of work judging by the number of enums.

How to fix func `#define fluid_atomic_int_set(_pi, _val) g_atomic_int_set(_pi, _val)`
I have type value `short`, but function accepts only int. Use C-cast?

### DIFF
--- a/include/fluidsynth/event.h
+++ b/include/fluidsynth/event.h
@@ -37,7 +37,7 @@ extern "C" {
 /**
  * Sequencer event type enumeration.
  */
-enum fluid_seq_event_type
+enum fluid_seq_event_type : unsigned char
 {
     FLUID_SEQ_NOTE = 0,		/**< Note event with duration */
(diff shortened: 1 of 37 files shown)


derselbst (changes requested):
Oh sry, but you have done way too much. Perhaps there was a misunderstanding. My previous comment was scoped to the structs I have mentioned, since you brought up this issue.

You have changed (and therefore broken) the entire public API. You have also introduced fixed-width enum notation as found in C23, even though fluidsynth's headers are in C90 for compatibility. You're not using scoped enums as the PR title suggests either.

What I meant was to e.g. change `fluid_event_t::type` from `int` to `enum fluid_seq_event_type`, which will leave the size of the underlying type up to the compiler, while also making the code more readable / more explicit.

PR #1565, opened 2025-05-02
Attempt to fix spurious CI failures

mmlr:
This PR attempts to fix two problems seen in CI, where it fails at pull request validation for unrelated reasons:

* The XCode CI build seems to be broken since the merge of #1494 due to what appears to be a name collision that is not handled at link time.
* The docker push should not run for PRs as it doesn't have permission to push, exclude it based on Build.Reason.


src/CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ set ( libfluidsynth_SOURCES
     rvoice/fluid_adsr_env.h
     rvoice/fluid_chorus.c
     rvoice/fluid_chorus.h
-    rvoice/fluid_iir_filter.cpp
+    rvoice/fluid_iir_filter_table.cpp
```
derselbst, line 140: Let's call it `fluid_iir_filter_impl`

PR #1566, opened 2025-05-03
Fixes for missing API Reference with Doxygen >= 1.9.8 and broken dark mode with >= 1.9.5

mmlr:
While preparing for documenting the proposed OSAL feature, I've stumbled across broken rendering with newer Doxygen versions. I've fixed the two most glaring issues that made the docs unusable in a browser set to dark mode with Doxygen 1.9.8.

I've lumped in a commit for an actual documentation fix and a cleanup commit, which I could also split into its own PR if desired.


derselbst (approved):
Thanks!

PR #1568, opened 2025-05-05
Update LGPLv2.1 license text and notices for remote-only FSF

musicinmybrain:
Update LGPL-2.1-or-later license notices
The FSF is now remote-only and no longer has a physical address; see
https://www.gnu.org/licenses/old-licenses/lgpl-2.1.en.html#SEC4.

----

Update LGPLv2.1 license text
Based on https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt

- The FSF is now remote-only, and no longer has a physical address
- There are minor changes to whitespace and placeholder text

----

Suggested in https://github.com/swami/libinstpatch/pull/81#issuecomment-2843598552.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,13 +2,13 @@
                        Version 2.1, February 1999
 
  Copyright (C) 1991, 1999 Free Software Foundation, Inc.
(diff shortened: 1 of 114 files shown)


derselbst (approved):
Thanks a bunch!

PR #1570, opened 2025-05-08
Add a C++11 based OS abstraction

mmlr:
This brings in the C++11 based OSAL and a lot of (possibly controversial) supporting changes.

The C++11 based abstraction uses threading and locking primitives from the C++11 standard library. When support for C++17 filesystem is also detected, fluid_file_test and fluid_stat are implemented on top of that, otherwise they are stubs. Atomics are implemented using a generic, mutex based implementation.

The changes here also serve to make the alternative OSALs more functional. Namely by providing fluid_shell_parse_argv_internal, a simple ~200 LoC, state machine based shell parser to take the place of g_shell_parse_argv. It's output is verified to be the same as the one from GLib by means of a unit test. Having this standalone parser makes the command line and TCP input work when using the embedded or cpp11 OSAL.

Lots of extra CI builds with osal=cpp11 are added here. This serves to check how broadly the C++11 based abstraction is supported, but should probably be reduced later on as it seems wasteful.

The CI builds on Windows already revealed some issues that were fixed. This concerns two remaining uses of GLib in `fluidsynth_priv.h` where alternatives are now provided:
* [v]snprintf on Windows with old VC runtimes is provided by wrapping calls to _[v]snprintf that fix their C99 non-conformance
* VLA declaration is implemented as a one-liner using [_]alloca

Finally, the possibly most controversial change is to the main fluidsynth binary. It uses fluid_test_file to check for the presence of configuration files before sourcing them. Generally the use of fluid_test_file in such a manner is a time-of-check-to-time-of-use race. In this context the direct use of the unexported fluid_test_file from the main binary is also problematic due to linking. Adding the OSAL sources to the binary does work, but seems problematic due to pulling in other symbols. Instead, I opted for removing the use of fluid_test_file in the main binary. The configuration fallback (via option -> user config -> system config) is now done by attempting to source the files directly and stopping once this is successful. This introduces a behaviour change when inaccessible configuration files are present. Previously the fallback would stop at that file and later sourcing it would fail, resulting in no configuration being applied. Now, such files fail to source and are skipped, allowing lower priority configuration files to be applied.

Some additional fixes are interspersed here. These came up while testing and verifying the OSAL implementation. If desired, these can be extracted and submitted separately.


mmlr:
I've added a unit test for the threading and locking primitives that revealed some issues. The thread private data was not set properly and condition waiting didn't adpot and release the lock and would deadlock. The issues have been fixed and the mutex locking has also been made more robust regarding interruption of locking.

Overall the CI situation looks promising, there are only a couple of failures:

It looks like the new threading test reveals a memory leak under GLib in Linux, which caused the ubsan=1 debug=1 build to fail.

The Windows 10 and XP builds fail with a linker error regarding the global atomic lock that is declared extern in fluid_sys_cpp11.h. I assume there needs to be some specific "export" qualifier added so that it finds the symbol?

The only actual build failure of the C++11/17 code seems to happen under MinGW x64 where it fails in a somewhat strange way. It somehow doesn't include the thread, mutex and condition_variable headers, then doesn't understand the types and then suggests to include exactly the files that are already included... As the chrono and map types seem to be fine, I guess it is not a general issue including things from the standard library. It looks like there is simply no C++11 threading support with the used GCC 8.1.

mmlr:
> The Windows 10 and XP builds fail with a linker error regarding the global atomic lock that is declared extern in fluid_sys_cpp11.h. I assume there needs to be some specific "export" qualifier added so that it finds the symbol?

It obviously just needed C linkage. The build and unit tests now work on the Windows 10 and XP targets with only the MinGW x64 build failing for actual lack of support.

src/CMakeLists.txt
```diff
@@ -187,6 +185,9 @@ set ( libfluidsynth_SOURCES
     bindings/fluid_ladspa.c
     bindings/fluid_ladspa.h
 )
+
+file ( GLOB fluid_osal_SOURCES utils/fluid_sys_${osal}.* )
```
derselbst, line 189: Please refrain from file globbing. This won't work as expected when git bisecting through the code.

src/utils/fluid_locked_atomics.h
```diff
@@ -0,0 +1,129 @@
...
+#define fluid_atomic_int_dec_and_test(_pi) \
+    _fluid_atomic_int_dec_and_test((fluid_atomic_int_t *)_pi)
+#define fluid_atomic_int_compare_and_exchange(_pi, _old, _new) \
+    _fluid_atomic_int_compare_and_exchange((fluid_atomic_int_t *)_pi, _old, _new)
+#define fluid_atomic_int_add(_pi, _add) \
+    _fluid_atomic_int_add((fluid_atomic_int_t *)_pi, _add)
+#define fluid_atomic_int_exchange_and_add fluid_atomic_int_add
+
+#define FLUID_ATOMIC_WRAPPER(op) \
+    extern fluid_mutex_t _atomic_lock; \
```
derselbst, line 46: You are using a global mutex here, rather than C++11 atomics. Is it because this approach allows you to inline everything into the C world?

src/utils/fluid_stub_functions.h
```diff
@@ -31,30 +31,30 @@
 #include "fluidsynth_priv.h"
 
 
-#define STUB_FUNCTION_VOID(function, args...) \
+#define STUB_FUNCTION_VOID(function, ...) \
     static FLUID_INLINE void \
-    function(args) \
+    function(__VA_ARGS__) \
```
derselbst, line 36: I think using `__VA_ARGS__` would require to bump `CMAKE_C_STANDARD` to `99`? I'm unsure if older version of MSVC would buy that.

src/utils/fluid_sys_cpp11.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+        std::thread *thread = new std::thread(thread_wrapper, func, data);
+        if (detach)
+            thread->detach();
+
+        return thread;
+    }
+    catch (const std::bad_alloc &)
+    {
+        FLUID_LOG(FLUID_ERR, "Out of memory");
```
derselbst, line 88:
```suggestion
        FLUID_LOG(FLUID_PANIC, "Out of memory");
```

src/utils/fluid_sys_cpp11.cpp
```diff
@@ -0,0 +1,319 @@
...
+{
+    static_cast<std::thread *>(thread)->join();
+    return FLUID_OK;
+}
+
+void _fluid_mutex_init(fluid_mutex_t *mutex)
+{
+    *mutex = new(std::nothrow) std::mutex();
+    if (*mutex == nullptr)
+        FLUID_LOG(FLUID_ERR, "Out of memory on mutex allocation");
```
derselbst, line 128:
```suggestion
        FLUID_LOG(FLUID_PANIC, "Out of memory on mutex allocation");
```

src/utils/fluid_sys_cpp11.cpp
```diff
@@ -0,0 +1,319 @@
...
+void fluid_mutex_unlock(fluid_mutex_t mutex)
+{
+    static_cast<std::mutex *>(mutex)->unlock();
+}
+
+void _fluid_rec_mutex_init(fluid_rec_mutex_t *mutex)
+{
+    *mutex = new(std::nothrow) std::recursive_mutex();
+    if (*mutex == nullptr)
+        FLUID_LOG(FLUID_ERR, "Out of memory on recursive mutex allocation");
```
derselbst, line 176:
```suggestion
        FLUID_LOG(FLUID_PANIC, "Out of memory on recursive mutex allocation");
```

src/utils/fluid_sys_cpp11.cpp
```diff
@@ -0,0 +1,319 @@
...
+void fluid_cond_mutex_unlock(fluid_cond_mutex_t *mutex)
+{
+    static_cast<std::mutex *>(mutex)->unlock();
+}
+
+fluid_cond_mutex_t *new_fluid_cond_mutex(void)
+{
+    std::mutex *mutex = new(std::nothrow) std::mutex();
+    if (mutex == nullptr)
+        FLUID_LOG(FLUID_ERR, "Out of memory on condition mutex allocation");
```
derselbst, line 208:
```suggestion
        FLUID_LOG(FLUID_PANIC, "Out of memory on condition mutex allocation");
```

src/utils/fluid_sys_cpp11.cpp
```diff
@@ -0,0 +1,319 @@
...
+    std::unique_lock<std::mutex> lock(*static_cast<std::mutex *>(mutex), std::adopt_lock);
+    static_cast<std::condition_variable *>(cond)->wait(lock);
+    lock.release();
+}
+
+fluid_cond_t new_fluid_cond(void)
+{
+    std::condition_variable *cond = new(std::nothrow) std::condition_variable();
+    if (cond == nullptr)
+        FLUID_LOG(FLUID_ERR, "Out of memory on condition variable allocation");
```
derselbst, line 238:
```suggestion
        FLUID_LOG(FLUID_PANIC, "Out of memory on condition variable allocation");
```

test/CMakeLists.txt
```diff
@@ -31,6 +31,14 @@ ADD_FLUID_TEST(test_seq_event_queue_remove)
 ADD_FLUID_TEST(test_jack_obtaining_synth)
 ADD_FLUID_TEST(test_utf8_open)
 
+if ( GLIB_SUPPORT AND GLib2_VERSION VERSION_GREATER_EQUAL 2.33.12 )
```
derselbst, line 34: Can you pls. add a comment to the code that explains why this version requirement is needed?

test/test_threading.c
```diff
@@ -0,0 +1,105 @@
...
+    data.atomic = 0;
+
+    for (i = 0; i < THREAD_COUNT; i++)
+    {
+        threads[i] = new_fluid_thread("increment", increment, &data, 0, FALSE);
+    }
+
+    for (i = 0; i < THREAD_COUNT; i++)
+    {
+        fluid_thread_join(threads[i]);
```
derselbst, line 91: I think `delete_fluid_thread` is missing here.

derselbst (commented):
First of all, many thanks again!

That one failing MinGW pipeline uses 8.1.0. There's another [newer pipeline](https://github.com/FluidSynth/fluidsynth/actions/runs/14950489543) that uses 15.1. I've [read this note](https://gcc.gnu.org/gcc-13/changes.html#windows), which confirms that C++11 threads  is now supported for both, the posix and the win32 threading model in MinGW. I think we could safely drop that old MinGW 8.1.0 pipeline.

I've got a few more questions / notes below. I will also bring this PR to the discussion as well as the C++17 filesystem topic, in the next days.

src/utils/fluid_locked_atomics.h
mmlr: The main issue is that the atomic types are declared as basic int, unsigned int and float in `fluidsynth_priv.h` and are used directly without any new_fluid_... and delete_fluid_... to manage their lifecycle. In `fluidsynth_priv.h` they can't be std::atomic and making them into opaque structs with pointers gives no good way of handling their lifetime.

After implementing it like this, I've discovered that [GLib does exactly the same thing](https://github.com/GNOME/glib/blob/314d63fcf7f309d2696480cba7d6c404d6027159/glib/gatomic.c#L858) when not using native atomics.

In C++20 one could use std::atomic_ref, but for C++11 I don't see how this could work outside of adding new lifecycle functions for the atomic types and then allocating an actual std::atomic.

src/CMakeLists.txt
mmlr: I have now added a version that selects .cpp or .c depending on osal == cpp11.

src/utils/fluid_locked_atomics.h
derselbst: > The main issue is that the atomic types are declared as basic int, unsigned int and float in fluidsynth_priv.h and are used directly without any new_fluid_... and delete_fluid_

Correct me if I'm missing smth., but this is totally valid. `std::atomic` is specialized for integral and floatingpoint types, making it trivially constructible (until C++20), trivially destructible and standard-layout, i.e. effectively becoming a POD-type. With additional static assertions that sizeof and alignof `std::atomic<T>` is the same as `T`, one could have just declared a `T` in a C struct. E.g. check the code below. It compiles fine if you stay below C++20, so I'm pretty sure this approach would have been well-defined.

Yet, no need to change the mutex approach you've chosen, I'm totally fine with either approach :)

```c++
#include <atomic>
#include <type_traits>

std::atomic_int i={0};

static_assert(std::is_trivially_copyable<decltype(i)>(), "is is not trivially copyable");
static_assert(std::is_trivially_default_constructible<decltype(i)>(), "i is not trivially constructible");
static_assert(std::is_trivially_destructible<decltype(i)>(), "i is not trivial destr");
static_assert(std::is_pod<decltype(i)>(), "i is not a POD");
static_assert(std::is_standard_layout<decltype(i)>(), "i is non standard layout");

static_assert(sizeof(i) == sizeof(int), "i doesn't match sizeof(int)");
static_assert(std::alignment_of<decltype(i)>() == std::alignment_of<int>(), "i doesn't match alignment of int");

int main()
{return 0;}
```

mmlr: I did actually try this and it boils down to simply coercing the type to a std::atomic_int during the operation. It does seem to actually work, i.e. it generates a `lock` prefixed operation in assembly and the tests pass. If the conditions outlined above hold, it implies that there is no extra information in the data itself and the type conversion should be valid. The actual type of the storage does then not matter at all, as long as it can hold the basic type and is aligned properly, so no struct should be necessary either.

Overall I find it rather unclean, as it involves simply casting away the original type. As I understand it, that's why C++20 introduced std::atomic_ref, as the semantic could not be done "properly" before, even if it would probably work. Then again, all the lock types being `void *` and then just static_casting them into their allocated type is not really proper either...

This is the diff to a version that uses std::atomic. Note how it looks more proper with the static_cast, but indeed only compiles when casting the incoming pointer to a `void *`, really making it a reinterpret_cast (which works as well).

```diff
diff --git a/src/utils/fluid_sys_cpp11.cpp b/src/utils/fluid_sys_cpp11.cpp
index 1e974eb7..8e43ceb4 100644
--- a/src/utils/fluid_sys_cpp11.cpp
+++ b/src/utils/fluid_sys_cpp11.cpp
@@ -20,6 +20,7 @@
 
 #include "fluid_sys.h"
 
+#include <atomic>
 #include <chrono>
 #include <condition_variable>
 #include <map>
@@ -264,6 +265,41 @@ void fluid_private_set(fluid_private_t priv, void *value)
     private_data[priv] = value;
 }
 
+void fluid_atomic_int_set(fluid_atomic_int_t *pi, int val)
+{
+    *static_cast<std::atomic_int *>((void *)pi) = val;
+}
+
+bool fluid_atomic_int_dec_and_test(fluid_atomic_int_t *pi)
+{
+    return --*static_cast<std::atomic_int *>((void *)pi) == 0;
+}
+
+bool fluid_atomic_int_compare_and_exchange(fluid_atomic_int_t *pi, int old, int _new)
+{
+    return static_cast<std::atomic_int *>((void *)pi)->compare_exchange_strong(old, _new);
+}
+
+int fluid_atomic_int_add(fluid_atomic_int_t *pi, int add)
+{
+    return static_cast<std::atomic_int *>((void *)pi)->fetch_add(add);
+}
+
+bool fluid_atomic_pointer_compare_and_exchange(void *pp, void *old, void *_new)
+{
+    return static_cast<std::atomic<void *> *>(pp)->compare_exchange_strong(old, _new);
+}
+
+void *fluid_atomic_pointer_get(void *pp)
+{
+    return *static_cast<std::atomic<void *> *>(pp);
+}
+
+void fluid_atomic_pointer_set(void *pp, void *val)
+{
+    *static_cast<std::atomic<void *> *>(pp) = val;
+}
+
 #if HAVE_CXX_FILESYSTEM
 
 #include <filesystem>
diff --git a/src/utils/fluid_sys_cpp11.h b/src/utils/fluid_sys_cpp11.h
index 9ef07252..6414d39b 100644
--- a/src/utils/fluid_sys_cpp11.h
+++ b/src/utils/fluid_sys_cpp11.h
@@ -117,7 +117,17 @@ void fluid_private_set(fluid_private_t priv, void *value);
 
 /* Atomic operations */
 
-#include "fluid_locked_atomics.h"
+#define fluid_atomic_int_inc(_pi) fluid_atomic_int_add(_pi, 1)
+#define fluid_atomic_int_get(_pi) fluid_atomic_int_add(_pi, 0)
+#define fluid_atomic_int_exchange_and_add fluid_atomic_int_add
+
+void fluid_atomic_int_set(fluid_atomic_int_t *pi, int val);
+bool fluid_atomic_int_dec_and_test(fluid_atomic_int_t *pi);
+bool fluid_atomic_int_compare_and_exchange(fluid_atomic_int_t *pi, int old, int _new);
+int fluid_atomic_int_add(fluid_atomic_int_t *pi, int add);
+bool fluid_atomic_pointer_compare_and_exchange(void *pp, void *old, void *_new);
+void *fluid_atomic_pointer_get(void *pp);
+void fluid_atomic_pointer_set(void *pp, void *val);
 
 
 /* Threads */
```

derselbst: Cool, thanks for giving it a try! I admit the casting looks... "gewöhnungsbedürftig", so let's keep the mutex based approach.
(PS: I will be offline for the next weeks and intending to merge it when I'm back, so that for the upcoming 2.5.0 we will have two choices for OSALs.)

derselbst (approved):
Since no additional feedback has been provided in the past weeks, I'm merging this now. As mentioned, this change will become part of fluidsynth 2.5.0 (no ETA yet).

ReinholdH:
@derselbst, @mmlr I'd like to join in the thanks. Using Visual Studio 2022 now w/o glib our tests have been very promising and are almost completed. We plan to release the fluidsynth libs of 2.5.0 in a next update. Thank you.

PR #1575, opened 2025-05-24
Add Big Endian support to embedded OSAL

Ghabry:
Changes:

1. Add big endian byte swapping to the embedded OSAL
2. Removed required Threads from the CMakeLists. My test platform (Wii Homebrew / devkitppc) has no Thread support _at all_ making the check fail

Tested this on a Wii (PowerPC BE) to verify that the patch works.

Great work on the abstraction layer. Can finally drop my no-glib patchset.

---

btw there is exactly one compiler error left but I was not sure how to fix this properly so I left it in:

``pthread_setschedparam`` is called which is not available on my platform:

```diff
diff --git a/src/utils/fluid_sys.c b/src/utils/fluid_sys.c
index a4388ac5..d5e80f75 100644
--- a/src/utils/fluid_sys.c
+++ b/src/utils/fluid_sys.c
@@ -420,7 +420,7 @@ fluid_thread_self_set_prio(int prio_level)
         memset(&priority, 0, sizeof(priority));
         priority.sched_priority = prio_level;
 
-        if(pthread_setschedparam(pthread_self(), SCHED_FIFO, &priority) == 0)
+        //if(pthread_setschedparam(pthread_self(), SCHED_FIFO, &priority) == 0)
         {
             return;
         }
```

---

For context: We use Fluidsynth in the app "EasyRPG Player" to play Midi files through soundfonts and the application exists as Homebrew for Wii and WiiU which are Big Endian (PowerPC) :).


Ghabry:
Thanks for the fix! Building works and sounds correct on the Wii.

PR #1581, opened 2025-06-08
Allow unknown INFO chunks

### LINKED ISSUE #1580: Fluidsynth should not reject unknown INFO chunks
### FluidSynth version

```sh

FluidSynth executable version 2.4.6
Sample type=double
```

### Describe the bug

SF2 2.04 specification, section 10.2:

> 10.2 Unknown Chunks
In parsing the RIFF structure, unknown but well formed chunks or sub-chunks may be encountered. Unknown chunks
within the INFO-list chunk should simply be ignored. Other unknown chunks or sub-chunks are illegal and should be
treated as structural errors.

I've created a [test soundfont for my DMOD proposal](https://github.com/spessasus/soundfont-proposals) and according to the specifications, it is still a valid file that should load. It only includes the extra DMOD chunk, yet fluidsynth rejects it, violating this section of the spec. I've specifically added it in the INFO-list because the spec allows to do so.

### Expected behavior

The program should load the bank and play it as is, ignoring unknown INFO chunks. The current behavior is not to spec and eliminates the possibility of any extensions, like my DMOD proposal.

### Steps to reproduce

1. [Download the DMOD test soundfont](https://github.com/spessasus/soundfont-proposals/blob/main/DMOD%20Test%20SoundFont_v2.sf2)
2. Attempt to load it with fluidsynth
3. `fluidsynth: error: Invalid chunk id in INFO chunk`

### Additional context

I'm working on a soundfont/dls editor which also writes this chunk to an sf2 file. Having fluidsynth incorrectly reject it isn't good.

spessasus:
Fixes #1580 

**Note:** I had to move the unknown logging code into the known INFO parser as otherwise the bytes don't get skipped over properly and fluid attempts to read sdta starting from the unknown chunk.

### DIFF
--- a/src/sfloader/fluid_sffile.c
+++ b/src/sfloader/fluid_sffile.c
@@ -731,7 +731,9 @@ static int process_info(SFData *sf, int size)
                  * within the INFO-list chunk should simply be ignored.
                  * Other unknown chunks or sub-chunks are illegal and should be
                  * treated as structural errors.*/
-                FLUID_LOG(FLUID_WARN, "Ignoring unknown chunk ID in INFO chunk");
+                unsigned char *p = (unsigned char *)&chunk.id;
+                FLUID_LOG(FLUID_WARN, "Ignoring unknown chunk ID '%c%c%c%c' in INFO chunk",
+                          p[0], p[1], p[2], p[3]);
             }
 
             /* alloc for chunk fcc and da chunk */


spessasus:
Thanks for the merge. Can I have a statement on the DMOD proposal as well? I can create a PR for that too if you'd like.

PR #1583, opened 2025-06-16
Fix debug builds of C++11 OSAL due to missing assert include

mmlr:
This one somehow got missed and breaks debug builds.


derselbst (approved):
Thanks!

PR #1586, opened 2025-06-20
Fix build for Windows on ARM64: '-mms-bitfields' unsupported

carlo-bramini:
Compiling Fluidsynth for Windows on ARM64 hangs with this error:
```
aarch64-w64-mingw32-gcc: error: unrecognized command-line option '-mms-bitfields'
```
Actually, `-mms-bitfields` is supported only by Intel target, so it's correct to receive an error when the target Aarch64 is selected. In my opinion, the best way for fixing this issue is to test if  `-mms-bitfields` is really supported, as included in the attached patch.


derselbst (approved):
LGTM Carlo, thanks!

PR #1587, opened 2025-06-23
Fix exclusion of fluid_thread_self_set_prio in embedded OSAL

mmlr:
This is a rework of the change in #1575. The embedded OSAL may be used on non-POSIX platforms as well, so the exclusion of the fluid_thread_self_set_prio definition needs to apply to all of them.

Also the declaration of the function doesn't need to be excluded as it is defined in all cases, just possibly as a stub.

### DIFF
--- a/src/utils/fluid_sys.c
+++ b/src/utils/fluid_sys.c
@@ -384,6 +384,8 @@ unsigned int fluid_curtime(void)
 }
 
 
+#if !OSAL_embedded
+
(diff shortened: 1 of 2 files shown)


derselbst:
Thanks Micha, LGTM! @Ghabry  Here's a slight change of our previous fix, feel free to test if it works for your use-case as well.

Ghabry (approved):
works for me

derselbst (approved):
Great, thank you both!

Issue #1574, opened 2025-05-21
Fix SIGFPE on OS/2


komh:
### FluidSynth version

```sh
FluidSynth runtime version 2.4.6
Copyright (C) 2000-2025 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.4.6
Sample type=double
```

### Describe the bug

When playing a midi file like:
  ```fluidsynth mid-file.mid sf2-file.sf2```

fluidsynth crashes due to SIGFPE.

### Expected behavior

It should be play a mid file without any problems at all.

### Steps to reproduce

Execute fluidsynth with DART audio drvier, ```-a dart``` like:
  ```fluidsynth mid-file.mid sf2-file.sf2 -a dart```

### Additional context

This is a regression of commit 1ef445e6d787c2c55dd739653c60deda0a36bedd.

Here is the patch.

[0001-Fix-SIGFPE-on-OS-2.patch](https://github.com/user-attachments/files/20366745/0001-Fix-SIGFPE-on-OS-2.patch)
```
From 1ffc2de2706ef7066aba75389769334a626433bb Mon Sep 17 00:00:00 2001
From: KO Myung-Hun <komh78@gmail.com>
Date: Wed, 21 May 2025 19:52:05 +0900
Subject: [PATCH] Fix SIGFPE on OS/2

This is due to a regression of commit 1ef445e.
---
 src/CMakeLists.txt | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index ed051a0e..395f8338 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -188,9 +188,9 @@ set ( libfluidsynth_SOURCES
     bindings/fluid_ladspa.h
 )
 if ( CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU" )
-  set_source_files_properties(rvoice/fluid_iir_filter.cpp PROPERTIES COMPILE_FLAGS "-fno-math-errno -ffast-math")
+  set_source_files_properties(rvoice/fluid_iir_filter_impl.cpp PROPERTIES COMPILE_FLAGS "-fno-math-errno -ffast-math")
[...]

PR #1597, opened 2025-07-02
Fix a build failure on OS/2

derselbst:
COMMENT is defined as macro on OS/2, see #1574.


komh:
Thanks! Build fine on OS/2.

derselbst (commented):
Thanks for confirming and reporting!

PR #1599, opened 2025-07-04
Add support for the DMOD chunk

### LINKED ISSUE #1582: Default Modulators Proposal v2
### Related discussion

#1384, this is an updated version of that proposal, I'll close it right after I post this one.

Currently, soundfont does not allow us to add default modulators. The only way to do so is by adding them to *every instrument*, which can be a very tedious task and heavily discourages users from adding extended CC support to their banks.

The default modulators are also ambiguous given the existence of the velToFc modulator, which fluidsynth ignores. What if a bank designer wants it? They have to add it to all instruments in the bank. Now imagine having to do that with a bank like [this one!](https://musical-artifacts.com/artifacts/2525)

## [My revised proposal](https://github.com/spessasus/soundfont-proposals/blob/main/default_modulators.md)

The most important change is inspired by [this discussion:](https://github.com/SFe-Team-was-taken/SFe/discussions/38) the DMOD list now overrides and replaces all the default modulators.

### Does not violate the SoundFont Specification

> 10.2 Unknown Chunks
In parsing the RIFF structure, unknown but well formed chunks or sub-chunks may be encountered. Unknown chunks
within the INFO-list chunk should simply be ignored. Other unknown chunks or sub-chunks are illegal and should be
treated as structural errors.

### Default Modulators are Cool
Currently it is discouraging for bank creators to use modulators for extending CC support (like brightness or filter resonance) since having to add them to all the instruments can be tedious.

CC: @mrbumpy409, as an experienced soundfont designer, would having custom default modulators be useful? I hope so :-)

### Easy To Implement
Since the structure of DMOD is exactly the same as PMOD or IMOD, the parsers can be reused.

### Unambiguous
When the bank has its own modulators specified, it's clear for them. No more of the confusing velToFc modulator! If the creator wants the modulator, it is directly declared in the bank and all the confusion is avoided.

### Simple behavior
No more of that confusing addition I've described before. Just a drop-in replacement with the complete default modulator definition!

### Encouraging more modulators
Bank editors, like the one I'm building, could automatically add an [extended list of modulators for a broader CC support](https://github.com/spessasus/spessasynth_core/wiki/Modulator-Class#default-modulators). If the creator doesn't want them or wants to change them, no problem! Just edit them, no need to mess with every instrument.

### Better CC Extensions
DMOD removes the ambiguity of modulator extensions: For example, spessasynth and BASSMIDI both support brightness CC by default, yet their curves and amount are a bit different, resulting in the same MIDI file sounding differently on both synths. If all synths implement DMOD, they will all use DMOD the same way and all the extensions coded in will work exactly the same. For example, XG inspired banks can have the XG Brightness CC curve baked in globally, making it easy to adjust if necessary.

### Additional context

Test Bank and a DMOD editing tool are also described in the proposal.

Also @derselbst, If you can point to the correct portion of the code, I could try to make a PR for this :-)

spessasus:
I figured it out! :D

This PR adds support for the DMOD chunk per the [extension specification](https://github.com/spessasus/soundfont-proposals/blob/main/default_modulators.md).
I've extracted the mod importing code from the zone_mod_import so it can be used for DMOD without duplicated coede.

Here's a quick DMOD sf2 tester that makes it instantly audible, as it has a single default modulator that is a pitch wheel with a switch, and a fixed range of 3 semitones. The MIDI tests pitch bend range, so we can hear and verify that it's always the same.

[Quick_and_dirty_dmod_test.zip](https://github.com/user-attachments/files/21068924/Quick_and_dirty_dmod_test.zip)

All regular SF2 files and DLS files seem to work fine.
Note that I've also added a few `FLUID_DBG` logs to the INFO chunk as they were very handy. I can remove them if needed.

Resolves #1582

### DIFF
--- a/src/sfloader/fluid_defsfont.c
+++ b/src/sfloader/fluid_defsfont.c
@@ -443,6 +443,10 @@ int fluid_defsfont_load_all_sampledata(fluid_defsfont_t *defsfont, SFData *sfdat
     return sample_parsing_result;
 }
 
+// declared here so it can be used for default modulators in fluid_defsfont_load
+static int
+fluid_mod_import_sfont(fluid_mod_t **mod, fluid_list_t *sfmod);
+
 /*
  * fluid_defsfont_load
  */
@@ -450,6 +454,7 @@ int fluid_defsfont_load(fluid_defsfont_t *defsfont, const fluid_file_callbacks_t
 {
     SFData *sfdata;
     fluid_list_t *p;
+    fluid_list_t *dmod_data;
     SFPreset *sfpreset;
     SFSample *sfsample;
     fluid_sample_t *sample;
@@ -478,6 +483,17 @@ int fluid_defsfont_load(fluid_defsfont_t *defsfont, const fluid_file_callbacks_t
     {
         FLUID_LOG(FLUID_ERR, "Couldn't parse presets from soundfont file");
         goto err_exit;
+    }
+
+    dmod_data = sfdata->default_mod_list;
+    if (dmod_data != NULL)
+    {
+        /* Load the default modulators*/
+        if (fluid_mod_import_sfont(&defsfont->default_mod_list, dmod_data) != FLUID_OK)
+        {
+            FLUID_LOG(FLUID_ERR, "Unable to load the default modulators");
+            goto err_exit;
+        }
     }
 
     /* Keep track of the position and size of the sample data because
@@ -1562,26 +1578,31 @@ fluid_zone_mod_source_import_sfont(unsigned char *src, unsigned char *flags, uns
     return TRUE;
 }
 
-/*
+/**
  * fluid_zone_mod_import_sfont
  * Imports modulators from sfzone to modulators list mod.
- * @param zone_name, zone name.
- * @param mod, address of pointer on modulators list to return.
- * @param sfzone, pointer on soundfont zone.
+ * @param mod -  address of pointer on modulators list to return.
+ * @param sfmod - pointer on the modulator list.
  * @return FLUID_OK if success, FLUID_FAILED otherwise.
  */
 static int
-fluid_zone_mod_import_sfont(char *zone_name, fluid_mod_t **mod, SFZone *sfzone)
+fluid_mod_import_sfont(fluid_mod_t **mod, fluid_list_t *sfmod)
 {
     fluid_list_t *r;
     int count;
 
     /* Import the modulators (only SF2.1 and higher) */
-    for(count = 0, r = sfzone->mod; r != NULL; count++)
+    for(count = 0, r = sfmod; r != NULL; count++)
     {
 
-        SFMod *mod_src = (SFMod *)fluid_list_get(r);
+        SFMod *mod_src = (SFMod*)(r->data);
         fluid_mod_t *mod_dest = new_fluid_mod();
+
+        if (mod_src == NULL)
+        {
+            // empty list (DMOD case) - nothing to do!
+            return FLUID_OK;
+        }
 
         if(mod_dest == NULL)
         {
@@ -1650,6 +1671,24 @@ fluid_zone_mod_import_sfont(char *zone_name, fluid_mod_t **mod, SFZone *sfzone)
 
         r = fluid_list_next(r);
     } /* foreach modulator */
+    return FLUID_OK;
+}
+
+/*
+ * fluid_zone_mod_import_sfont
+ * Imports modulators from sfzone to modulators list mod.
+ * @param zone_name, zone name.
+ * @param mod, address of pointer on modulators list to return.
+ * @param sfzone, pointer on soundfont zone.
+ * @return FLUID_OK if success, FLUID_FAILED otherwise.
+ */
+static int
+fluid_zone_mod_import_sfont(char *zone_name, fluid_mod_t **mod, SFZone *sfzone)
+{
+    if (fluid_mod_import_sfont(mod, sfzone->mod) != FLUID_OK)
+    {
+        return FLUID_FAILED;
+    }
 
     /* checks and removes invalid modulators in modulators list*/
     fluid_zone_check_mod(zone_name, mod);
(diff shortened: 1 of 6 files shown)


src/sfloader/fluid_defsfont.h
```diff
@@ -107,16 +107,17 @@ struct _fluid_defsfont_t
     unsigned int samplesize;  /* the size of the sample data in bytes */
     short *sampledata;        /* the sample data, loaded in ram */
 
-    unsigned int sample24pos;		/* position within sffd of the sm24 chunk, set to zero if no 24 bit sample support */
```
spessasus, line 110: FYI these were tabs and my IDE suggested that I turn them into spaces

src/sfloader/fluid_sffile.c
```diff
@@ -712,6 +721,43 @@ static int process_info(SFData *sf, int size)
...
+            for(; count > 0; count--)
+            {
+
+                if((dmod = FLUID_NEW(SFMod)) == NULL)
+                {
+                    FLUID_LOG(FLUID_ERR, "Out of memory");
+                    return FALSE;
+                }
+
+                dmod_list->data = dmod;
```
derselbst, line 748: `dmod` is currently also leaked, will be fixed when you deallocate the list correctly as mentioned.

src/sfloader/fluid_defsfont.h
```diff
@@ -147,6 +148,7 @@ struct _fluid_defpreset_t
     unsigned int num;                     /* the preset number */
     fluid_preset_zone_t *global_zone;        /* the global zone of the preset */
     fluid_preset_zone_t *zone;               /* the chained list of preset zones */
+    fluid_mod_t *default_mod;             /* default mods for the soundfont */
```
derselbst, line 151: This looks like a leftover.

src/sfloader/fluid_defsfont.h
derselbst: nvm

src/sfloader/fluid_defsfont.c
```diff
@@ -1562,27 +1578,32 @@ fluid_zone_mod_source_import_sfont(unsigned char *src, unsigned char *flags, uns
...
     int count;
 
     /* Import the modulators (only SF2.1 and higher) */
-    for(count = 0, r = sfzone->mod; r != NULL; count++)
+    for(count = 0, r = sfmod; r != NULL; count++)
     {
 
-        SFMod *mod_src = (SFMod *)fluid_list_get(r);
+        SFMod *mod_src = (SFMod*)(r->data);
         fluid_mod_t *mod_dest = new_fluid_mod();
```
derselbst, line 1599: The modulator allocated in this line and later appended to the `defsfont->default_mod_list` is never freed. You need to call `delete_fluid_list_mod(defsfont->default_mod_list)` in `delete_fluid_defsfont()`.

src/sfloader/fluid_sffile.c
```diff
@@ -712,6 +721,43 @@ static int process_info(SFData *sf, int size)
...
+             */
+            if (chunk.size % SF_MOD_SIZE != 0 || size == 0)
+            {
+                FLUID_LOG(FLUID_ERR, "DMOD chunk has invalid size (%d bytes)", chunk.size);
+                return FALSE;
+            }
+
+
+            // read the modulators sequentially
+            count = (int)(chunk.size / SF_MOD_SIZE) - 1; // minus the terminal record
```
derselbst, line 737: Here, you're casting to int, but `count` is declared `uint`, pls. decide for one.

src/synth/fluid_synth.c
```diff
@@ -5298,12 +5298,24 @@ fluid_synth_alloc_voice_LOCAL(fluid_synth_t *synth, fluid_sample_t *sample, int
...
+        {
+            default_mod = synth->default_mod;
+        }
 
         while(default_mod != NULL)
         {
             if(
-                /* See if default_mod is the velocity_to_attenuation modulator */
+                /* if SF explicitly wants the default vel2att modulator, keep it*/
+                override_default_modulators != TRUE &&
```
derselbst, line 5317: This is an implementation detail I'm not too sure about. When fluidsynth has been configured for using a breath controller for triggering notes, it replace the default velocity to attenuation modulator. IMO, this logic should remain unchanged, i.e. if the soundfont contains a vel-to-atten modulator in its DMOD chunk and fluidsynth is in breath mode, it should still be replaced. Otherwise the breath logic of triggering and attenuating notes would be broken.

derselbst (commented):
Sry, I only find this for this during weekends. So here we are. I've tested it locally and the difference is clearly noticeable with your provided test file. There are a few more details below.

src/sfloader/fluid_sffile.c
```diff
@@ -649,11 +662,15 @@ static int process_info(SFData *sf, int size)
         uint32_t *fcc;
     } item;
     unsigned short ver;
+    unsigned char *p;
+    sf->default_mod_list = new_fluid_list();
```
derselbst, line 666:
```suggestion
    sf->default_mod_list = NULL;
```
Otherwise there would an empty list element at the end of the list.

src/sfloader/fluid_defsfont.c
derselbst: I think you've missed this one.

derselbst: I think you've missed this one

spessasus:
Sorry for missing these two, I've fixed them now :-)

spessasus:
> Pls. do not sell this as an SF3 feature. It's not. It's a custom but backward compatible extension to SF2.

I know, I didn't mean it as an SF3 feature! I meant mentioning it in the wiki, much like the SF3 format is mentioned there, _not that it's a part of SF3_. So thanks for that!

> If you feel it needs to be mentioned / documented / announce anywhere else, let me know.

I'm not sure where, but the idea is to get this adopted in other sf2 players, similarly to the way SF3 format was universally adopted.

derselbst:
Ok. You can drop a mail to our [mailing list](https://lists.nongnu.org/mailman/listinfo/fluid-dev), or create a [post](https://github.com/FluidSynth/fluidsynth.github.io/tree/main/_posts) that will appear on fluidsynth.org if you want.

PR #1620, opened 2025-08-03
Replace make_tables by C++11 constexpr arrays

derselbst:
Bye bye `make_tables`. This PR proposes to replace the auto-generated lookup tables (using the hacky ExternalProject CMake approach) with C++11 lookup tables.

References [gcem](https://github.com/kthohr/gcem) as submodule for C++11 constexpr math, making it necessary to update all CI pipeline to check out submodules.

When this PR is merged, a C++11 compliant compiler will be required to compile fluidsynth!

Currently, this PR is in a PoC state as it's incomplete:

- [x] Port the 2D DSP lookup tables
- [x] Compare the contents of the new array with the old array.

Yet, it compiles fine on all platforms and passes the unit tests. For the latter point, I wrote a unit test to compare with the old lookup tables. It passed with a maximum difference of `1e-4`. Since I don't want to add 8000 lines of code, I'll revert it again, as the C++ lookuptable implementation is unlikely to change. For future reference, the commit was b263b244edcc0ca2444fb2964643648ad9eadef2.


komh:
Hi/2.

I've encountered the following error on OS/2:

```
[ 10%] Building CXX object src/CMakeFiles/libfluidsynth-OBJ.dir/gentables/fluid_ct2hz.cpp.o
F:/lang/work/fluidsynth/fluidsynth.git/src/gentables/fluid_ct2hz.cpp:5:10: fatal error: gcem.hpp: No such file or directory
    5 | #include "gcem.hpp"
      |          ^~~~~~~~~~
compilation terminated.
make[2]: *** [src/CMakeFiles/libfluidsynth-OBJ.dir/build.make:152: src/CMakeFiles/libfluidsynth-OBJ.dir/gentables/fluid_ct2hz.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

komh:
Thanks, it fixed. This step is automated later ?

Anyway, unfortunately, this pr does not produce any sounds at all on my machine.

derselbst:
> Thanks, it fixed. This step is automated later ?

No. I have never seen this automated via e.g. hooks or CMake. It's the developers responsibility to deal with the repository and its content. Triggering `git submodule update` automatically might unintentionally override / revert changes made to the submodule.

Since the content of the gcem submodule is unlikely to change, using git subtree might be a reasonable alternative. git submodule would still be used for versioning the testdata of fluidsynth's manual testsuite. So for me personally it wouldn't make a difference. If I ever figure out a reliable way to automatically validate the rendering results of the manual testsuite, I would still have to enable submodule checkout in all CI pipelines.

> unfortunately, this pr does not produce any sounds at all on my machine

I will investigate next weekend.

white-axe:
Is there a way to do this without the compile-time `long double` arithmetic in fluid_cb2amp.cpp, fluid_concave.cpp, fluid_convex.cpp and fluid_pan.cpp? Why not just use regular `double`? This pull request has caused FluidSynth to fail to build with GCC on PowerPC targets that use the IBM floating-point ABI, such as Debian and Ubuntu, because [GCC doesn't have compile-time `long double` division implemented for this ABI](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=26374). Clang still works, though.

Issue #1524, opened 2025-03-30
XG Bank Selection should not hardcode drum bank 128


spessasus:
### FluidSynth version
Latest master (2.4.4)

### Describe the bug

The provided soundfont, when played with the MIDI file doesn't use drum presets but uses musical presets.
This is because it assumes the `MT` variations of the presets at the bank 127. While one can say that this is a fault of the soundfont, people are still probably going to blame the player...

### Expected behavior

Drums used correctly.

### Steps to reproduce
1. Build the latest master
2. [Get the soundfont](https://raw.githubusercontent.com/edge-classic/EDGE-classic/5fa1e0867e1ef71e260f45204888df85ada4be1b/soundfont/Default.sf2)
3. [Get that megaman 6 MIDI](https://github.com/user-attachments/files/19239114/MM6_-_MrX2010XG.zip)
4. Don't hear drums

### Additional context

One way to approach this (the way I did) is to perform an "XG validation" check. If it fails, all drums are forced to 128, regardless of presets existing in other XG valid drum banks:

https://github.com/spessasus/SpessaSynth/blob/5b936c32593515ab4925ad3598a0d2433bcf666c/src/spessasynth_lib/soundfont/basic_soundfont/basic_soundfont.js#L161-L190

derselbst:
The problem with your workaround I see is that it's only feasible for a single loaded Soundfont. But Fluidsynth supports multiple stacked soundfonts, in which case it would become too complex and obscure what's happening under the hood. So if you do believe this new behavior could cause issues, I would rather fallback to hardcoded bank 128 as before.

spessasus:
Speesasynth also supports stacked soundfonts and it doesn't have this problem. It just does the XG check not on the entire stack, but just on a single preset instance. And when it searches for a preset on the stack, it calls each soundfont's getPreset which individually handles its own XG hacks. Same with BASSMIDI. It supports SFLIST and it has the before mentioned fix (try running the bank in the falcosoft midi player with BASS mode)

Not sure if that's possible with the current code base though. Still, that sound bank is definitely a problem. If you look at the link to the download, it's a file in a GitHub repo which actually uses fluidsynth, so it was not a Soundfont made for testing bugs, but a real use case.

However, if you think that it's not a bug or won't fix, feel free to close this issue.

spessasus:
[...]

PR #1626, opened 2025-08-13
Implement native DLS parser

### LINKED ISSUE #1373: Incorrect DLS Support
### FluidSynth version
```cmd

Sample type=double
```

### Describe the bug
The DLS support in fluidsynth is missing critical features. The MIDI attached uses bank selects and attached `gm.dls` has them (like Square or Sine Wave) but fluidsynth defaults to bank 0 (Square Wave). 

Another very important thing are the drums. They use 0 tick length notes (quite common) but there's no sound, indicating that it doesn't support volEnvDecay in DLS!

I've attached an sf2 converted version of this soundfont, and playing `apple.mid` with it works fine.

### Expected behavior
Playing `apple.mid` through `gm.dls` should sound exactly like playing it with `GS Sound Set (16-bit).sf2`.

### Steps to reproduce

1. Download the attached file
2. `fluidsynth gm.dls -o synth.device-id=16 -o synth.gain=0.5 apple.mid`

### Additional context

Since fluidsynth seems to ignore articulators entirely, here's some helpful stuff I found out:

I've made my own DLS to SF2 converter (which I used to convert the file) and all parameters are stored using DLS articulators (think SF2 Modulators).
My converter checks if source, control (secondary source in SF2) and transform are all 0. [That usually means a soundFont generator.](https://github.com/spessasus/SpessaSynth/blob/89a24252375d339b40a14e2973915b62e2463fe7/src/spessasynth_lib/soundfont/dls/read_articulation.js#L89)

If not, there are a few [special cases:](https://github.com/spessasus/SpessaSynth/blob/89a24252375d339b40a14e2973915b62e2463fe7/src/spessasynth_lib/soundfont/dls/read_articulation.js#L192) for example source is VibLfo and destination is pitch, then it's VibLfoToPitch (and sometimes control is modulation wheel, so that's a modulator).
Or another case, source is keyNum and destination is VolEnvDecay, that's keyNumToVolEnvDecay.
**Note that in this case, (keyNum to something) there's a [special correction](https://github.com/spessasus/SpessaSynth/blob/master/src/spessasynth_lib/soundfont/dls/read_articulation.js#L247) that I needed to apply.**

Other than that, it tries to [transform the remaining articulators into SF2 modulators.](https://github.com/spessasus/SpessaSynth/blob/master/src/spessasynth_lib/soundfont/dls/articulator_converter.js)

I hope this helps!

```
// I can paste the entire DLS parser here if you want ;-)
```

### LINKED ISSUE #1426: Fluidsynth fails to load a valid DLS file
### FluidSynth version
```
FluidSynth runtime version 2.4.0

FluidSynth executable version 2.4.0
Sample type=double
```

### Describe the bug
The provided DLS file does not load correctly:
```
fluidsynth: error: Not a SoundFont file
fluidsynth: error: ipatch_dls_reader_new() failed with error: 'DLS Reader error: Invalid data (ofs=14188, traceback ['fmt ' ofs=0x14186, size=18 <= 'wave' ofs=0x1417A, size=328 <= 'wvpl' ofs=0x1416E, size=112350 <= 'DLS ' ofs=0x0, size=194774])'
fluidsynth: error: Failed to load SoundFont "/home/spessasus/the_good_stuff/soundfonts/downloadable_sounds/Nokia_S30.dls"
Failed to load the SoundFont /home/spessasus/the_good_stuff/soundfonts/downloadable_sounds/Nokia_S30.dls
Type 'help' for help topics.
```

### Expected behavior
The file should load. I know that this is a valid file since it opens in:
- SynthFont Viena
- Winamp in the MIDI Decoder configuration
- Awave Studio
- SpessaSynth

### Steps to reproduce
1. Grab this: [Nokia_S30.dls.zip](https://github.com/user-attachments/files/17689672/Nokia_S30.dls.zip) NOTE: Remove `.zip`, this is just the raw DLS file.
2. `fluidsynth Nokia_S30.dls`
3. See the error

### Additional context
no

### LINKED ISSUE #1584: Segmentation Fault with a corrupted DLS file
### FluidSynth version

```sh
FluidSynth runtime version 2.4.6

FluidSynth executable version 2.4.6
Sample type=double
```

### Describe the bug

Loading a corrputed DLS file (no wavepool) causes a SIGSEGV instead of falling back to the default soundfont, like it usually does with broken files.

```sh
FluidSynth runtime version 2.4.6

fluidsynth: error: Not a SoundFont file

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Overdrive Guitar' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Picked Bass' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Synth Bass 1' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Synth Bass 2' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Strings' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Square Wave' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Saw Wave' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Bass & Lead' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Reverse Cymbal' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Seashore' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Telephone 2' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Lazergun' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Bubble' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Synth Bass 4' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Orchestral Pad' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Synth Strings 3' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Synth Brass 3' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Sine Wave' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Power' (index=-1)

** (process:55332): WARNING **: 23:04:13.732: Failed to fixup sample for inst 'Electronic' (index=-1)
fish: Job 1, 'fluidsynth ~/Downloads/GM\ GS\ …' terminated by signal SIGSEGV (Address boundary error)
```

### Expected behavior

It should fallback to the default soundfont or exit gracefully.

### Steps to reproduce

[GM GS Bank.zip](https://github.com/user-attachments/files/20764611/GM.GS.Bank.zip)

Load this into fluidsynth and see it crash.

### Additional context

Also happens with SWAMI (isn't it related to fluidsynth?)

If corrupted files are allowed to cause a sigsegv, I'll close this issue and won't report more if I find any.

rsp4jack:
## Description

This PR introduces a native DLS loader, written in C++17. Using C++17 makes it more concise and have good readability.

It provides almost full support for DLS. Current DLS support is provided by libinstpatch, and it lacks some key DLS functionality. (see #1373)

When this PR is merged, completely getting rid of GLib will be possible! (with C++17 OSAL)

## Issues (not the GitHub one)

* [x] With this PR, we will have two DLS parser. Should we just say bye-bye and throw instpatch parser away? And if it is kept here, what is it for?
The instpatch parser perhaps supports GIG and SLI but it doesn't seem to work (see #493, I'm not sure if libinstpatch is still the same).

* [x] If the problem above is clear, report.cmake should be changed (I have not add enable-native-dls to it yet)

Another problem I found about CXX_STANDARD:

https://github.com/FluidSynth/fluidsynth/blob/8d26042026d950184e302abf2cf2df1b98502ff8/CMakeLists.txt#L565-L572

https://github.com/FluidSynth/fluidsynth/blob/8d26042026d950184e302abf2cf2df1b98502ff8/CMakeLists.txt#L708-L716

https://github.com/FluidSynth/fluidsynth/blob/8d26042026d950184e302abf2cf2df1b98502ff8/CMakeLists.txt#L783-L786

If OSAL is cpp17 and OBOE is enabled, CMAKE_CXX_STANDARD will become 14, even if the compiler support 17 or 23. 
I think target_compile_features should be used here.

## Linked issues (this is the GitHub one)

Fixes #1373
Fixes #1426
Fixes #1584 (when libinstpatch is disabled)

#847


rsp4jack:
About CrysDLS v1.23.dls: I just found 'crs1' chunks which have wrong size. Probably Crystal just don't want people read it (as it is encrypted originally).

rsp4jack:
see also https://github.com/spessasus/spessasynth_core/issues/5

spessasus:
You can try opening it in [spessafont](https://spessasus.github.io/SpessaFont/), converting to SF2 and comparing Fluid's sf structure to the one your parser produces?

src/sfloader/fluid_dls.cpp
```diff
@@ -2409,11 +2409,15 @@ static int fluid_dls_preset_get_banknum(fluid_preset_t *preset) noexcept
...
-            return DRUM_INST_BANK;
-        }
-        return inst->banklsb;
+        // if (inst->is_drums)
+        // {
+        //     return DRUM_INST_BANK;
+        // }
+        // return inst->banklsb;
+        
+        // Yamaha's DLS seems to use the same bank select mode as GS
```
spessasus, line 2418: Well, no, actually. This DLS has *both* XG drums (bank 127) and GS drums (bank 128).

Ideally in XG mode you'd select the XG drums and in GS mode, GS drums, since they have different drum maps, especially kick in XG being metronome in GS.

And it also has bank LSB numbers whereas GS only uses MSB.

rsp4jack: Oh no XG uses 127 as drum bank?

spessasus: Yes, [fluid has an active issue about that](https://github.com/FluidSynth/fluidsynth/issues/1524)

spessasus:
The difference between DLS 1 and DLS 2 is that DLS 2 allows articulators in every region, not just the global one. If the chunk was "lart" instead of "lar2", then it wouldn't be allowed. You can load the file with DirectMusic Producer or Audio Compositor which support DLS level 2.

Please see section 3.3 DLS Level 1 Compatibility notes of the DLS 2 specification. It clearly states that melodic instrument restrictions only apply to Level 1, not 2.

rsp4jack:
So DLS-2 allows region-level articulators nevertheless the instument is drums or melodic. But the problem is:

> DLS-2 2.10 <art2-ck>
> The <art2-ck> chunk may exist at **either** the instrument level, in which case it contains
> global articulation data, **or** at the region level, in which case it contains local articulation data.

But the Saw Lead contains both instrument-level and region-level level 2 articulators!

rsp4jack:
> "Or" means does not mean "mutually exclusive" :-)

I don't think so :confused:   
Does not "either ... or ..." mean that?

But anyway. If it exists, we had better support it. So what rules should apply here, for the "global zone" in DLS-2? Does it supersede or add?

rsp4jack:
Also DirectMusic produces the same sound.

derselbst (commented):
What a surprise - Many thanks for this contribution! Allow me to start with some high-level feedback for now:

* I would prefer if we leave the sample decoding to libsndfile entirely. I.e. fluidsynth should only support pure PCM out of the box (i.e. s16,...and u8 if you want to), similar to as we do for SF2.
* Is there a specific reason for using placement new? Otherwise, I would have preferred to just go with regular new or even better `std::make_unique()` (at least as long we're living in the C++ world, i.e. `unique_ptr::release()`ing before sending the object into the C world)
* Fluidsynth has a sample cache `fluid_samplecache.h`, though it's pretty bound to SF2. Would it make sense to (ab-)use it anyway? (It's totally fine if you decide not to.)
* C++17 is a pretty big jump. I'm fine with that esp since it's optional, but also because I haven't heard back from our Windows XP users for quite a while. If someone still requests WinXP compatibility, it appears to me that backporting to C++11 should be pretty straightforward (i.e. custom implementation of `std::optional`, a few syntactical changes here and there, etc).

I was hoping for fluidsynth 2.5.0 to be released in September or October. I am not sure if this PR should already be part of 2.5.0 or better wait for 2.6.0. Let me know your preference! Given the velocity that you two already have (much appreciated), it seems doable to get it into a shape that we can release it with 2.5.0. If I can help out, ideally, mention me directly, as it's hard for me to keep up with all the comments you've posted in the past days :)

rsp4jack:
> I would prefer if we leave the sample decoding to libsndfile entirely

Good idea! I just forgot that. DLS is a very extensible format, it supports every sample format that .wav supports ([even mp3](https://github.com/pedrolcl/sonivox/pull/6#issuecomment-2566860198))

> Is there a specific reason for using placement new?

Actually no. I will change it to normal new/delete.

rsp4jack:
@derselbst Plz check out the "Issues" part of this PR, and I would like to have your opinions about it. 

> Fluidsynth has a sample cache fluid_samplecache.h, though it's pretty bound to SF2. 

That's the problem. It is designed for SF2 reader, reusing it isn't easy.

> it seems doable to get it into a shape that we can release it with 2.5.0.

Definitely, and I'd like to see that :)

rsp4jack:
@spessasus 

> [File failed to upload.]

where is the file?

spessasus:
Ah, sorry. I currently have really slow internet and I guess 26MB is too much for GitHub.

Here is the file uploaded on catbox:

https://files.catbox.moe/1rnsy9.dls

derselbst (commented):
I completed a first review. Looking good so far, only there's this issue of throwing `std::exception` and catching `std::runtime_error` throughout the file that doesn't seem right to me, pls. have a look.

Tomorrow I'll try to see what can be done to fix the failing CI builds, so I might push to your branch.

rsp4jack:
CI is fixed now.

It is possible to convert keyNumToPitch into scaletuning and a fine tune, as I have written the opposite conversion in libsonivox (https://github.com/pedrolcl/sonivox/blob/77af924f1826af5016e6c6fa8d4aadeac33e3483/arm-wt-22k/lib_src/eas_sf2.c#L1764)

The above two equation can be transformed into:

$$
tuning = scaletuning \times key - scaletuning \times rootkey
$$
$$
tuning = \frac{keynumtopitch}{128} \times key - 100 \times rootkey
$$

So:

$$
scaletuning =  \frac{keynumtopitch}{128}
$$
$$
finetune = scaletuning \times rootkey - 100 \times rootkey
$$

@spessasus Check out your scaletuning conversion at https://github.com/spessasus/spessasynth_core/blob/d81bf8a1319d3b58e35d7f340a8a77e6c4ff6a7e/src/soundbank/basic_soundbank/write_dls/modulator_converter.ts#L219-L225.

rsp4jack:
@spessasus I have never heard that the 0.4 quirk does not exist in DLS. Here is the first region of Grand Piano of the DLS you provided in DirectMusic Producer:

image

The atten field in the region's wsmp is -3145728 = -48 * 65536 = -4.8 dB. Probably it is a bug of Audio Compositor.

spessasus:
> @spessasus I have never heard that the 0.4 quirk exists in DLS. Here is the first region of Grand Piano of the DLS you provided in DirectMusic Producer:
> 
> image
> 
> The atten field in the region's wsmp is -3145728 = -48 * 65536 = -4.8 dB. Probably it is a bug of Audio Compositor.

And take a look at the original SF2. In the first region of the first region of the Grand Piano preset has 0 attenuation set,
And the corresponding instrument's first region (both of these get merged into one) has 4,8 dB of attenuation set. So 4,8 + 0 = 4,8 dB of attenuation, exactly what it is in DLS :-)

rsp4jack:
Ahh I found the problem: I didn't keep -g the same when testing. Sry for that.

rsp4jack:
@spessasus 

> yet it's sometimes inaudible in the fork when playing the HYBRID collage.

I can't reproduce it. 
I can't really tell the difference between SF2 and DLS outputs. If I subtract one by the other, there is difference indeed, but I think it is acceptable.

rsp4jack:
I just found out that RPN 2 is implemented through overriding generator value, therefore region selection process won't be affected. Also fine tune generator cannot be used for https://github.com/FluidSynth/fluidsynth/pull/1626#issuecomment-3217693461 because of this.

rsp4jack:
> Hence I assumed that when you change RPN2, it would also affect keyNumToPitch somehow.

The Key Number Generator is a mechanism to modify the note number before region selection. It's output value is used as the source of Key Num to Pitch.

rsp4jack:
An idea: why not just multiply final key number by keyNumToPitch / 12800.0? It should works together with subtonal tuning.

spessasus:
@rsp4jack 

> @spessasus Check out your scaletuning conversion at https://github.com/spessasus/spessasynth_core/blob/d81bf8a1319d3b58e35d7f340a8a77e6c4ff6a7e/src/soundbank/basic_soundbank/write_dls/modulator_converter.ts#L219-L225.

What's wrong with it?

spessasus:
I get it now. The github mobile app doesn't render the math markdown so I was very confused about those dollar characters. I can see them now that I'm using a browser. But still, thanks!

spessasus:
@rsp4jack 

I have another question. What if the keyNumToPitch connection is defined in the `art1` in the instrument (global region)? It doesn't have a root key. How would you convert that to SF2? Do I assume the root key of 60, or do I have to apply the fineTune generator to all local instrument zones separately, based on their root key?

rsp4jack:
> or do I have to apply the fineTune generator to all local instrument zones separately, based on their root key?

Yes. AFAIK, this is the only way to implement keyNumToPitch in SF2. Assume a root key is bad, it may produce a different pitch from original DLS.

rsp4jack:
Yes, [Airfont](http://www.ronimusic.com/sf2/Airfont_340.dls). I think it is a natively DLS.

rsp4jack:
And also CrysDLS v3.11.dls

spessasus:
Which presets of airfont have it? It has quite a few of them.

rsp4jack:
I checked the DLS implementations I have on my computer: 

* Extreme Sample Converter: normal
* libsonivox: inaudible low in frequency
* DMUSProducer: audible low in frequency
* the fork: audible low in frequency

Probably AirFont is converted from SF2 (the DLS version only appears at ronimusic.com)

spessasus:
@rsp4jack 

So, what do we do about it? 
I think that if the fine tune reaches a large value (let's say, abs(fine tune) >= 8000 cents), we discard it and just set the scale tuning. What do you think?

And also, crysDLS seems to be a native bank, which presets in it have the keyNumToPitch?

@derselbst 

If also like to let you know that rsp4jack's parser contains fixes for the semi-corrupted DLS files. While I know that you prefer to keep fluidsynth very strict (for example, discarding a MIDI file with invalid meta length that's otherwise ignored), I suggest that we keep it that way. Nearly all other DLS software is a very old, proprietary Windows-only application, which makes it a pain to work with on other platforms or integrating it into anything. Making fluidsynth throw an error on even a slightest mismatch of the DLS structure will only make it harder for users with some obscure DLS files to actually use them.

rsp4jack:
If I disable the finetune produced by keyNumToPitch in the fork, it sounds fine.

So, DLS specification does not have a clear description about it, and DMUSProducer seems not caring it.
libsonivox implements keyNumToPitch with a finetune. But DLS on my hand that has keyNumToPitch is using it as scaleTuning.

> I think that if the fine tune reaches a large value (let's say, abs(fine tune) >= 8000 cents), we discard it and just set the scale tuning. What do you think?

Sounds good. I will implement it soon.

spessasus:
keyNumToPitch of 0 is not good for testing as it's just scaleTuning of 0 without any pitch correction AFAIK.
A good test would be some other value.


ESC and sonivox probably reject the file because of the corrupted RIFF chunk our parsers have a fix for.

rsp4jack:
I think that, the pitch calculation is modulated through articulation in DLS. The keyNumToPitch represents `pitch = scale * norm(key)`, and it has a default value 12800. It does `fluid_voice_calculate_pitch`'s job.

spessasus:
Alright, so I'm currently working on rewriting my DLS parser/writer with the keyNumToPitch like sonivox.

Once it works,
I'll create a simple sf2 with some scale tuning and convert it to DLS. Then, if DMUS producer and sonivox sound like the SF2 version, it means that the conversion is correct.

The reason I'm focusing so much on DMUSProd so much, is because it seems to use directMusic directly, which is actually the same synthesizer that powers the infamous MSGS wavetable synth. 
And given that MS was very involved the DLS specification, I'm quite sure that the MS DLS synth is the reference implementation of the DLS standard.

One more thing:
Other programs that use the MS synthesizer are Winamp (the Nullsoft MIDI player, allows custom DLS) and FL Studio's "Fruity LSD" plugin. So if you have those on hand, it may be easier to test than the clunky DMUS producer :-)

spessasus:
@rsp4jack 

Another weird thing I've found:

Fury.dls from [ronimusic](https://www.ronimusic.com/smp_ios_dls_files.htm) has key to attenuation articulators which look like this:

```
Source: velocity linear unipolar positive,
Control: none linear unipolar positive,
Scale: -62914560 >> 16 = -960,
Output transform: concave
Destination: gain
```
Which results in a modulator that looks like this:
image

Which in term results in inaudible sound. I've confirmed that your parser loads them like that (by adding a simple if statement):
```
fluidsynth: atten articulator src invert: 0, src bipolar: 0, scale: 960.000000, ctl invert: 0, ctl bipolar: 0
fluidsynth: atten articulator src invert: 0, src bipolar: 0, scale: 960.000000, ctl invert: 0, ctl bipolar: 0
fluidsynth: atten articulator src invert: 0, src bipolar: 0, scale: 960.000000, ctl invert: 0, ctl bipolar: 0
fluidsynth: atten articulator src invert: 0, src bipolar: 0, scale: 960.000000, ctl invert: 0, ctl bipolar: 0
...
```
[Audio compositor](https://web.archive.org/web/20040121121412/http://home.att.net/~audiocompositor/) also reports that articulator:
image

I had to add a special handling for this to manually invert the transform for vel, cc7 and cc10 as specified in table 3.

Yet I can't see any special code in your fork for manually inverting the source for it to work and it seems to be perfectly audible. Any idea why?

derselbst (commented):
I gave this a listening test today and I'm really delighted! The audible difference compared to libinstpatch is totally negligible. Excellent job, thanks a lot! I have merged against master and fixed that one conflict. Let me know when you think it's ready to go! I would give probably it a final `make format` before merging it.

spessasus:
> The audible difference compared to libinstpatch is totally negligible. 

I'd say it's quite significant (in a good way), given libinstpatch's lack of envelope and keygroup support.

Speaking of which, is libinstpatch getting removed with this PR too? I'd say that yes, given that not only having two DLS parsers could be confusing and unnecessary, but the native parser is better in every single way. But what do you think? @derselbst @rsp4jack

src/sfloader/fluid_dls.cpp
```diff
@@ -0,0 +1,2851 @@
...
+        // velocity -> filter cutoff disabled
+        // actually it is already disabled by fluidsynth
+        mods.push_back(fluid_mod_t{
+        GEN_FILTERFC, FLUID_MOD_VELOCITY, FLUID_MOD_GC, 0, 0, FLUID_MOD_TRANSFORM_LINEAR, 0, nullptr, nullptr, nullptr });
+        // vib lfo --(midi channel pressure)-> pitch disabled
+        mods.push_back(fluid_mod_t{
+        GEN_VIBLFOTOPITCH, FLUID_MOD_CHANNELPRESSURE, FLUID_MOD_GC, 0, 0, FLUID_MOD_TRANSFORM_LINEAR, 0, nullptr, nullptr, nullptr });
+        // pitch wheel --(rpn 0)-> pitch 12800 cents
+        // see also the comment in gen.h about GEN_PITCH
+        mods.push_back(fluid_mod_t{ GEN_FINETUNE,
```
spessasus, line 274: While DLS specifies 12800 cents instead of 12700, I think that given fluidsynth uses the SF2 synthesis model we should stick with the SF2's 12700 cents to make sure that everything is in tune.

In fact,
[this file for testing RPN Range](https://github.com/jazz-soft/test-midi-files/blob/main/midi/test-rpn-00-00-pitch-bend-range.mid)

Is slightly out of tune when compared to an SF2 conversion of gm.dls. Render the two, invert one in audacity and mix them down (essentially a PCM difference). then you can hear that they are silent when no pitch bend is applied (meaning no difference in PCM), but the more pitch bend is applied, the louder it gets meaning more detune.

spessasus: @derselbst, what do you think? Does fluid's internal sound bank structure directly replicate the soundfont model? If so, then changing the default pitch modulator for DLS will result in the DLS files being detuned from their SF2 counterparts.

rsp4jack:
I just found a bug so all modulators different from default modulators are ignored. I will push a fix.

src/sfloader/fluid_dls.cpp
derselbst: My naïve thinking is, that a DLS accurately converted to SF2 should yield the same synthesized audio output. Hence, I would vote to indeed specify the SF2 default of 12700 cents here, i.e. removing the modulator override here.

The same thought could be used to argue against overriding chorus and reverb modulators as well.

If a DLS or "DLS converted to a SoundFont" is serious about its modulator amounts, the DLS / Soundfont is free to specify / override the modulator(s) in the file itself. Otherwise it's subject to the defaults imposed by the synthesis model it will be used in.

Thinking this further, an additional argument comes to my mind: Fluidsynth provides an API for manipulating default modulators. If someone wanted to use this API to e.g. change the amount of the default pitch modulator to any other value, they would only be able to do so in the SF2 world, but not for DLS, since the modulator-override is hard-coded here.

I don't know about @rsp4jack, but I'd vote for not overriding any SF2 default modulators with DLS specific ones. (PS: You can still keep this code, but `#if 0`ed-out for reference and documentation purposes.)

spessasus: I definitely agree with rsp4jack, I think that if a user uploads a DLS file, they will probably expect accurate playback and won't mess with fluidsynth API. If they really need that, they can just convert the file to an SF2 file and mess with it. The documentation could explicitly say that DLS doesn't work with the API so the sound bank needs to be converted into an SF2 one. Sacrificing the most likely use-case (just using a DLS file) for something relatively obscure (messing with modulators on an API level, something which most users, probably using something like Qsynth, will not do) is a bad trade IMO.

Though derselbst is the one who has write access to this repo so our opinions don't really matter...

rsp4jack:
'pgal' chunk in MobileBAE DLS banks is supported now. I tested Nokia LLoyd Bank DLS and it sounds good.

src/sfloader/fluid_dls.cpp
derselbst: > Though derselbst is the one who has write access to this repo so our opinions don't really matter

Your opinions do matter, I appreciate them and will account for them when proposing an alternative approach to this.

rsp4jack:
@SmithGoll Thanks for your help! I have pushed a fix.

I think this PR is complete now.

src/sfloader/fluid_dls.cpp
```diff
@@ -0,0 +1,3012 @@
...
+        {
+            keyToPitch.amount = mod.amount;
+            return;
+        }
+
+        for (auto &existing_mod : mods)
+        {
+            if (fluid_mod_test_identity(&existing_mod, &mod))
+            {
+                existing_mod.amount += mod.amount;
```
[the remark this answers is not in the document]
rsp4jack: Oh this is definitely a bug. According to DLS spec it should be overwriting. Thanks!

src/sfloader/fluid_dls.cpp
```diff
@@ -0,0 +1,3013 @@
...
+              dlsid.Data4[2],
+              dlsid.Data4[3],
+              dlsid.Data4[4],
+              dlsid.Data4[5],
+              dlsid.Data4[6],
+              dlsid.Data4[7]);
+    return std::nullopt;
+}
+
+inline bool fluid_dls_font::execute_cdls(fluid_long_long_t offset, int size)
```
spessasus, line 1797: @rsp4jack 
Are there actually any DLS files that make use of conditional chunk? Is there a point in parsing it at all?

To me it sounds like a possibility of some DLS being extracted from some proprietary software, which could declare the entire collection as only playable on some random manufacturer ID even though fluidsynth could play it just fine, but CDL would tell it to ignore it. And from what I can see in the code, if CDL fails, it throws an error. Unless I missed something, that is.

spessasus:
> I'm not a fan of breaking functionality that some users might be used to from one release to another. 

And how is that a breaking change exactly? From my understanding, libinstpatch was used for DLS only so no regressions are introduced.

derselbst:
Fluidsynth 2.4.8 can be "easily" compiled for Windows9x - _with_ libinstpatch support. Fluidsynth 2.5.0 will require a C++11 compiler at least. This native DLS parser will even require a C++17 compiler. That means you can no longer target those legacy platforms like Win9x. Is that a problem? I don't know. But I don't want to find out by scratching all that away from now to then. Esp. since there is no technical necessity that justifies removing of either glib or libinstpatch right now.

Topic settled.

rsp4jack:
@spessasus It is reasonable to ignore CDLs but I would like to keep it for now. If there are problems reported with it, I will consider create a patch.

PR #1658, opened 2025-09-30
Add public API for dynamic real-time MIDI playback control

nehochupechatat:
This pull request adds support for:
- A public API that would allow the end user to control MIDI playback realtime by muting/unmuting tracks and channels, transposing them and sending pitch bends/progrsm changes/control changes. Very useful for video game developers in particular.
- 'loopstart'/'loopend' markers in MIDI files (case-insensitive). Many MIDIs use this.

### DIFF
--- a/src/midi/fluid_midi.c
+++ b/src/midi/fluid_midi.c
@@ -26,8 +26,7 @@
 static int fluid_midi_event_length(unsigned char event);
 static int fluid_isasciistring(char *s);
 static long fluid_getlength(const unsigned char *s);
-
-
+unsigned char casestrcmp(const char *s, const char *t);
 /* Read the entire contents of a file into memory, allocating enough memory
  * for the file, and returning the length and the buffer.
  * Note: This rewinds the file to the start before reading.
@@ -54,7 +53,8 @@ static int fluid_player_load(fluid_player_t *player, fluid_playlist_item *item);
 static void fluid_player_advancefile(fluid_player_t *player);
 static void fluid_player_playlist_load(fluid_player_t *player, unsigned int msec);
 static void fluid_player_update_tempo(fluid_player_t *player);
-
+int fluid_player_reset_channel_mute(fluid_player_t *player);
+int fluid_player_reset_track_mute(fluid_player_t *player);
 static fluid_midi_file *new_fluid_midi_file(const char *buffer, size_t length);
 static void delete_fluid_midi_file(fluid_midi_file *mf);
 static int fluid_midi_file_read_mthd(fluid_midi_file *midifile);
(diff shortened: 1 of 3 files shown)


src/midi/fluid_midi.c
```diff
@@ -1719,8 +1816,8 @@ new_fluid_player(fluid_synth_t *synth)
     player->end_pedals_disabled = 0;
     player->last_callback_ticks = -1;
     fluid_atomic_int_set(&player->seek_ticks, -1);
-    fluid_player_set_playback_callback(player, fluid_synth_handle_midi_event, synth);
-    fluid_player_set_tick_callback(player, NULL, NULL);
+    fluid_player_set_playback_callback(player, fluid_synth_handle_midi_event, player);
```
derselbst, line 1819: This effectively passes the `player` object to `fluid_synth_handle_midi_event`, which however expects a `synth` object. This will effectively crash fluidsynth, which our failing CI builds also prove. It also means that your statement to have "thoroughly tested this" is not credible!

src/midi/fluid_midi.c
```diff
@@ -1719,8 +1816,8 @@ new_fluid_player(fluid_synth_t *synth)
...
+    fluid_player_set_tick_callback(player, fluid_player_handle_loops, player);
```
derselbst, line 1820: It should be documented in `fluid_player_set_tick_callback` that loops will no longer work if the user registers his own tick_callback. Which is why this function was invented.

src/midi/fluid_midi.c
```diff
@@ -2884,3 +3005,414 @@ fluid_midi_event_length(unsigned char event)
...
+
+int fluid_player_handle_loops(void *data, int tick)
+{
+	fluid_player_t *player = (fluid_player_t *)data;
+
+	int total_ticks = fluid_player_get_total_ticks(player) - 1;	/* subtract one for safety */
+	
+	/* return if loopstart tick is undefined in MIDI or */
+	/* loopend tick is undefined in MIDI and player hasn't reached song end tick */
+	if(!player->loopstart_tick || (!player->loopend_tick && tick < total_ticks))
```
derselbst, line 3398: So this means that if your loopstart is at the very beginning, the very first tick of the MIDI file, your looping logic won't work, wouldn't it?

src/midi/fluid_midi.c
```diff
@@ -1617,23 +1661,63 @@ fluid_track_send_events(fluid_track_t *track,
         }
         else
         {
-            if(player->playback_callback)
+			
+			/* Transpose MIDI channels by user-defined amount of semitones */
+			if ((event->type == NOTE_ON || event->type == NOTE_OFF) && player->transposed_ch[event->channel] && player->transpose != 0)
+			{
+				/* Limit transposition amount to avoid sending invalid MIDI note numbers */
+				if(player->transpose < -24) {player->transpose = -24;} else if (player->transpose > 24) {player->transpose = 24;}
```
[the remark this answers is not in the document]
nehochupechatat: Because there are MIDI files that use basses with note range lower than 36 (octave 24-36).

nehochupechatat:
While I have fixed most of the glitches there's just one problem.

Even if I remove the "check if loopstart tick is zero" function, my testing with fluidsynth.c still shows that if there are no markers, it will still just loop with the "playlist" mechanism, resetting everything in the process.

The track muting is now determined through player->ntracks, but that's not very convinient because it only gets it upon starting. And since it's asynchronous, you can't just call play and then get the track number... I was perhaps thinking of moving the ntrack declaration somewhere earlier?

Yes, I changed the APIs to your suggestions, but I kept the midi Ctrl change enums so game developers could send custom Control Change events. Will commit soon

derselbst:
I'm closing this now. For the future, pls. consult the [Contributing guideline](https://github.com/FluidSynth/fluidsynth/blob/master/CONTRIBUTING.md) of a project, before actually contributing anything. I admit, our guidelines are a bit out of date. Yet, it would have prevented at least 40% of code-related problems in this PR, if you had read it.

PR #1661, opened 2025-10-01
Convert gcem from submodule to subtree

derselbst:
We currently distribute the source code for fluidsynth releases via the auto-generated GitHub tarballs. Unfortnately, those do not include submodules. I.e. they currently do not include gcem, which however, is required to compile fluidsynth.

To solve this, we could convert gcem from submodule to subtree. I'm not sure if it's wise to do this, though.


derselbst:
I thought about it and changed my mind. I don't like it. It essentially adds 12000 lines to our codebase. Their typos now become our typos and break codespell CI.

PR #1662, opened 2025-10-02
Automatically download missing gcem during build

derselbst:
This is an alternative approach to #1661. This PR will allow people to continue compiling fluidsynth from the auto-generated source archive, by downloading gcem from Github during the build.

If that fails, it prints an error telling the user what to do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -362,6 +362,8 @@ if ( WIN32 )
   endif  ( MINGW )
 endif ( WIN32 )
 
+find_package ( GCEM REQUIRED )
+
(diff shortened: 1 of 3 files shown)


derselbst:
> In some Linux distros file(DOWNLOAD) is forbidden to run at build time, for instance: in Debian

In this case the packager is supposed to  either checkout with submodules, or to manually download gcem from the provided path, put it into the source tree next to the fluidsynth tarball and add a prepare step to unpack gcem.zip into the `/gcem` directory as advised by the error message. That should be pretty lightweight.

No need to have a packaged version of gcem available for the distro. However, if so, `find_path` should pick it up automatically.

PR #1663, opened 2025-10-04
Fix drum kits not being selected after SysEx GS Rythm Part

### LINKED ISSUE #1579: Drum kits are not being selected after SysEx GS Rythm Part
### FluidSynth version

```sh
2.4.5
```

### Describe the bug

I have created a file for testing Microsoft GS Wavetable Synth (MSGS). It uses GS SysEx messages to set channels 0 and 1 to drum mode. When I run the file with Windows Media Player, it plays correctly: I'm hearing an orchestral crash cymbal and a gated snare. But, when I play it with FluidSynth, I hear strings and an organ instead.

### Expected behavior

The attached file must play the same way as in Windows Media Player, that is connected to MSGS.

### Steps to reproduce

Download attached file, unpack it and play it with FluidSynth.

[D_DM2TTL - GS Drums.zip](https://github.com/user-attachments/files/20609714/D_DM2TTL.-.GS.Drums.zip)

### Additional context

Here are relevant parts from running FluidSynth with `-v`:
```console
fluidsynth: === systemreset ===
fluidsynth: Processing SysEX GS DT1 message, bank selection mode might have been changed.
fluidsynth: debug: SysEx DT1: setting MIDI channel 0 to type 1
fluidsynth: prog                0       0       0
fluidsynth: Processing SysEX GS DT1 message, bank selection mode might have been changed.
fluidsynth: prog                0       0       48
fluidsynth: debug: Deselected preset 'Chateau Grand' from channel 0
fluidsynth: debug: Selected preset 'Fast Strings' on channel 0
```
and
```console
fluidsynth: cc          0       7       100
fluidsynth: debug: SysEx DT1: setting MIDI channel 1 to type 1
fluidsynth: prog                1       0       0
fluidsynth: Processing SysEX GS DT1 message, bank selection mode might have been changed.
fluidsynth: prog                1       0       16
fluidsynth: debug: Deselected preset 'Chateau Grand' from channel 1
fluidsynth: debug: Selected preset 'Drawbar Organ' on channel 1
```

derselbst:
Previously, fluidsynth changed the channel type and sent a progchange. This was not sufficient to switch to the drum bank, as the previously selected bank was still cached. This PR now explicitly resets the bank, causing a subsequent progchange to select the correct preset from the Soundfont.

Note that this "subsequent progchange" is expected to be sent by the user, cf. SC-88Pro/8850 owner's manual, site 60:

> "To select a drum set after setting the part mode, transmit a program change [...]"

Fixes #1579.

### DIFF
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -2606,15 +2606,15 @@ fluid_synth_sysex_gs_dt1(fluid_synth_t *synth, const char *data, int len,
             // used in the selected mode. This behavior is not replicated here.
             // Issue 1579: The preset selected for the channel needs to be forcibly changed. Therefore it is not sufficient
             // to send a prog change, as the old bank is still active in the channel.
-            // Also, do not explicitly send a prog change here. It must be sent by the user, see note on site 60:
+            // MSGS selects the standard drum kit right after this message (which is equivalent to sending a prog change 0).
+            // However, this behavior is insonsistent with the note on site 60:
             // "To select a drum set after setting the part mode, transmit a program change [...]"
+            fluid_synth_cc_LOCAL(synth, chan, ALL_CTRL_OFF);
             fluid_channel_set_sfont_bank_prog(synth->channel[chan],
                                               -1,
                                               type == CHANNEL_TYPE_DRUM ? DRUM_INST_BANK : 0,
                                               -1);
-            fluid_synth_cc_LOCAL(synth, chan, ALL_CTRL_OFF);
-            // Unset the currently selected program, so that this channel remains silent, in case the user doesn't send a progchange afterwards
-            fluid_synth_unset_program(synth, chan);
+            fluid_synth_program_change(synth, chan, 0);
         }
         return FLUID_OK;
     }
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -2200,7 +2200,7 @@ fluid_synth_sysex(fluid_synth_t *synth, const char *data, int len,
                                           handled, dryrun);
         if(synth->verbose)
         {
-            FLUID_LOG(FLUID_INFO, "Processing SysEX GS DT1 message, bank selection mode might have been changed.");
+            FLUID_LOG(FLUID_INFO, "Processed SysEX GS DT1 message, bank selection mode might have been changed.");
         }
         FLUID_API_RETURN(result);
     }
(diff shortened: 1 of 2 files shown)


RoqueDeicide:
Hello there.

The proposed behavior is inconsistent with MSGS. The latter will always switch the channel to the `Standard` drum kit immediately after receiving `Use for Rhythm Part` SysEx message, even if no `Program Change` message is ever received. This means, that MIDI files, that were only tested with MSGS may break with FluidSynth.

I've looked at the manual's page 60, and I feel there is an ambiguity here: the statement "To select a drum set after setting the part mode, transmit a program change to part 11" doesn't specify, what happens, if no "Program Change" message is sent. Considering, that Roland made both SC-8850 and MSGS, I believe, that MSGS's behavior is the one, that's correct here.

derselbst:
Ok, so you're suggesting to send a progchange 0 right afterwards, instead of unsettling the program?

RoqueDeicide:
As far as MSGS behavior is concerned, it doesn't apply any program change, as "Program Change" message triggers an update to the bank number. For instance, here is a MIDI file in pseudocode:

```
SysEx GS Reset
Bank Select MSB 50
SysEx GS Rhythm Part that sets channel 0 to drum mode
Play a note 59
SysEx GS Rhythm Part that sets channel 0 to melodic mode
Play a note 59
SysEx GS Rhythm Part that sets channel 0 to drum mode
Play a note 59
SysEx GS Rhythm Part that sets channel 0 to melodic mode
Play a note 59
```
The above MIDI file, when played with MSGS, will produce: 1 crash cymbal, followed by 1 note 59 of Piano 1, followed by 1 crash cymbal again, followed by a piano note again.

On MSGS, and all compliant GS and XG synths, `Bank Select` messages don't have any effect, until `Program Change` message is received. All SysEx drum mode messages are supposed to take effect immediately, by themselves and without any implied `Program Change` messages.

My suggestion, is to load Standard drum kit / Piano 1 preset upon the switch of the channel to and from drum mode, with as few side effects as possible.

RoqueDeicide:
Here is some more detail:

On MSGS, the bank number is first provided by `Bank Select` messages, it is then set in stone by `Program Change` message. The bank number will then persist through drum mode changes, but only have an effect, when the channel is in melodic mode.

To illustrate this behavior, I've made a MIDI file, on MSGS it plays:

 1) Cymbal
 2) Piano key
 3) Cymbal
 4) Sine wave
 5) Cymbal
 6) Silence

The "Sine wave" is selected, because the bank number, that was specified by the `Bank Select MSB` message, that was sent before the change to drum mode, gets committed by the `Program Change 80` message. The 6th note is silent, because `Program Change 0`, that was sent before the switch to melodic mode, commits the `Bank Select MSB 64` message, that was sent right after the sine wave was selected, causing MSGS to try to load a preset at program 0 and bank MSB 64, which doesn't exist in `gm.dls` file, and MSGS has no "Capital Tone Fallback" feature. Cymbals keep playing throughout the file, as the bank number is ignored in drum mode.

[GS Drums Bank Select and Program Change interaction on MSGS.zip](https://github.com/user-attachments/files/22702799/GS.Drums.Bank.Select.and.Program.Change.interaction.on.MSGS.zip)

derselbst:
Thanks for the test, I've added it to our other GS DT 1 test files and link it to your comments.

> The 6th note is silent, because Program Change 0, that was sent before the switch to melodic mode, commits the Bank Select MSB 64 message, that was sent right after the sine wave was selected, causing MSGS to try to load a preset at program 0 and bank MSB 64, which doesn't exist in gm.dls file, and MSGS has no "Capital Tone Fallback" feature.

This behavior is unfortunately not possible with fluidsynth for two reasons:

* Fluidsynth currently doesn't "remember" the bank that the channel had when it was melodic.
* Even it if had remembered bank 64, Fluidsynth's preset fallback logic would fall back to bank 0 as soon as it receives the very last prog 0.

So I guess this will remain a detail where fluidsynth differs from MSGS.

PR #1677, opened 2025-10-15
Add build script for iOS

withSang:
This PR enables building a FluidSynth dynamic framework for iOS.
- The script downloads the iOS CMake toolchain (https://github.com/leetal/ios-cmake) to generate the XCode project to build the framework. Since the toolchain sets `CMAKE_SYSTEM_NAME` to `iOS`, some parts of CMakeLists.txt were modified to support that case.
- Since `<CoreAudio/AudioHardware.h>` is not available for iOS, SDL3 instead of CoreAudio is built alongside Fluidsynth as the audio driver.

There have been forks that made iOS build possible (like https://github.com/DominusExult/fluidsynth-sans-glib), but they do not support FluidSynth v2.5.0+ since they modified the sources heavily.

I'm an Android / iOS developer working on a music-playing app and have successfully incorporated your library in my app with the build script. I want to contribute back to enable other iOS developers as well to build and use the library easily.

An example app can be provided if needed.
Sadly, I haven't tried Azure CI before. If you can help me a little, I can provide iOS framework CI checks as well.

### DIFF
--- a/contrib/ios_build.sh
+++ b/contrib/ios_build.sh
@@ -59,7 +59,8 @@ cmake ../.. \
     -DSDL3_DIR="${SDL3_CMAKE_DIR_IOS}" \
     "${CMAKE_COMMON_FLAGS[@]}"
 
-xcodebuild -project FluidSynth.xcodeproj -target libfluidsynth -configuration Release -sdk iphoneos build 2>&1
+xcodebuild -project FluidSynth.xcodeproj -target libfluidsynth -configuration Release -sdk iphoneos \
+    build DEBUG_INFORMATION_FORMAT="dwarf-with-dsym" GCC_GENERATE_DEBUGGING_SYMBOLS=YES 2>&1
 cd ..
 
 # Build for iOS simulator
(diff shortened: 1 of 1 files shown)


withSang:
> That looks pretty good, thanks! I really need a CI pipeline for this, otherwise this build script will just become some unmaintained junk of code.
> 
> I'm migrating away from AzureCI and prefer to use GitHub Actions by now. Have you had experience with GitHub Actions and could you provide a CI pipeline on your own?
> 
> @DominusExult FYI

Thank you for responding! I also agree with your point.
I'm familiar with GitHub Actions. I'll let you know again when I submit the CI pipeline as well 😄

withSang:
> Another note on xcframework, you need to make sure to include the dSYM files for distribution.
> See the discussion at https://github.com/libsdl-org/SDL/issues/13573 and my accepted pull request for SDL3.

Thanks! I've added it as well :-)

derselbst (approved):
Many thanks for your efforts! LGTM, I'll merge it now to the `ios` branch, to add a build badge to the readme and publish the binaries as mentioned.

PR #1718, opened 2025-12-08
[RFC] Add OSAL_posix (mostly ported from glib)

neoharp-dev:
Glib is a heavy dependency
cpp11 have performance issues on Android

So I yanked futex code from glib.  Posix threads and sync primitives are used if futexes are not available.

It is used in our app, and works in both Android and iOS (although currently iOS app is still in testing), perforamce issues we had previously with cpp11 were not observed with OSAL_posix.


derselbst (commented):
Thanks for this enormous work! I am not yet sure, if I want to merge this though. For me, the underlying use-case (solving performance issues on Android) is too specific. I'd keep this open a bit as an RFC allowing others to provide feedback.

rsp4jack:
setenv is used in pulse audo driver but not implemented in cpp17 osal, so implementing it here is good.

PR #1719, opened 2025-12-08
Add limiter support (with limiter from signalsmith-audio-basics)

neoharp-dev:
This adds limiter support by using Signalsmith Audio Basics C++ header-only library (see https://signalsmith-audio.co.uk/code/ and https://github.com/Signalsmith-Audio/basics)

Path to Signalsmith basics lib have to be supplied to cmake in the environment variable SIGNALSMITH_AUDIO_BASICS_PATH

The limiter is applied in fluid_rvoice_mixer_process_fx after chorus and reverb effects. In the audio drivers that mix effects and dry audio in one buffer, the limiting will be applied to the resulting buffer. In audio drivers that do this separately (for example, Jack driver), the limiting will be applied only to the dry buffer


src/rvoice/fluid_limiter.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+
+#ifndef _FLUID_LIMITER_H
+#define _FLUID_LIMITER_H
+
+#ifdef LIMITER
+
+// how many channel buffers to process at once
+// maybe if we want to paralellize, we might set this to 1
```
derselbst:
```suggestion
// maybe if we want to parallelize, we might set this to 1
```

derselbst (commented):
I played around with that in the past days a bit and it works quite fine. I've made several changes but were unable to push them to your branch, so I put them [here](https://github.com/FluidSynth/fluidsynth/tree/add-limiter-support). Pls. merge it, I'll need to think about which commits will survive in the end (not quite done yet).

Basically, signalsmith-audio is now a submodule to simplify building it. They require C++20 because of `std::atomic_flag::test()`, fixed that as well. I have documented (and renamed) the new settings, pls. have a look. I have removed `synth.limiter.input-gain` , because that's the job of `synth.gain`.

neoharp-dev:
macos builds seem to fail because of some intermittent error unrelated to the code

geraintluff:
> They require C++20 because of std::atomic_flag::test()

Sorry about this!  Fixed now, so it should work with C++17 again.

PR #1720, opened 2025-12-08
Add iOS support to coreaudio driver

neoharp-dev:
Previously coreaudio driver was working only on MacOSX, and only way to get sound on iOS was with SDL3, which is a heavy dependency

This adds support for iOS in coreaudio driver.

It introduces `COREAUDIO_SUPPORT_HAL` preprocessor variable that indicated if coreaudio has HAL (that is, this is MacOSX), or not

If presense of HAL is not detected:

there is only one possible audio output, so `audio.coreaudio.device` is not populated.

`componentSubType` is set to `kAudioUnitSubType_RemoteIO`. This name is confusing (like, what is 'remote' about it ?), but apparently this is the way to go on iOS.

Works in test version of our iOS app (app is not yet live on app store)

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -39,15 +39,6 @@ jobs:
     - uses: actions/checkout@v5
       with:
         submodules: recursive
-    - name: Build Dependencies
-      run: |
-        git clone https://github.com/libsdl-org/SDL.git && \
-        pushd SDL && \
-        git checkout preview-3.3.2 && \
-        xcodebuild -project Xcode/SDL/SDL.xcodeproj -target "SDL3.xcframework" -configuration Release SYMROOT="build/" && \
-        mkdir build && \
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ elseif(NOT BUILD_SHARED_LIBS)
   message(WARNING "Your version of CMake is very old. This may cause linking issues if your dependencies are not in your compiler's default search paths.")
 endif()
 
-project ( FluidSynth C CXX )
+project ( FluidSynth C CXX OBJC)
```
derselbst: This unconditionally enables objective-C globally. Instead, pls. conditionally `enable_language` somewhere in `if (enable-coreaudio)` below.

neoharp-dev: fixed by enabling it in src/CMakeLists.txt right where the .m file is added to fluid_coreaudio_SOURCES

PR #1736, opened 2025-12-23
Update API docs about possible hazards when changing no-realtime settings

derselbst:
Following the proposal of #1726.


derselbst:
@jimhen3ry Feel free to have a look. Let me know if you had other places in mind where the docs should be updated.

PS: I cannot / will not change the behavior of the existing `fluid_settings_set* functions`.

doc/fluidsettings.xsl
jimhen3ry: I don’t think a global wording change for <realtime/> is really needed. The existing text already communicates the intended “runtime-changeable” contract. Since this boilerplate applies to all settings, leaving it as-is feels safer.

jimhen3ry (commented):
These changes are a useful clarification of the settings lifecycle and the distinction between construction-time and runtime behavior. Thank you for taking the time to prepare these to address my concern. 

My suggestions above are intended as wording polish to make your changes a bit more clear, not to change the underlying intent.

Separately, I have two small documentation changes that I made as I was documenting the audio.sample-format extensions that seem aligned with the goal of this PR:
– a brief lifecycle/ordering clarification in audio_driver.txt, and
– an explicit note in the synth.sample-rate setting entry in fluidsettings.xml about avoiding synth/driver desynchronization when changing the sample rate after synth creation.

Those files aren’t touched in this PR, so I didn’t leave inline suggestions, but I’m happy to paste the diffs here or open a small companion PR if you want to see these files. Either way works for me.

PR #1756, opened 2026-02-07
Fix SoundFonts failing to load with old MinGW versions

### LINKED ISSUE #1755: Unable to load Soundfonts with MinGW 8.1
### FluidSynth version

```sh
<=2.5.2
```

### Describe the bug

When compiling fluidsynth with MinGW 8.1, fluidsynth will be unable to load SoundFonts. (Possibly also MIDIs and DLS). That's because when fluidsynth seeks in a file using `SEEK_CUR`, `_fseeki64` is used, which in turn calls `_ftelli64` to seek from the current offset. However, `_ftelli64` reports incorrect values in this toolchain, see: https://sourceforge.net/p/mingw-w64/bugs/864/

This causes fluidsynth to seek to the wrong offset, read wrong data, and ultimately fail.

### Expected behavior

SEEK_CUR uses the correct file offset, as reported by `ftell()` or `fgetpos()`.

### Steps to reproduce

See unit test and MinGW8.1 CI build in linked PR.

### Additional context

_No response_

derselbst:
This PR fixes #1755. Particularly:

* Add a unit test for reading and seeking within a file, similar to what the SF2 load does.
* Add a unit test for printing `sizeof` and `alignof` various datatypes and asserting some of them.
* Add MinGW CI builds that would have previously failed.
* Split the `READCHUNK` into two separate read calls to guard against any possibly introduced extra padding in the `SFChunk` struct.
* Treat Soundfonts, DLS and MIDIs as binary files, to prevent git from having to detect their file type automatically, prevent diffing and auto crlf.

### DIFF
--- a/test/test_file_seek_tell.c
+++ b/test/test_file_seek_tell.c
@@ -10,7 +10,7 @@ void noop(void)
 }
 
 // A unit test that fails when compiled with MinGW32 8.1.0, due to inconsistencies when using fseek(SEEK_CUR).
-// Mimick the simple behavior of the sfont loader to test seek and tell are working as expected.
+// Mimic the simple behavior of the sfont loader to test seek and tell are working as expected.
 int main(void)
 {
     fluid_long_long_t pos, size;


src/utils/fluid_sys.c
```diff
@@ -1641,6 +1641,64 @@ fluid_long_long_t fluid_file_tell(FILE* f)
...
+
+        return FLUID_FAILED;
+    }
+
+    return FLUID_OK;
+}
+
+int fluid_file_seek(FILE *fd, fluid_long_long_t ofs, int whence)
+{
+#if (defined(__MINGW32__) || defined(__MINGW64__)) && defined(__GNUC__) && (__GNUC__ < 15)
```
sezero: gcc version has nothing to do with mingw version. That additional `__MINGW64__` check is also redundant, because `__MINGW32__` is always defined even for win64. I suggest that you guys simplify the condition to `#ifdef __MINGW32__`

derselbst: I don't quite understand your comment. As said in the PR description, I wrote a unit test and added a CI build that would have without this logic. After adding this macro `#ifdef` logic, the unit test passed for that particular build. If the gcc version was truly unrelated to the mingw version, the test should have continued to fail. I therefore conclude that MinGW does indeed define `__GNUC__`. This is backed by the fact that MinGW 8 reports as GCC: https://github.com/FluidSynth/fluidsynth/actions/runs/22018244095/job/63623147113#step:4:74

Solely using `#ifdef __MINGW32__` is insufficient, as I want to apply this logic only for old MinGW. I chose version 15, because the other MinGW-based CI builds we have use version 15 and do not suffer from this problem.

I do not mind that `__MINGW64__` might be redundant.

sezero: > The history of both files show that they were added in 2025. 

The history goes older: Follow the link at the end on github ui.

>  By the way: what is the macro to know the MINGW version/runtime version?

https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/_mingw_mac.h#L17-L42

We usually detect like this:

```c
#ifdef __MINGW32__
#include <_mingw.h>
  #ifdef __MINGW64_VERSION_MAJOR /*  this is mingw-w64 */
    /* do stuff.. */
  #else                          /* this is old/ancient mingw */
    /* do other stuff */
  #endif
#endif
```

derselbst: Apologize Pedro, here's the correct SF link, I'll fix it in the code later (can't explain how to got wrong there): https://sourceforge.net/p/mingw-w64/bugs/864/

I understand that solely relying on the MinGW's GCC compiler version for this workaround is not sufficiently accurate. I can only tell from the two different types of MinGW based CI builds and therefore I assumed that the entire MinGW toolchain would be usually kept in some sort of version-sync.

However, I'm not that familiar with MinGW. If you think it's better to always apply for MinGW I'm open to change it.

PR #1758, opened 2026-02-10
Automate manual audio regression comparisons

derselbst:
The manual rendering tests previously required human inspection and could not be executed in CI. Inspired by the discussion in #1747, I fired up copilot to introduce an automated workflow that compiles a reference FluidSynth revision, renders the audio with the reference and current version of FluidSynth, and compares SNR/RMS/ABS metrics, without requiring versioned "golden reference" files.

  - Allow `check_rendering` to run against a caller-specified `fluidsynth` binary via `MANUAL_TEST_FLUIDSYNTH`.
  - Add `test/run-manual-regression.sh` to build current + reference revisions, render both sets, and compare with SoX metrics and thresholds.
  - New Linux workflow `rendering-regression.yml` to run the regression comparison with configurable reference ref and thresholds.

Example usage:

```bash
REFERENCE_REF=HEAD~1 SNR_MIN=60 RMS_MAX=0.0001 ABS_MAX=0.01 test/run-manual-regression.sh
```

TODO:

- [x] test suite currently fails when and old version of fluidsynth is being passed new / unknown commandline options, like `-o synth.limiter.active=1`
- [x] SNR, RMS and abs diff need to be evaluated to get meaningful values

### DIFF
--- a/.github/workflows/rendering-regression.yml
+++ b/.github/workflows/rendering-regression.yml
@@ -18,7 +18,7 @@ permissions:
 
 env:
   BUILD_TYPE: RelWithDebInfo
-  REFERENCE_REF: 'v2.4.8'
+  REFERENCE_REF: 'v2.5.1'
   SNR_MIN: '60'
   RMS_MAX: '0.0001'
   ABS_MAX: '0.01'


derselbst:
Why did I close this?

PR #1772, opened 2026-03-21
Increase compatibility with MTS by implementing MIDI RP-020

derselbst:
This implements MIDI RP-020, by creating an implicit-default tuning bank/program 0/0. This allows MIDI files to mess around with tuning 0/0 directly and without having to activate it via RPN 0x4 / 0x3.

### DIFF
--- a/test/test_mts_cc_tuning.c
+++ b/test/test_mts_cc_tuning.c
@@ -6,7 +6,7 @@
 #include "fluid_chan.h"
 #include "fluid_tuning.h"
 
-void verify_equal_temperament(fluid_synth_t *synth)
+void verify_equal_temperament(const fluid_synth_t *synth)
 {
     int i;
     /* Verify that the default tuning is equal-tempered (pitch[key] == key * 100 cents) */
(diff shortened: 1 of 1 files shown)


derselbst:
Problem 2 is fixed by #1774. Problem 1 is actually not a problem. `fluid_voice_calculate_pitch()` should not take the sample fine-tune into account.

Regression tests now report the following diff between c36b8e5202f28ed1645402d163a8b6df8253f560 and this PR:

```
File                                                                            SNR          RMS          ABS
mts/issue102_marcus.wav                                                        1.55     0.045736     0.628815
  SNR below threshold (60)
  RMS above threshold (0.0001)
  ABS above threshold (0.01)
mts/issue102_orig.wav                                                        -35.96     0.114550     0.786896
  SNR below threshold (60)
  RMS above threshold (0.0001)
  ABS above threshold (0.01)
rpn/SA2 - Prison Lane -- MIDIMan - FluidSynth Test.mid.wav                    24.49     0.006092     0.058685
  SNR below threshold (60)
  RMS above threshold (0.0001)
  ABS above threshold (0.01)
```

The diff in the MTS test files is expected. The difference in Prison Lane results from the Teremin (?) instrument having a slightly different pitch now, presumably due to the previously mentioned rounding errors. It's otherwise inaudible.

PR #1775, opened 2026-04-03
Implement RPN 0x05 Modulation Depth Range

### LINKED ISSUE #1355: Modulation Depth Range
This RPN is not implemented in fluidsynth.
[This document, section 3.4.4](http://www.shclemen.com/download/GM2.pdf#page=17)

### Expected behavior
[this MIDI file](https://github.com/user-attachments/files/16323312/test-rpn-00-05-modulation-depth-range.zip) should have more and more modulation range depth with each note played.

My proposed solution is to use a _modulation multiplier_. The sf2 spec assumes 50 cents depth by default, so when the RPN gets called, calculate a modulation multiplier by dividing the target depth range by 50. Then when running the DSP chain, simply multiply `vibLfoToPitch` or `modLfoToPitch` (or both) by that amount. That way, if the soundfont creator wanted a more subtle vibrato effect on a given instrument (25 cents for example) and the MIDI composer would want to double the modulation depth (by setting it via RPN to 100 cents, which is the double of MIDI spec default), the final depth would be 50 cents, so double the initial depth, instead of suddenly jumping from 25 cents to 100. 

Another solution could be finding all modulators that use mod wheel as source and vibLfoToPitch as destination and overriding the amount with the set range, but I think it could be quite destructive because of the problem mentioned above.

### Additional context
[My implementation of the solution 1](https://github.com/spessasus/SpessaSynth/blob/4345a32d161720d7f2706a50f92a1c82251c0f1b/src/spessasynth_lib/synthetizer/worklet_system/worklet_methods/tuning_control.js#L86)

derselbst:
Following @spessasus proposal in #1355, this PR implements GM2 RPN 0x5 by using a modulation multiplier: 

*    Implemented RPN 0x05 handler to store and update the current modulation depth range in cents (default: 50 cents per GM2 spec)
*   When processing modulators targeting `GEN_VIBLFOTOPITCH` or `GEN_MODLFOTOPITCH` AND source1 is set to the modulation wheel, the modulator output is scaled by `modulation_depth_range / 50.0`
*    RPN 0x05 is ignored for drum/rhythm channels, following GM2 spec
*    Added test MIDI files and soundfonts to verify the implementation.

### Why This Approach?

Rather than replacing SoundFont-defined vibrato depths with absolute values, this implementation uses a modulation multiplier to preserve the composer's original instrument voice characteristics. This means if a SoundFont defines a subtle 25-cent vibrato and the MIDI composer sets RPN 0x05 to 100 cents (double the default), the result is 50 cents—a proper doubling of the original depth—instead of overriding it to 100 cents. This maintains artistic intent across different instruments and SoundFonts. It also maintains backward compatibility for Soundfonts that override / disable the [default mod2viblfo modulator](https://github.com/derselbst/fluidsynth/blob/cd045d6719fc3e7d8006ae98cc3787d655e8e55c/src/synth/fluid_synth.c#L409-L417) and choose to use `GEN_MODLFOTOPITCH` instead.

Resolves #1355

### DIFF
--- a/src/synth/fluid_chan.c
+++ b/src/synth/fluid_chan.c
@@ -191,6 +191,9 @@ fluid_channel_init_ctrl(fluid_channel_t *chan, int is_all_ctrl_off)
 
         chan->pitch_wheel_sensitivity = 2; /* two semi-tones */
 
+        /* Modulation depth range (GM2 RPN 0x05): default is 50 cents (half a semitone) */
+        chan->modulation_depth_range = 50.0f;
+
         /* Just like panning, a value of 64 indicates no change for sound ctrls */
         for(i = SOUND_CTRL1; i <= SOUND_CTRL10; i++)
         {
--- a/src/synth/fluid_chan.h
+++ b/src/synth/fluid_chan.h
@@ -110,6 +110,7 @@ struct _fluid_channel_t
 
     unsigned char channel_pressure;                 /**< MIDI channel pressure from [0;127] */
     float pitch_wheel_sensitivity;          /**< Current pitch wheel sensitivity */
+    float modulation_depth_range;          /**< Modulation depth range in cents (RPN 0x05) */
     short pitch_bend;                      /**< Current pitch bend value */
     /* Sostenuto order id gives the order of SostenutoOn event.
      * This value is useful to known when the sostenuto pedal is depressed
@@ -176,6 +177,10 @@ fluid_real_t fluid_channel_get_key_pitch(fluid_channel_t *chan, int key);
   ((chan)->pitch_wheel_sensitivity)
 #define fluid_channel_set_pitch_wheel_sensitivity(chan, val) \
   ((chan)->pitch_wheel_sensitivity = (val))
+#define fluid_channel_get_modulation_depth_range(chan) \
+  ((chan)->modulation_depth_range)
+#define fluid_channel_set_modulation_depth_range(chan, val) \
+  ((chan)->modulation_depth_range = (val))
 #define fluid_channel_get_num(chan)             ((chan)->channum)
 #define fluid_channel_set_interp_method(chan, new_method) \
   ((chan)->interp_method = (new_method))
--- a/src/synth/fluid_mod.c
+++ b/src/synth/fluid_mod.c
@@ -315,7 +315,7 @@ fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid
      */
     unsigned char mod_src = is_src1 ? mod->src1 : mod->src2;
     unsigned char mod_flags = is_src1 ? mod->flags1 : mod->flags2;
-    unsigned char applyModulationDepth = (mod_flags & FLUID_MOD_CC) && (mod_src == MODULATION_MSB) && (mod->dest == GEN_VIBLFOTOPITCH);
+    unsigned char applyModulationDepth = (mod_flags & FLUID_MOD_CC) && (mod_src == MODULATION_MSB) && ((mod->dest == GEN_VIBLFOTOPITCH) || (mod->dest == GEN_MODLFOTOPITCH));
     mod_flags &= ~FLUID_MOD_CC;
 
     if(mod_src == FLUID_MOD_NONE)
--- a/src/synth/fluid_mod.c
+++ b/src/synth/fluid_mod.c
@@ -301,7 +301,7 @@ fluid_mod_get_source_value(const unsigned char mod_src,
  * transforms the initial value retrieved by \c fluid_mod_get_source_value into [0.0;1.0]
  */
 fluid_real_t
-fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid_real_t range, int is_src1)
+fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid_real_t range, int is_src1, fluid_voice_t *voice)
 {
     /* normalized value, i.e. usually in the range [0;1] */
     const fluid_real_t val_norm = val / range;
@@ -315,6 +315,7 @@ fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid
      */
     unsigned char mod_src = is_src1 ? mod->src1 : mod->src2;
     unsigned char mod_flags = is_src1 ? mod->flags1 : mod->flags2;
+    unsigned char applyModulationDepth = (mod_flags & FLUID_MOD_CC) && (mod_src == MODULATION_MSB) && (mod->dest == GEN_VIBLFOTOPITCH);
     mod_flags &= ~FLUID_MOD_CC;
 
     if(mod_src == FLUID_MOD_NONE)
@@ -405,6 +406,11 @@ fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid
         }
     }
 
+    if(applyModulationDepth && voice != NULL && voice->channel != NULL)
+    {
+        val *= fluid_channel_get_modulation_depth_range(voice->channel) / 50.0f;
+    }
+
     return val;
 }
 
@@ -483,13 +489,13 @@ fluid_mod_get_value(fluid_mod_t *mod, fluid_voice_t *voice)
     v1 = fluid_mod_get_source_value(mod->src1, mod->flags1, &range1, voice);
 
     /* transform the input value */
-    v1 = fluid_mod_transform_source_value(mod, v1, range1, TRUE);
+    v1 = fluid_mod_transform_source_value(mod, v1, range1, TRUE, voice);
 
     /* get the second input source */
     v2 = fluid_mod_get_source_value(mod->src2, mod->flags2, &range2, voice);
 
     /* transform the second input value */
-    v2 = fluid_mod_transform_source_value(mod, v2, range2, FALSE);
+    v2 = fluid_mod_transform_source_value(mod, v2, range2, FALSE, voice);
 
     /* it indeed is as simple as that: */
     final_value = (fluid_real_t) mod->amount * v1 * v2;
--- a/src/synth/fluid_mod.c
+++ b/src/synth/fluid_mod.c
@@ -315,7 +315,7 @@ fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid
      */
     unsigned char mod_src = is_src1 ? mod->src1 : mod->src2;
     unsigned char mod_flags = is_src1 ? mod->flags1 : mod->flags2;
-    unsigned char applyModulationDepth = (mod_flags & FLUID_MOD_CC) && (mod_src == MODULATION_MSB) && ((mod->dest == GEN_VIBLFOTOPITCH) || (mod->dest == GEN_MODLFOTOPITCH));
+    unsigned char applyModulationDepth = is_src1 && (mod_flags & FLUID_MOD_CC) && (mod_src == MODULATION_MSB) && ((mod->dest == GEN_VIBLFOTOPITCH) || (mod->dest == GEN_MODLFOTOPITCH));
     mod_flags &= ~FLUID_MOD_CC;
 
     if(mod_src == FLUID_MOD_NONE)
--- a/src/synth/fluid_mod.h
+++ b/src/synth/fluid_mod.h
@@ -60,7 +60,7 @@ fluid_real_t fluid_mod_get_value(fluid_mod_t *mod, fluid_voice_t *voice);
 int fluid_mod_check_sources(const fluid_mod_t *mod, char *name);
 
 fluid_real_t fluid_mod_get_source_value(const unsigned char mod_src, const unsigned char mod_flags, fluid_real_t *range, const fluid_voice_t *voice);
-fluid_real_t fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid_real_t range, int is_src1);
+fluid_real_t fluid_mod_transform_source_value(fluid_mod_t* mod, fluid_real_t val, const fluid_real_t range, int is_src1, fluid_voice_t *voice);
 
 void delete_fluid_list_mod(fluid_mod_t *mod);
 
--- a/src/synth/fluid_synth.c
+++ b/src/synth/fluid_synth.c
@@ -3227,21 +3227,7 @@ fluid_synth_update_pitch_wheel_sens_LOCAL(fluid_synth_t *synth, int chan)
 static int
 fluid_synth_update_modulation_depth_range_LOCAL(fluid_synth_t *synth, int chan)
 {
-    fluid_voice_t *voice;
-    int i;
-
-    for(i = 0; i < synth->polyphony; i++)
-    {
-        voice = synth->voice[i];
-
-        if(fluid_voice_get_channel(voice) == chan && fluid_voice_is_playing(voice))
-        {
-            fluid_voice_update_param(voice, GEN_VIBLFOTOPITCH);
-            fluid_voice_update_param(voice, GEN_MODLFOTOPITCH);
-        }
-    }
-
-    return FLUID_OK;
+    return fluid_synth_modulate_voices_LOCAL(synth, chan, 1, MODULATION_MSB);
 }
 
 /**
@@ -7964,7 +7950,7 @@ static void fluid_synth_process_awe32_nrpn_LOCAL(fluid_synth_t *synth, int chan,
         case GEN_REVERBSEND:
             fluid_clip(data, 0, 255);
             /* transform the input value */
-            converted_sf2_generator_value = fluid_mod_transform_source_value(&default_reverb_mod, data, 256, TRUE);
+            converted_sf2_generator_value = fluid_mod_transform_source_value(&default_reverb_mod, data, 256, TRUE, NULL);
             FLUID_LOG(FLUID_DBG, "AWE32 Reverb: %f", converted_sf2_generator_value);
             converted_sf2_generator_value*= fluid_mod_get_amount(&default_reverb_mod);
             break;
@@ -7972,7 +7958,7 @@ static void fluid_synth_process_awe32_nrpn_LOCAL(fluid_synth_t *synth, int chan,
         case GEN_CHORUSSEND:
             fluid_clip(data, 0, 255);
             /* transform the input value */
-            converted_sf2_generator_value = fluid_mod_transform_source_value(&default_chorus_mod, data, 256, TRUE);
+            converted_sf2_generator_value = fluid_mod_transform_source_value(&default_chorus_mod, data, 256, TRUE, NULL);
             FLUID_LOG(FLUID_DBG, "AWE32 Chorus: %f", converted_sf2_generator_value);
             converted_sf2_generator_value*= fluid_mod_get_amount(&default_chorus_mod);
             break;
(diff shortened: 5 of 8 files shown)


spessasus (changes requested):
This implements the original solution, not the v2 I proposed later.
It multiplies the *generators* by the multiplier, and I proposed mutiplying the *modulator's output* of a modulator that has mod wheel as a source.

This causes issues. Take a listen:
[modulation-range.zip](https://github.com/user-attachments/files/26474476/modulation-range.zip)

This MIDI is the original test one, except it has all mod wheel changes removed (it stays at 0). The soundfont is a simple saw wave with a vibrato set by default (soundfonts commonly do that).

However, rendering it with the current PR changes the depth of the vibrato, despite there not being any modulation wheel applied at all! Mod wheel range should only affect the modulation wheel, not instrument's internal parameters.

Note that spessasynth also exhibits this behavior, so it's not a good point of reference right now.

As for a fix,
The change would probably go in the modulator computing code. It would check if the modulator's source is modulation wheel, and if so, multiply it by the modulation multiplier just before returning the value.
That way we also get a bonus of allowing modulation to be something other than vibrato (for example lfo to volume -> rotary effect for organ)

derselbst:
>  It multiplies the generators by the multiplier, and I proposed mutiplying the modulator's output of a modulator that has mod wheel as a source.

Thanks for the feedback, that wasn't clear to me. I could tweak this into the modulator logic. Yet, I'm struggling to implement this correctly. ~~The problem: When you change RPN 0x5 you'd expect to change the sound as well (even while the mod wheel isn't changed at all). That means, all modulators that have a MODULATION_MSB set as source are supposed to be re-evaluated. But fluidsynth re-evaluates by destination generator only. One would need to go through all voices and all their modulators, scanning for MODULATION_MSB modulators, memorize all destination generator they manipulate and then call `fluid_voice_update_param()` on those generators.~~ 

~~Or am I overengineering this, and should we only restrict it to modulators that have `GEN_MODLFOTOPITCH` set as destination AND MODULATION_MSB set as source?~~

~~And why wouldn't it be sufficient to apply the modulation multiplier to only the modulation part of `GEN_MODLFOTOPITCH`? (=excluding initial and NRPN amount)~~

Edit: I think I found a solution.

---

And how to handle modulators that use two sources with MODULATION_MSB? Would we apply the modulation multiplier to both sources?

Or should we apply this hack solely to the default mod2VibLfo modulator?

spessasus:
> And how to handle modulators that use two sources with MODULATION_MSB? Would we apply the modulation multiplier to both sources?

I can't think of a valid reason to have a modulator in which both sources are the same thing, but I gues technically it would be the correct solution to multiply it twice (or by 2 * multiplier).
I suggest a very straightforward implementation, by simply checking if either one is cc + mod wheel. Then maybe tag the modulator as modWheelModulator (during parse time?) and in calculation just check that flag and multiply.

> Or should we apply this hack solely to the default mod2VibLfo modulator?

Absolutely not, `gm.dls` and its sf2 conversion use modLfoToPitch instead. This would break one of the most commonly used sound banks pretty much.

derselbst:
> > Or should we apply this hack solely to the default mod2VibLfo modulator?
> 
> Absolutely not, `gm.dls` and its sf2 conversion use modLfoToPitch instead. This would break one of the most commonly used sound banks pretty much.

Ok, but in this case I would restrict this "hack" to apply to only those modulator sources that have `MODULATION_MSB` set as source and either `GEN_MODLFOTOPITCH` or `GEN_VIBLFOTOPITCH` as destination. Otherwise it would break all modulators that abuse `MODULATION_MSB` for something else.

spessasus:
> Ok, but in this case I would restrict this "hack" to apply to only those modulator sources that have `MODULATION_MSB` set as source and either `GEN_MODLFOTOPITCH` or `GEN_VIBLFOTOPITCH` as destination. Otherwise it would break all modulators that abuse `MODULATION_MSB` for something else.

Yes, that's good.

PS: GS has system exclusive messages (predating GM2) that can set mod wheel depth and pitch bend range:

image

Can these be supported? I could open a new issue and provide more details on how they work.

derselbst:
I'm still struggling to implement this and need a third opinion.

@mrbumpy409 In case you have some time: We would like to implement RPN 5 (Modulation Depth Range). The GM2 spec is attached below.

The most SF2-like way would be to use the currently unused Src2 of the default `mod2VibLfo` modulator for this (just like it's done for default pitch modulator + PitchBend Range). But we can't as modifying a default modulator would be a breaking change.

Instead, the idea is to check if the modulator targets either `GEN_MODLFOTOPITCH` or `GEN_VIBLFOTOPITCH` and uses CC#1 as source. And if so apply the following factor to the modulator: `modulation_depth_range / 50.0`, i.e.:

`modulator_output = src1 * src2 * amount * (modulation_depth_range / 50.0)`

Would that make sense to you? And should it be applied once per modulator, or once per source (= a modulator with two CC#1 sources would have the multiplier applied twice)?


> 3.4.4 00H / 05H Modulation Depth Range (Vibrato Depth Range)
> [Refer to Complete MIDI 1.0 Detailed Specification 1999 or later - "Modulation Depth Range RPN" ]
> [required]
> Default Value: 00H/40H (+/-50 cents)
> Sets the peak value of Vibrato or LFO Pitch change amount from the basic pitch set by the Modulation Depth controller (cc#1). 
> The value "1" of MSB of the Data Entry corresponds to a semitone, and "1" of LSB corresponds to 100/128 Cents. For example, MSB = 01H, LSB = 00H means that the Mod Wheel will modulate a maximum of +/- one semitone of vibrato depth (that is, two semitones peak to peak, or one semitone from the center frequency in either direction). Another example, MSB = 00H, LSB = 08H means that the vibrato depth will be 6.25 cents in either direction from the center frequency.
> [recommended]
> Rhythm Channels shall not respond to this message.
> Devices should be able to pitch change a minimum of 6 semitones (+/-600 cents).
> (Note: This range is only intended to apply to GM2 devices. In the case of other device specifications that are based upon and reference this specification, implementers should consult the referencing specification for the appropriate range value).


---

> I could open a new issue and provide more details on how they work.

Sure go ahead.

mrbumpy409:
> @mrbumpy409 In case you have some time: We would like to implement RPN 5 (Modulation Depth Range). The GM2 spec is attached below.

@derselbst, please hold off on implementing this until I've had time to review and test the expected behavior. I see that the GM2 RPN 0x05 definition (which is also shared by GS) is very different than what @spessasus is proposing, so I want to look into this a bit first, make test files, etc.

mrbumpy409:
> The most SF2-like way would be to use the currently unused Src2 of the default `mod2VibLfo` modulator for this (just like it's done for default pitch modulator + PitchBend Range). But we can't as modifying a default modulator would be a breaking change.

I agree that modifying the default modulator would break with SF2 spec. But, using the secondary source for this wouldn't work anyway, since both sources are mapped to -1–1 before being multiplied together. In other words, you could not use this method to increase the modulation amount beyond the default modulator's amount value (50 cents)—only subtract from it.

> Instead, the idea is to check if the modulator targets either `GEN_MODLFOTOPITCH` or `GEN_VIBLFOTOPITCH` and uses CC#1 as source. And if so apply the following factor to the modulator: `modulation_depth_range / 50.0`, i.e.:
> 
> `modulator_output = src1 * src2 * amount * (modulation_depth_range / 50.0)`
> 
> Would that make sense to you?

Not if you are trying to implement RPN 0x05 as described in the GS/GM2 spec. What @spessasus is suggesting appears to be a very different interpretation of this controller:

* **GS/GM2 spec:** RPN 0x05 would simply override the default mod2VibLfo modulator amount with the value in cents specified by the RPN data messages.
* **spessassus' method:** RPN 0x05 would determine when CC#1 is being used to modulate any LFO destination (including volume & filter) and use multiplication to scale the existing modulation amount.

@spessasus, do you have a historic source for your proposed RPN 0x05 behavior? Otherwise, it would seem best to follow the GS/GM2 defined behavior.

I tested both the Roland Sound Canvas (SC-8820, emulated) and BASSMIDI, and both behave according to the GS/GM2 spec, i.e., the maximum effect of CC#1 on LFO depth is overridden, *not* multiplied against the existing value. My Sound Blaster Audigy 2 ZS ignores RPN 0x05 so cannot be used as a reference point.

To implement the GS/GM2 RPN 0x05 behavior, FluidSynth could simply override the amount value in any modulators that meet the following criteria:

1. CC#1 is used as the primary or both modulation sources and no other source besides CC#1 is used.
2. The modulator destination is either "Vib LFO → pitch (c)" or "Mod LFO → pitch (c)".

spessasus:
Hi @mrbumpy409 ,

Firstly, the proposal does not involve multiplying the sources. It involves multiplying *the final output value of the modulator.* Like so:

```
value = source1Transformed * source2Transformed * amount * modulationMultiplier
```
And @derselbst asked if fluidsynth should treat a modulator with 2 modulation wheel sources like so:

```
value = source1Transformed * source2Transformed * amount * (modulationMultiplier * modulationMultiplier)
```

Secondly,
What you described is my alternative proposal of finding a modulation modulator and overriding it.
While SCVA you presumably used to test it does match the GM2 spec exactly, though keep in mind that all of its presets have the "default vibrato depth" set to 50 cents.
And we can't use BASSMIDI as a reference since it doesn't fully implement SF2 modulators (it seems to use a completely different LFO for mod wheel, as vibLfoDelay doesn't affect it).

The GM2 spec assumes that the default modulation depth is 50 cents, but that's not the case with soundfonts. They can override the default modulator to any value they wish. The famous example being `gm.dls`, which overrides this modulator for many presets, for example `Piano 1`.

The idea I had is for the depth to *scale* with what soundfont has set.
If we leave it at the default SF2 value, which also is the default GM2 value, we get accurate behavior:
- MIDI requests 100 cents as mod depth: `modulation multiplier = 100 / 50 = 2`
- the default modulator gets multiplied by 2, so `50 * 2 = 100`, the exact amount requested.

If the soundfont engineer decides that the instrument should have a more subtle vibrato (say, 25 cents):
- MIDI requests 100 cents as mod depth: `modulation multiplier = 100 / 50 = 2`
- `25 * 2 = 50` cents of detune.

In other words, MIDI requested twice the regular amount, so it gets twice the regular amount.
I think that this is a neat way to allow soundfonts to maintain their fine-tuned parameters while responding to the RPN message.

Many MIDI files perform a manual reset at the start (setting pitch bend ranges, resetting controllers and RPN tunings). While granted, I haven't seen a MIDI that actually uses this exact RPN so far, I have seen MIDIs that use the GS variant of it (proposed in https://github.com/FluidSynth/fluidsynth/issues/1776). So this means that if a MIDI sets the default range at the beginning, any custom mod wheel range is overriden with 50 cents, despite the MIDI file requesting the "default", which in case of a MIDI without such message (still can be GM2!) the modulation depth is preserved as defined in the soundfont.

And remember,
Most sf2s will leave it at the default, so the exact GM2 behavior is preserved. But for those that don't, their programmed behavior is nicely preserved that way.

What do you think?

mrbumpy409:
> Secondly, What you described is my alternative proposal of finding a modulation modulator and overriding it. While SCVA you presumably used to test it does match the GM2 spec exactly, though keep in mind that all of its presets have the "default vibrato depth" set to 50 cents.

Based on the GS/GM2 RPN 0x05 spec, the instrument's default vibrato depth is irrelevant. RPN 0x05 is designed to be used to set an exact, desired modulation depth in cents. If you want CC#1 to modulate the pitch by +/- a semitone, you would send the following sequence:

* CC#101: 0
* CC#100: 5
* CC#6: 1
* CC#38: 0

And this should work by overriding the instrument's default vibrato depth regardless of what it was previously set to. Note that this would only override CC#1's modulation of the LFO pitch amount, not other sources.


>And we can't use BASSMIDI as a reference since it doesn't fully implement SF2 modulators (it seems to use a completely different LFO for mod wheel, as vibLfoDelay doesn't affect it).

I think we can use BASSMIDI as a reference in the sense that it incorporates GS behaviors (e.g., RPNs) that are not normally found in SoundFont synths, as we are attempting to do here. In this case BASSMIDI has implemented this RPN in agreement with GS/GM2 and in the same way I am proposing here.

> The GM2 spec assumes that the default modulation depth is 50 cents, but that's not the case with soundfonts. They can override the default modulator to any value they wish. The famous example being `gm.dls`, which overrides this modulator for many presets, for example `Piano 1`.

The GS/GM2 spec for RPN 0x05 is written in a way where the default modulation depth is irrelevant.

> The idea I had is for the depth to _scale_ with what soundfont has set. If we leave it at the default SF2 value, which also is the default GM2 value, we get accurate behavior:
> 
>     * MIDI requests 100 cents as mod depth: `modulation multiplier = 100 / 50 = 2`
> 
>     * the default modulator gets multiplied by 2, so `50 * 2 = 100`, the exact amount requested.
> 
> 
> If the soundfont engineer decides that the instrument should have a more subtle vibrato (say, 25 cents):
> 
>     * MIDI requests 100 cents as mod depth: `modulation multiplier = 100 / 50 = 2`
> 
>     * `25 * 2 = 50` cents of detune.
> 
> 
> In other words, MIDI requested twice the regular amount, so it gets twice the regular amount. I think that this is a neat way to allow soundfonts to maintain their fine-tuned parameters while responding to the RPN message.

But RPN 0x05 values are defined in cents, not multiplication factors, so how do I know which RPN values to send to, e.g, double the current vibrato depth? As a musician, I actually find the GS/GM2 definition more logical to work with. If I want 72 cents of modulation from CC#1, I know exactly what data to send. If I'm having to deal with multiplication, how would I figure out how to get to 72 cents of modulation? I'd first have to open the SoundFont in an editor to see what the instrument's programmed CC#1 modulation depth is.

> Many MIDI files perform a manual reset at the start (setting pitch bend ranges, resetting controllers and RPN tunings). While granted, I haven't seen a MIDI that actually uses this exact RPN so far, I have seen MIDIs that use the GS variant of it (proposed in #1776). So this means that if a MIDI sets the default range at the beginning, any custom mod wheel range is overriden with 50 cents, despite the MIDI file requesting the "default", which in case of a MIDI without such message (still can be GM2!) the modulation depth is preserved as defined in the soundfont.

This sounds like a separate issue, if GM reset messages are overriding SoundFont modulators from what is programmed (assuming I am understanding correctly).

> And remember, Most sf2s will leave it at the default, so the exact GM2 behavior is preserved. But for those that don't, their programmed behavior is nicely preserved that way.

The whole point of RPN 0x05 is to override the programmed behavior, so I don't understand the need to factor it in at all. And as I mentioned earlier, this RPN only affects the CC#1 modulation of LFO pitch. Any other programming of LFO pitch (e.g., generator value in the instrument) is left untouched.

derselbst:
Thanks for your input Chris.

> The whole point of RPN 0x05 is to override the programmed behavior, so I don't understand the need to factor it in at all. And as I mentioned earlier, this RPN only affects the CC#1 modulation of LFO pitch. Any other programming of LFO pitch (e.g., generator value in the instrument) is left untouched.

When I find time, I'll propose an alternative implementation that overrides the amount of all modulators that target either `GEN_MODLFOTOPITCH` or `GEN_VIBLFOTOPITCH` and use CC#1 as source1. Limited to the MIDI channel of RPN5, ofc.

The tricky part will be to determine when RPN5 is not set, in which case the amount is not overridden. This could be in case of a system reset, or after All Ctrl Off. The problem: When an application attempts to "reset" RPN5 by setting it to its default value, this will be treated 50 cents override amount, which would be fine for most Soundfonts, other than those attempting to override it via default modulators.

spessasus:
@mrbumpy409 (CC: @derselbst)

What I'm trying to model is exactly how the SF2 pitch wheel modulator behaves.
Because when you think about it, both pitch wheel range and modulation depth range are the same thing. Both set by RPN and the spec describes them as "absolute value". If this was the case, you wouldn't be able to override the pitch wheel modulator. 

Yet fluidsynth is primarily a *soundfont* synthesizer. I chooses the SF2 spec over the MIDI spec when they overlap.

As a soundfont designer, I can override the pitch wheel modulator. If I change its amount to 6400, it will work. If the RPN requests pitch wheel range to be 2 semitones, it'll be 1 instead. I can also disable the modulator entirely if you wish, which is something you do in the SF Spec Test, it only makes sense to allow disabling the vibrato modulator as well. 
With the multiplier proposal, any value multiplied by 0 is still 0, so no matter what the MIDI requests, it will be disabled.
With overriding proposal, the soundfont value is ignored, which is inconsistent with current behavior:
- Pitch wheel modulator ultimately decides the range, even though GM specifies that it's an absolute value
- Volume and Expression CCs are defined by modulators, even though GM specifies the exact equation (`Gain in dB = (40 * log10(cc7/127)) + (40 * log10(cc11/127))`, section 3.3.6 in GM2 spec)

You can disable or change all MIDI behavior with modulators if you wish, so it doesn't make sense for modulation depth to be different. Of course, ideally we'd create another source like mod wheel depth, but that's not an option in SF2 (that'll be fixed in SFe4, but that's another topic entirely)

Also the overriding behavior has one flaw: what if two modulators using mod wheel are present, one to disable vibLfo and another to drive the modLfo? This is the case with `gm.dls` again (its SF2 conversion and how fluidsynth loads it internally), as it has two:
- vibLfoToPitch set to 0 to disable it (as DLS1 only has mod LFO and not vib LFO)
- modLfoToPitch with the actual 47 cents set as the modulation wheel.

If we override all modulators that:

> 1. CC#1 is used as the primary or both modulation sources and no other source besides CC#1 is used.
> 2. The modulator destination is either "Vib LFO → pitch (c)" or "Mod LFO → pitch (c)".

Both match the description, so both get overriden.
We now get two LFOs which have different rates and delays (vib has none since DLS1 does not use it), and mod which has the actual rate set. In other words, two triangle LFOs out of phase that try to modulate the sound at once. Mutliplier doesn't have this issue as it only multiplies modulators, not forces disabled modulators (amount 0) to start working again.
So I still believe that my proposal is the correct solution here.

TL;DR; the overriding behavior is inconsistent with the entire idea of soundfont modulators, since they allow to override default behaviors.

mrbumpy409:
> Also the overriding behavior has one flaw: what if two modulators using mod wheel are present, one to disable vibLfo and another to drive the modLfo? This is the case with `gm.dls` again (its SF2 conversion and how fluidsynth loads it internally), as it has two:
> 
>     * vibLfoToPitch set to 0 to disable it (as DLS1 only has mod LFO and not vib LFO)
> 
>     * modLfoToPitch with the actual 47 cents set as the modulation wheel.
> 
> 
> If we override all modulators that:
> 
> > 1. CC#1 is used as the primary or both modulation sources and no other source besides CC#1 is used.
> > 2. The modulator destination is either "Vib LFO → pitch (c)" or "Mod LFO → pitch (c)".
> 
> Both match the description, so both get overriden.

Aah, shoot. You're right. That is an oversight on my part. I see that implementing RPN 0x05 as defined in the spec is not as simple as I had originally thought. Your proposal does more elegantly handle cases where the default modulator is overridden, canceled, replaced, or added to via another modulator. For SoundFonts that do not touch or add to the default CC#1-to-LFO-pitch modulator, the math works out just fine for expected RPN 0x05 behavior, which will cover 99% of cases.

> We now get two LFOs which have different rates and delays (vib has none since DLS1 does not use it), and mod which has the actual rate set. In other words, two triangle LFOs out of phase that try to modulate the sound at once. Mutliplier doesn't have this issue as it only multiplies modulators, not forces disabled modulators (amount 0) to start working again. So I still believe that my proposal is the correct solution here.

You've convinced me. 🙂

@derselbst: Sorry for the back and forth. I finally understand the reasons that @spessasus has implemented this the way that he has, and I think it will work better to handle all possible modulator complexities. I see no reason not to proceed with his implementation.

derselbst:
Thanks for the feedback Chris! I need to write up a summary of the implementation and will then proceed merging this... once I find time :)

PR #1777, opened 2026-04-12
Migrate wiki to Zensical

spessasus:
This PR migrates the wiki into the repo as described in https://github.com/FluidSynth/fluidsynth/discussions/1764

This has several advantages:
- The wiki is now indexable (very important!!!)
- Way better search than github (instant clientside search that has better results that GH wiki search from my experience)
- TOC is generated automatically, not a nested list markdown
- It looks nicer (in my opinion of course)
- The wiki is now a part of the VCS, meaning that it can be updated with a PR, updating the wiki automatically when the PR gets merged.

Also a few notes:
- I added the `view source` and `edit this page` buttons, to address the point @derselbst made in the original discussion. It still requires the editor to submit a PR, but it now should directly link to the file the user needs to edit (and GH even suggests forking with one button). So I think it's worth it in the end.
- I'm not sure if the pages are configured to use `fluidsynth.org` or `fluidsynth.github.io`. So I can't check whether the workflow I included will work. The workflow is a copy of the one i use, which uses the `.io` version.
- Adding to the point above, I set `site_url = "https://fluidsynth.org/wiki"`, so that might need to change if `.io` ends up as the GH pages domain.
- I fixed a lot of errors (like repeated words `to to`, typos, stuff like `lets -> let's` and grammar errors (missing commas, etc.). I applied the suggestions manually, so I believe that I didn't change the meaning. Feel free to git diff with the original `.md` wiki files though.
- I added titles to each page, matching their titles in `_Sidebar.md`, but with spaces for a cleaner look.
- I reformatted the tables so they align properly. 
- I replaced the `:emoji_name:` syntax with actual emojis as that syntax is a GH-only thing.
- I added spessafont to the soundfont editor list :-)
- I added a few `cmake` labels to code blocks where they were missing (`Using-prebuilt-libraries-on-Android.md`)
- I added a Bug Tracker tab which points to the GitHub issues as an easy way to find them from the wiki.
- I replaced `**Note:**` with the `!!! Note` [admonition syntax](https://zensical.org/docs/authoring/admonitions/)
- The `[[Page name]]` syntax is a GH-only thing so I replaced it with the standard `[Page name](Page-Name.md)` syntax. It works exactly the same.
- I removed [BuildingFluidSynth](https://github.com/FluidSynth/fluidsynth/wiki/BuildingFluidSynth) as the page itself was said to be removed soon. That "soon" was 8 years ago, so I think it makes sense to remove it?
- I moved `ladspa.md` to the `wiki` directory as it was only used in the wiki. If that's not actually the case, feel free to change that.

To test, install zensical (`pip install zensical`) and `zensical serve` to check the website,  `zensical build --clean` to build the complete html files (out dir is `site`, which is `.gitignored`.)

@derselbst I hope that seeing Zensical in action with this PR will convince you to switch the fluidsynth wiki to it :-)


derselbst:
That is indeed very convincing. Thanks for taking the time implementing this and fixing all those nasty little bugs!

I'm also just realizing that having the Wiki here allows our CodeSpell workflow to nitpick every typo. I've fixed those. Also moved the wiki/ into doc/ and integrated the new workflow into the existing api_doc_build workflow.

`www.fluidsynth.org` is a CNAME that points to `fluidsynth.github.io`. Attempting to access `fluidsynth.github.io` redirects you to back to the CNAME. The github page is hosted in that repo: https://github.com/FluidSynth/fluidsynth.github.io

There's one nitpick I wasn't able to fix yet: On the  start page / index.html, the breadcrumbs / nav-path shows:

    🎹 FluidSynth > 🎹 FluidSynth

That's odd. I mean I know why it does that, but it doesn't seem right to me. I messed around with the navigation but couldn't find a solution that works. Maybe you have an idea?

spessasus:
@derselbst 

I think that the core issue is that zensical expects a separate "home" page (at the top level) and fluidsynth doesn't have that.
So I separated index.md from the `FluidSynth` section and renamed it to `Home` and the section to `About FluidSynth`. thoughts?

EDIT: another thing to note is that https://www.fluidsynth.org doesn't support HTTPS, so firefox (at least in my case) warns about it and calls it unsafe. Will the wiki be affected by this?

spessasus:
Also another thing: This is a bit of a stretch, but I wonder if you'd be open to the idea of migrating the API docs to zensical as well?

The current site looks... *less than ideal:*
image

And zensical supports [automatic documentation generation](https://zensical.org/docs/setup/extensions/mkdocstrings/) if fluidsynth uses that. What do you think?

derselbst:
> So I separated index.md from the FluidSynth section and renamed it to Home and the section to About FluidSynth. thoughts?

That looks nice, thanks!

> EDIT: another thing to note is that https://www.fluidsynth.org/ doesn't support HTTPS, so firefox (at least in my case) warns about it and calls it unsafe. Will the wiki be affected by this?

You should be getting the warning only when accessing https://fluidsynth.org - because for some reason it uses a TLS certificate from www.elementsofsound.org and fluidsynth.org isn't listed as SAN on this certificate. Accessing https://www.fluidsynth.org/ works correctly though, because it has its dedicated TLS certificate.

> I wonder if you'd be open to the idea of migrating the API docs to zensical as well?

Generally, yes. The fact that doxygen's search indexing duplicates every single API function also strikes me and I was unable to solve this so far. I have two caveats though:

* I am pretty much used the doxygen-style syntax of documenting this API-specific stuff for years now. I am very conservative in changing all the comments in the code to become e.g. docstrings... unless I'd really see a huge benefit from it.
* The main "documentation-commands" (like `@note`, `@warning`, `@defgroup`, `@startlifecycle`, etc.) should remain to be supported. However, looking at https://mkdocstrings.github.io/c/ makes me think the support for C documentation seems to be very minimalistic. Unless I'm missing something?

PR #1780, opened 2026-04-19
fix(sfloader): move INFO chunk FLUID_LOG into the branch where item.chr is allocated

heitbaum:
The debug log referencing item.chr was placed at the end of the outer
else block, reachable by code paths where item.fcc was never allocated,
causing a potential NULL pointer dereference and a -Wmaybe-uninitialized
warning. Move the log to immediately after the successful fread(), where
item.chr is guaranteed to be valid.

Fixes -Wmaybe-uninitialized warning in fluid_sffile.c:685

### DIFF
--- a/src/sfloader/fluid_sffile.c
+++ b/src/sfloader/fluid_sffile.c
@@ -855,6 +855,7 @@ static int process_info(SFData *sf, int size)
 
                     /* force terminate info item */
                     item.chr[chunk.size] = '\0';
+                    FLUID_LOG(FLUID_DBG, "INFO chunk %c%c%c%c (%d bytes) -> %s", p[0], p[1], p[2], p[3], chunk.size, item.chr);
                 }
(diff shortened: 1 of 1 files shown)


derselbst (changes requested):
Sry, your change would still cause a NULL ptr. deref in worst case. The FLUID_LOG in question should simply be moved into the else branch, where item.fcc is alloced, ~after it's successfully alloced, i.e. line 844~. All other cases will print a warning anyway.

derselbst:
My bad: log at line 858, after the buffer was read in and NULL terminated of course.

heitbaum:
> My bad: log at line 858, after the buffer was read in and NULL terminated of course.

Reworked addressing the LOG directly.

derselbst (approved):
LGTM, thanks! Waiting for CI.

PR #1781, opened 2026-04-19
fix(midi): replace deprecated API calls in fluid_player_set_bpm

heitbaum:
fluid_player_set_bpm is deprecated in favour of fluid_player_set_tempo, as is fluid_player_set_midi_tempo which it called internally. Replace the internal call with fluid_player_set_tempo() using FLUID_PLAYER_TEMPO_EXTERNAL_BPM to suppress the -Wdeprecated-declarations warning without introducing internal workarounds.


derselbst (approved):
LGTM, thanks!

PR #1782, opened 2026-04-19
Correct wiki links to the new wiki

spessasus:
This small PR changes all links that still point to the GH wiki to the new wiki
Also adds the output generated `site` dir to `.gitignore` because it forgot to add it before and we don't want to accidentally commit the generated files :-]


derselbst (commented):
Thanks! Btw: You did add `site` to gitignore and I changed it to `.zensical`, because I've overridden the `site_dir` ;)

spessasus:
Ah, fair point. I removed it from gitignore!

derselbst (approved):
Thanks!

PR #1792, opened 2026-05-25
Port API documentation to Zensical

derselbst:
Previously, fluidsynth's API documentation was generated by doxygen to static HTML. Doxygens default HTML theme looks like 90s age. Because of that in the past @mawe42 made an effort to give it a better looking. Yet, the design lacks a dark theme and is not very friendly to mobile devices. Also, the fact that doxygen's search index kepts duplicating every single function and type bothers me a lot.

This PR ports fluidsynth's API documentation to Zensical. It does that by enabling doxygen to generate XML. This XML is then translated via XSLT to markdown, which is organized into a directory structure below `/doc/wiki` so it can then be consumed by Zensical to build a static HTML website from it.

The old doxygen usage text files will remain for now. Also, no structural changes in the source code comments are required by this approach.

A preview of the new website can be obtained as build artifact from the API Docs workflow here (for the next 7 days): https://github.com/FluidSynth/fluidsynth/actions/runs/26392445262/artifacts/7195085643

This change was made possible with Claude Sonnet 4.6.

Edit: Credits to @spessasus for getting this idea stuck in my head :)

TODO

- [x] Fluidsynth's API is currently accessible via https://www.fluidsynth.org/api - with the new website, the API docs will be located below the wiki i.e. at https://www.fluidsynth.org/wiki/api . Figure out a solution that does not break the old link.

---

PS: Why using XSLT rather than e.g. [moxygen](https://0state.com/moxygen)? Because moxygen failed. It cut off some typedefed function signatures and it's markdown output cannot be customized.


spessasus:
I've taken a look at the compiled website and I have one important suggestion: I think that the API reference should have its own tab, next to "Documentation". That way developers have a single click access without having to navigate through the documentation section, and both are separate, like before (one documentation is from GH wiki and the other is the API reference). Other than that, it seems fine.

derselbst:
> I've taken a look at the compiled website and I have one important suggestion: I think that the API reference should have its own tab, next to "Documentation". That way developers have a single click access without having to navigate through the documentation section, and both are separate, like before (one documentation is from GH wiki and the other is the API reference). Other than that, it seems fine.

Done, thanks for the feedback!

PR #1793, opened 2026-05-30
GitHub Actions: Run acceptance-test on multiple Windows versions

cclauss:
https://github.com/actions/runner-images#available-images
```
os: [windows-2022, windows-2025-vs2026, windows-latest, windows-11-arm]
runs-on: ${{ matrix.os }}
```
`windows-11-arm` is from https://github.com/actions/partner-runner-images

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -566,7 +566,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:  # https://github.com/actions/runner-images#available-images
-        os: [windows-2022, windows-2025-vs2026, windows-latest, windows-11-arm]
+        os: [windows-2022, windows-2025]
     runs-on: ${{ matrix.os }}
     name: 🧪 Acceptance Test
     steps:
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -563,7 +563,11 @@ jobs:
     - msvc-build
     - msys2-build
     - mingw-build
-    runs-on: windows-2022
+    strategy:
+      fail-fast: false
+      matrix:  # https://github.com/actions/runner-images#available-images
+        os: [windows-2022, windows-2025-vs2026, windows-latest, windows-11-arm]
+    runs-on: ${{ matrix.os }}
     name: 🧪 Acceptance Test
     steps:
     - name: Download all artifacts


derselbst (changes requested):
This won't work. The acceptance test consumes build artifacts from the previous builds and tries to execute them. We're only compiling for x86 & x64, so running on an ARM will definitely fail.

I'm open to run the acceptance tests on `windows-2022` and `windows-2025-*`. I prefer to not use the -latest images as they change unpredictably.

cclauss:
I too was confused by the results, but...
> ✅ [Fluidsynth Windows / 🧪 Acceptance Test (windows-11-arm) (pull_request)](https://github.com/FluidSynth/fluidsynth/actions/runs/26687053654/job/78657503644?pr=1793) Successful in 8s

---
`windows-11-arm` ___always___ [fails](https://github.com/nwhitehead/pyfluidsynth/actions/runs/26686456428) on [`pyFluidSynth`](https://github.com/nwhitehead/pyfluidsynth):
https://github.com/nwhitehead/pyfluidsynth/blob/357b3a19908dea895a83e12b70a440e6f9a44158/.github/workflows/ci.yml#L56

However, here it ___always___ passes.  Do we know why??

* https://docs.python.org/3/library/ctypes.html#loading-shared-libraries

derselbst:
Aha, so there seems to be an emulation feature called Prism that I didn't knew before:
https://learn.microsoft.com/en-us/windows/arm/apps-on-arm-x86-emulation

So the fluidsynth x86 binaries are emulated by the Windows 11 ARM OS. Yet, this is no meaningful test for fluidsynth, since we're not testing Prism here. Only if we added Windows ARM support to the build pipeline it would become a meaningful test.

> windows-11-arm always [fails](https://github.com/nwhitehead/pyfluidsynth/actions/runs/26686456428) on [pyFluidSynth](https://github.com/nwhitehead/pyfluidsynth):
https://github.com/nwhitehead/pyfluidsynth/blob/357b3a19908dea895a83e12b70a440e6f9a44158/.github/workflows/ci.yml#L56

I think it's because you're trying to load an x64 fluidsynth.dll into a Win32 python process.

.github/workflows/windows.yml
```diff
@@ -563,7 +563,11 @@ jobs:
     - msvc-build
     - msys2-build
     - mingw-build
-    runs-on: windows-2022
+    strategy:
+      fail-fast: false
+      matrix:  # https://github.com/actions/runner-images#available-images
+        os: [windows-2022, windows-2025-vs2026, windows-latest, windows-11-arm]
```
cclauss:
```suggestion
        os: [windows-2022, windows-2025]
```
https://github.com/actions/runner-images/issues/14017

cclauss:
`matrix: os: [windows-2022, windows-2025]`

Thanks for finding Prism. macOS has a similar capability called [Rosetta](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment) that allows modern Macs with Apple Silicon (ARM) to run code compiled for legacy Intel Macs.

I will read up on Prism to see if I can fix pyFluidSynth on Windows on ARM.  Thanks for your work on FluidSynth!!

derselbst (approved):
LGTM, thanks!

PR #1795, opened 2026-06-08
Change the log level of stub functions

white-axe:
We're having problems at https://github.com/mkxp-z/mkxp-z/pull/321 with the "function is a stub" log messages sometimes emitted by FluidSynth when configured with `-Dosal=cpp11` or `-Dosal=embedded` being annoying to end users. Given that they don't indicate an error that occurred within FluidSynth, maybe they shouldn't be error messages?

### DIFF
--- a/src/utils/fluid_stub_functions.h
+++ b/src/utils/fluid_stub_functions.h
@@ -30,18 +30,24 @@
 #include "fluidsynth_priv.h"
 
 
+#ifdef NDEBUG
+#define STUB_FUNCTION_LOG_LEVEL FLUID_DBG
+#else
+#define STUB_FUNCTION_LOG_LEVEL FLUID_ERR
+#endif
+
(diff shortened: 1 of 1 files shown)


derselbst:
That depends on how you interpret a "STUB". I could argue, that those STUBs are to be implemented downstream, rather than being used as is, which would be an error. Error, because fluidsynth is trying to use functionality, which your OSAL doesn't provide.

Compromise: conditional log level depending on whether it's a debug or release build. I.e. if `NDEBUG` is defined (=release build), use `FLUID_DBG`, otherwise keep using `FLUID_ERR`.

derselbst:
Oh, and pls. rebase your changes against `release/2.5` branch, if you want to see this in 2.5.5

white-axe:
Alright. That seems fair.

derselbst (approved):
LGTM, thanks!

PR #1805, opened 2026-07-18
Fix some bugs related to DLS 'pgal' (#1804)

### LINKED ISSUE #1804: PGAL Aliasing bug with DLS files
### FluidSynth version

```sh
2.5.5
```

### Describe the bug

There are aliasing issues when loading PGAL DLS banks. Drums play at a higher pitch and the acoustic piano does not play at all. [Here's a WAV recording of the song.](https://github.com/user-attachments/files/30139786/Copacabana_FluidSynth-mDLS.wav)

### Expected behavior

And [here is a render of how it's supposed to sound.](https://github.com/user-attachments/files/30139513/Copacabana_MobileBAE.wav)

### Steps to reproduce

The problem can be reproduced by playing [the following MIDI](https://3.onj.me/phonetones/unzipped/Sony%20Ericsson/T610/Copacabana.mid) on [the following DLS soundbank](https://3.onj.me/phonetones/Software%20and%20Soundbanks/Soundbanks/Beatnik%20mobileBAE/mobile60.dls)

### Additional context

_No response_

rsp4jack:
Fix #1804

* Ignore drums when parsing instrument aliasing
* Apply drum note aliasing to not only region selection but also generators

### DIFF
--- a/src/sfloader/fluid_dls.cpp
+++ b/src/sfloader/fluid_dls.cpp
@@ -1487,6 +1487,11 @@ fluid_dls_font::fluid_dls_font(fluid_synth_t *synth,
         throw std::runtime_error{ "Rewind to start of file failed" };
     }
 
+    if(filesize < 12)
+    {
+        throw std::runtime_error{ "File too short! (< 12 bytes)" };
+    }
+
     // Parse DLS
     // chunk: RIFF[DLS ]
     // subchunk: ...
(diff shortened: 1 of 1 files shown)


derselbst (approved):
Sounds correct to me now, thanks a lot! I'll directly cherry-pick this to release/2.5 branch and create to follow up PRs to a) change the error message a bit and b) make fluidsynth fail when it fails to load a soundfont.

PR #1809, opened 2026-07-18
Fix DLS sample validation

rsp4jack:
2354c2a9acdb26de7cdcd37c903ee108f46c0a7d introduces validation of DLS samples, but when a sample is invalid, it won't be add into samples_fluid, causing mismatch between array index and sample ID to following samples.

This can cause out-of-bound reading beyond the samples_fluid (a std::vector)'s length, but not beyond its capacity.

### DIFF
--- a/src/sfloader/fluid_dls.cpp
+++ b/src/sfloader/fluid_dls.cpp
@@ -1707,7 +1707,7 @@ fluid_dls_font::fluid_dls_font(fluid_synth_t *synth,
     bool invalid_loops_were_sanitized = false;
     for(auto &sample : samples)
     {
-        fluid_sample_t fluid{};
+        auto& fluid = samples_fluid.emplace_back();
         fluid.start = sample.start;
         fluid.end = sample.end - 1;
         fluid.samplerate = sample.samplerate;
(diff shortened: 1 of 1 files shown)


rsp4jack:
Hmm, is this bug a CVE? I just realized that. @derselbst

derselbst:
Oh, nice spot! I need to double check that fluidsynth behaves correctly when being confronted with a NULL-sample.

Regarding CVE: It may read or write to uninitialized memory, so technically, it would be worth a CVE. However, given that
* it took 2 weeks last time to only get the CVE-IDs issued by Github, AND
* that the CVEs have still not been published at [MITRE](https://www.cve.org/CVERecord?id=CVE-2026-58264), AND
* there's only version 2.5.6 affected,

I would refrain from requesting another CVE for this (that might be only scored 5 to 6). And rather release a fix in the next days.

derselbst (approved):
> I need to double check that fluidsynth behaves correctly when being confronted with a NULL-sample.

There is a check for zero length samples deep down in `fluid_rvoice_check_sample_sanity()`. So while this works well indeed, it's quite of an overhead. I'll cherry-pick your commit and add a sample NULL guard to before calling `fluid_synth_alloc_voice_LOCAL`, to avoid passing NULL samples into fluidsynth's DSP loop.

Issue #1808, opened 2026-07-18
Heavy glitches under heavy workloads using PipeWire audio driver


rsp4jack:
### FluidSynth version

```sh
FluidSynth runtime version 2.5.6
Copyright (C) 2000-2026 Peter Hanappe and others.
Distributed under the LGPL license.
SoundFont(R) is a registered trademark of Creative Technology Ltd.

FluidSynth executable version 2.5.6
Sample type=double
```

### Describe the bug

Under heavy workloads, when using PipeWire audio driver, there are enormous glitches than other (JACK, PulseAudio, ALSA) drivers.

Here are the test videos with audio, which demonstrates the pipewire performance data and the glitches, including PW, PA, JACK and ALSA (the image below is a thumbnail)

<img width="1572" height="672" alt="Thumbnail" src="https://github.com/user-attachments/assets/25ed36b9-e5df-4436-a8d9-a6ab15690fe8" />

[fluidsynth audio driver test.7z](https://buzzheavier.com/dqnhgn80qggi)

### Expected behavior

not so much glitches

### Steps to reproduce

MIDI: (This is not a GZ! Remove the .gz suffix) [Voyage 1.26 million wh.mid.gz](https://github.com/user-attachments/files/30151253/Voyage.1.26.million.wh.mid.gz)
Soundfont: (This is a GZ) [OmegaPiano.sf2.gz](https://github.com/user-attachments/files/30151269/OmegaPiano.sf2.gz)

Command:
```
fluidsynth -a pipewire -g 1 -z 64 -c 32 -o synth.sample-rate=48000 -o synth.polyphony=1024 -o audio.realtime-prio=1 -o audio.sample-format=float ./OmegaPiano.sf2 ./Voyage\ 1.26\ million\ wh.mid
```

### Additional context

Kernel: Linux 7.1.3-arch2-1 #1 SMP PREEMPT_DYNAMIC Tue, 14 Jul 2026 23:03:30 +0000 x86_64 GNU/Linux
FluidSynth build: 2.5.6-1 by Arch Linux

derselbst:
I can look into it, just to be sure the bug is not on fluidsynth's side. I need a few clarifications:

* When you say "heavy workload", do you just mean fluidsynth being busy to rendering a high-polyphonic tune, or do you mean that your CPU is heavily busy e.g. computing PI while executing fluidsynth?
* What was your motivation for overriding `audio.realtime-prio` (a vaule of 99 would be highest prio)?
[...]

PR #1811, opened 2026-07-19
Minor performance tweaks and profiling logs for pipewire

derselbst:
This PR implements some minor performance tweaks found while digging into #1808. After these changes, the MIDI file of the reference issue renders about 8 seconds faster for me (when using the fast file renderer). So a step in the right direction. Yet the big (but known) problems remain to be `fluid_rvoice_dsp_interpolate()` and `fluid_iir_filter_apply()`. These functions use inherently scalar arithmetics according to Intel VTune (see below) while dealing with lots of data, making them very slow.

Bildschirmfoto_20260719_175050

Apparently, due to the tiny arithmetic change in the IIR filter, MIDI that make use of the filter are now rendered slightly differently, causing some of our regression tests to hit the threshold and fail. This is expected though.


derselbst (commented):
Regression tests rehearsed, difference minimal and acceptable.

PR #1813, opened 2026-07-21
Some optimizations for PipeWire audio driver

rsp4jack:
* Support audio.periods
* Async node support (default on): 
  > If TRUE, creates an asynchronous PipeWire stream. This isolates synthesizer from PipeWire's synchronous processing graph, preventing system-wide audio dropouts (xruns) during heavy workload (such as high polyphones).
  > Requires PipeWire >= 0.3.73.
  > Enabling this automatically doubles the actual latency.

### DIFF
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -800,7 +800,7 @@ Developers:
             <desc>
                 If TRUE, creates an asynchronous PipeWire stream. This isolates synthesizer from PipeWire's synchronous processing graph, preventing system-wide audio dropouts (xruns) during heavy workload (such as high polyphones).
                 Requires PipeWire &gt;= 0.3.73.
-                <note>Enabling this automatically increases latency by exactly one period.</note>
+                <note>Enabling this automatically doubles the actual latency.</note>
             </desc>
         </setting>
         <setting>
--- a/doc/fluidsettings.xml
+++ b/doc/fluidsettings.xml
@@ -794,6 +794,16 @@ Developers:
             </desc>
         </setting>
         <setting>
+            <name>pipewire.async</name>
+            <type>bool</type>
+            <def>1 (TRUE)</def>
+            <desc>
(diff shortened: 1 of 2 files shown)


rsp4jack:
#1808
Based on #1811

derselbst (commented):
LGTM. I do see the benefit of the async mode. In my quick testing I was not able to confirm that changing `audio.periods` actually makes a difference. E.g. `-z 64 -c 2` just sounds as bad as `-z 64 -c 32`. Not sure if that's expected.

```bash
$ pipewire --version
pipewire
Compiled with libpipewire 1.6.8
Linked with libpipewire 1.6.8
```

rsp4jack:
After some profiling, I found that what [a0e7410](https://github.com/FluidSynth/fluidsynth/pull/1813/commits/a0e7410df41c531fc57c693c2dffaabe98dc1f23) actually do is force FS to render at driver (basically the sound card)'s quantum (in my case it is 2048).

As https://github.com/FluidSynth/fluidsynth/pull/982#issuecomment-921635253:

> the PipeWire API is severely undocumented

So I have no idea about how to manipulate PW's buffer pool. I think it is not what we can control directly.

But anyway, now FS requires latency at periods * period-size instead of max-latency, and I can confirm that now changing periods make a difference:

```
fluidsynth -a pipewire -g 1 -z 128 -c <...> -o audio.pipewire.async=1 -o synth.sample-rate=48000 -o synth.polyphony=2000 -o audio.realtime-prio=88 -o synth.cpu-cores=4 ...

-c 2  : 106 xruns
-c 64 :  66 xruns
```

rsp4jack:
> So I have no idea about how to manipulate PW's buffer pool. I think it is not what we can control directly.

Yes. See https://gitlab.freedesktop.org/pipewire/pipewire/-/work_items/2845. The current implementation is a low-hanging fruit, and an ideal implementation is to maintain our own buffers (like https://github.com/PipeWire/pipewire/blob/master/src/modules/module-vban/audio.c)

derselbst (approved):
Sounds good to me now, thanks!

PR #1836, opened 2026-08-29
Fix regression on pthread detection with FreeBSD

lhondareyte:
Since 1400ac9, cmake failed to detect threads on FreeBSD 14 + CLANG 19: 

```
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Performing Test HAVE_CXX_THREAD_PRIMITIVES
-- Performing Test HAVE_CXX_THREAD_PRIMITIVES - Failed
CMake Error at CMakeLists.txt:567 (message):
  Your compiler does not support C++11 threading primitives (<mutex>,
  <condition_variable>, <thread>).  This is required when building with
  osal=cpp11.
```

FreeBSD requires explicit linking with libpthread (-lpthread). This patch fixes the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -554,6 +554,11 @@ if ( osal STREQUAL "cpp11" )
     # Will silently fall back to a lower standard version if not available
     set( CMAKE_CXX_STANDARD 17 )
     include( CheckCXXSourceCompiles )
+    # Some OS - ie. FreeBSD - require explicitly linking pthread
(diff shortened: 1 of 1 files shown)


derselbst (approved):
LGTM, thanks! Unfortunately, we've lost our previous FreeBSD CI pipeline as it was based on Cirrus CI. If somebody wants to contribute a BSD CI pipeline I'd happily welcome it!

PR #1838, opened 2026-08-29
Freebsd workflow

pedrolcl:
This is a FreeBSD CI workflow based on the Solaris one, using https://github.com/vmactions/freebsd-vm


derselbst (approved):
Wow, that was quick. Thanks Pedro!

PR #1842, opened 2026-09-01
Allow setting pipewire node name

anupam-git:
Pipewire supports setting a device name when creating a node. Currently there is no way to set the node name when using pipewire audio driver backend. 

This patch introduces a new settings entry named ~`audio.pipewire.node-name`~`audio.pipewire.id`, which sets the pipewire node name property when using the pipewire audio backend driver.
If node name is not set, the device name defaults to `fluidsynth`

Update: As discussed below, the property name has been updated to `audio.pipewire.id` to be consistent with other audio backend drivers

### DIFF
--- a/src/drivers/fluid_pipewire.c
+++ b/src/drivers/fluid_pipewire.c
@@ -202,6 +202,7 @@ new_fluid_pipewire_audio_driver2(fluid_settings_t *settings, fluid_audio_func_t 
     char *media_role = NULL;
     char *media_type = NULL;
     char *media_category = NULL;
+    char *node_name = NULL;
     float *buffer = NULL;
     const struct spa_pod *params[1];
     struct pw_properties *props;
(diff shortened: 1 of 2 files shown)


anupam-git:
Implements #1843

derselbst (commented):
A good idea in general. However, I'd vote to name that setting `audio.pipewire.id`, to be consistent with oboe and jack audio drivers. And pls. document the new setting in `/doc/fluidsettings.xml`.

anupam-git:
Sure. I'll make the requested changes.

src/drivers/fluid_pipewire.c
derselbst, line 375:
```suggestion
    FLUID_FREE(client_name);

```

derselbst, line 367:
```suggestion
    FLUID_FREE(client_name);

```

derselbst (commented):
Two missing frees and then it should be good.

anupam-git (commented):
Applied the suggested changes. Variables are now freed when not in use

derselbst (approved):
LGTM now, thanks!

PR #1845, opened 2026-09-05
Reduce pulled in packages for FreeBSD workflow

diizzyy:
cmake is a meta package, cmake-core is cmake itself
git-lite has fewer dependencies and is enough for workflow


derselbst (approved):
LGTM, thanks!

PR #1846, opened 2026-09-05
Remove FreeBSD 13 from workflow

diizzyy:
Support for 13 ended April 30, 2026


diizzyy:
Reference: https://www.freebsd.org/security/unsupported/

derselbst (changes requested):
Even though support has ended, I'd prefer to keep it for now as it still works well, sry.

diizzyy:
Ok, thanks

